PR #1418, opened 2020-01-09
Simplify granule & PDR ingest & discovery

yjpa7145:

PR #1442, opened 2020-01-23
CUMULUS-1040: Refactor ingest package

markdboyd:
- Refactor `ingest` package to use new `aws-client` and `errors` packages

Should merge after #1418


packages/ingest/aws.js
```diff
@@ -48,340 +13,82 @@ function getEndpoint(local = false, port = 8000) {
...
-    };
-
-    const attr = await queue.getQueueAttributes(params).promise();
-    attr.Attributes.name = name;
-    return attr.Attributes;
-  }
-}
-
-
-class ECS {
```
markdboyd: This whole class wasn't used anywhere that I could see, so I removed it. Thoughts?

yjpa7145: +1 for removing it

packages/ingest/lock.js
```diff
@@ -18,12 +22,12 @@ async function checkOldLocks(bucket, list) {
   if (list) {
     let count = list.length;
 
-    list.forEach((item) => {
+    list.forEach(async (item) => {
```
yjpa7145: There is a bug here, there is nothing waiting for each Promise to resolve.

packages/aws-client/index.js
```diff
@@ -1,14 +1,12 @@
 exports.client = require('./client');
-exports.cloudformation = require('./cloudformation');
-exports.dynamo = require('./dynamo');
+exports.cloudformation = require('./CloudFormation');
```
yjpa7145: Seems like the exports in this file should match the case of the things that they're exporting. `s3` -> `S3`, for example.

packages/ingest/lock.js
```diff
@@ -1,72 +1,73 @@
...
+* @param {Array} list - The list of locks in the bucket
+* @returns {integer} - Number of locks remaining in bucket
 **/
 async function checkOldLocks(bucket, list) {
-  if (list) {
-    let count = list.length;
-
-    list.forEach((item) => {
-      const date = list[item].LastModified;
+  if (list.length > 0) {
```
yjpa7145: Suggested refactor of this function:

```js
async function checkOldLocks(bucket, locks = []) {
  const fiveMinutesAgo = Date.now() - (5 * 60 * 1000);
  const expiredLocks = locks.filter((lock) => lock.LastModified < fiveMinutesAgo);
  await Promise.all(expiredLocks.map((lock) => deleteS3Object(bucket, lock.Key)));

  return locks.length - expiredLocks.length;
}
```

packages/ingest/lock.js
```diff
@@ -1,72 +1,73 @@
...
+* Counts the number of locks in a bucket.
 *
 * @param {Object} bucket - The AWS S3 bucket to check
-* @param {string} pName - The provider name
+* @param {string} providerName - The provider name
 * @returns {integer} - Number of current locks in the bucket
 **/
-async function countLock(bucket, pName) {
-  const list = aws.s3().listObjectsV2({
+async function countLock(bucket, providerName) {
```
yjpa7145: 👍

PR #1446, opened 2020-01-28
CUMULUS-1684 Use KMS for provider credentials encryption

yjpa7145:
This PR updates the provider model to use KMS when encrypting provider credentials. This allows us to get rid of the S3 keypair generation (using openssl) in the Terraform deployment and, in a future release, remove the S3KeyPairProvider code.

**Changes**
- Update the version of Localstack to support kms
- Add a `providerSecretsMigration` lambda function. After deployment, integrators will need to invoke this function, which will migrate their encrypted provider credentials from plaintext or S3KeyPair encrypted to KMS encrypted.
- Add a `verifyProviderSecretsMigration` lambda function. This is not used now but, in the following release, it will be invoked as part of the Terraform plan stage and fail the deployment if it finds provider credentials which have _not_ been migrated to KMS. This would indicate that the integrator had not run the migration Lambda.
- Create a `@cumulus/sftp-client` package
- Update the `SftpClient` to be able to handle a plaintext password, an S3 keypair-encrypted password, or a KMS encrypted password. Once the provider credentials have been migrated to KMS, we can remove support for plaintext and keypair-encrypted credentials.
- Update the `FtpProviderClient` to be able to handle a plaintext password, an S3 keypair-encrypted password, or a KMS encrypted password. Once the provider credentials have been migrated to KMS, we can remove support for plaintext and keypair-encrypted credentials.
- Update the Providers model code to store passwords using KMS
- Add a `@cumulus/aws-client/KMS` module with KMS-related utility functions
- Add a `@cumulus/aws-client/S3.getTextObject()` function
- Mark the `S3KeyPairProvider` as deprecated
- Mark the `KMS` key pair provider as deprecated in favor of `@cumulus/aws-client/KMS`
- Deprecate the `Sftp` class in `@cumulus/common` in favor of the `@cumulus/sftp-client` package
- Remove the `rsa_keys` null resource from the TF deployment

**Remaining To Do**
- [x] Manual testing
- [ ] Update Changelog
- [ ] Document migration steps for providers


packages/api/tests/lambdas/test-verifyProviderSecretsMigration.js
```diff
@@ -0,0 +1,150 @@
...
+  });
+
+  await dynamodbDocClient().put({
+    TableName: process.env.ProvidersTable,
+    Item: { ...s3Provider, createdAt: Date.now() }
+  }).promise();
+
+  const err = await t.throwsAsync(handler());
+
+  t.is(
```
markdboyd: So this test works because you're only creating one provider that has KMS encryption, which will not throw an error, and the one using S3 key pair encryption that will throw an error. If you had a test where you setting up two failing cases, you wouldn't be able to guarantee which one throws an error first. 

I'm assuming you know this and are fine with the lambda working this way, but the test says `verifies all providers`, but with the Lambda as written it will stop verification as soon as it hits the first failing case

packages/api/tests/models/test-providers-model.js
```diff
@@ -139,3 +167,36 @@ test('Providers.update() throws a ValidationError if an invalid host is used', a
...
+
+test('Providers.update() encrypts the credentials using KMS', async (t) => {
+  const providersModel = new Provider();
+
+  const provider = fakeProviderFactory({
+    username: 'my-username-1',
+    password: 'my-password-1'
+  });
+
+  await providersModel.create(provider);
```
markdboyd: Would it be better for this initial provider record to be written with plaintext credentials just to prove the process of updating plaintext -> encrypted? It ultimately doesn't matter much I think because you're updating the credentials that get encrypted and verifying that the updated values come back

packages/aws-client/S3.js
```diff
@@ -271,9 +271,26 @@ exports.getS3Object = improveStackTrace(
...
+exports.getTextObject = (bucket, key) =>
   exports.getS3Object(bucket, key)
-    .then(({ Body }) => JSON.parse(Body.toString()));
+    .then(({ Body }) => Body.toString());
+
+/**
+ * Fetch JSON stored in an S3 object
+ * @param {string} bucket - the S3 object's bucket
+ * @param {string} key - the S3 object's key
+ * @returns {Promise<*>} the contents of the S3 object, parsed as JSON
```
markdboyd: The type of the return object should be a string, right?

packages/aws-client/tests/test-KMS.js
```diff
@@ -0,0 +1,41 @@
...
+    kms().describeKey({
+      KeyId: createKeyResponse.KeyMetadata.KeyId
+    }).promise()
+  );
+});
+
+test('encrypt() properly encrypts a value', async (t) => {
+  const ciphertext = await KMS.encrypt(t.context.KeyId, 'asdf');
+
+  const plaintext = await kms().decrypt({
```
markdboyd: I assume you're not using `KMS.decryptBase64String` here so that the test is only unit testing one function

packages/aws-client/tests/test-S3.js
```diff
@@ -41,6 +43,28 @@ test.after.always(async (t) => {
...
+test('getTextObject() returns the contents of an S3 object', async (t) => {
+  const { Bucket } = t.context;
+  const Key = randomString();
+
+  await awsServices.s3().putObject({ Bucket, Key, Body: 'asdf' }).promise();
+
+  t.is(await getTextObject(Bucket, Key), 'asdf');
+});
+
+test('getJsonS3Object() returns the JSON-parsed contents of an S3 object', async (t) => {
```
markdboyd: 👍

packages/common/key-pair-provider.js
```diff
@@ -25,6 +26,8 @@ class S3KeyPairProvider {
    * @returns {Promise} the encrypted string
    */
   static async encrypt(str, keyId = 'public.pub', bucket = null, stack = null) {
+    deprecate('@cumulus/common/key-pair-provider', '1.17.0', '@cumulus/aws-client/KMS');
```
markdboyd:
```suggestion
    deprecate('@cumulus/common/key-pair-provider', '1.17.0', '@cumulus/aws-client/KMS.encrypt');
```

packages/common/key-pair-provider.js
```diff
@@ -45,6 +48,8 @@ class S3KeyPairProvider {
    * @returns {Promise.<string>} the decrypted string
    */
   static async decrypt(str, keyId = 'private.pem', bucket = null, stack = null) {
+    deprecate('@cumulus/common/key-pair-provider', '1.17.0', '@cumulus/aws-client/KMS');
```
markdboyd:
```suggestion
    deprecate('@cumulus/common/key-pair-provider', '1.17.0', '@cumulus/aws-client/KMS.decryptBase64String');
```

packages/common/kms.js
```diff
@@ -4,9 +4,12 @@ const AWS = require('aws-sdk');
 const { createErrorType } = require('@cumulus/errors');
 
 const KMSDecryptionFailed = createErrorType('KMSDecryptionFailed');
+const { deprecate } = require('./util');
 
 class KMS {
   static async encrypt(text, kmsId) {
+    deprecate('@cumulus/common/key-pair-provider', '1.17.0', '@cumulus/aws-client/KMS');
```
markdboyd:
```suggestion
    deprecate('@cumulus/common/key-pair-provider', '1.17.0', '@cumulus/aws-client/KMS.encrypt');
```

packages/common/kms.js
```diff
@@ -18,6 +21,8 @@ class KMS {
   }
 
   static async decrypt(text) {
+    deprecate('@cumulus/common/key-pair-provider', '1.17.0', '@cumulus/aws-client/KMS');
```
markdboyd:
```suggestion
    deprecate('@cumulus/common/key-pair-provider', '1.17.0', '@cumulus/aws-client/KMS.decryptBase64String');
```

packages/ingest/FtpProviderClient.js
```diff
@@ -1,54 +1,63 @@
...
-      }
+  async getPassword() {
+    if (!this.plaintextPassword) {
+      this.plaintextPassword = await decrypt(this.providerConfig.password);
     }
+
+    return this.plaintextPassword;
+  }
+
+  async buildFtpClient() {
```
markdboyd: 👍

packages/ingest/test/test-FtpProviderClient.js
```diff
@@ -14,6 +17,85 @@ const { s3 } = require('@cumulus/aws-client/services');
...
+test('FtpProviderClient supports plaintext usernames and passwords', async (t) => {
+  const ftpClient = new FtpProviderClient({
+    host: '127.0.0.1',
+    encrypted: false,
+    username: 'testuser',
+    password: 'testpass',
+    useList: true
+  });
+
+  const files = await ftpClient.list('/');
```
markdboyd: Is the FTP client mocked somehow that I'm not seeing? How is the unit tests listing results?

packages/ingest/test/test-FtpProviderClient.js
```diff
@@ -14,6 +17,85 @@ const { s3 } = require('@cumulus/aws-client/services');
...
+  const fileNames = files.map((f) => f.name);
+
+  t.true(fileNames.includes('index.html'));
+});
+
+test('FtpProviderClient supports S3-keypair-encrypted usernames and passwords', async (t) => {
+  const username = await S3KeyPairProvider.encrypt('testuser');
```
markdboyd: So we're deprecating our S3 key pair provider but it is still supported by the FTP client?

packages/sftp-client/index.js
```diff
@@ -0,0 +1,267 @@
...
+    this.config = config;
+    this.connected = false;
+
+    if (get(config, 'encrypted', false) === false) {
+      this.plaintextUsername = config.username;
+      this.plaintextPassword = config.password;
+    }
+  }
+
+  async getUsername() {
```
markdboyd: This might be overkill, but any way to share these functions with the base provider client from above using composition or inheritance?

packages/sftp-client/tests/test-index.js
```diff
@@ -0,0 +1,214 @@
...
+    Bucket: process.env.system_bucket,
+    Key: `${process.env.stackName}/crypto/${privateKeyName}`,
+    Body: await KMS.encrypt(t.context.kmsKeyId, unencryptedPrivateKey)
+  });
+
+  const sftpClient = new SftpClient({
+    host: '127.0.0.1',
+    port: '2222',
+    username: 'user',
+    cmKeyId: 'yes',
```
markdboyd: I guess any truthy value will work, but for the sake of a test as documentation, should this be `true?

tf-modules/archive/bootstrap.tf
```diff
@@ -1,12 +1,4 @@
-resource "null_resource" "rsa_keys" {
```
markdboyd: Awesome 💯

tf-modules/archive/bootstrap.tf
```diff
@@ -1,12 +1,4 @@
...
-  triggers = { x = uuid() }
-  provisioner "local-exec" {
-    command = "${path.module}/fetch_or_create_rsa_keys.sh ${var.system_bucket} ${var.prefix}"
```
markdboyd: Does that mean this `fetch_or_create_rsa_keys.sh` script can be deleted as well?

tf-modules/archive/provider_credentials_encryption.tf
```diff
@@ -0,0 +1,67 @@
...
+    resources = [aws_kms_key.provider_kms_key.arn]
+  }
+}
+
+resource "aws_iam_role_policy" "provider_secrets_encryption" {
+  role   = aws_iam_role.lambda_api_gateway.id
+  policy = data.aws_iam_policy_document.provider_secrets_encryption.json
+}
+
+resource "aws_lambda_function" "provider_secrets_migration" {
```
markdboyd: To state the obvious: these lambdas are only run manually? Are we going to document for users that they need to do this? Or more precisely: when do users need to run this?

packages/api/lambdas/providerSecretsMigration.js
```diff
@@ -0,0 +1,38 @@
...
+const { isNonEmptyString } = require('@cumulus/common/string');
+const { isNil } = require('@cumulus/common/util');
+const Provider = require('../models/providers');
+
+const getDecryptedField = async (provider, field) => {
+  if (isNil(provider[field])) return null;
+  if (provider.encrypted === false) return provider[field];
+
+  return KMS.decryptBase64String(provider[field])
+    .catch(() => S3KeyPairProvider.decrypt(provider[field]));
```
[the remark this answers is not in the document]
yjpa7145: So passwords (and usernames) can be in four states:

1) empty
2) plaintext (`encrypted = false`)
3) encrypted using S3 keypair encryption (`encrypted = true` and provider created _before_ this deployment)
4) encrypted using KMS encryption (`encrypted = true` and provider created _after_ this deployment)

If `encrypted = true`, then there isn't a way to tell what method was used to encrypt the password. Here we're trying to decrypt it using KMS and, if that fails, we assume that it was encrypted using S3 keypair instead.

packages/api/lambdas/providerSecretsMigration.js
```diff
@@ -0,0 +1,38 @@
...
+  const providerModel = new Provider();
+  return providerModel.update({ id: provider.id }, updates);
+};
+
+const handler = async () => {
+  const scanResponse = await dynamodbDocClient().scan({
+    TableName: process.env.ProvidersTable
+  }).promise();
+
+  await Promise.all(scanResponse.Items.map(migrateProvider));
```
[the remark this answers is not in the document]
yjpa7145: Provider usernames and passwords stored in the database are encrypted using [S3KeypairProvider](https://github.com/nasa/cumulus/blob/master/packages/common/key-pair-provider.js#L15), which this ticket is deprecating. Going forward, those values will be encrypted using KMS. Before we can remove that S3 key pair support, we need to re-encrypt the values using KMS.

packages/api/lambdas/verifyProviderSecretsMigration.js
```diff
@@ -0,0 +1,29 @@
...
+const KMS = require('@cumulus/aws-client/KMS');
+const { dynamodbDocClient } = require('@cumulus/aws-client/services');
+const { isNil } = require('@cumulus/common/util');
+
+const verifyProvider = async (provider) => {
+  if (provider.encrypted === true) {
+    try {
+      await KMS.decryptBase64String(provider.username);
+      await KMS.decryptBase64String(provider.password);
+    } catch (_) {
```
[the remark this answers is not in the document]
yjpa7145: The errors that are thrown don't give you a specific issue, they say "invalid key or maybe you don't have access to the key or ...". In this case, we're just using the Lambda function to stop a deployment that has _not_ yet run the migration script. My vote is to leave it the way it is.

packages/api/lib/ems.js
```diff
@@ -108,21 +108,20 @@ async function getExpiredS3Objects(bucket, prefix, retentionInDays) {
...
+    log.debug('EMS reports are not configured to be sent');
+    return reports;
+  }
+
   const emsConfig = {
     username: process.env.ems_username,
     host: process.env.ems_host,
     port: process.env.ems_port,
-    privateKey: process.env.ems_privateKey || 'ems-private.pem',
-    submitReport: process.env.ems_submitReport === 'true' || false
```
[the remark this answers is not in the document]
yjpa7145: Correct.

packages/api/models/providers.js
```diff
@@ -1,15 +1,17 @@
...
 const { isNil } = require('@cumulus/common/util');
 const { isValidHostname } = require('@cumulus/common/string');
 
 const Manager = require('./base');
-const providerSchema = require('./schemas').provider;
 const Rule = require('./rules');
+const schemas = require('./schemas');
 const { AssociatedRulesError } = require('../lib/errors');
 
+const encrypt = (x) => KMS.encrypt(process.env.provider_kms_key_id, x);
```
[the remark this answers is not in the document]
yjpa7145: Fixed in f0180788e

packages/api/models/providers.js
```diff
@@ -51,50 +53,30 @@ class Provider extends Manager {
...
+    if (item.username || item.password) record.encrypted = true;
 
-  async update(key, _item, keysToDelete = []) {
-    const item = _item;
-    // encrypt the password
-    if (item.password) {
-      item.password = await this.encrypt(item.password);
-      item.encrypted = true;
-    }
+    if (item.username) record.username = await encrypt(item.username);
```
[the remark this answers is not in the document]
yjpa7145: Yes. It's also called any time a provider's username or password are updated using the API.

packages/api/tests/lambdas/test-providerSecretsMigration.js
```diff
@@ -0,0 +1,232 @@
...
+  await dynamodbDocClient().put({
+    TableName: process.env.ProvidersTable,
+    Item: { ...provider, createdAt: Date.now() }
+  }).promise();
+
+  await handler();
+
+  const fetchedProvider = await providerModel.get({ id: provider.id });
+  t.is(fetchedProvider.encrypted, false);
+  t.is(await fetchedProvider.username, undefined);
```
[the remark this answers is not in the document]
yjpa7145: Copy / Paste error, will fix.

packages/api/tests/lambdas/test-providerSecretsMigration.js
```diff
@@ -0,0 +1,232 @@
...
+  const kmsProvider = fakeProviderFactory({
+    protocol: 'ftp',
+    username: 'my-username',
+    password: 'my-password'
+  });
+
+  await providerModel.create(kmsProvider);
+
+  // Create and store the S3KeyPairProvider provider
+  const username = await S3KeyPairProvider.encrypt('my-username');
```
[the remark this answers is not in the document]
yjpa7145: Yes. This unit test is verifying that all providers are updated (if necessary).

packages/api/tests/lambdas/test-providerSecretsMigration.js
```diff
@@ -0,0 +1,232 @@
...
+  }).promise();
+
+  // Create and store the KMS provider
+  const kmsProvider = fakeProviderFactory({
+    protocol: 'ftp',
+    username: 'my-username',
+    password: 'my-password'
+  });
+
+  await providerModel.create(kmsProvider);
```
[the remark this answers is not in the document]
yjpa7145: Yes. This unit test is verifying that all providers are updated (if necessary).

packages/api/tests/lambdas/test-providerSecretsMigration.js
```diff
@@ -0,0 +1,232 @@
...
+
+  await handler();
+
+  const fetchedProvider = await providerModel.get({ id: provider.id });
+  t.is(fetchedProvider.encrypted, true);
+  t.is(await KMS.decryptBase64String(fetchedProvider.username), 'my-username');
+  t.is(await KMS.decryptBase64String(fetchedProvider.password), 'my-password');
+});
+
+test.serial('Empty, Plaintext, KMS-encrypted, and S3KeyPairProvider-encrypted credentials are all handled properly', async (t) => {
```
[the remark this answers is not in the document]
yjpa7145: Yes. This unit test is verifying that all providers are updated (if necessary).

packages/api/lambdas/verifyProviderSecretsMigration.js
```diff
@@ -0,0 +1,29 @@
...
+    throw new Error(`Provider ${provider.id} has plaintext username or password. Must invoke the providerSecretsMigration Lambda function.`);
+  }
+};
+
+const handler = async () => {
+  const scanResponse = await dynamodbDocClient().scan({
+    TableName: process.env.ProvidersTable
+  }).promise();
+
+  await Promise.all(scanResponse.Items.map(verifyProvider));
```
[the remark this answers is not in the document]
yjpa7145: Yes. This isn't being used right now, we're just deploying it so that it will be available in future releases. In the release _after_ the one that includes this migration, we are going to invoke this Lambda as part of the Terraform deployment. It will serve as a circuit breaker to make sure that someone can't deploy a later version of Cumulus without having invoked the migration lambda.

packages/api/tests/lambdas/test-verifyProviderSecretsMigration.js
```diff
@@ -0,0 +1,150 @@
...
+  await dynamodbDocClient().put({
+    TableName: process.env.ProvidersTable,
+    Item: { ...provider, createdAt: Date.now() }
+  }).promise();
+
+  const err = await t.throwsAsync(handler());
+
+  t.is(
+    err.message,
+    `Provider ${provider.id} has plaintext username or password. Must invoke the providerSecretsMigration Lambda function.`
```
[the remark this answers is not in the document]
yjpa7145: Maybe, but the expectation is that this Lambda function is going to serve one purpose and then be removed in a couple versions. If we're maintaining it, then something has gone horribly wrong.

packages/api/tests/lambdas/test-verifyProviderSecretsMigration.js
yjpa7145: As soon as we find a single provider that's not encrypted using KMS, we know that the migration has not been run, so we stop everything right there.

packages/api/tests/models/test-providers-model.js
yjpa7145: Yeah, whatever values are there are going to be overwritten, and this test only cares about what those new values are. Incidentally, the providers model doesn't provide a way to store plaintext values anyway. They're always encrypted.

packages/aws-client/S3.js
yjpa7145: It's being sent through `JSON.parse`, so it could be an object, array, string, number, boolean, etc.

packages/aws-client/tests/test-KMS.js
yjpa7145: Correct.

packages/ingest/test/test-FtpProviderClient.js
yjpa7145: It's hitting the FTP server running in Docker.

packages/ingest/test/test-FtpProviderClient.js
yjpa7145: We have to keep support for it for the time being because, when this release is rolled out, we need to make sure that things work before and after the migration has been run. When this is initially released, the provider credentials will still be encrypted using S3. Once the migration has been run, they will be encrypted using KMS. In a future release, we will remove this S3 support and just use KMS.

packages/sftp-client/index.js
yjpa7145: I thought about it and came down on the "overkill" side of the line, especially since they're in two separate npm packages.

packages/sftp-client/tests/test-index.js
yjpa7145: According to the provider schema, it should be a string.

```js
cmKeyId: {
  type: 'string',
  description: 'AWS KMS Customer Master Key arn or alias'
}
```

That being said, the value of `cmKeyId` is never actually used. There's definitely some refactoring there that should be done but, since I didn't add that code as part of this ticket, I thought that it was out of scope.

tf-modules/archive/bootstrap.tf
yjpa7145: Good catch, I thought I had.

tf-modules/archive/provider_credentials_encryption.tf
yjpa7145: First question: "yes"

Second question: Yeah, there's still a documentation part of this ticket that needs to be finished. I wanted to make sure that it all made sense and that there weren't going to be any major changes before I wrote the user docs / changelog.

packages/api/lambdas/providerSecretsMigration.js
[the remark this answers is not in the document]
markdboyd: But if it's already encrypted with KMS and the provider was created after the deployment, then it doesn't need to be migrated... correct?

packages/api/lambdas/verifyProviderSecretsMigration.js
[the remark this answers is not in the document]
markdboyd: Hmm, but what if this script is only failing because of some weirdness with KMS or a deployment error that **we** caused? Wouldn't we want the error to specifically identify the problem?

packages/api/tests/lambdas/test-verifyProviderSecretsMigration.js
markdboyd: So by `verifies all providers`, we mean `throws error if any provider is not encrypted using KMS`. I'm not suggesting that you rename the test, just clarifying

packages/aws-client/S3.js
markdboyd: TIL: JSON.parse can return other values. OK.

tf-modules/archive/provider_credentials_encryption.tf
markdboyd: OK. I'm fine with docs being a separate ticket but I'd like to see the CHANGELOG entries in this PR. I think it's nice for historical review and it's also easier imo to include the CHANGELOG entries in the PR with the actual changes because it reminds me what changed

packages/sftp-client/tests/test-index.js
markdboyd: I see. Because `cmKeyId` actually will come from a provider object when the `SftpClient` is initialized?

https://github.com/nasa/cumulus/pull/1446/files?file-filters%5B%5D=.js&file-filters%5B%5D=.tf&file-filters%5B%5D=.yml#diff-bc5d232e73466353582fee67f9820ca8R21

packages/sftp-client/index.js
markdboyd: I think it could make some sense to put both of the FTP clients into a single package, but not blocking for this PR

packages/api/tests/lambdas/test-verifyProviderSecretsMigration.js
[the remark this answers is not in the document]
markdboyd: And there wouldn't be any additional utility in an error class named `ProviderPlaintextCredentials` or something?

packages/api/lambdas/providerSecretsMigration.js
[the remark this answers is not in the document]
yjpa7145: That's correct. The problem is that, if a password is encrypted, there's no way to tell which encryption algorithm was used to encrypt it. The only way to tell is to try to decrypt it using both algorithms and see which one works.

packages/api/lambdas/verifyProviderSecretsMigration.js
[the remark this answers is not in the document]
yjpa7145: Updated in 404b9f4e7

packages/api/tests/lambdas/test-verifyProviderSecretsMigration.js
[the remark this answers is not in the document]
yjpa7145: I don't think so, since this exception is never being caught anywhere, so there won't be any code deciding what to do based on the type of error.

packages/api/tests/lambdas/test-verifyProviderSecretsMigration.js
yjpa7145: Correct.

packages/sftp-client/tests/test-index.js
yjpa7145: Correct.

tf-modules/archive/provider_credentials_encryption.tf
yjpa7145: I'm going to include the Changelog entries and migration instructions in this PR, I just wanted to make sure we were all in agreement with the changes before I bothered writing the migration instructions.

yjpa7145: Added migration notes in the Changelog

CHANGELOG.md
```diff
@@ -6,6 +6,37 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### BREAKING CHANGES
+
+- **CUMULUS-1684**
```
markdboyd: Not all of the contents of this list are breaking changes. The stuff that is just additions should be moved under the `Added` header, etc

packages/api/lambdas/providerSecretsMigration.js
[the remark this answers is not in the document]
markdboyd: So the bottom line is that as long as a non-empty string value comes back from `getDecryptedField`, it is used to update the provider? Even if the value came back decrypted from KMS and doesn't actually *need* to be updated?

packages/api/tests/lambdas/test-verifyProviderSecretsMigration.js
```diff
@@ -0,0 +1,145 @@
...
+  );
+});
+
+test.serial.only('verifyProviderSecretsMigration fails if S3 keypair credentials are found', async (t) => {
```
markdboyd: Need to remove `.only` from here. And you should ask @flamingbear about the git secret he uses to avoid committing these

CHANGELOG.md
```diff
@@ -6,6 +6,37 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
+  - In previous releases, a provider's username and password were encrypted
+    using a custom encryption library. That has now been updated to use KMS.
+    This release includes a Lambda function named
+    `<prefix>-ProviderSecretsMigration`, which will re-encrypt existing
+    provider credentials to use KMS. After this release has been deployed, you
+    will need to manually invoke that Lambda function using either the AWS CLI
+    or AWS Console. It should only need to be run once.
+  - Future releases of Cumulus will invoke a
```
markdboyd: These first two points should really be under a "Please note" or "Migration steps" header

CHANGELOG.md
```diff
@@ -6,6 +6,37 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
+### BREAKING CHANGES
+
+- **CUMULUS-1684**
+  - In previous releases, a provider's username and password were encrypted
+    using a custom encryption library. That has now been updated to use KMS.
+    This release includes a Lambda function named
+    `<prefix>-ProviderSecretsMigration`, which will re-encrypt existing
+    provider credentials to use KMS. After this release has been deployed, you
+    will need to manually invoke that Lambda function using either the AWS CLI
+    or AWS Console. It should only need to be run once.
```
markdboyd: >It should only need to be run once.

Assuming it doesn't fail?

packages/api/lambdas/providerSecretsMigration.js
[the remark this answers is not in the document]
yjpa7145: Yes.

CHANGELOG.md
yjpa7145: Fixed in fc0dbd4d7

packages/api/tests/lambdas/test-verifyProviderSecretsMigration.js
yjpa7145: Fixed in fc0dbd4d7

CHANGELOG.md
yjpa7145: Fixed in fc0dbd4d7

CHANGELOG.md
yjpa7145: Fixed in fc0dbd4d7

packages/api/tests/lambdas/test-verifyProviderSecretsMigration.js
flamingbear: for the record it's simply adding `	patterns = test\\.only` to your .git/config file under `[secrets]`

PR #1455, opened 2020-01-31
update aws-client deps to version 1.0-alpha6

markdboyd:
@Jkovarik and I noticed on a previous PR that Lerna is sometimes happy if the version number for a Lerna package is not the same across all packages, so this PR fixes all packages to depend on the same version of `aws-client`


markdboyd:
This is no longer relevant after #1460

PR #1465, opened 2020-02-05
Cumulus 1741

laurenfrederick:
Addresses [CUMULUS-1741](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1741)

## Changes

* Allow users to pass in a security group id to use with their ES instance
* Attach the ES security group to the EC2 instance (this fixes the ES index from database async operation timing out)

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


tf-modules/cumulus/ecs_cluster_instance_autoscaling_cf_template.yml.tmpl
```diff
@@ -4,6 +4,7 @@ Resources:
     Properties:
       SecurityGroups:
         - ${security_group_id}
+        - ${es_security_group_id}
```
nemreid: When passing in a security group ID, users will need to be aware that this will apply to all EC2 instances in their cluster, and will be [most permissive](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-security-groups.html#security-group-rules):

>If there is more than one rule for a specific port, we apply the most permissive rule. For example, if you have a rule that allows access to TCP port 22 (SSH) from IP address 203.0.113.1 and another rule that allows access to TCP port 22 from everyone, everyone has access to TCP port 22.

Perhaps the changelog note will suffice for this purpose.
We may want to consider it naming it something more descriptive, since it does not apply just to ElasticSearch now.

tf-modules/data-persistence/terraform.tfvars.example
```diff
@@ -19,4 +19,4 @@ enable_point_in_time_tables = [
 ]
 include_elasticsearch  = true
 subnet_ids             = ["subnet-123456"]
-security_groups        = ["sg-123456"]
```
nemreid: While you're at it you may want to delete `create_service_linked_role` from, and add `custom_domain_name` to, this example file.

tf-modules/data-persistence/variables.tf
```diff
@@ -63,3 +63,9 @@ variable "subnet_ids" {
   type        = list(string)
   default     = []
 }
+
+variable "security_group" {
+  description = "Security Group Ids for Elasticsearch"
```
markdboyd:
```suggestion
  description = "Security Group Id for Elasticsearch"
```

PR #1466, opened 2020-02-05
Add the ability to set a custom backend api url in the archive module (from top level Cumulus)

ifestus:
**Summary:** Currently there doesn't appear to be a way to set a custom archive api from the top level Cumulus module. That is, there isn't a way to set [api_url](https://github.com/nasa/cumulus/blob/master/tf-modules/archive/variables.tf#L124) from [cumulus/archive.tf](https://github.com/nasa/cumulus/blob/master/tf-modules/cumulus/archive.tf). This PR adds that capability. Analogous to the `distribution_url` for the distribution api.

The use-case this attempts to solve is if a user wants to sit their backend-api behind a Cloudfront endpoint.

## Changes

* Add `api_url` definition in `tf-modules/cumulus/archive.tf`
* Add `archive_api_url` variable in `tf-modules/cumulus/variables.tf`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+
+- **CUMULUS-1459**
+  - Async Operations are indexed in Elasticsearch
+  - To index any existing async operations you'll need to perform an index from database function.
+
 ### Fixed
 
 - **CUMULUS-1740** - `cumulus_meta.workflow_start_time` is now set in Cumulus
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 - **CUMULUS-1740** - `cumulus_meta.workflow_start_time` is now set in Cumulus
   messages
-- **Ability to set custom backend api url in the archive module**
+
+### Added
+
+- **Ability to set custom backend API url in the archive module**
   - Add `api_url` definition in `tf-modules/cumulus/archive.tf`
   - Add `archive_api_url` variable in `tf-modules/cumulus/variables.tf`
 - **Fix default values for urs_url in variables.tf files**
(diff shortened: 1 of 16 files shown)


tf-modules/cumulus/variables.tf
```diff
@@ -487,3 +487,9 @@ variable "non_complete_execution_payload_timeout" {
   type        = number
   default     = 30
 }
+
+variable "archive_api_url" {
```
markdboyd: There should be an associated CHANGELOG entry for this. Also, would you mind adding one for the change to the default `urs_url` variables that you made?

ifestus: :+1: sounds good to me, will do

ifestus: 4bc0855

tf-modules/cumulus/variables.tf
```diff
@@ -487,3 +487,9 @@ variable "non_complete_execution_payload_timeout" {
...
+  type        = string
+  default     = null
+  description = "If not specified, the value of the API Gateway endpoint is used"
```
ifestus:
```suggestion
  description = "If not specified, the value of the Backend (Archive) API Gateway endpoint is used"
```

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 - **CUMULUS-1740** - `cumulus_meta.workflow_start_time` is now set in Cumulus
   messages
+- **Ability to set custom backend api url in the archive module**
```
markdboyd: Super nitpick: can we capitalize api to API in this sentence? And move the notes here about what was added under an `###Added` header?

ifestus: d29a1e3

CHANGELOG.md
```diff
@@ -11,6 +11,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - **CUMULUS-1740** - `cumulus_meta.workflow_start_time` is now set in Cumulus
   messages
 
+### Added
+
+- **Ability to set custom backend API url in the archive module**
+  - Add `api_url` definition in `tf-modules/cumulus/archive.tf`
+  - Add `archive_api_url` variable in `tf-modules/cumulus/variables.tf`
+- **Fix default values for urs_url in variables.tf files**
```
markdboyd: Sorry, this one should still be under the "Fixed" header. but the other ones should stay here.

ifestus: my bad - updated in a95c093

PR #1468, opened 2020-02-07
Cumulus 1610 - Backup and Recovery Docs

laurenfrederick:
**Summary:** Backup and Recovery docs for contractual deliverable

Addresses [CUMULUS-1610](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1610)

## Changes

* Moved backup and recovery to its own page
* Added instructions on how to recover using Terraform
* Fixed Cumulus backup and restore - fixed unit test and removed int test
* Added `ignore_changes = [ name ]` to tables, which will let you recover to a different table name if you so choose

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


example/spec/backupRestore/BackupRestoreSpec.js
```diff
@@ -1,84 +0,0 @@
-'use strict';
```
kkelly51: So did we delete this whole file because it tests the same thing as the api unit test?

packages/api/tests/cli/test-back-restore.js
```diff
@@ -61,27 +61,10 @@ test.serial('backup records from DynamoDB', async (t) => {
...
-  }
-  fs.writeFileSync(tempRestoreFile, fileContent);
+  await restore(tempBackupFile, tableName, 1);
 
-  await restore(tempRestoreFile, tableName, 1);
-
-  // count the records
-  const resp = await gModel.scan(null, null, 0, 'COUNT');
-  t.is(resp.Count, limit);
+  // verify records are the same
```
kkelly51: This is fine

tf-modules/data-persistence/dynamo.tf
```diff
@@ -30,6 +30,7 @@ resource "aws_dynamodb_table" "access_tokens_table" {
 
   lifecycle {
     prevent_destroy = true
+    ignore_changes = [ name ]
```
kkelly51: What does this do?

docs/features/backup_and_restore.md
```diff
@@ -0,0 +1,116 @@
...
+
+PITR gives your DynamoDB tables continuous protection from accidental writes and deletes. With PITR, you do not have to worry about creating, maintaining, or scheduling backups. You enable PITR on your table and your backup is available for restore at any point in time from the moment you enable it, up to a maximum of the 35 preceding days. For example, imagine a test script writing accidentally to a production DynamoDB table. You could recover your table to any point in time within the last 35 days.
+
+On-demand backups help with long-term archival requirements for regulatory compliance. On-demand backups give you full-control of managing the lifecycle of your backups, from creating as many backups as you need to retaining these for as long as you need.
+
+## Enabling PITR during deployment
+
+By default, the Cumulus [data-persistence module](https://github.com/nasa/cumulus/blob/master/tf-modules/data-persistence) enables PITR on the default tables listed in the [module's variable defaults](https://github.com/nasa/cumulus/blob/master/tf-modules/data-persistence/variables.tf) for `enable_point_in_time_tables`. At the time of writing, that list includes:
+
+- CollectionsTable
```
kkelly51: Do we want to add Async Ops to this list?

example/spec/backupRestore/BackupRestoreSpec.js
laurenfrederick: Yeah I did. Both the unit and integration tests were broken, since the thing didn't work and I didn't see us testing anything differently in integration vs. unit tests.

tf-modules/data-persistence/dynamo.tf
laurenfrederick: In the docs, if you want to restore a table to an earlier version you have to restore to a different table name. To get that new table into your deployment you import it into Terraform with `terraform import`. When you go to redeploy our table names are hardcoded to something like `prefix-CollectionsTable`. Without this it'll try to create `prefix-CollectionsTable` even if you've imported `collectionstable-2` which will create you a new table and then you lose all your data. With this line, if you import a table with a different name it'll ignore those changes and not try to completely redeploy/recreate your table if only the name has changed.

docs/features/backup_and_restore.md
laurenfrederick: Yeah I can add it

docs/features/backup_and_restore.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+the backup will be stored at `backups/<table-name>.json`
+
+#### Restore with the CLI
+
+To restore data from a json file run the following command:
+
+```bash
+     ./node_modules/.bin/cumulus-api restore backups/<table-name>.json --table <new-table-name>
```
markdboyd: This can just be `cumulus-api` instead of `./node_modules/.bin/cumulus-api`, correct?

docs/features/data_in_dynamodb.md
```diff
@@ -13,71 +13,8 @@ hide_title: true
...
-- **Collections:** collection records
-- **Providers:** provider records
-- **Rules:** rules for managing and running workflows
-- **Executions:** workflow executions (step function executions)
-- **Granules:** granules processed by the Cumulus instance
-- **PDRs:** PDRs processed in Cumulus
-
 Amazon DynamoDB stores three geographically distributed replicas of each table to enable high availability and data durability. Amazon DynamoDB runs exclusively on solid-state drives (SSDs). SSDs help AWS achieve the design goals of predictable low-latency response times for storing and accessing data at any scale.
 
-## Backup and Restore with AWS
```
markdboyd: The header on line 9 should be removed as well if this selection is deleted

docs/features/backup_and_restore.md
```diff
@@ -0,0 +1,123 @@
...
+id: backup_and_restore
+title: Cumulus Backup and Restore
+hide_title: true
+---
+
+# Cumulus Backup and Restore
+
+## Deployment Backup and Restore
+
+Most of your Cumulus deployment can be recovered by redeploying via Terraform. The Cumulus metadata including providers, collections, granules, rules, and executions are stored in [DynamoDB](./data_in_dynamodb) and can be setup to be backed up and restored. If a deployment is lost, logs and Step Function executions in the AWS console will be irrecoverable.
```
markdboyd:
```suggestion
Most of your Cumulus deployment can be recovered by redeploying via Terraform. However, the Cumulus metadata including providers, collections, granules, rules, and executions that is stored in [DynamoDB](./data_in_dynamodb) can only be restored if backup was configured or enabled. If a deployment is lost, logs and Step Function executions in the AWS console will be irrecoverable.
```

docs/features/backup_and_restore.md
```diff
@@ -0,0 +1,123 @@
...
+
+At this point the data will be in DynamoDB, but not in Elasticsearch, so nothing will be returned on the Operator dashboard or through Operator API calls. To get the data into Elasticsearch, run an [`index-from-database` operation](https://nasa.github.io/cumulus-api/#index-from-database) via the Operator API. The status of this operation can be viewed on the dashboard. When Elasticsearch is switched to the recovery index the data will be visible on the dashboard and available via the Operator API.
+
+#### Restoring an individual table
+
+A table can be restored to a previous state using PITR. This is easily achievable via the AWS Console by visiting the `Backups` tab for the table.
+
+A table can only be recovered to a new table name. Following the restoration of the table, the new table must be imported into Terraform.
+
+First, remove the old table by peforming
```
markdboyd:
```suggestion
First, remove the old table from the Terraform state:
```

docs/features/backup_and_restore.md
```diff
@@ -0,0 +1,123 @@
...
+
+```sh
+terraform state rm module.data_persistence.aws_dynamodb_table.collections_table
+```
+replacing `collections_table` with the table identifier in the [DynamoDB Terraform table definitions](https://github.com/nasa/cumulus/blob/master/tf-modules/data-persistence/dynamo.tf).
+
+Then import the new table into Terraform using
```
markdboyd:
```suggestion
Then import the new table into the Terraform state:
```

docs/features/backup_and_restore.md
markdboyd: Also, do you have to restore a new table name? If so, do we need the note about updating Terraform to use the new table?

packages/api/tests/cli/test-back-restore.js
```diff
@@ -61,27 +61,10 @@ test.serial('backup records from DynamoDB', async (t) => {
...
-
-  // create a backup file with 200 records
-  let fileContent = '';
-  for (let i = 0; i < limit; i += 1) {
-    const granule = fakeGranuleFactory();
-    fileContent += `${JSON.stringify(granule)}\n`;
-    granuleIds.push(granule.granuleId);
-  }
-  fs.writeFileSync(tempRestoreFile, fileContent);
+  await restore(tempBackupFile, tableName, 1);
```
markdboyd: So this test is restoring to the same table that the data was initially populated in? Doesn't that mean it is possible that even if the restoration fails, the test would still pass from the initially populated data? Should the restore be to a new table since that also simulates the real world behavior?

tf-modules/data-persistence/dynamo.tf
markdboyd: Hmm, that's difficult. Ordinarily, I would think that we *do want* the old table to be deleted if you have changed the table name, for example if we update the default table names in the future. But that's an edge case that may never really happen. And as you say, I don't see how we can support the backup to a new table name without this change

packages/api/bin/backup.js
```diff
@@ -33,7 +33,7 @@ async function backup(table, region = 'us-east-1', folder = 'backups') {
   const dyno = Dyno(dynoParams);
   const stringify = new stream.Transform({ objectMode: true });
   stringify._transform = function _transform(record, enc, callback) {
-    const line = Dyno.serialize(record);
+    const line = JSON.stringify(record);
```
markdboyd: serialization and JSON.stringify are very different I presume, so I assume testing demonstrated that it wasn't working with `Dyno.serialize`?

laurenfrederick: It was not and did work with the stringify

tf-modules/data-persistence/dynamo.tf
laurenfrederick: If we updated table names and then deleted everyone's table and created a new one with the new table name our users would have a problem. Really I don't think we ever want to be regenerating the table since it would have data in it. We'd have to have a complex migration scenario.

markdboyd: Yeah, you're right. Makes sense. Definitely a good change then 👍

docs/features/backup_and_restore.md
laurenfrederick: Yeah I fixed it. And no you do not need a new table. I made updates to the docs to specify.

docs/features/data_in_dynamodb.md
laurenfrederick: Good catch. Thank you!

packages/api/tests/cli/test-back-restore.js
laurenfrederick: I updated that test and added a new one. Very good point.

docs/features/backup_and_restore.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+the backup will be stored at `backups/<table-name>.json`
+
+#### Restore with the CLI
+
+To restore data from a json file run the following command:
+
+```bash
+cumulus-api restore backups/<table-name>.json --table <new-table-name>
```
markdboyd: So you said this doesn't need to be a new table right? If so, it should just say `<table-name>`

PR #1473, opened 2020-02-11
CUMULUS-1717: Update granule reporting to remove race condition

markdboyd:

Addresses [CUMULUS-1717: Fix reporting race conditions in granule reporting](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1717)

## Changes

* Added new granule model methods to handle creating/updating granule records from Cumulus messages
* Added unit tests / int tests
* Removed granule functionality from publish-reports Lambda
* Added publish-granules Lambda which publishes granule record messages to SNS based on DynamoDB event stream
* Updated interfaces diagram for docs

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


tf-modules/archive/ingest-reporting.tf
```diff
@@ -122,32 +122,29 @@ resource "aws_lambda_event_source_mapping" "publish_executions" {
 
 # Report granules
 
-resource "aws_iam_role" "report_granules_lambda_role" {
+resource "aws_iam_role" "publish_granules_lambda_role" {
   name                 = "${var.prefix}-ReportGranulesLambda"
```
nemreid:
```suggestion
  name                 = "${var.prefix}-PublishGranulesLambda"
```

packages/api/lambdas/publish-reports.js
```diff
@@ -135,10 +47,7 @@ async function handlePdrMessage(eventMessage) {
  * @returns {Promise}
  */
 function publishReportSnsMessages(eventMessage) {
```
yjpa7145: Can we get rid of the `publishReportSnsMessages` function and just have the handler call `handlePdrMessage` directly?

packages/api/models/granules.js
```diff
@@ -471,6 +478,111 @@ class Granule extends Manager {
...
       super.delete({ granuleId: granule.granuleId })));
   }
+
+  /**
+   * Get the set of fields which are mutable based on the granule status.
+   *
+   * @param {Object} record - A granule record
+   * @returns {Array} - The array of mutable field names
+   */
+  _getMutableFieldNames(record) {
```
yjpa7145: This is another candidate for a class method that could really just be a stand-alone function.

packages/api/models/granules.js
```diff
@@ -471,6 +478,111 @@ class Granule extends Manager {
...
+    if (record.status === 'running') {
+      return ['createdAt', 'updatedAt', 'timestamp', 'status', 'execution'];
+    }
+    return Object.keys(record);
```
yjpa7145: Does this mean that, for a record with a "completed" status, we can overwrite `createdAt`? Also, does this mean that `status` can be overwritten for by a "running" record?

yjpa7145:
This isn't a criticism of your PR, just a general observation that it triggered. I have no idea what these API Model classes are supposed to be. Over the course of the last couple years, we seem to have just piled stuff into them. I feel like, for any definition of "this is what a model class represents", we'd be able to look at half of them and say "well, this one doesn't meet that."

packages/api/tests/lambdas/test-cw-sf-event-to-db-records.js
```diff
@@ -0,0 +1,202 @@
...
+    await saveGranulesToDb(cumulusMessage);
+    t.pass();
+  } catch (err) {
+    t.fail(`Exception should not have been thrown, but caught: ${err}`);
+  } finally {
+    storeGranuleStub.restore();
+  }
+});
+
+test('The cw-sf-event-to-db-records Lambda function creates execution, granule, and PDR records', async (t) => {
```
yjpa7145: It only creates execution and granule records, not PDRs.

packages/api/tests/lambdas/test-cw-sf-execution-event-to-db.js
```diff
@@ -1,121 +0,0 @@
-'use strict';
```
yjpa7145: I'm not sure it can be undone now (at least my git-fu isn't that strong), but it's easier to track changes in a case like this if you do a `git mv test-cw-sf-execution-event-to-db.js test-cw-sf-event-to-db-records.js` instead of deleting one file and creating the other. That way you still get the git history instead of it looking like all of `event-to-db-records` was created in this PR.

yjpa7145: Exactly like was done for `packages/api/tests/lambdas/publish-reports/test-getGranuleRecordsFromCumulusMessage.js → packages/api/tests/models/granules/test-getGranuleRecordsFromCumulusMessage.js`. Eh, stuff happens.

packages/api/lambdas/publish-reports.js
markdboyd: I didn't bother because this whole thing is going to get deleted in Menno's PR, but I can change it now if you prefer

packages/api/models/granules.js
markdboyd: This felt like a class method to me because it is specific to behavior of storing granule records, which seems like a reasonable if loose definition for this model

packages/api/tests/lambdas/test-cw-sf-event-to-db-records.js
markdboyd: Yeah, got ahead of myself since it will also create PDRs. I can update

packages/api/tests/lambdas/test-cw-sf-execution-event-to-db.js
markdboyd: Yeah, I should have done that but not sure if it's easy to change now. I can check

packages/api/models/granules.js
markdboyd: For a `completed` or `failed` record, you can overwrite anything, which this test is meant to show:

https://github.com/nasa/cumulus/blob/CUMULUS-1717-update-granule-reporting/packages/api/tests/models/granules/test-granule-reporting.js#L63

For a `running` record, you can update the `status` **but only when the execution has changed**:

https://github.com/nasa/cumulus/blob/CUMULUS-1717-update-granule-reporting/packages/api/models/granules.js#L562

This allows the scenario where a granule is being re-ingested or the execution re-run from the AWS console and we want the granule to go from "completed" to "running", but the execution should also have changed.

packages/api/lambdas/publish-reports.js
yjpa7145: Don't bother.

packages/api/tests/lambdas/test-cw-sf-execution-event-to-db.js
yjpa7145: Since it's only a test file, don't spend too much time trying to change it.

packages/api/models/granules.js
yjpa7145: So if a "completed" event is processed first, then a "running" event is processed for the same execution, the "running" event can make no updates to the record.

If a record is in the "completed" state, then a "running" event is received from a different execution, the fields listed on line 490 can be updated. In that case, do we want to be updating the "createdAt" field?

markdboyd: > So if a "completed" event is processed first, then a "running" event is processed for the same execution, the "running" event can make no updates to the record.

Correct

> If a record is in the "completed" state, then a "running" event is received from a different execution, the fields listed on line 490 can be updated.

Correct. I'm not sure about `createdAt` in that case. It is normally set from the workflow_start_time, and that value would be different from the new execution. But I guess we may not want to change it so that the time that the record was originally `createdAt` is preserved. So - maybe not?

PR #1475, opened 2020-02-12
Add S3 DiscoverGranules integration test

yjpa7145:


CHANGELOG.md
```diff
@@ -25,8 +25,28 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
   - Change variable `saml_launchpad_metadata_path` to
     `saml_launchpad_metadata_url` in the `tf-modules/cumulus` Terraform module.
 
+- **CUMULUS-1446**
+  - `@cumulus/integration-tests/index.rulesList()` updated to only take
+    `rulesDirectory` parameter
+  - Remove unused `@cumulus/integration-tests/api.addCollectionApi()` function
```
markdboyd: I think we need to deprecate all these functions, since they are publicly exported?

example/spec/parallel/discoverGranules/DiscoverGranulesS3FailureSpec.js
```diff
@@ -0,0 +1,96 @@
...
+  it('fails', () => {
+    if (!beforeAllCompleted) fail('beforeAll() failed');
+    else expect(workflowExecution.status).toEqual('FAILED');
+  });
+
+  it('can be fetched from the API', async () => {
+    if (!beforeAllCompleted) fail('beforeAll() failed');
+    else {
+      try {
+        await pWaitFor(
```
markdboyd: I realize that the function is somewhat different, but we already have a `waitForModelStatus` function in `example/spec/helpers/apiUtils.js` that we're using throughout our tests. Is there some reason to not use it here? I kind of like that pattern better than doing `expect().nothing()` below

example/spec/parallel/discoverGranules/DiscoverGranulesS3SuccessSpec.js
```diff
@@ -0,0 +1,156 @@
...
+
+  it('executes successfully', () => {
+    if (!beforeAllCompleted) fail('beforeAll() failed');
+    else expect(workflowExecution.status).toEqual('SUCCEEDED');
+  });
+
+  it('can be fetched from the API', async () => {
+    if (!beforeAllCompleted) fail('beforeAll() failed');
+    else {
+      try {
```
markdboyd: Same comment as above about `waitForModelStatus`

example/spec/parallel/discoverGranules/DiscoverGranulesS3SuccessSpec.js
```diff
@@ -0,0 +1,156 @@
...
+        );
+        expect().nothing();
+      } catch (error) {
+        fail();
+      }
+    }
+  });
+
+  it('results in a successful IngestGranule workflow execution', async () => {
+    if (!beforeAllCompleted) fail('beforeAll() failed');
```
markdboyd: I really dislike Jasmine for this

example/spec/parallel/discoverGranules/DiscoverGranulesS3FailureSpec.js
```diff
@@ -0,0 +1,96 @@
...
+  beforeAll(async () => {
+    ({ stackName, bucket } = await loadConfig());
+
+    const testId = randomString();
+
+    // Create the provider
+    provider = await loadProvider({
+      filename: './data/providers/s3/s3_provider.json',
+      postfix: testId,
+      s3Host: randomString()
```
markdboyd: So this is what causes the workflow failure? The use of a non-existent bucket?

example/spec/parallel/kinesisTests/KinesisReplaySpec.js
```diff
@@ -62,7 +62,7 @@ describe('The Kinesis Replay API', () => {
 
   async function cleanUp() {
     // delete rules
-    const rulesToDelete = await rulesList(testConfig.stackName, testConfig.bucket, ruleDir);
+    const rulesToDelete = await readJsonFilesFromDir(ruleDir);
```
markdboyd: So you're reading the definitions from the local directory instead of the bucket? Does that introduce a possible problem if not everything from the local directory got synced up to the bucket?

packages/integration-tests/api/executions.js
```diff
@@ -10,16 +10,15 @@ const { callCumulusApi } = require('./api');
  * @param {string} params.arn - an execution arn
  * @returns {Promise<Object>} - the execution fetched by the API
  */
-async function getExecution({ prefix, arn }) {
-  return callCumulusApi({
+const getExecution = ({ prefix, arn }) =>
```
markdboyd: I realize that the function doesn't need to be marked as `async`, but is there some reason to refer the `const` style definition for functions?

packages/integration-tests/index.js
```diff
@@ -253,51 +264,126 @@ function setProcessEnvironment(stackName, bucketName) {
   process.env.stackName = stackName;
   process.env.messageConsumer = `${stackName}-messageConsumer`;
   process.env.KinesisInboundEventLogger = `${stackName}-KinesisInboundEventLogger`;
-  process.env.CollectionsTable = `${stackName}-CollectionsTable`;
```
markdboyd: Is the idea that individual tests will set this up as needed? And that it unnecessarily couples table names to the stack names?

packages/integration-tests/index.js
```diff
@@ -253,51 +264,126 @@ function setProcessEnvironment(stackName, bucketName) {
...
+const addCustomUrlPathToCollectionFiles = (collection, customFilePath) =>
+  collection.files.map((file) => {
     let urlPath;
-    if (Object.is(file.url_path, undefined)) {
-      urlPath = '';
-      if (!Object.is(collection.url_path, undefined)) {
-        urlPath = `${collection.url_path}/`;
-      }
-    } else {
+    if (isString(file.url_path)) {
```
markdboyd: Yes this is much clearer

packages/integration-tests/index.js
```diff
@@ -253,51 +264,126 @@ function setProcessEnvironment(stackName, bucketName) {
...
+    };
   });
-}
+
+/**
+ * Update a collection with a custom file path, duplicate handling, and name and
+ * dataType updated with the postfix.
+ *
+ * @param {Object} params
+ * @param {Object} params.collection
```
markdboyd: Need descriptions for these params

packages/integration-tests/index.js
```diff
@@ -308,41 +394,26 @@ function addCustomUrlPathToCollectionFiles(collection, customFilePath) {
...
- * @param {string} dataDirectory - the directory of collection json files
- * @returns {Promise.<Array>} list of collections
- */
-async function listCollections(stackName, bucketName, dataDirectory) {
-  return setupSeedData(stackName, bucketName, dataDirectory);
+  await Promise.all(
+    collections.map((collection) => addCollection(stackName, collection))
+  );
+
+  return rawCollections.length;
```
markdboyd: If any of the `addCollection` calls in the above `Promise.all` failed, then you would only have added some amount of collections which is less than `rawCollections.length`?

packages/integration-tests/index.js
```diff
@@ -355,20 +426,16 @@ async function listCollections(stackName, bucketName, dataDirectory) {
...
+    collections.map(
+      ({ name, version }) => {
+        const realName = postfix ? `${name}${postfix}` : name;
+        return collectionsApi.deleteCollection(stackName, realName, version);
+      }
+    )
+  );
 
-  return Promise.all(promises).then((cs) => cs.length);
+  return collections.length;
```
markdboyd: Same comment as above regarding the length of the result

packages/integration-tests/index.js
```diff
@@ -449,54 +553,39 @@ async function addProviders(stackName, bucketName, dataDirectory, s3Host, postfi
...
-  return Promise.all(promises).then((ps) => ps.length);
+  return providers.length;
 }
 
 /**
  * Delete all collections listed from a collections directory
  *
  * @param {string} stackName - CloudFormation stack name
- * @param {string} bucket - S3 internal bucket name
+ * @param {string} _bucket - S3 internal bucket name
```
markdboyd: Why the underscore?

packages/integration-tests/index.js
```diff
@@ -536,7 +625,7 @@ async function addRulesWithPostfix(config, dataDirectory, overrides, postfix) {
         ...config
       }));
 
-      const r = new Rule();
+      const r = new RulesModel();
```
markdboyd: If you're going to bother changing this line, I'd go all the way and change the single letter variable names

packages/integration-tests/index.js
```diff
@@ -653,16 +731,26 @@ async function buildWorkflow(
...
-    ? await new Collection().get({ name, version })
-    : {};
-  const providerInfo = provider
-    ? await new Provider().get({ id: provider.id })
-    : {};
-
-  template.meta.collection = collectionInfo;
-  template.meta.provider = providerInfo;
+
+  if (collection) {
```
markdboyd: It looks like the previous code didn't handle this case either: but what if `collection` is an object with a `name` but not a `version`? Then the API request to get the collection will fail and `template.meta.collection` won't have the expected value. Should we check for both properties?

example/spec/parallel/discoverGranules/DiscoverGranulesS3FailureSpec.js
yjpa7145: Correct. I asked Chuck about it (since he wrote the ticket), and he was happy with that for the failure test.

example/spec/parallel/discoverGranules/DiscoverGranulesS3SuccessSpec.js
yjpa7145: Agreed.

packages/integration-tests/index.js
yjpa7145: If any of them fail, it's going to throw an exception and never return a length, which was the case before this change as well.

packages/integration-tests/index.js
yjpa7145: See https://github.com/nasa/cumulus/pull/1475/files#r379627630

packages/integration-tests/index.js
yjpa7145: The bucket variable is never used. The `no-unused-vars` eslint rule complains about that but, in our eslint config, (we allow overriding that for variables that begin with an underscore)[https://github.com/nasa/cumulus/blob/master/.eslintrc.json#L71].

In this case the better solution would be to remove that variable entirely, but this function is called all over the place and, to keep the PR from exploding, I chose to just not update all of them.

example/spec/parallel/kinesisTests/KinesisReplaySpec.js
yjpa7145: They were never being read from the bucket. On master, `rulesList` is calling `setupSeedData` [here](https://github.com/nasa/cumulus/blob/master/packages/integration-tests/index.js#L607). `setupSeedData` was just loading the rules from files [here](https://github.com/nasa/cumulus/blob/master/packages/integration-tests/index.js#L272). The bucket was only being used because `setupSeedData` was passing it to [`setProcessEnvironment`](https://github.com/nasa/cumulus/blob/master/packages/integration-tests/index.js#L251), which was using it to set the `process.env.system_bucket` environment variable, which had no impact on actually returning the list of rules.

example/spec/parallel/discoverGranules/DiscoverGranulesS3FailureSpec.js
yjpa7145: To improve on `expect().nothing()`, I refactored this test to use `await expectAsync(...).toBeResolved();` instead.

As for using `waitForModelStatus`, my preference is to use the API to get the status of the execution (like I'm doing here) rather than query the DB directly. My understanding is that, as much as possible, the integration tests should be testing Cumulus from the perspective of an outside user. By using the API to get the status, we're using Cumulus's public interface rather than our behind-the-scenes knowledge of how Cumulus works.

yjpa7145: Note: that refactor hasn't been pushed yet.

example/spec/parallel/discoverGranules/DiscoverGranulesS3SuccessSpec.js
yjpa7145: https://github.com/nasa/cumulus/pull/1475/files#r380949806

packages/integration-tests/api/executions.js
yjpa7145: In this case it's personal preference because it allows for the more compact "arrow function without an explicit return" syntax.

packages/integration-tests/index.js
yjpa7145: The `setProcessEnvironment` function is only called from inside of the `integration-tests/index.js` file, so no test files were invoking it. After taking a closer look at it, the entire function can actually be deleted. I'll fix that.

packages/integration-tests/index.js
yjpa7145: Will fix

packages/integration-tests/index.js
yjpa7145: Will do.

packages/integration-tests/index.js
yjpa7145: A collection with a name but without a version is an invalid collection and would only be caused by a broken test. I'm fine leaving this the way it is and assuming that our integration tests are passing a valid collection config here.

packages/integration-tests/index.js
```diff
@@ -616,14 +763,19 @@ async function rulesList(stackName, bucketName, rulesDirectory) {
...
   setProcessEnvironment(stackName, bucketName);
-  const promises = rules.map((rule) => {
-    if (postfix) {
-      rule.name += postfix;
-    }
-    return limit(() => _deleteOneRule(rule.name));
-  });
-  return Promise.all(promises).then((rs) => rs.length);
+
+  process.env.RulesTable = `${stackName}-RulesTable`;
```
markdboyd: It seems like this is already set by `setProcessEnvironment`?

packages/integration-tests/index.js
```diff
@@ -616,14 +763,19 @@ async function rulesList(stackName, bucketName, rulesDirectory) {
  * @returns {Promise.<number>} - Number of rules deleted
  */
 async function deleteRules(stackName, bucketName, rules, postfix) {
+  // setProcessEnvironment is not needed by this function, but other code
```
markdboyd: Is this correct? `setProcessEnvironment` is setting the env var for the rules table name, which is needed by this function

yjpa7145: `RulesTable` is the only env var that is required by this function so, in an ideal world, only that one would be set. Because there is other code (mostly our spec tests) that are implicitly depending on `deleteRules` and the other functions in here to set those other environment variables, `setProcessEnvironment` must be run.

Until it can be refactored, I'd feel better leaving this the way it is.

packages/integration-tests/index.js
yjpa7145: See https://github.com/nasa/cumulus/pull/1475/files#r382790636

PR #1477, opened 2020-02-12
Cumulus 1684 fix ems

laurenfrederick:
**Summary:** Fix issue with EMS reports

## Changes

* SFTP changed and was accepting the key, not the file name in S3. Pull the key from S3 and pass that to sftp.

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


packages/api/lib/ems.js
```diff
@@ -101,6 +103,22 @@ async function getExpiredS3Objects(bucket, prefix, retentionInDays) {
     .map((s3Object) => ({ Bucket: bucket, Key: s3Object.Key }));
 }
 
+async function retrievePrivateKey() {
+  const privateKeyFile = process.env.ems_privateKey || 'ems-private.pem';
+
+  const keyExists = await s3ObjectExists(
```
yjpa7145: Rather than have an explicit check to see if the key exists, can you just call `getS3Object`, catch an exception if it's thrown, and throw an "object does not exist" error? That would save an API call that isn't really needed.

packages/api/lib/ems.js
```diff
@@ -101,6 +103,22 @@ async function getExpiredS3Objects(bucket, prefix, retentionInDays) {
...
+    { Bucket: process.env.system_bucket, Key: `${process.env.stackName}/crypto/${privateKeyFile}` }
+  );
+  if (!keyExists) {
+    return Promise.reject(new Error(`${privateKeyFile} does not exist in S3 crypto directory`));
```
yjpa7145:
```suggestion
    throw new Error(`${privateKeyFile} does not exist in S3 crypto directory`);
```

packages/api/lib/ems.js
```diff
@@ -101,6 +103,22 @@ async function getExpiredS3Objects(bucket, prefix, retentionInDays) {
...
+  }
+
+  log.debug(`Reading Key: ${privateKeyFile} bucket:${process.env.system_bucket}, stack:${process.env.stackName}`);
```
yjpa7145: Do we need this debug statement? It's worth including the bucket and key in the exception that's thrown when the key is not found but, if it is found, this just clutters up the logs.

packages/api/lib/ems.js
```diff
@@ -101,6 +103,22 @@ async function getExpiredS3Objects(bucket, prefix, retentionInDays) {
...
+  const privateKey = await getS3Object(process.env.system_bucket, `${process.env.stackName}/crypto/${privateKeyFile}`);
```
yjpa7145: You can use the [getTextObject()](https://github.com/nasa/cumulus/blob/master/packages/aws-client/S3.js#L281) function here:

```js
return getTextObject(
  process.env.system_bucket,
  `${process.env.stackName}/crypto/${privateKeyFile}`
);
```

packages/api/lib/ems.js
laurenfrederick: Ok changed it.

packages/api/lib/ems.js
laurenfrederick: Removed. Idk I took all of what was in there before.

packages/api/lib/ems.js
laurenfrederick: Cool I didn't know that existed. Thanks!

packages/api/lib/ems.js
```diff
@@ -101,6 +102,22 @@ async function getExpiredS3Objects(bucket, prefix, retentionInDays) {
     .map((s3Object) => ({ Bucket: bucket, Key: s3Object.Key }));
 }
 
+async function retrievePrivateKey() {
+  const privateKeyFile = process.env.ems_privateKey || 'ems-private.pem';
+
+  let privateKey;
+  try {
+    privateKey = await getTextObject(process.env.system_bucket, `${process.env.stackName}/crypto/${privateKeyFile}`);
```
yjpa7145: This line could just be `return await getTextObject(...)`, which would simplify it a bit.

yjpa7145: Not worth holding up the approval over, it's a small thing. Or, as Menno would say, something about 🐜 .

packages/api/lib/ems.js
```diff
@@ -101,6 +102,22 @@ async function getExpiredS3Objects(bucket, prefix, retentionInDays) {
...
+  } catch (e) {
+    if (e.code === 'NoSuchKey') {
+      throw new Error(`${privateKeyFile} does not exist in S3 crypto directory`);
```
yjpa7145: It might be worth including the `s3://bucket/key` here, so it's clear what file didn't exist.

packages/api/lib/ems.js
laurenfrederick: I tried `return getTextObject` and it wouldn't catch the error. Did it need an `await` in there?

yjpa7145: Yeah, the `await` is what makes the difference.

PR #1478, opened 2020-02-12
Cumulus 1713 Sending Collection info to an SNS message

kkelly51:

Addresses [CUMULUS-1713: sending Collection info to SNS Topic](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1713)

## Changes

* Creates an SNS Topic for collections
* Sends a record with { collectionRecord, event: 'Create' } when collection is created or updated
* Sends a record with { event: 'Delete', {name, version} } when collection is deleted


packages/api/models/collections.js
```diff
@@ -141,6 +169,15 @@ class Collection extends Manager {
       ? item : await this.get(item).catch(() => item);
     await this.collectionConfigStore.delete(dataType || name, version);
 
+    const record = {
+      event: 'Delete',
+      record: {
+        name,
+        version
```
jennyhliu: Can you add `deletedAt`?

kkelly51: Yeah, good idea

kkelly51:
> Shouldn't we post to the SNS topic on update too?

PR #1480, opened 2020-02-13
Update the s3-replicator readme to more closely reflect which variables are valid and necessary

ifestus:
## Changes

* Remove `aws_profile` and `aws_region` from `s3-replicator/README.md`
* Move `security_groups` to an `optional` area in the example configuration.


tf-modules/s3-replicator/README.md
```diff
@@ -27,14 +27,14 @@ Configuration variables are shown in `terraform.tfvars.sample`, and are explaine
...
 vpc_id               = "vpc-123456"                                       # VPC ID (for Lambda)
 subnet_ids           = ["subnet-123456"]                                  # Subnet IDs (for Lambda)
-security_groups      = ["sg-123456"]                                      # Security Group IDs (for Lambda)
 source_bucket        = "source-bucket-name"                               # Source S3 bucket
 source_prefix        = "source-prefix"                                    # Source object prefix e.g. 'path/to/filedir'
 target_bucket        = "target-bucket-name"                               # Target S3 bucket e.g. 'mybucket'
 target_prefix        = "target-prefix"                                    # Target object prefix path
+
+# Optional
+security_groups      = ["sg-123456"]                                      # Security Group IDs (for Lambda)
```
markdboyd: It seems like the variable name is actually `security_group_ids` in `variables.tf`?

markdboyd: Also, can we move `subnet_ids` down to the optional section, since it also has a default value? Same with `vpc_id` I guess

ifestus: good catch 😄 
e90897b

markdboyd: Can we move subnet_ids down to the optional section, since it also has a default value? Same with vpc_id I guess

ifestus: Good point. I hadn't moved them down because I'm not really sure how that affects the deployment in practice, especially in an NGAP environment. Looking more closely, it appears that those truly are optional and I'll move them.

ifestus: 54b9897

ifestus: OOPS 7615eba

PR #1483, opened 2020-02-14
Allow customization of AWS tags in Terraform deployments

yjpa7145:


CHANGELOG.md
```diff
@@ -8,6 +8,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
 ### BREAKING CHANGES
 
+- **CUMULUS-1672**
+  - The `cumulus` Terraform module in previous releases set a
+    `Deployment = var.prefix` tag on all resources that it managed. In this
+    release, a `tags` input variable has been added to the `cumulus` Terraform
+    module to allow resource tagging to be customized. No default tags will be
+    applied to Cumulus-managed resources. To replicate the previous behavior,
+    set `tags = { Deployment: var.prefix }` as an input variable for the
+    `cumulus` Terraform module.
```
chuckwondo: Can we avoid this breaking change by setting a default value that matches the previous behavior? If so, I would suggest doing so, not only to avoid a breaking change, but also to avoid having users be unaware that there is no tagging by default, and then deploy everything with _no_ tags at all.

yjpa7145: We talked it over in the standup and the consensus was that, as long as it was documented, this was an acceptable change.

docs/data-cookbooks/run-tasks-in-lambda-or-docker.md
```diff
@@ -90,7 +90,6 @@ Given it has been discovered this task can no longer run in AWS Lambda, you can
 ```hcl
 resource "aws_sfn_activity" "queue_granules" {
   name = "${var.prefix}-QueueGranules"
-  tags = local.default_tags
```
markdboyd: I assume we aren't including these in the docs since `local.default_tags` is gone now. Is there value in showing use of `var.tags` so people's resources would be tagged?

docs/workflows/lambda.md
```diff
@@ -46,8 +44,6 @@ resource "aws_lambda_function" "myfunction" {
 
 Make sure to include a `vpc_config` that matches the information you've provided the `cumulus` module if intending to integrate the lambda with a Cumulus deployment.
 
-Also note that for this example to work, you will need to have `default_tags` defined as in the [example](https://github.com/nasa/cumulus/blob/master/example/cumulus-tf/main.tf), or in the [template-deploy-repo](https://github.com/nasa/cumulus-template-deploy/blob/master/cumulus-tf/main.tf).
```
markdboyd: You should also delete the note below

CHANGELOG.md
markdboyd: Should we have a corresponding PR to the template-deploy repo updating `{ Deployment: var.prefix }` to be the default value for the `tags` variable?

docs/data-cookbooks/run-tasks-in-lambda-or-docker.md
yjpa7145: I'd say not in this case, since it isn't really related to what this doc is trying to show.

docs/workflows/lambda.md
yjpa7145: Fixed in a4239a08e

PR #1484, opened 2020-02-17
Modify localAPI testAuth.js to more closely follow auth.js

flamingbear:

Addresses [CUMULUS-1749: Datepicker filter on Dashboard](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1749)

## Changes

* Updates testAuth to validate presented JWT.

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


packages/api/app/testAuth.js
```diff
@@ -93,17 +100,34 @@ async function ensureAuthorized(req, res, next) {
...
+  }
 
   if (!jwtToken) {
     return res.boom.unauthorized('Missing token');
   }
 
-  if (jwtToken === jwt) {
+  let userName;
+  try {
+    ({ username: userName, accessToken } = verifyJwtToken(jwtToken));
```
markdboyd: We're not doing any checks to ensure that the `username` or `accessToken` match the values that we set up for this test auth? Maybe we should? Otherwise possibly returning a "User not authorized" error is a bit confusing

flamingbear: Actually I could test for auth.  before I was setting the auth user on the dashboard, but took that out because we weren't testing it anywhere.  But don't really have strong feelings.

markdboyd: Yeah, I'm saying that we could actually verify that `userName === username` to ensure that the username unpacked from the JWT matches the one it should have been created with, so more closely mimicking real auth.  We'd have to rename `accessToken` that we're destructuring here so it doesn't conflict with `accessToken` in the global scope if we wanted to compare them. 

It's definitely not necessary for test auth, the only reason I advocate it is that, as currently written, there's no way this test auth should ever produce a "User not authorized" error because it's never checking the username. So it's a bit confusing for it to possibly return that error for a request.

flamingbear: Well, that's the default error if it falls through all other errors. even in the real auth.

flamingbear: also the inconsistent use of `username` and `userName` is a total PITA

markdboyd: Yeah, I guess the way it works in real auth doesn't really make sense either, because it immediately returns that error if the user is wrong:

https://github.com/nasa/cumulus/blob/master/packages/api/app/auth.js#L60

But then it's also treated as the default error message, even though it wouldn't be possible that the error is "User not authorized" at that point

https://github.com/nasa/cumulus/blob/master/packages/api/app/auth.js#L82

Maybe for test auth we just replace `res.boom.unauthorized` below with `res.boom.badImplementation`, which should return a general 500 error?

flamingbear: really it might should be in both auth and testAuth.  https://github.com/nasa/cumulus/blob/master/packages/api/app/auth.js#L82

markdboyd: OK. I'll approve and leave it up to you whether to change it for the test auth

PR #1485, opened 2020-02-19
CUMULUS-1748: support relative paths in recursion.js

nemreid:

Addresses [CUMULUS-1748: relative paths in recursion](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1748)

## Changes

* Support for relative paths
* Also fix CUMULUS-1750

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


packages/ingest/recursion.js
```diff
@@ -40,7 +28,7 @@ async function recurOnDirectory(fn, currentPath, segments, position) {
       if (['-', 0].includes(item.type)) {
         files.push(item);
       } else if (['d', 1].includes(item.type)) {
-        const nextDir = (currentPath === '' ? item.name : `${currentPath}/${item.name}`);
+        const nextDir = (currentPath === '/' ? `/${item.name}` : `${currentPath}/${item.name}`);
```
nemreid: If somebody knows a more elegant way of doing this I am all 👂 👂

packages/ingest/util.js
```diff
@@ -2,21 +2,6 @@
...
 const mime = require('mime-types');
 
-/**
- * Ensure provider path conforms to expectations.
- * Removes any/all leading forward slashes.
- *
- * @param {string} provPath - provider path
- * @returns {string} path, updated to conform if necessary.
- */
-function normalizeProviderPath(provPath) {
```
nemreid: This function was a bad idea way back when. Turns all relative paths absolute. Boooo. Who committed that?? ... 😰

packages/ingest/test/test-FtpProviderClientFailure.js
```diff
@@ -0,0 +1,45 @@
...
+  executeRawFtpCommand('chmod 400 -R forbidden/file.txt');
+});
+
+test.after.always(async () => {
+  // await when fixed
+  executeRawFtpCommand('chmod 644 -R forbidden/file.txt');
+});
+
+test.skip('FtpProviderClient throws an error when listing a non-permitted directory', async (t) => {
+  // TODO: update cumuluss/vsftpd to allow `chmod` to work in the setup for this test.
```
nemreid: Confirmed that this works in ad-hoc testing, at least.

packages/ingest/FtpProviderClient.js
```diff
@@ -51,13 +52,31 @@ class FtpProviderClient {
...
 
   async buildFtpClient() {
-    return new JSFtp({
-      host: this.host,
-      port: get(this.providerConfig, 'port', 21),
-      user: await this.getUsername(),
-      pass: await this.getPassword(),
-      useList: get(this.providerConfig, 'useList', false)
-    });
+    if (isNil(this.ftpClient)) {
```
markdboyd: Is `buildFtpClient` somehow called multiple times on the same class instance now?

packages/ingest/FtpProviderClient.js
```diff
@@ -51,13 +52,31 @@ class FtpProviderClient {
...
+        port: get(this.providerConfig, 'port', 21),
+        user: await this.getUsername(),
+        pass: await this.getPassword(),
+        useList: get(this.providerConfig, 'useList', false)
+      });
+    }
+    return this.ftpClient;
+  }
+
+  errorHandler(rejectFn, e) {
```
markdboyd: I would prefer a better variable name here than `e` 🐜

packages/ingest/FtpProviderClient.js
```diff
@@ -51,13 +52,31 @@ class FtpProviderClient {
...
+    let err = e;
```
markdboyd: Maybe change `err` to `formattedError` 🐜

packages/ingest/FtpProviderClient.js
```diff
@@ -74,38 +93,44 @@ class FtpProviderClient {
     const client = await this.buildFtpClient();
 
     return new Promise((resolve, reject) => {
-      client.on('error', reject);
+      client.on('error', this.errorHandler.bind(this, reject));
```
markdboyd: Man I find the `.bind` syntax confusing. So the `reject` argument we're passing will *precede* the argument that would get passed in by default to the callback here, which is an error. 👌

packages/ingest/FtpProviderClient.js
```diff
@@ -74,38 +93,44 @@ class FtpProviderClient {
...
+          const message = err.message || err.text;
+          if (message && message.includes('Timed out') && counter < 3) {
             log.error(`Connection timed out while listing ${path}. Retrying...`);
             counter += 1;
             return this._list(path, counter).then((r) => {
-              log.info(`${counter} retry suceeded`);
+              log.info(`${counter} retry succeeded`);
               return resolve(r);
-            }).catch((e) => reject(e));
+            }).catch(this.errorHandler.bind(this, reject));
```
markdboyd: So the net improvement of moving to the new error handling is improved error messages and ensuring that the client gets destroyed before rejecting?

packages/ingest/recursion.js
markdboyd: So this avoids there being a double `//` between the `currentPath` and `item.name` if the `currentPath` **is not** `/`?

packages/ingest/recursion.js
```diff
@@ -54,28 +42,37 @@ async function recurOnDirectory(fn, currentPath, segments, position) {
...
+async function recursion(fn, configuredPath) {
+  const normalizedPath = path.normalize(configuredPath);
+  const isAbsolutePath = path.isAbsolute(normalizedPath);
   try {
-    const segments = normalizedPath.split('/'); // split on divider
-    return await recurOnDirectory(fn, segments[0], segments, 0);
+    const segments = normalizedPath
+      .split('/') // split on divider
+      .filter((segment) => segment.trim() !== ''); // filter out empty strings from split
+    const startingPath = isAbsolutePath ? '/' : '.';
```
markdboyd: I'm not sure that I understand this behavior and/or how it differs from the previous behavior. If I pass in `/path/to/dir` as my `configuredPath`, then why do we start by trying to list files in `/`? In other words, why would we start in parent directories of the directory that I've specified?

packages/ingest/test/test-FtpProviderClient.js
```diff
@@ -132,7 +133,7 @@ test('useList is present and true when assigned', async (t) => {
 
   await myFtpProviderClient.list('');
 
-  t.true(jsftpSpy.callCount > 1);
+  t.true(jsftpSpy.callCount > 0);
```
markdboyd: I tried grokking the code, but why did the expected call count drop?

packages/ingest/test/test-FtpProviderClientFailure.js
```diff
@@ -0,0 +1,45 @@
...
+    jsftp.raw(cmd, (err, data) => {
+      if (err) reject(err);
+      resolve(data);
+    });
+  });
+};
+
+test.before(async () => {
+  // await when fixed
+  executeRawFtpCommand('chmod 400 -R forbidden/file.txt');
```
markdboyd: I fear that the answer is obvious, but why aren't we `await`ing?

packages/ingest/test/test-FtpProviderClientFailure.js
```diff
@@ -0,0 +1,45 @@
...
+});
+
+test.after.always(async () => {
+  // await when fixed
+  executeRawFtpCommand('chmod 644 -R forbidden/file.txt');
+});
+
+test.skip('FtpProviderClient throws an error when listing a non-permitted directory', async (t) => {
```
markdboyd: I assume you don't want to `skip` this test?

packages/ingest/test/test-FtpProviderClientFailure.js
```diff
@@ -0,0 +1,45 @@
...
+  // TODO: update cumuluss/vsftpd to allow `chmod` to work in the setup for this test.
+  const myFtpProviderClient = new FtpProviderClient({
+    host: '127.0.0.1',
+    username: 'testuser',
+    password: 'testpass',
+    useList: true
+  });
+
+  await t.throwsAsync(myFtpProviderClient.list('forbidden/file.txt'),
+    'FTP Code 451: Could not retrieve a file listing for forbidden/file.txt.'
```
markdboyd: Do we want to test against the full error message? You could test against a regex that ensures the error code and file name show up in the error message 🐜

packages/ingest/test/test-recursion.js
```diff
@@ -4,113 +4,175 @@ const test = require('ava');
...
-  t.deepEqual(files, dirs.path);
+  const files = await recursion(fn, '/path');
+  t.deepEqual(files, dirs['/path']);
 });
 
-test('recursion lists all files in a simple directory structure with regex path', async (t) => {
+test('recursion lists all files in a simple relative regex path', async (t) => {
   const dirs = {
-    '': [
+    '.': [
```
markdboyd: Might be a bad idea, but do we need/want to include files at the starting directory in this test fixture to ensure that those **aren't** picked up as well?

packages/ingest/test/test-recursion.js
```diff
@@ -4,113 +4,175 @@ const test = require('ava');
...
 
-test('recursion lists all files in a complex directory structure with text path', async (t) => {
+test('recursion lists all files in a complex relative text path', async (t) => {
   const dirs = {
-    '': [
+    '.': [
       { type: 'd', name: 'path' }
     ],
-    path: [
+    './path': [
```
markdboyd: Same comment about including files in the parent directories in the test fixtures

packages/ingest/util.js
markdboyd: According to the "rules", we would need to deprecate this instead of just deleting it

packages/sftp-client/index.js
```diff
@@ -134,12 +134,10 @@ class SftpClient {
    * @returns {Promise<Array<Object>>} list of file objects
    */
   async list(remotePath) {
-    const actualPath = remotePath.trim() === '' ? '/' : remotePath;
```
markdboyd: So if the `removePath` was empty, start at `/`. Is this behavior preserved?

bamboo/bootstrap-unit-tests.sh
```diff
@@ -45,10 +45,11 @@ while ! $docker_command  'curl --connect-timeout 5 -sS -o /dev/null http://127.0
 done
 echo 'HTTP service is available'
 
+$docker_command "mkdir /keys;cp $UNIT_TEST_BUILD_DIR/packages/test-data/keys/ssh_client_rsa_key /keys/; chmod -R 400 /keys"
```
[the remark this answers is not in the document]
nemreid: It is both that and the fact that SFTP requires the key permissions to be `400`.
Changing the way we react to running into a directory we can't access seemed out of scope for this PR but would be required for everything not to bomb out on encountering a directory that does not allow reads.

packages/ingest/FtpProviderClient.js
nemreid: It is called in `_list`, which is the function that is passed to `recursion` and called repeatedly to recur through the file system. We were not doing any caching of our client in this client, even though the SftpProviderClient _does_ already work this way.

packages/ingest/FtpProviderClient.js
nemreid: `.bind` is pretty simple if you think of it as a function where:
1. arg #1 is what the bound function's context will have as `this`.
2. args #2-infinity onwards are just pre-set into the args the function accepts. So `errorHandler.bind(this, reject);` just returns a function that is functionally identical to `errorHandler` but with `rejectFn` pre-set to `reject`, and not effectively in the arguments list. Which is why it only takes one argument, the error, like you said, even though that is argument 2 on the original function.

This gets harder with parameter objects. 🙂

packages/ingest/FtpProviderClient.js
nemreid: Yeah, it's DRYing up a bunch of additional `client.destroy` calls, and ensuring we throw a standardized error out of the FtpProviderClient.

packages/ingest/recursion.js
nemreid: Because with the regex filtering, you could pass in `/path/to/dir` or you could pass in `/(p.*)/to/dir`. In the latter case you want to list everything in `/` and recur into everything that matches `(p.*)`.

In the previous implementation it kind of haphazardly tried to figure out which part is text and which part is regex and it was highly confusing and not maintainable, as well as broken. I realize there may be problems with permissions when trying to list parent directories so that's why the fallback _does_ start by listing your `configuredPath`.

packages/ingest/test/test-FtpProviderClient.js
nemreid: It actually went up. In my previous PR I changed this because the recursive calls caused the callCount to go from `1` to `13`, so I set `> 1`. Then I made a change here and the number is back to one, _I think_. It's been a minute since I changed that. It's a number that may or may not increase as the complexity of `test-data` increases, since we mount that as a volume on the ftp server, but in all cases should be > 0.

packages/ingest/test/test-FtpProviderClientFailure.js
nemreid: This was the only way to allow this file to satisfy both `ava` and `eslint`. If we `await` this command throws an error because our server's `vsftpd.conf` does not have `chmod_enabled=true`.

packages/ingest/test/test-FtpProviderClientFailure.js
nemreid: I do because we can't actually set the permissions on `forbidden` to set up this test correctly.

packages/ingest/test/test-FtpProviderClientFailure.js
nemreid: `throwsAsync` takes an error message to compare with, though. I guess I could catch it and test it manually.

packages/sftp-client/index.js
nemreid: This was a fix Marc put in after I merged my previous PR. It has been consolidated into `recursion.js` which uses `path.normalize` to ensure there are never empty paths passed down the chain.

packages/ingest/util.js
nemreid: I don't know who else would use a function that is specifically written for messing with provider paths, but I would rather break them than have them continue to use a function that "normalizes" by making all the paths absolute, which is simply a terrible idea.

packages/ingest/test/test-FtpProviderClientFailure.js
markdboyd: Yeah, looks like the AVA docs discuss that

```
const promise = Promise.reject(new TypeError('🦄'));

test('rejects', async t => {
	const error = await t.throwsAsync(promise);
	t.is(error.message, '🦄');
});
```

Definitely not necessary

packages/ingest/recursion.js
markdboyd: >  it was highly confusing and not maintainable, as well as broken

Well when you put it _that_ way...

packages/ingest/test/test-FtpProviderClientFailure.js
nemreid: 7e9d0ed26

packages/ingest/recursion.js
nemreid: No it is specifically for fixing the case of listing `/`. If we list `/`, find `item`, and recur into it, we want to list `/item`, not `//item`. Any other `currentPath` will be fine, as it will just be `<path>/item`.

packages/ingest/util.js
nemreid: But I can deprecate it and put some massive warning comments in.

packages/ingest/test/test-recursion.js
nemreid: f92ab5faa

packages/ingest/test/test-recursion.js
nemreid: f92ab5faa

packages/ingest/util.js
nemreid: 4a1835ded

packages/ingest/recursion.js
nemreid: A more elegant solution came to me: dcc4286a1

PR #1489, opened 2020-02-24
SIT deployments

nemreid:

## Changes

* Enable SIT deployments

## PR Checklist

- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


example/deployments/sit.tfvars
```diff
@@ -0,0 +1,49 @@
...
+s3_replicator_config = {
+  source_bucket = ""
+  source_prefix = ""
+  target_bucket = ""
+  target_prefix = ""
+}
+system_bucket="cumulus-sit-internal"
+
+launchpad_api = "https://api.launchpad.nasa.gov/icam/api/sm/v1"
+launchpad_certificate = "launchpad.pfx"
```
laurenfrederick: Does this actually exist in the crypto dir along with the ems keys?

nemreid: Since creating the PR, I've uploaded the keys and created (and then corrected) the TEA JWT secrete. Build just passed, I think everything that should be there is there now.

PR #1490, opened 2020-02-24
CUMULUS-1718: PDR race condition fix & reporting changes

nemreid:

Addresses [CUMULUS-1718: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1718)

## Changes

- **CUMULUS-1718**
  - The `@cumulus/sf-sns-report` task for reporting mid-workflow updates has been retired.
  This task was used as the `PdrStatusReport` task in our ParsePdr example workflow.
  If you have a ParsePdr or other workflow using this task, use `@cumulus/sf-sqs-report` instead.
  Trying to deploy the old task will result in an error as the cumulus module no longer exports `sf_sns_report_task`.
  - Migration instruction: In your workflow definition, for each step using the old task change:
  ```"Resource": "${module.cumulus.sf_sns_report_task.task_arn}"```
  to
  ```"Resource": "${module.cumulus.sf_sqs_report_task.task_arn}"```
  - Added `@cumulus/sf-sqs-report` task to allow mid-workflow reporting updates.
  - Added `stepfunction_event_reporter_queue_url` and `sf_sqs_report_task` outputs to the `cumulus` module.
  - Added `publishPdrs` Lambda to handle publishing PDR messages to SNS when PDR records are written to DynamoDB.
  - Added `@cumulus/api/models/Pdr.storePdrsFromCumulusMessage` to store PDRs from a Cumulus message to DynamoDB.
  - Updated `cwSfEventToDbRecords` to write PDR records to DynamoDB from the incoming Cumulus message
  - Moved `data-cookbooks/sns.md` to `data-cookbooks/ingest-notifications.md` and updated it to reflect recent changes.
  - Deprecate `@cumulus/sf-sns-report`.
    - This task has been updated to always throw an error directing the user to use `@cumulus/sf-sqs-report` instead. This was done because there is no longer an SNS topic to which to publish, and no consumers to listen to it.
  - Removed the `publishReports` lambda, which was replaced by `cwSfEventToDbRecords`, due to a refactor in the way we process step function events.
  - Removed the `sf_sns_report_task` output from the `cumulus` module.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


packages/api/models/pdrs.js
```diff
@@ -100,24 +100,44 @@ class Pdr extends Manager {
...
-
-    if (!pdrName) return Promise.resolve();
-
-    const pdrRecord = Pdr.generatePdrRecord(payload);
+  async storePdrsFromCumulusMessage(cumulusMessage) {
+    const pdrRecord = this.generatePdrRecord(cumulusMessage);
+    if (!pdrRecord) return null;
+    const updateParams = await this.generatePdrUpdateParamsFromRecord(pdrRecord);
+    if (pdrRecord.status === 'running') {
+      updateParams.ConditionExpression = 'execution <> :execution OR progress < :progress';
```
markdboyd: So only update if the execution is different (re-run case) or the progress currently stored is **less than the value we're trying to update it to**

packages/api/models/pdrs.js
```diff
@@ -100,24 +100,44 @@ class Pdr extends Manager {
...
+          return null;
+        }
+        throw err;
+      }
+    }
+    return this.dynamodbDocClient.update(updateParams).promise();
+  }
 
-    return this.create(pdrRecord);
+  async generatePdrUpdateParamsFromRecord(pdrRecord) {
```
markdboyd: Do we need a docstring here? 🐜

packages/api/models/pdrs.js
```diff
@@ -100,24 +100,44 @@ class Pdr extends Manager {
...
+    const mutableFieldNames = Object.keys(pdrRecord);
```
markdboyd: So all fields are mutable even for "running" PDRs because we have the conditional expression to defend against unwanted updates? From what I can see, the only values that could change besides `progress` and `stats` (which we want to change) are `timestamp` and `duration`, but I think we also would want those to change

packages/api/models/pdrs.js
```diff
@@ -100,24 +100,44 @@ class Pdr extends Manager {
...
-   * Create a new PDR record from incoming SNS messages
+   * Try to update a PDR record from a cloudwatch event.
+   * If the record already exists, only update if the execution is different (re-run case).
    *
-   * @param {Object} payload - SNS message containing the output of a Cumulus Step Function
-   * @returns {Promise<Object>} a PDR record
+   * @param {Object} cumulusMessage - cumulus message object
    */
-  createPdrFromSns(payload) {
-    const pdrObj = get(payload, 'payload.pdr', get(payload, 'meta.pdr'));
```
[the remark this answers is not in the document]
markdboyd: Also, as much as I want to delete this method, don't we need to deprecate it?

packages/api/models/pdrs.js
```diff
@@ -100,24 +100,44 @@ class Pdr extends Manager {
...
-    const pdrName = get(pdrObj, 'name');
-
-    if (!pdrName) return Promise.resolve();
-
-    const pdrRecord = Pdr.generatePdrRecord(payload);
+  async storePdrsFromCumulusMessage(cumulusMessage) {
```
markdboyd: So the relationship of a cumulus message to a PDR is 1:1, correct? Not sure if we want to change the name here to `storePdrFromCumulusMessage` 🐜

packages/api/tests/lambdas/test-cw-sf-event-to-db-records.js
```diff
@@ -170,33 +180,103 @@ test.serial('saveGranulesToDb() does not throw an exception if storeGranulesFrom
   }
 });
 
-test('The cw-sf-event-to-db-records Lambda function creates execution and granule records', async (t) => {
-  const { cumulusMessage, executionModel, granuleModel } = t.context;
+test.serial('savePdrsToDb() saves a PDR record', async (t) => {
```
markdboyd: The function is named `savePdrToDb` 🐜

packages/api/tests/lambdas/test-cw-sf-event-to-db-records.js
```diff
@@ -170,33 +180,103 @@ test.serial('saveGranulesToDb() does not throw an exception if storeGranulesFrom
...
+
+  const collectionId = (() => {
+    const { name, version } = cumulusMessage.meta.collection;
+    return constructCollectionId(name, version);
+  })();
+
+  const fetchedPdr = await pdrModel.get({ pdrName: pdr.name });
+  const expectedPdr = {
+    pdrName: pdr.name,
+    collectionId,
```
markdboyd: Oh, this works because it's an immediately invoked function?

packages/api/tests/models/test-pdrs-model.js
```diff
@@ -26,12 +25,14 @@ const createPdrMessage = ({
   numFailedExecutions = 0,
   numRunningExecutions = 0,
   createdAtTime = Date.now(),
+  execution = randomId('execution'),
   providerId = 'prov1',
+  stateMachine = randomId('pdr'),
```
markdboyd: It doesn't matter greatly, but I assume you meant to do something like `randomId('stateMachine')` here

tasks/sf-sns-report/index.js
```diff
@@ -1,33 +1,14 @@
...
-  await publishReportSnsMessages(message);
-
-  return get(message, 'payload', {});
-}
-
-exports.publishSnsMessage = publishSnsMessage;
+const throwDisabledError = () => {
+  throw new Error(
+    '@cumulus/sf-sns-report has been deprecated due to a change in reporting architecture. As the'
+    + ' reporting SNS topic and any consumers to listen to it have been removed, this task has'
```
markdboyd: The topic hasn't been deprecated, has it? But pushing to the topic no longer results in records being saved/updated

tf-modules/archive/cw_sf_event_to_db_records.tf
```diff
@@ -66,6 +80,41 @@ resource "aws_iam_role_policy" "cw_sf_event_to_db_records_lambda_role_policy" {
...
+    }
+  }
+}
+
+resource "aws_sqs_queue_policy" "cw_sf_event_to_db_records_input_queue_policy" {
+  queue_url = aws_sqs_queue.cw_sf_event_to_db_records_input_queue.id
+  policy = data.aws_iam_policy_document.cw_sf_event_send_message_policy.json
+}
+
+resource "aws_lambda_event_source_mapping" "start_sf_mapping" {
```
markdboyd: Why is this called `start_sf_mapping`?

tf-modules/archive/cw_sf_event_to_db_records.tf
```diff
@@ -66,6 +80,41 @@ resource "aws_iam_role_policy" "cw_sf_event_to_db_records_lambda_role_policy" {
   policy = data.aws_iam_policy_document.cw_sf_event_to_db_records_lambda.json
 }
 
+
+
+resource "aws_sqs_queue" "cw_sf_event_to_db_records_input_queue" {
+  name = "${var.prefix}-cwSfEventToDbRecordsInputQueue"
+  receive_wait_time_seconds  = 20
+  visibility_timeout_seconds = (aws_lambda_function.cw_sf_event_to_db_records.timeout * 6)
```
markdboyd: Very clever, but why this specific multiplier? Don't we just want to ensure that the visibility timeout is at least greater than the lifespan of the Lambda?

tf-modules/ingest/message_template.tf
```diff
@@ -43,11 +43,12 @@ resource "aws_iam_role_policy" "lambda_processing_role_get_secrets" {
 
 locals {
   default_queues = {
-    triggerLambdaFailure      = aws_sqs_queue.trigger_lambda_failure.id
-    startSF                   = aws_sqs_queue.start_sf.id
     backgroundProcessing      = aws_sqs_queue.background_processing.id
     kinesisFailure            = aws_sqs_queue.kinesis_failure.id
+    reporting                 = var.cw_sf_event_to_db_records_sqs_queue_url
```
markdboyd: I'm probably forgetting something, but is there a value/necessity in including the queue URL in the list of queues in the message?

docs/data-cookbooks/ingest-notifications.md
```diff
@@ -1,30 +1,38 @@
...
+The `cwSfEventToDbRecords` Lambda function reads from the `cwSfEventToDbRecordsInputQueue` queue and updates DynamoDB. The DynamoDB events for the `ExecutionsTable`, `GranulesTable` and `PdrsTable` are streamed on DynamoDBStreams, which are read by the `publishExecutions`, `publishGranules` and `publishPdrs` Lambda functions, respectively.
+
+These Lambda functions publish to the three SNS topics both when the workflow starts and when it reaches a terminal state (completion or failure). The following describes how many message(s) each topic receives **both on workflow start and workflow completion/failure**:
 
 - `reportExecutions` - Receives 1 message per workflow execution
 - `reportGranules` - Receives 1 message per granule in a workflow execution
 - `reportPdrs` - Receives 1 message per PDR
 
-![Diagram of architecture for reporting workflow ingest notifications from AWS Step Functions](assets/workflow_reporting_diagram.png)
+![Diagram of architecture for reporting workflow ingest notifications from AWS Step Functions](assets/interfaces.svg)
```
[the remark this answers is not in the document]
nemreid: I don't think so. It doesn't look like the doc versioning saves a copy of the assets, so that would break all previous versions.

docs/data-cookbooks/ingest-notifications.md
```diff
@@ -1,30 +1,38 @@
...
 
-The `publishReports` Lambda is triggered via a [Cloudwatch rule for any Step Function execution state transitions](https://docs.aws.amazon.com/step-functions/latest/dg/cw-events.html). Both the `publishReports` Lambda and Cloudwatch rule and are included by default in a Cumulus deployment.
+The  Lambda is triggered via a [Cloudwatch rule for any Step Function execution state transitions](https://docs.aws.amazon.com/step-functions/latest/dg/cw-events.html). The `cwSfEventToDbRecords` Lambda is included in the `cumulus module` and the Cloudwatch rule in the `workflow` module and are included by default in a Cumulus deployment.
```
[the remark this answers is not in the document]
nemreid: Good catch 👍

docs/data-cookbooks/ingest-notifications.md
```diff
@@ -1,30 +1,38 @@
...
 
-## Sending SNS messages to report status
+## Sending SQS messages to report status
```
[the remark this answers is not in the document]
nemreid: Gooder catch 👍

tf-modules/archive/cw_sf_event_to_db_records.tf
nemreid: > To allow your function time to process each batch of records, set the source queue's visibility timeout to at least 6 times the timeout that you configure on your function. The extra time allows for Lambda to retry if your function execution is throttled while your function is processing a previous batch.

https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html

tf-modules/ingest/message_template.tf
nemreid: If user tasks wanted to use it, they could, I suppose.

nemreid: I don't have a real strong justification other than it seemed like the same kind of thing we already expose in this block.

tasks/sf-sns-report/index.js
nemreid: Hmm.. Arguably. I guess you are right in that a topic with the same name still exists, but it no longer serves the same purpose or have the same producers/consumers.

packages/api/tests/lambdas/test-cw-sf-event-to-db-records.js
nemreid: Yeah, hence the `();` at the end of that.

packages/api/models/pdrs.js
nemreid: Yeah, I guess so. Makes me think about whether to support creating a record for each PR in the output from `DiscoverPdrs`... 🤔

packages/api/models/pdrs.js
nemreid: Yeah, disallowing meaningless updates through the conditional entirely. Not much value in updating just timestamps in my opinion, as we discussed previously.

packages/api/models/pdrs.js
nemreid: Correct. 💯

CHANGELOG.md
```diff
@@ -30,14 +35,19 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     deployment, which will cause the deployment to fail if the migration
     Lambda has not been run.
 
+- **CUMULUS-1718**
+  - The `@cumulus/sf-sns-report` task for reporting mid-workflow updates has been retired.
+  This task was used as the `PdrStatusReport` task in our ParsePdr example workflow.
+  If you have a ParsePdr or other workflow using this task, use `@cumulus/sf-sqs-report` instead.
+  Trying to deploy the old task will result in an error as the cumulus module no longer exports `sf_sns_report_task`.
+  - Migration instruction: In your workflow definition, for each step using the old task change:
+  ```"Resource": "${module.cumulus.sf_sns_report_task.task_arn}"```
```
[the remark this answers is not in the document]
nemreid: I guess not. Will update. Might just be a formatting thing within a list item, though.

example/spec/parallel/ingest/ingestFromPdrSpec.js
```diff
@@ -290,11 +290,9 @@ describe('Ingesting from PDR', () => {
       describe('PdrStatusReport lambda function', () => {
         let lambdaOutput;
         beforeAll(async () => {
-          lambdaOutput = await lambdaStep.getStepOutput(parsePdrExecutionArn, 'SfSnsReport');
+          lambdaOutput = await lambdaStep.getStepOutput(parsePdrExecutionArn, 'SfSqsReport');
         });
 
-        // SfSnsReport lambda is used in the workflow multiple times, apparantly, only the first output
```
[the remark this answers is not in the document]
nemreid: I guess that's true.. I'm not really sure how loops like that are reflected in the `executionHistory`.

example/lambdas/snsS3Test/index.js
```diff
@@ -42,7 +42,28 @@ async function handleGranules(event) {
...
+ * @returns {Promise} confirmation of added message
+ */
+async function handlePdrs(event) {
+  const s3 = new S3();
+  const messageString = event.Records[0].Sns.Message;
+  const pdr = JSON.parse(messageString);
+  if (!pdr.name) {
+    return Promise.resolve();
+  }
+  return s3.putObject({
```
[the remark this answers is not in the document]
nemreid: I was wondering about having an integration test using this consumer lambda.

packages/api/lambdas/cw-sf-event-to-db-records.js
```diff
@@ -24,21 +33,28 @@ const saveGranulesToDb = async (cumulusMessage) => {
...
 };
 
 const handler = async (event) => {
-  const cumulusMessage = await getCumulusMessageFromExecutionEvent(event);
-
-  await Promise.all([
-    saveExecutionToDb(cumulusMessage),
-    saveGranulesToDb(cumulusMessage)
-  ]);
+  const sqsMessages = event.Records;
```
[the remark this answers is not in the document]
nemreid: I mean, the messages from the queue are still expected to contain cloudwatch step function events..

I'd be surprised if the response from SQS somehow returned an empty object, but I guess there is no harm in using `get`.

docs/data-cookbooks/ingest-notifications.md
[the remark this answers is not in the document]
markdboyd: Good point

example/lambdas/snsS3Test/index.js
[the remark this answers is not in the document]
markdboyd: FWIW I added tests to `IngestGranuleSuccessSpec.js` for granules

https://github.com/nasa/cumulus/blob/master/example/spec/parallel/ingestGranule/IngestGranuleSuccessSpec.js#L635
https://github.com/nasa/cumulus/blob/master/example/spec/parallel/ingestGranule/IngestGranuleSuccessSpec.js#L651

example/spec/parallel/ingest/ingestFromPdrSpec.js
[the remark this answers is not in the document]
markdboyd: The step itself will appear listed in the execution history multiple times, but `lambdaStep.getStepOutput` will just get the output from the first invocation

packages/api/models/pdrs.js
markdboyd: Wouldn't each PDR in `DiscoverPdrs` trigger a separate `ParsePdr` workflow and the processing of those workflows would create the individual PDR records?

tasks/sf-sns-report/index.js
markdboyd: OK. Your call if you want to update the language to only mention there are no consumers reporting output from the topics anymore.

tf-modules/ingest/message_template.tf
markdboyd: Yeah, I can't remember if there's a reason why queues need to be here in the message, except in the case of throttle queues which are looked up from the message.

The only concern I guess would be if this queue URL is in the Cumulus message then tasks might publish to it directly rather than using the `sf-sqs-report` task. Is that a bad thing?

nemreid: Not as long as they follow the stipulations laid out in `ingest-notifications.md`

nemreid: Not as long as they follow the stipulations laid out in `ingest-notifications.md`

packages/api/models/pdrs.js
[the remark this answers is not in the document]
nemreid: Out of curiosity: On what basis? I had thought we'd settled on not wanting ourselves to use the models outside of the API package, presumably the same would apply to users.

If it's just a general "don't delete code without deprecating ever" then we already have an issue with all of the `report-*` lambda code we've deleted. 🙂

packages/api/models/pdrs.js
nemreid: Yeah.. I was thinking about what if `QueuePdrs` fails, but I guess we wouldn't want a bunch of `running` records with the wrong execution and no information on stats.

packages/api/models/pdrs.js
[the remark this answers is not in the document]
markdboyd: I guess the basis was my understanding of the convo we had, but I wasn't actually in that meeting. Given that technically any exported code could be being used by someone else. But your call.

Do we need to care/worry about no longer falling back to look for a PDR on `meta.pdr`?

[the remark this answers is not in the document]
nemreid: > generatePdrRecord does not fallback to looking for a PDR on meta.pdr
I don't think we want to check `meta.pdr` and here's why:
If an `ingestGranule` execution has a `meta.PDR` field, then it would trigger a record update. However, there would not be any `stats` values in the payload, so the `progress` value would be set to `0`. Because this is a different execution, an update *would* occur but the progress would be reset to `0`. Presumably the `ParsePdr` execution would finish later and everything would end up correct in the end, but I don't think it's a valuable use case.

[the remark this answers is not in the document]
markdboyd: OK. Aren't we relying on the convention there that the IngestGranule execution will always have the PDR on `meta.pdr` though? That could change, right?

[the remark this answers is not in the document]
markdboyd: Discussed offline and determined that support for `meta.pdr` had already been deprecated, which also addresses the concerns about the IngestGranule workflow

packages/api/lambdas/cw-sf-event-to-db-records.js
```diff
@@ -24,21 +33,28 @@ const saveGranulesToDb = async (cumulusMessage) => {
...
+
+  return Promise.all(sqsMessages.map(async (message) => {
+    const executionEvent = JSON.parse(get(message, 'Body', get(message, 'body', '{}')));
```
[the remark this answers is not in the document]
nemreid: b538730

packages/api/lambdas/cw-sf-event-to-db-records.js
```diff
@@ -24,21 +33,28 @@ const saveGranulesToDb = async (cumulusMessage) => {
...
+    const cumulusMessage = await getCumulusMessageFromExecutionEvent(executionEvent);
+
+    await Promise.all([
```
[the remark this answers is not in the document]
nemreid: Not really. Probably error prone to use this pattern outside of a `try-catch` because it breaks `.then` followups in the general case.

packages/api/lambdas/cw-sf-event-to-db-records.js
```diff
@@ -24,21 +33,28 @@ const saveGranulesToDb = async (cumulusMessage) => {
...
+      saveExecutionToDb(cumulusMessage),
+      saveGranulesToDb(cumulusMessage),
+      savePdrToDb(cumulusMessage)
```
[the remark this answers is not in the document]
nemreid: Correct.

example/spec/parallel/ingest/ingestFromPdrSpec.js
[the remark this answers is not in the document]
nemreid: restored 👍

example/lambdas/snsS3Test/index.js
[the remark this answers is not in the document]
nemreid: Added tests. Helped me find some serious bugs with my TF configs, and an existing pointless resource. 👍 1f1c72d

PR #1491, opened 2020-02-24
Cumulus-1713 Removing integration tests error message

kkelly51:
Turns off SNS collection reporting for integration tests. Should no longer get error about missing SNS Topic ARN

### DIFF
--- a/packages/api/models/collections.js
+++ b/packages/api/models/collections.js
@@ -117,11 +117,13 @@ class Collection extends Manager {
     await this.collectionConfigStore.put(dataType || name, version, item);
 
     const collectionRecord = await super.create(item);
-    const publishRecord = {
-      event: 'Create',
-      record: collectionRecord
-    };
-    await publishCollectionSnsMessage(publishRecord);
+    if (!process.env.noSNS) {
+      const publishRecord = {
+        event: 'Create',
+        record: collectionRecord
+      };
+      await publishCollectionSnsMessage(publishRecord);
+    }
 
     return collectionRecord;
   }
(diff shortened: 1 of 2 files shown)


packages/api/models/collections.js
```diff
@@ -169,15 +171,17 @@ class Collection extends Manager {
...
-    const record = {
-      event: 'Delete',
-      deletedAt: Date.now(),
-      record: {
-        name,
-        version
-      }
-    };
-    await publishCollectionSnsMessage(record);
+    if (!process.env.noSNS) {
```
markdboyd: I have mixed feelings about using env vars specific to testing in our real API package. See suggestion below

packages/integration-tests/index.js
```diff
@@ -256,6 +256,7 @@ function setProcessEnvironment(stackName, bucketName) {
   process.env.CollectionsTable = `${stackName}-CollectionsTable`;
   process.env.ProvidersTable = `${stackName}-ProvidersTable`;
   process.env.RulesTable = `${stackName}-RulesTable`;
+  process.env.noSNS = true;
```
markdboyd: Is there any reason to not go the other way and set the actual env var that will allow the publish requests to the collections topic to succeed? That would also eliminate the error message

```
process.env.collection_info_sns_topic_arn = `${stackName}-collection-information-topic`
```

nemreid:
I think the issue here for integration tests is that we are calling `collectionModel.delete` in `addCollections`, instead of invoking the API lambda to delete the collection instead. That's why the `process.env.collection_sns_topic_arn` variable is unset, and an error is thrown because the collection fails to add.

We could set the env, as Mark suggested, but it would be cleaner to refactor so that the API is called instead of the models in our tests. This would also remove the error because the API has the env var set.

I understand if that is considered out of scope, though, as it may be a fair bit of work.

yjpa7145:
> I think the issue here for integration tests is that we are calling `collectionModel.delete` in `addCollections`, instead of invoking the API lambda to delete the collection instead. That's why the `process.env.collection_sns_topic_arn` variable is unset, and the warning is thrown because the collection delete call can't publish.
> 
> We could set the env, as Mark suggested, but it would be cleaner to refactor so that the API is called instead of the models in our tests. This would also remove the error because the API has the env var set.
> 
> I understand if that is considered out of scope, though, as it may be a fair bit of work.

It's actually a pretty small change: https://github.com/nasa/cumulus/pull/1494

kkelly51:
Closing because better fix in PR [#1494](https://github.com/nasa/cumulus/pull/1494)

PR #1497, opened 2020-02-26
Cumulus 1768 stats type fix

laurenfrederick:
**Summary:** Fix an issue with the stats endpoints (average, aggs, and histogram) not using the correct type

Addresses [CUMULUS-1768](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1768)

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


packages/api/tests/es/test-stats.js
```diff
@@ -0,0 +1,132 @@
+'use strict';
```
laurenfrederick: This is by no means exhaustive, but there are currently no tests for `es/stats` and I started the debugging by starting to write this test file. I plan to continue writing tests, but would like to get the fix merged in.

packages/api/tests/es/test-stats.js
```diff
@@ -0,0 +1,132 @@
...
+
+const test = require('ava');
+const rewire = require('rewire');
+const get = require('lodash.get');
+
+const awsServices = require('@cumulus/aws-client/services');
+const s3 = require('@cumulus/aws-client/S3');
+const { randomString } = require('@cumulus/common/test-utils');
```
flamingbear: It's you!  do you mind using randomId instead?  It really helps debug.  Only gotcha is for buckets on localstack they should be lowercase with no special characters.  e.g.`randomId('systembucket')`

packages/api/tests/es/test-stats.js
```diff
@@ -0,0 +1,132 @@
...
+    { key: 'failed', count: 2 }
+  ]);
+
+  const collectionStats = new Stats({}, 'collection', process.env.ES_INDEX);
+  const collectionCountResult = await collectionStats.count();
+
+  t.is(collectionCountResult.meta.count, 1);
+
+  const req = { query: { type: 'granule '} };
+  console.log(get(req, 'query.type', null));
```
flamingbear: should this be a test here?

packages/api/tests/es/test-stats.js
flamingbear: I don't know where my last comment went, but will you use randomId() instead?  (just be sure to use all lowercase on your bucket name e.g. `randomId('systembucket')`.  Thanks.  It makes it easier to debug.

laurenfrederick: Yep maybe all my fault. I had copy/pasted. But I did fix.

packages/api/tests/es/test-stats.js
laurenfrederick: No I moved those specific tests to that other file. I mean yes I am still sloppy, but that was quick testing my fix before I deployed, then I accidentally left it in. I added all of the tests in the other file to test what I had been testing here.

packages/api/tests/es/test-stats.js
```diff
@@ -0,0 +1,127 @@
...
+const { Search } = require('../../es/search');
+const models = require('../../models');
+const { fakeGranuleFactoryV2, fakeCollectionFactory } = require('../../lib/testUtils');
+const { bootstrapElasticSearch } = require('../../lambdas/bootstrap');
+const Stats = require('../../es/stats');
+
+const collectionTable = randomId('collectionsTable');
+const granuleTable = randomId('granulesTable');
+
+process.env.system_bucket = randomId('systemBucket');
```
flamingbear: I'm not 100% sure but I think this gets you in trouble with localstack.  try `systembucket`

flamingbear: Probably doesn't matter in this case but 
```
>  awslocal s3 mb s3://LocalStack
make_bucket: LocalStack
>  awslocal s3 ls
2020-02-26 15:38:40 localbucket
2020-02-26 21:38:34 localstack
```

flamingbear: It looks like it's fine since it lowercases both directions and and can delete the bucket after.

PR #1499, opened 2020-02-26
Fix TF cumulus module to support non-private API Gateway

chuckwondo:
Fixes deployment error caused by setting private_archive_api_gateway
to false by allowing vpc_id to be null and `lambda_subnet_ids` to be an
empty list ([]).  The `subnet_ids` var is replaced by the 2 vars
`lambda_subnet_ids` and `ecs_cluster_instance_subnet_ids`, which now
surfaces this distinction that already exists within the `cumulus`
module.

This still requires the policy for the API Gateway to be manually
edited or removed, if necessary, because the policy will not be
updated automatically due to NGAP requirements.


tf-modules/cumulus/ecs_cluster_instance_autoscaling_cf_template.yml.tmpl
```diff
@@ -95,10 +95,18 @@ Resources:
       AutoScalingRollingUpdate:
         MinInstancesInService: ${min_size}
     Properties:
+%{ if length(availability_zones) > 0 ~}
+      AvailabilityZones:
```
yjpa7145: Is there a time that we would want to create an autoscaling group without specifying the subnets that it should use? From my reading of the AWS docs, it looks like `AvailabilityZones` is only required for EC2-Classic instances.

* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html#cfn-as-group-availabilityzones
* https://www.terraform.io/docs/providers/aws/r/autoscaling_group.html

tf-modules/archive/main.tf
```diff
@@ -3,3 +3,15 @@ terraform {
     aws = ">= 2.31.0"
   }
 }
+
+locals {
+  subnet_ids = var.lambda_subnet_ids == null ? [] : var.lambda_subnet_ids
```
yjpa7145: This variable was called `lambda_subnet_ids` instead of `subnet_ids` so that we would have the ability to specify one set of subnets for Lambda functions and another set for other things (like EC2 instances or ES domains). I'd rather see that kept instead of the more generic `local.subnet_ids`.

tf-modules/cumulus/ecs_cluster_instance_autoscaling_cf_template.yml.tmpl
chuckwondo: This change was required because I got an error when neither `AvailabilityZones` nor `VPCZoneIdentifier` were supplied. The error indicated that one or the other must be specified.

tf-modules/archive/main.tf
chuckwondo: Ah, sure thing. I'll update.

tf-modules/cumulus/ecs_cluster_instance_autoscaling_cf_template.yml.tmpl
markdboyd: The real issue here was that since Chuck had no `subnet_ids` set, the `VPCZoneIdentifier` block was empty and deploying this ECS cluster was failing. As the docs say:

> A list of Availability Zones for the group. You must specify one of the following properties: VPCZoneIdentifier or AvailabilityZones. 

So dding the ability to set the availability zones was just a workaround.

@yjpa7145 - Is there a better solution here? Is there any way to make the whole ECS cluster optional

yjpa7145:
The `archive` Terraform module's `lambda_subnet_ids` input variable currently defaults to `null`. Given these changes, that needs to be updated to default to `[]`.

example/cumulus-tf/main.tf
```diff
@@ -45,10 +45,10 @@ module "cumulus" {
   prefix = var.prefix
 
   vpc_id            = var.vpc_id
-  lambda_subnet_ids = var.subnet_ids
+  lambda_subnet_ids = var.lambda_subnet_ids
 
   ecs_cluster_instance_image_id   = data.aws_ssm_parameter.ecs_image_id.value
-  ecs_cluster_instance_subnet_ids = var.subnet_ids
+  ecs_cluster_instance_subnet_ids = var.ecs_cluster_instance_subnet_ids
```
yjpa7145: The `example/cumulus-tf` module doesn't have a `ecs_cluster_instance_subnet_ids` input variable. Given that these tests are only run in NGAP, it's safe to leave this the way it was.

example/cumulus-tf/terraform.tfvars.example
```diff
@@ -13,19 +13,21 @@ buckets = {
...
     name = "PREFIX-public"
     type = "public"
   }
 }
-subnet_ids    = ["subnet-12345"]
-system_bucket = "PREFIX-internal"
-vpc_id        = "vpc-12345"
+
+lambda_subnet_ids               = ["subnet-12345"]
+ecs_cluster_instance_subnet_ids = ["subnet-12345"]
```
yjpa7145: Same comment as here: https://github.com/nasa/cumulus/pull/1499/files?utf8=%E2%9C%93&diff=split&w=1#r385733658

yjpa7145:
> The `archive` Terraform module's `lambda_subnet_ids` input variable currently defaults to `null`. Given these changes, that needs to be updated to default to `[]`.

The same goes for the `ingest` module.

example/cumulus-tf/main.tf
chuckwondo: ok

example/cumulus-tf/terraform.tfvars.example
chuckwondo: ok

chuckwondo:
> > The `archive` Terraform module's `lambda_subnet_ids` input variable currently defaults to `null`. Given these changes, that needs to be updated to default to `[]`.
> 
> The same goes for the `ingest` module.

I also found a `null` default for it in `tf-modules/cumulus/variables.tf`, so I'll change it to `[]` in all 3.

yjpa7145:
> I also found a `null` default for it in `tf-modules/cumulus/variables.tf`, so I'll change it to `[]` in all 3.

I would be inclined to leave that one as `null`. If the `cumulus` module passes `null` into the `ingest` or `archive` modules, they are each going to use `[]` as a result. If you change the default in the `cumulus` module, you're also going to want to check all of the other modules that value is passed into and make sure that passing in `[]` instead of `null` doesn't change their behavior. If any of them have `var.lambda_subnet_ids == null ? x : y`, then that'll break them. The same goes for the `cumulus` module itself.

chuckwondo:
> > I also found a `null` default for it in `tf-modules/cumulus/variables.tf`, so I'll change it to `[]` in all 3.
> 
> I would be inclined to leave that one as `null`. If the `cumulus` module passes `null` into the `ingest` or `archive` modules, they are each going to use `[]` as a result. If you change the default in the `cumulus` module, you're also going to want to check all of the other modules that value is passed into and make sure that passing in `[]` instead of `null` doesn't change their behavior. If any of them have `var.lambda_subnet_ids == null ? x : y`, then that'll break them. The same goes for the `cumulus` module itself.

ah, got it

PR #1504, opened 2020-02-28
Refactor use of collection.dataType in discover-granules

yjpa7145:
This addresses CUMULUS-1736.

- The `@cumulus/discover-granules` task now sets the `dataType` of discovered granules based on the `name` of the configured collection, not the `dataType`.
- The config schema of the `@cumulus/discover-granules` task now requires that collections contain a `version`.
- The `@cumulus/sync-granule` task now sets the `dataType` of granules based on the `name` of the configured collection, not the `dataType`.
- The `collections` model in the API package now determines the name of a collection based on the `name` property, rather than using `dataType` and then falling back to `name`.
- The `@cumulus/integration-tests.loadCollection()` function no longer appends the postfix to the end of the collection's `dataType`.
- The `@cumulus/integration-tests.addCollections()` function no longer appends the postfix to the end of the collection's `dataType`.


tasks/queue-granules/tests/index.js
```diff
@@ -102,10 +102,10 @@ test.serial('The correct output is returned when granules are queued without a P
   const { event } = t.context;
   event.input.granules = [
     {
-      dataType: dataType, version: version, granuleId: randomString(), files: []
+      dataType, version, granuleId: randomString(), files: []
```
markdboyd: Not a question about this test but about the logic in queue-granules. So queue-granules does this

```
const collectionConfig = await collectionConfigStore.get(granule.dataType, granule.version);
```

Why is it pulling from granule.dataType and granule.version instead of getting those values from the configured collection name and version? I realize that right now queue-granules does not expect a collection as task config, but it could. It also feels odd to keep using `granule.dataType` when we mean `granule.collectionName`

Is the idea that queue-granules could handle queueing granules from multiple collections? Is that even possible though, given that discover-granules expects a single collection definition and triggers queue-granules?

CHANGELOG.md
```diff
@@ -8,6 +8,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### BREAKING CHANGES
 
+- **CUMULUS-1736**
+  - The `@cumulus/discover-granules` task now sets the `dataType` of discovered
+    granules based on the `name` of the configured collection, not the
+    `dataType`.
+  - The config schema of the `@cumulus/discover-granules` task now requires that
+    collections contain a `version`.
+  - The `@cumulus/sync-granule` task now sets the `dataType` of granules based
```
markdboyd: This is in some case where `granule.dataType` was not set during the discovery?

CHANGELOG.md
```diff
@@ -107,6 +116,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- **CUMULUS-1736**
+  - The `collections` model in the API package now determines the name of a
```
markdboyd: Should we mention that items in the collection in the collection config store are also now stored by name/version instead of dataType/version?

packages/common/collection-config-store.js
```diff
@@ -33,12 +33,12 @@ class CollectionConfigStore {
...
    *
-   * @param {string} dataType - the name of the collection config to fetch
-   * @param {string} dataVersion - the version of the collection config to fetch
+   * @param {string} name - the name of the collection config to fetch
+   * @param {string} version - the version of the collection config to fetch
    * @returns {Object} the fetched collection config
    */
-  async get(dataType, dataVersion) {
-    const collectionId = constructCollectionId(dataType, dataVersion);
+  async get(name, version) {
```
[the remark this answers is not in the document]
yjpa7145: From talking to Menno and the DAACs, they should always be the same.

CHANGELOG.md
yjpa7145: I'll update it to read:
```md
- The `@cumulus/sync-granule` task will set the `dataType` and `version` of a
    granule based on the configured collection if those fields are not already
    set on the granule. Previously it was using the `dataType` field of the
    configured collection, then falling back to the `name` field of the
    collection. This update will just use the `name` field of the collection to
    set the `dataType` field of the granule.
```

CHANGELOG.md
yjpa7145: I don't think so, the CollectionConfigStore is an internal implementation detail of the API, and isn't exposed to the world.

packages/api/models/schemas.js
```diff
@@ -77,7 +77,7 @@ module.exports.collection = {
     },
     dataType: {
       title: 'DataType',
-      description: 'This is used to identify a granule in a PDR',
+      description: 'This field is deprecated and unused',
```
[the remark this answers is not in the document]
yjpa7145: Will do.

packages/api/tests/endpoints/collections/update-collection.js
```diff
@@ -90,14 +90,13 @@ test.todo('CUMULUS-912 PUT with pathParameters and with an unauthorized user ret
 test('PUT replaces an existing collection', async (t) => {
   const { testCollection, testCollection: { name, version } } = t.context;
   const expectedCollection = {
-    ...omit(testCollection, ['dataType', 'duplicateHandling']),
+    ...omit(testCollection, ['duplicateHandling']),
```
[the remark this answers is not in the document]
yjpa7145: Fixed.

tasks/discover-granules/index.js
```diff
@@ -157,7 +157,7 @@ const buildGranule = curry(
 
     return {
       granuleId,
-      dataType: config.collection.dataType,
+      dataType: config.collection.name,
```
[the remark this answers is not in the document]
yjpa7145: It's unfortunately named. The `granule.dataType` field is really the name of the collection that the granule belongs to. That value is needed.

tasks/queue-granules/tests/index.js
yjpa7145: > Is the idea that queue-granules could handle queueing granules from multiple collections?

That's my understanding, yes.

yjpa7145:
> We also have several test fixtures in `example/data/collections` where the `name` and `dataType` are different. Should we remove `dataType` from those if we're retiring that value to ensure that our integration tests are validating against our updated model changes?

I have a separate branch that does that. I didn't want to include that change in this PR because it contains about 100 near-identical changes, and it would have made reviewing this PR confusing.

tasks/discover-granules/index.js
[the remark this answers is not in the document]
yjpa7145: I created CUMULUS-1788 to look into this.

CHANGELOG.md
yjpa7145: Looking at the the `SyncGranule` code, it's very confusingly written. I'll write another ticket to take a closer look at that and make sure that it's behaving as expected. This PR doesn't substantively change that behavior, though, so I think it should be tackled in another PR.

yjpa7145: Created CUMULUS-1789 for this.

CHANGELOG.md
yjpa7145: The changes in that file are purely renaming the variables, there are no functional changes in how those are stored or fetched.

DATA_MODEL_CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
```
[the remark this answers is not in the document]
yjpa7145: Fixed.

packages/api/models/schemas.js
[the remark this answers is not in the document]
yjpa7145: Done.

DATA_MODEL_CHANGELOG.md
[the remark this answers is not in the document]
markdboyd: Thanks!

tasks/queue-granules/tests/index.js
nemreid: @markdboyd `queue-granules` is also used by the `ParsePdr` workflow, and a PDR _can_ contain multiple granules of differing collections.

PR #1520, opened 2020-03-10
Remove Files model from API and create GranuleFilesCache library

yjpa7145:
This addresses CUMULUS-1725:

* Access to the files table has been moved out of the models code
* It has been documented that the files table is purely a lookup cache and is eventually consistent
* The functionality to update the files table has been moved to its own lambda


packages/api/lib/DistributionEvent.js
```diff
@@ -155,19 +158,23 @@ class DistributionEvent {
...
-      .then((granule) =>
-        (granule
-          ? {
-            collectionId: granule.collectionId,
-            ...deconstructCollectionId(granule.collectionId),
-            granuleId: granule.granuleId,
-            fileType: this.getFileType(this.bucket, this.key, granule)
-          }
-          : {}));
+    if (isNil(this.productInfo)) {
```
laurenfrederick: It feels like maybe the test for this should have been updated?

packages/api/lib/GranuleFilesCache.js
```diff
@@ -0,0 +1,193 @@
...
+ * @param {Array<Object>} params.puts - a list of new files to be added to the
+ *   cache
+ * @param {Array<Object>} params.deletes - a list of files to be removed from
+ *   the cache
+ */
+const batchUpdate = async (params = {}) => {
+  const { puts = [], deletes = [] } = params;
+
+  // Perform input validation
+  puts.forEach((put) => {
```
laurenfrederick: You don't need to change it feels like this one and the next one could be extracted into a reusable function or also a utility function outside this file

PR #1522, opened 2020-03-11
CUMULUS-1774: Intermittent failure in ruleSpec

markdboyd:

Addresses [CUMULUS-1774: Intermittent failure in ruleSpec](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1774)

## Changes

* Adds `waitForRuleInList` helper to wait for a rule to exist in rules listing
* Update ruleSpec to use `waitForRuleInList` to avoid race conditions

## PR Checklist

- [x] Integration tests


example/spec/helpers/ruleUtils.js
```diff
@@ -0,0 +1,28 @@
...
+ *   Query params to use for search in rules listing
+ * @returns {Promise}
+ */
+const waitForRuleInList = async (stackName, ruleQueryParams) => pWaitFor(
+  async () => {
+    const resp = await listRules({
+      prefix: stackName,
+      query: ruleQueryParams
+    });
+    return JSON.parse(resp.body).results.length > 0;
```
yjpa7145: Since the function is called `waitForRuleInList`, returning `undefined` or throwing an exception feels a little more intuitive than `true/false`.

yjpa7145: I take this comment back, I didn't see that this was for the `pWaitFor` function.

markdboyd: `pWaitFor` returns a promise that only resolves if the condition returns `true`

https://github.com/sindresorhus/p-wait-for#pwaitforcondition-options

PR #1523, opened 2020-03-11
Update access tokens model to use strongly consistent reads for get operations

markdboyd:
DynamoDB is eventually consistent by default:

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadConsistency.html

So this PR fixes a race condition exposed by our tests (but also possible in real world usage) where there is an attempt to read an access token from the table before the record has propagated to all DynamoDB servers. This read was occurring in the context of our API authentication code, but could in theory happen anywhere. The solution is to enforce strong consistency for reads for getting access tokens, which is possible using the SDK.


packages/api/models/access-tokens.js
```diff
@@ -11,5 +12,25 @@ class AccessToken extends Manager {
...
+  /**
+   * Gets the item if found. If the record does not exist
+   * the function throws RecordDoesNotExist error
+   *
+   * Enforces strongly consistent reads for the DynamoDB get operation.
+   *
+   * @param {Object} item - the item to search for
+   * @returns {Promise} The record found
+   */
+  async get(item) {
```
yjpa7145: The `async` is unnecessary here.

PR #1524, opened 2020-03-11
add report_collections_sns_topic_arn output

ifestus:
## Changes

* add `report_collections_sns_topic_arn` output to `archive` and `cumulus` modules.
* make use of that new `report_collections_sns_topic_arn` in `example/cumulus-tf`


example/cumulus-tf/outputs.tf
```diff
@@ -20,6 +20,10 @@ output "distribution_redirect_uri" {
 
 # Workflow reporting SNS topics
```
nemreid: With this change, this is no longer exactly a _Workflow_ reporting SNS topics section because collections are created differently.

ifestus: Solid point - I can move this out of under the comment. Do you think this is still the right way to expose the SNS topic arn?

tf-modules/archive/outputs.tf
```diff
@@ -22,6 +22,10 @@ output "publish_reports_lambda_function_arn" {
   value = aws_lambda_function.publish_reports.arn
 }
 
+output "report_collections_sns_topic_arn" {
+  value = aws_sns_topic_report_collections_topic.arn
```
nemreid: I think you mean
```suggestion
  value = aws_sns_topic.report_collections_topic.arn
```

example/cumulus-tf/outputs.tf
nemreid: You can also just remove the `workflow` at the start. This seems correct, yes.

ifestus:
```suggestion
# Reporting SNS topics
```

PR #1525, opened 2020-03-11
Handle S3 eventual consistency when getting readable streams

markdboyd:


packages/aws-client/S3.js
```diff
@@ -313,10 +313,41 @@ exports.putJsonS3Object = (bucket, key, data) =>
...
+ * Use `getS3Object()` before fetching stream to deal
+ * with eventual consistency issues by checking for object
+ * with retries.
+ *
+ * @param {string} bucket - the S3 object's bucket
+ * @param {string} key - the S3 object's key
+ * @returns {ReadableStream}
+ * @throws {Error} if S3 object cannot be found
+ */
+exports.getS3ObjectReadStreamAsync = async (bucket, key) =>
```
yjpa7145: Could be simplified to:
```js
exports.getS3ObjectReadStreamAsync = (bucket, key) =>
  exports.getS3Object(bucket, key, { retries: 3 })
    .then(() => exports.getS3ObjectReadStream(bucket, key));
```

packages/aws-client/S3.js
```diff
@@ -313,10 +313,34 @@ exports.putJsonS3Object = (bucket, key, data) =>
...
```
yjpa7145: The `async` here is not necessary.

PR #1528, opened 2020-03-12
CUMULUS-1784 - Fix EMS distribution reports source IP fetch

markdboyd:

Addresses [CUMULUS-1784: Fix EMS distribution reports source IP fetch.](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1784)

## Changes

* Updated `DistributionEvent.remoteIP()` to get IP address of request from `A-sourceip` then fallback to original behavior

## PR Checklist

- [ ] Update CHANGELOG
- [X] Unit tests


packages/api/lib/DistributionEvent.js
```diff
@@ -120,12 +122,24 @@ class DistributionEvent {
...
     return username && username !== 'None' ? username : '-';
   }
 
+  /**
+   * Get a query parameter value from the request URI in the distribution event.
+   *
+   * @param {string} queryParam - name of query parameter value to get
+   * @returns {string} - value of query parameter
+   */
+  getRequestQueryParamValue(queryParam) {
```
laurenfrederick: It would be nice to have unit tests just testing this function, the `username()` function, and `remoteIp` in addition  to `DistributionEvent.toString()`.

laurenfrederick: Also, really nitpicky here, but the function `getRequestQueryParamValue` is also setting the `parsedRequestUri`, which is a bit confusing at first glance. Maybe move 136-139 into a separate function or at least update the docstring on this one.

markdboyd: `username()` and `remoteIp()` are [Javascript getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get). I could write a test for a single property, like so:

```
test.serial('DistributionEvent.remoteIP() returns correct IP for download via proxy', async (t) => {
  const distributionEvent = new DistributionEvent(t.context.proxyDownloadLogLine);
  t.is(
    distributionEvent.remoteIP,
    '111.222.333.444' // from A-sourceip query param
  );
});
```

But testing for the output of `toString()` includes that test by default, since `toString()` is getting all the class values via the getters and producing a string with them. I'm happy to add it if you think it's necessary though.

And I'll refactor `getRequestQueryParamValue`

PR #1538, opened 2020-03-20
Improve fake provider reliability

yjpa7145:
This addresses CUMULUS-1683. It adds a CloudFormation template that creates an AutoScaling group that will manage a fake provider EC2 instance. Because the IP of that instance could change when a new instance is spun up in the ASG, it stores its private IP address to S3. The tests have been updated to fetch that IP from S3.


website/versioned_docs/version-v1.10.1/workflows/docker.md
```diff
@@ -66,7 +66,7 @@ To create a new processing stream for a data collection, a Dockerfile is used to
 ```
 # cumulus processing Dockerfile: docker-hs3-avaps
 
-FROM 985962406024.dkr.ecr.us-east-1.amazonaws.com/cumulus-base:latest
+FROM 000000000000.dkr.ecr.us-east-1.amazonaws.com/cumulus-base:latest
```
yjpa7145: Not sure why this changed, I suspect it's related to @markdboyd's recent changes.

nemreid: Hmm, looks like #1535 did this before it was deleted: https://github.com/nasa/cumulus/commit/38b541db87c35bd589018f29e7430dab4fa2be5e#diff-24881383b42db6ca0b0504d347133246
@markdboyd, input?

markdboyd: That has to be coming from rewritten history. I'm 100% sure I didn't directly edit a really old versioned documentation file

nemreid: I'm assuming that the BFG would have gotten rid of this so I'm not quite following how it ended up on master.. 🤔

PR #1540, opened 2020-03-20
Remove leading slashes from S3 paths in S3ProviderClient

yjpa7145:


packages/ingest/S3ProviderClient.js
```diff
@@ -6,6 +6,8 @@ const errors = require('@cumulus/common/errors');
 const isString = require('lodash.isstring');
 const { basename, dirname } = require('path');
 
+const sanitizeS3Key = (key) => key.replace(/^\/*/, '');
```
markdboyd: Shouldn't this be `/^\/+/`, since `*` means zero to unlimited times, but we only care if there is at least one leading slash?

yjpa7145: They're both functionally equivalent since you're replacing it with nothing.

markdboyd: I don't know the internals of how `replace` actually operates, but wouldn't this mean it's actually doing *something* when it doesn't need to?

yjpa7145: Sure. I'll change it.

PR #1544, opened 2020-03-24
[CUMULUS-1787] Add endpoint for collections with active granules

dopeters:
**Summary:**

Addresses [CUMULUS-1787: Add collection list endpoint to filter collections by collections with active granules](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1787)

## Changes

* Adds the /granules endpoint for collections which will return collections with granules

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


packages/api/endpoints/collections.js
```diff
@@ -161,5 +180,6 @@ router.put('/:name/:version', put);
 router.delete('/:name/:version', del);
 router.post('/', post);
 router.get('/', list);
+router.get('/granules', granuleList);
```
laurenfrederick: Since we're returning collections and not granules, maybe we should call it `collectionsWithGranules` or `activeCollections`?

dopeters: My thinking was that since the endpoint itself would end up being `/collections/granules`, it would indicate that it's collections with granules. However, now I think you're right, it might imply that it's returning the actual granules rather than the collections themselves. I originally had it as `/active`, so maybe I'll change it back or use one of those suggestions

dopeters: Changed it back to `/active`, added test and updated changelog

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 - **CUMULUS-1732**
   - Added Python task/activity workflow and integration test (`PythonReferenceSpec`) to test `cumulus-message-adapter-python`and `cumulus-process-py` integration.
+- **CUMULUS-1787**
+  - Added `/active` endpoint to collections for returning collections with active granules
```
laurenfrederick:
```suggestion
  - Added `collections/active` endpoint for returning collections with active granules in `@cumulus/api`
```

PR #1546, opened 2020-03-26
Query collections with active granules

laurenfrederick:
**Summary:** Query collections with active granules - ES changes

Addresses [CUMULUS-1787](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1787)

## Changes

* Added functions for querying collections with active granules
* Fixed a bug in collection stats where collection stats were not respecting the collection version

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [ ] Integration tests


packages/api/es/collections.js
```diff
@@ -76,12 +80,83 @@ class Collection extends BaseSearch {
...
+        }
+      }
+    };
+
+    const res = await super.query(searchParams);
+
+    // get aggregations for results
+    res.results = await this.addStatsToCollectionResults(res.results);
+
+    return res;
```
dopeters: Potentially worth making this a call to `this.query` to keep things DRY

packages/api/es/collections.js
```diff
@@ -76,12 +80,81 @@ class Collection extends BaseSearch {
...
+        }
+      }
+    };
+
+    const res = await this.query(searchParams);
+
+    return res;
+  }
+
+  async query(searchParamsOverride = null) {
```
markdboyd, line 153: I think this default assignment to `null` is unnecessary. It would default to `undefined` if not specified and the logic in `Basesearch.query` would behave the same

packages/api/es/collections.js
```diff
@@ -76,12 +80,81 @@ class Collection extends BaseSearch {
...
+    const res = await super.query(searchParamsOverride);
 
     // get aggregations for results
-    const names = res.results.map((r) => r.name).filter((name) => (name));
-    res.results = await this.getStats(res.results, names);
+    res.results = await this.addStatsToCollectionResults(res.results);
```
markdboyd, line 157: So the point of this change is to filter out collections without a name?

packages/api/tests/endpoints/collections/active-collections.js
```diff
@@ -0,0 +1,161 @@
...
+      version: '1',
+      updatedAt: new Date(2020, 0, 29)
+    }), esAlias),
+    indexer.indexGranule(esClient, fakeGranuleFactoryV2({
+      updatedAt: new Date(2020, 1, 29),
+      collectionId: 'coll3___1'
+    }), esAlias)
+  ]);
+
+  stub.restore();
```
markdboyd, line 92: You should surround this `await` with a `try/finally` block and place `stub.restore` in the `finally` block since stubbing `Date.now` is probably used all over the place and you want to make the stub gets reverted before any operations in the test happen. Otherwise you might get tests failing/passing when they shouldn't

packages/api/tests/es/test-collections.js
```diff
@@ -0,0 +1,361 @@
...
+});
+
+test.serial('getStats correctly adds stats', async (t) => {
+  const collectionSearch = new Collection({}, null, process.env.ES_INDEX);
+  const stats = await collectionSearch.getStats([{ name: 'coll1', version: '1' }], ['coll1___1']);
+
+  t.is(stats.length, 1);
+  t.deepEqual(stats[0].stats, {
+    running: 0,
+    completed: 2,
```
markdboyd, line 137: So this works because `fakeGranuleFactoryV2` generates a granule with a `completed` status by default? Should we make that explicit for the test?

packages/api/tests/es/test-collections.js
```diff
@@ -0,0 +1,361 @@
...
+      version: '1',
+      updatedAt: new Date(2020, 0, 29)
+    }), esAlias),
+    indexer.indexGranule(esClient, fakeGranuleFactoryV2({
+      updatedAt: new Date(2020, 1, 29),
+      collectionId: 'coll3___1'
+    }), esAlias)
+  ]);
+
+  stub.restore();
```
markdboyd, line 85: Should also surround this stub and restore in try/finally

packages/api/tests/es/test-collections.js
```diff
@@ -0,0 +1,361 @@
...
+        running: 0,
+        completed: 1,
+        failed: 0,
+        total: 1
+      }
+    }
+  ]);
+});
+
+test.serial('queryCollectionsWithActiveGranules respects granule update times, but not collection', async (t) => {
```
markdboyd, line 350: The test proves what it says but I couldn't figure out why granule update times are respected and not collection

markdboyd: Ah, because `queryCollectionsWithActiveGranules` is actually searching for a type of `granule`, not `collection`

packages/api/es/collections.js
```diff
@@ -62,8 +63,11 @@ class Collection extends BaseSearch {
         total: 0
       };
 
+      // Can't aggregate on the _id but can on the _uid which is collection#_id
+      const esUid = `collection#${constructCollectionId(updatedRecord.name, updatedRecord.version)}`;
```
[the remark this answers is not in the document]
laurenfrederick: The _id is collection___version, which is great but you can't aggregate on it. Which reminds me, I'd love to upgrade ES in the IP sprint.

packages/api/es/collections.js
laurenfrederick: I had moved these lines out to a new function to add to stats that I was going to call separately. But then I went a different route and left it as a separate function because I like it for testing.

laurenfrederick: The function adds the collection stats to the returned result.

packages/api/tests/es/test-collections.js
laurenfrederick: I updated it to make it easier to understand

PR #1551, opened 2020-03-31
Use the full lodash package

yjpa7145:
Because the full lodash package is being installed as a dependency of other packages anyway, we aren't saving ourselves anything by using the `lodash.xxx` packages. This updates our code to rely on `lodash`. By importing lodash functions using `require('lodash/xxx')`, it keeps the webpack size small.


example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js
```diff
@@ -1,5 +1,5 @@
 const fs = require('fs');
-const difference = require('lodash.difference');
+const { difference } = require('lodash/difference');
```
nemreid, line 2: Shouldn't this be ...?
```suggestion
const difference = require('lodash/difference');
```

tasks/move-granules/index.js
```diff
@@ -146,7 +145,7 @@ async function moveFileRequest(
   };
 
   // the file moved to destination
-  const fileMoved = clonedeep(file);
+  const fileMoved = { ...file };
```
nemreid, line 148: Same note as the other file.

example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js
nemreid: The export is `export default difference`

.eslintrc.json
```diff
@@ -121,22 +121,14 @@
...
-    {
-      "files": ["**/bin/**/*.js"],
-      "rules": {
-        "no-console": "off"
-      }
+      "files": [
+        "**/bin/**/*.js",
+        "**/spec/**/*.js",
+        "packages/api/migrations/**/*.js",
+        "packages/deployment/**/*.js",
```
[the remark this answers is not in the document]
yjpa7145: Yup.

tasks/test-processing/index.js
```diff
@@ -14,11 +13,12 @@ async function uploadFakeBrowse(input) {
   input.granules.forEach((granule) => {
     granule.files.forEach((file) => {
       if (file.type === 'data') {
-        const browseFile = cloneDeep(file);
+        const browseFile = { ...file };
```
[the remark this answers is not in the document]
yjpa7145: Agreed, they do behave differently. In this case, though, that isn't a problem.

tasks/move-granules/index.js
yjpa7145: Yeah, I checked first and mad sure that this behaves as expected.

PR #1560, opened 2020-04-07
CUMULUS-1257: S3ProviderClient unit tests

nemreid:

Addresses [CUMULUS-1257: S3ProviderClient unit tests](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1257)

## Changes

* Increased test coverage of S3ProviderClient from 0% to 100%

### DIFF
--- a/packages/ingest/S3ProviderClient.js
+++ b/packages/ingest/S3ProviderClient.js
@@ -7,7 +7,7 @@ const isString = require('lodash/isString');
 const { basename, dirname } = require('path');
 
 class S3ProviderClient {
-  constructor({ bucket }) {
+  constructor({ bucket } = {}) {
     if (!isString(bucket)) throw new TypeError('bucket is required');
     this.bucket = bucket;
   }
--- a/packages/ingest/test/test-S3ProviderClient.js
+++ b/packages/ingest/test/test-S3ProviderClient.js
@@ -68,7 +68,7 @@ test.serial('S3ProviderClient.sync syncs a file to a target S3 location', async 
   const targetKey = 'target.json';
 
   await s3ProviderClient.sync(t.context.sourceKey, t.context.targetBucket, targetKey);
-  t.true(await S3.s3ObjectExists({ Bucket: t.context.targetBucket, Key: targetKey }));
+  t.is(await S3.getTextObject(t.context.targetBucket, targetKey), t.context.fileContent);
 });
 
 test.serial('S3ProviderClient.sync throws an error if the source file does not exist', async (t) => {
--- a/packages/ingest/test/test-S3ProviderClient.js
+++ b/packages/ingest/test/test-S3ProviderClient.js
@@ -1,9 +1,10 @@
 'use strict';
 
 const fs = require('fs');
-const { basename } = require('path');
+const { basename, dirname } = require('path');
 const test = require('ava');
 const S3 = require('@cumulus/aws-client/S3');
+const errors = require('@cumulus/errors');
 const { randomString } = require('@cumulus/common/test-utils');
 const S3ProviderClient = require('../S3ProviderClient');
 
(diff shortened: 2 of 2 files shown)


packages/ingest/S3ProviderClient.js
```diff
@@ -7,7 +7,7 @@ const isString = require('lodash/isString');
 const { basename, dirname } = require('path');
 
 class S3ProviderClient {
-  constructor({ bucket }) {
+  constructor({ bucket } = {}) {
```
nemreid, line 10: This change allows `new S3ProviderClient()` (as opposed to `new S3ProviderClient({})`) to throw the error instead of complaining about not being able to destructure undefined.

packages/ingest/test/test-S3ProviderClient.js
```diff
@@ -0,0 +1,81 @@
...
+
+test.serial('S3ProviderClient.list lists objects under a path in a bucket', async (t) => {
+  const s3ProviderClient = new S3ProviderClient({ bucket: t.context.sourceBucket });
+
+  const files = await s3ProviderClient.list(t.context.sourcePrefix);
+  t.is(files.length, 1);
+  t.is(files[0].name, basename(t.context.sourceKey));
+});
+
+test.serial('S3ProviderClient.downloads downloads a file to local disk', async (t) => {
```
markdboyd, line 58:
```suggestion
test.serial('S3ProviderClient.download downloads a file to local disk', async (t) => {
```

🐜

packages/ingest/test/test-S3ProviderClient.js
```diff
@@ -0,0 +1,81 @@
...
+
+  await s3ProviderClient.sync(t.context.sourceKey, t.context.targetBucket, targetKey);
+  t.is(await S3.getTextObject(t.context.targetBucket, targetKey), t.context.fileContent);
+});
+
+test.serial('S3ProviderClient.sync throws an error if the source file does not exist', async (t) => {
+  const s3ProviderClient = new S3ProviderClient({ bucket: t.context.sourceBucket });
+
+  await t.throwsAsync(
+    () => s3ProviderClient.sync('non-existent', t.context.targetBucket, 'target.json'),
```
markdboyd, line 78: There's nothing wrong with this - I'm just still not clear when throwAsync requires you to put your async function in an anonymous callback

packages/ingest/test/test-S3ProviderClient.js
```diff
@@ -0,0 +1,81 @@
...
+    () => new S3ProviderClient(),
+    'bucket is required'
+  );
+  t.throws(
+    () => new S3ProviderClient({}),
+    'bucket is required'
+  );
+});
+
+test.serial('S3ProviderClient.list lists objects under a path in a bucket', async (t) => {
```
markdboyd, line 50: Do we care about testing the case of listing objects at the root of the bucket?

packages/ingest/test/test-S3ProviderClient.js
```diff
@@ -0,0 +1,81 @@
...
+  await s3ProviderClient.sync(t.context.sourceKey, t.context.targetBucket, targetKey);
+  t.is(await S3.getTextObject(t.context.targetBucket, targetKey), t.context.fileContent);
+});
+
+test.serial('S3ProviderClient.sync throws an error if the source file does not exist', async (t) => {
+  const s3ProviderClient = new S3ProviderClient({ bucket: t.context.sourceBucket });
+
+  await t.throwsAsync(
+    () => s3ProviderClient.sync('non-existent', t.context.targetBucket, 'target.json'),
+    `Source file not found s3://${t.context.sourceBucket}/non-existent`
```
markdboyd, line 79: Would it be better to test against the `FileNotFound` error class? 🐜

packages/ingest/test/test-S3ProviderClient.js
nemreid: I could alternatively pass in `s3ProviderClient.sync.bind(null, 'non-existent', t.context.targetBucket, 'target.json');`. The point is to have a function that `throwsAsync` can call.

markdboyd: Yeah , I don't think this should change. I just get confused bc in some cases our `throwsAsync` calls pass in an async function directly, in other cases we wrap it in a function

For example:

https://github.com/nasa/cumulus/blob/master/packages/api/tests/lambdas/test-verifyProviderSecretsMigration.js#L139

packages/ingest/test/test-S3ProviderClient.js
nemreid: I guess I can add that.

packages/ingest/test/test-S3ProviderClient.js
nemreid: I see. So if you look at https://github.com/avajs/ava/blob/master/docs/03-assertions.md#throwsasyncthrower-expectation-message you'll see that `throwsAsync` accepts *either* of a function, *or* a promise.

Since calling an async function without `await` returns a promise, either way works.

PR #1562, opened 2020-04-09
Cumulus 1682 complexity 15

nemreid:
This PR is not the final PR as I am working on a complexity 10 branch, but it is a first installment meant to place a limiter on the current state and ensure we don't make it worse.

Addresses [CUMULUS-1682: Use ESLint complexity functionality to find highly complex code to simplify and simplify it.](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1682)

## Changes

* Enforce max 15 complexity
* whitelist very readable switch in ems-ingest-report
* pull checksum validation in parse-pdr into separate function

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


nemreid:
Turns out we used to have tests but due to a misunderstanding of what they were testing they were deleted. I've restored the old ingest package tests in the task package and moved the `parse-pdr` code in there because it has become very tightly coupled with that handler.

nemreid:
I hear you, but setting up test mocks that represent half-parsed PDRs feel really fragile to me. 🤔

PR #1565, opened 2020-04-10
CUMULUS-1852: Update POST error handling

markdboyd:
- Add error logging to POST endpoints
- Update docs for clarity


packages/api/endpoints/providers.js
```diff
@@ -55,25 +59,33 @@ async function get(req, res) {
...
-        await addToLocalES(record, indexProvider);
+    const data = req.body;
+    const id = data.id;
+
+    const providerModel = new Provider();
+
+    try {
+      // make sure the record doesn't exist
+      await providerModel.get({ id });
+      return res.boom.badReqest(`A record already exists for ${id}`);
```
laurenfrederick, line 71:
```suggestion
      return res.boom.badRequest(`A record already exists for ${id}`);
```

laurenfrederick: I know you didn't write that part, but if you could fix it that would be great and if you could put a test, that would be extra great.

packages/api/endpoints/rules.js
```diff
@@ -55,24 +59,32 @@ async function get(req, res) {
...
-      return res.send({ message: 'Record saved', record });
+      throw e;
     }
-    throw e;
+  } catch (e) {
+    if (e.name === 'SchemaValidationError') {
+      return res.boom.badRequest(e.message);
+    }
+    log.error('Error occurred while trying to create rule:', e);
+    return res.boom.badImplementation(e.message);
```
laurenfrederick, line 87: Returning `badImplementation` obscures the errors from the dashboard since for some reason it ignores the message. We should change this to another type or just throw the error.

packages/api/endpoints/providers.js
markdboyd: Wow, good catch

packages/api/endpoints/rules.js
markdboyd: Right - that's why I'm adding the `log.error` statements so that people will at least be able to see what went wrong in the Cloudwatch logs. I think `badImplementation` is what we want though, since a generic 500 error is appropriate when something unspecific went wrong on the server

laurenfrederick: The problem is the dashboard is not seeing the error and is unable to display it on the front end. We don't want to force people to have to go to the logs and see it when they should be able to see error details on the dashboard.

So I don't know why this one is the only one that doesn't send through the message: https://hapi.dev/module/boom/api/?v=9.1.0#boombadimplementationmessage-data---alias-internal

But maybe we could use expectationFailed to pass the message back to the dashboard? https://hapi.dev/module/boom/api/?v=9.1.0#boomexpectationfailedmessage-data

markdboyd: Expectation failed is a specific error code for when the conditions in an Expect header cannot be met

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/417

The closest thing to a "generic" error which would return the message to the user is 400/badRequest, but that is specifically for when the user's request is something invalid that the server can't handle. But we could get to this line because collection.create has a bug and throws an error, which would not be a bad request but a general server error. I think that's why Boom obscures the error message from the user for 5xx errors - because we don't want to leak implementation details in our error responses if the fault is in our code.

The genesis of this ticket was that Brian B didn't get any meaningful error from the API when creating a collection and he had an invalid `item.granuleIdExtraction`:

https://github.com/nasa/cumulus/blob/master/packages/api/models/collections.js#L43 

That throws a general Error, which is handled by the endpoint as returning a 500 error, but really it was a bad request. So maybe a better solution would be ensuring that anything during the record creation process which could be caused by a bad request is properly caught and returned to the user as a 400/badRequest? And then we keep 500/badImplementation with the logging for unexpected errors.

laurenfrederick: I'll have to test this to see what errors we're not handling then because I have seen this on the dashboard. i think then we get into a situation where we have to specifically handle every case, which I guess is not a bad thing. 

But it's making me feel like we need some kind of model error handler utility function where we could DRY 
```
if (e.name === 'SchemaValidationError') {
      return res.boom.badRequest(e.message);
    }
    log.error('Error occurred while trying to create rule:', e);
    return res.boom.badImplementation(e.message);
```
and not have this same code throughout our model endpoints.

markdboyd: What do you think about this?

019ac966f29242c9cf078bcf7ade78c125d9e1c2

It ensures that all the error cases are handled by the endpoint. It also didn't make sense to me that the model should throw a  "BadRequestError" error, so I created more specific error classes.

This handling would be easily to replicate for the other POST endpoints.

packages/api/endpoints/rules.js
```diff
@@ -55,24 +59,32 @@ async function get(req, res) {
...
-      if (inTestMode()) {
-        await addToLocalES(record, indexRule);
+    const data = req.body;
+    const name = data.name;
+
+    const model = new models.Rule();
+
+    try {
+      await model.get({ name });
+      return res.boom.conflict(`A record already exists for ${name}`);
```
laurenfrederick, line 70: We're doing `badRequest` in collections and providers. They should all be the same for consistency, whichever way you want to go.

packages/api/endpoints/rules.js
laurenfrederick: Yep I like that better. Thanks!

packages/api/endpoints/collections.js
```diff
@@ -102,12 +110,14 @@ async function post(req, res) {
...
   } catch (e) {
-    if (e.name === 'SchemaValidationError') {
-      return res.boom.badRequest(e.message);
-    }
-    if (e instanceof BadRequestError) {
+    if (
+      e.name === 'SchemaValidationError'
+      || e instanceof InvalidRegexError
+      || e instanceof UnmatchedRegexError
+    ) {
```
chuckwondo, line 117: I know this is not code that you've touched, but since you're in here, there are 2 other problems with the POST responses in this method:
1. line 93 should be changed from `notFound` to `badRequest` (400 is not appropriate because there was no attempt to find anything, rather the request is bad because it is missing information, and is thus badly formed)
2. line 99 should be changed from `badRequest` to `conflict` (409 [conflict] is standard practice for attempting to add something that already exists. 400 is not appropriate here because the request isn't "bad", meaning it is not an improperly formed request)

packages/api/endpoints/providers.js
```diff
@@ -55,25 +59,33 @@ async function get(req, res) {
...
-        await addToLocalES(record, indexProvider);
+    const data = req.body;
+    const id = data.id;
+
+    const providerModel = new Provider();
+
+    try {
+      // make sure the record doesn't exist
+      await providerModel.get({ id });
+      return res.boom.badRequest(`A record already exists for ${id}`);
```
chuckwondo, line 71: Use `conflict` instead of `badRequest`.

packages/api/endpoints/rules.js
chuckwondo: `conflict` is the appropriate way to go, not `badRequest`

packages/api/endpoints/rules.js
chuckwondo: Instead of using `badImplementation` which obscures the error from the dashboard, how about `serverUnavailable`?

markdboyd: It wouldn't make a difference:

> All 500 errors hide your message from the end user.

https://hapi.dev/module/boom/api/?v=9.1.0#http-5xx-errors

chuckwondo: Yep, makes sense. I'd keep it as `badImplementation` then, and the dashboard code has to address things properly. Not sure if it's the 500 responses that cause the dashboard to fail silently, but that's a matter of fixing the dashboard, I believe, not with changing the 500 response itself.

markdboyd: It's the 500 errors that are hiding the error message from the user. For 500 errors, the dashboard will only get a response that looks like:

```json
{
  "status": 500,
  "message": "Internal server error"
}
```

That's why we're trying to catch and handle any possible errors that shouldn't return a 500 response so the dashboard will get a helpful error message.

chuckwondo: Yep, I think we're in agreement here. I suppose I'm trying to clarify that sending a 500 response in the absence of being able to provide something more meaningful is perfectly valid.

markdboyd: OK cool. Sorry if I misunderstood 👍

packages/api/endpoints/providers.js
```diff
@@ -55,25 +59,33 @@ async function get(req, res) {
...
+        if (inTestMode()) {
+          await addToLocalES(record, indexProvider);
+        }
+        return res.send({ record, message: 'Record saved' });
       }
-      return res.send({ record, message: 'Record saved' });
+      throw e;
+    }
+  } catch (err) {
+    if (isBadRequestError(err)) {
```
markdboyd, line 84: I'm debating whether creating a helper to catch the common "bad request" errors is a good or a bad thing

packages/api/tests/endpoints/test-rules.js
```diff
@@ -305,6 +293,89 @@ test('POST returns a record exists when one exists', async (t) => {
...
+  const response = await request(app)
+    .post('/rules')
+    .set('Accept', 'application/json')
+    .set('Authorization', `Bearer ${jwtAuthToken}`)
+    .send(newRule)
+    .expect(400);
+  t.is(response.status, 400);
+});
+
+test.serial('POST returns a 500 response if workflow definition file does not exist', async (t) => {
```
markdboyd, line 339: I debated whether this should be a 400 error instead of 500. But ultimately a workflow definition file could be missing because of a deployment failure (or accidentally deleting files from S3) even if the request is perfectly valid, so 500 felt like the appropriate response

packages/api/tests/endpoints/test-rules.js
```diff
@@ -290,8 +276,10 @@ test('POST creates a rule', async (t) => {
   t.deepEqual(record, newRule);
 });
 
-test('POST returns a record exists when one exists', async (t) => {
-  const newRule = { ...testRule };
+test('POST returns a 409 response if record already exists', async (t) => {
+  const newRule = fakeRuleFactoryV2();
```
markdboyd, line 280: My goal with some of these changes to make these tests a little less inter-dependent

PR #1566, opened 2020-04-10
Build package docs

markdboyd:
Blocked by #1548 

- Add npm script to re-build auto-generated package API docs 
- Add step to release process for updating auto-generated package API docs


package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Cumulus Framework for ingesting and processing NASA Earth data streams",
   "scripts": {
     "audit": "audit-ci --config ./audit-ci.json && lerna exec -- audit-ci --config $(pwd)/audit-ci.json",
+    "docs-build-packages": "for d in packages/*; do cd $d && npm run build-docs && cd -; done;",
```
yjpa7145, line 7:
```suggestion
    "docs-build-packages": "lerna run build-docs",
```

yjpa7145: If we use `lerna run build-docs`, it will run it in the packages that have that command, and not run them in ones that don't. That way we won't need all of the `"build-docs": "true"` lines below.

PR #1567, opened 2020-04-11
CUMULUS-1853 Fix race condition in DiscoverGranules duplicate handling tests

yjpa7145:
Address a race condition in the `DiscoverGranules` duplicate handling tests. In the process, those two tests ("skip" and "error") were split into two more targeted test files. They were also updated to use only the Cumulus API.

**Other changes:**

- Added [API docs](https://github.com/nasa/cumulus/tree/CUMULUS-1853/packages/integration-tests#api) to the `@cumulus/integration-tests` README
- Added `@cumulus/integration-tests/collections.createCollection()`
- Added `@cumulus/integration-tests/executions.findExecutionArn()`
- Added `@cumulus/integration-tests/executions.getCompletedExecution()`
- Added `@cumulus/integration-tests/executions.getFailedExecution()`
- Added `@cumulus/integration-tests/granules.getCompletedGranule()`
- Added `@cumulus/integration-tests/providers.createProvider()`
- Added `@cumulus/integration-tests/rules.createOneTimeRule()`


packages/integration-tests/collections.js
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  if (createResponse.statusCode !== 200) {
+    throw new Error(`Failed to create collection: ${JSON.stringify(createResponse)}`);
+  }
+
+  return collection;
+};
+
+module.exports = {
```
markdboyd, line 48: If this is a function library, shouldn't the file be named `Collections.js`?

packages/integration-tests/collections.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+const CollectionsApi = require('@cumulus/api-client/collections');
+const { randomId } = require('@cumulus/common/test-utils');
+
+const buildCollection = (overrides = {}) => ({
```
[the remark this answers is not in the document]
markdboyd: This function also seems to have a lot of overlap with `fakeCollectionFactory`?

packages/integration-tests/executions.js
```diff
@@ -0,0 +1,143 @@
...
+
+      try {
+        execution = await executionsApi.getExecution(
+          pick(params, ['prefix', 'arn', 'callback'])
+        );
+      } catch (err) {
+        throw new pRetry.AbortError(err);
+      }
+
+      if (execution.status === 'failed') return execution;
```
markdboyd, line 117: I think this function and the one above should be merged and renamed to `getExecutionByStatus` or something. And add some logic like:

```
      if (execution.status === expectedStatus) {
        throw new pRetry.AbortError(
          new Error(`Execution ${params.arn} was ${execution.status}, not ${expectedStatus}`)
        );
      }
```

packages/integration-tests/granules.js
```diff
@@ -0,0 +1,60 @@
...
+ * @param {string} params.prefix - the prefix configured for the stack
+ * @param {string} params.granuleId - the granule id
+ * @param {Function} params.callback - an async function to invoke the api
+ *   lambda that takes a prefix / user payload. Defaults to
+ *   cumulusApiClient.invokeApifunction to invoke the api lambda
+ * @param {integer} params.timeout - the number of seconds to wait for the
+ *   granule to reach a terminal state. Defaults to 30.
+ * @returns {Promise<Object>} the completed granule
+ */
+const getCompletedGranule = async (params = {}) =>
```
markdboyd, line 21: Why not have this be generic for at least a failed or completed execution?

packages/integration-tests/package.json
```diff
@@ -12,7 +12,7 @@
   "publishConfig": {
     "access": "public"
   },
-  "homepage": "https://github.com/nasa/cumulus/tree/master/packages/integration-tests",
+  "homepage": "https://github.com/nasa/cumulus/tree/master/packages/integration-tests/README.md",
```
markdboyd, line 15: As far as I know, linking to a folder root on Github will display the README for that directory

packages/integration-tests/providers.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+const providersApi = require('@cumulus/api-client/providers');
+const { randomId } = require('@cumulus/common/test-utils');
+
+const buildProvider = (overrides = {}) => ({
```
markdboyd, line 6: `buildProvider` in the index should be deprecated/removed

packages/integration-tests/rules.js
```diff
@@ -0,0 +1,43 @@
...
+ * Build a onetime rule and create it using the Cumulus API
+ *
+ * See the `@cumulus/integration-tests` README for more information
+ *
+ * @param {string} prefix - the Cumulus stack name
+ * @param {Object} overrides - properties to set on the rule, overriding the
+ *   defaults
+ * @returns {Promise<Object>} the generated rule
+ */
+const createOneTimeRule = async (prefix, overrides = {}) => {
```
markdboyd, line 29: Why create a function specific to creating a "one-time" rule? Then you're going to have to create separate ones for SQS, Kinesis, etc? Why not have the type be an argument?

packages/integration-tests/collections.js
markdboyd: Same for all the other new files in this package in the PR

example/spec/parallel/discoverGranules/DiscoverGranulesDuplicateHandlingErrorSpec.js
```diff
@@ -0,0 +1,177 @@
...
+      );
+    } catch (err) {
+      beforeAllFailed = true;
+      console.error(err);
+      throw err;
+    }
+  });
+
+  afterAll(async () => {
+    await deleteS3Object(sourceBucket, existingGranuleKey).catch(noop);
```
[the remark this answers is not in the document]
yjpa7145: I can refactor it to use `pAll`, which has an option to not fail on the first exception. I'll just need to be careful because there _is_ an order to some of these operations. (You can't delete a collection with an associated rule, for example.)

example/spec/parallel/discoverGranules/DiscoverGranulesDuplicateHandlingErrorSpec.js
```diff
@@ -0,0 +1,177 @@
...
+        (execution) =>
+          get(execution, 'originalPayload.testExecutionId') === discoverGranulesRule.payload.testExecutionId,
+        { timeout: 15 }
+      );
+    } catch (err) {
+      beforeAllFailed = true;
+      console.error(err);
+      throw err;
+    }
+  });
```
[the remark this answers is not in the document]
yjpa7145: There were a few motivations here:

1. The original test file was actually testing two separate things: duplicate handling set to `skip` and duplicate handling set to `error`. Breaking those apart into two separate tests made it easier to see what was being tested, makes the tests run a bit faster since they can now run in parallel, and, most importantly, prevents one test from interfering with another.
2. The original test was using hard-coded granule IDs, collections, and providers. Refactoring it like this allowed for randomized test data.
3. Jonathan and I talked it over and agreed that the integration tests should be more from the user's perspective than they are now. Operators wouldn't be querying and writing to the DB, and they wouldn't be starting workflows through the AWS API, they would be doing it through the Cumulus API. By writing the tests at that level, we have already uncovered a number of bugs and issues.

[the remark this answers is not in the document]
yjpa7145: 4. The original test was using the `DiscoverGranules` workflow to populate the initial test granules into Cumulus. That eventually triggers `IngestGranule`, which actually creates the granules. It was more efficient to just invoke `IngestGranule` directly to stage that test data.

example/spec/parallel/discoverGranules/DiscoverGranulesDuplicateHandlingErrorSpec.js
```diff
@@ -0,0 +1,177 @@
...
+      existingGranuleId = randomId('existing-granule-');
+      existingGranuleKey = `${sourcePath}/${existingGranuleId}.txt`;
+      await s3PutObject({
+        Bucket: sourceBucket,
+        Key: existingGranuleKey,
+        Body: 'asdf'
+      });
+
+      // Ingest the existing granule
+      ingestGranuleRule = await createOneTimeRule(
```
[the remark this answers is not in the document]
yjpa7145: Reason number 2 in this comment: https://github.com/nasa/cumulus/pull/1567/files#r407560122

yjpa7145:
Replying to [this comment](https://github.com/nasa/cumulus/pull/1567#discussion_r407521143):

The names are the same, but they do slightly different things. I think that, as we refactor the integration tests to further "uniqify" the data and use the Cumulus API, `@cumulus/integration-tests.buildCollection()` will be phased out.

yjpa7145:
Replying to [this comment](https://github.com/nasa/cumulus/pull/1567#discussion_r407524043):

I agree that there's some overlap there that we may want to consider consolidating. I've made a note to look into it but, if we make that change, I think it should be a separate PR to keep it from exploding the size of this PR.

docs/data-cookbooks/setup.md
```diff
@@ -82,6 +82,7 @@ An example of an SQS rule configuration is [here](https://github.com/nasa/cumulu
 |workflow|`"CNMExampleWorkflow"`|Yes|Name of the workflow to be run. A list of available workflows can be found on the Workflows page|
 |provider|`"PODAAC_SWOT"`|No|Configured provider's ID. This can be found on the Providers dashboard page|
 |collection|`<JSON Object>` collection object shown [below](#collection-object)|Yes|Name and version of the collection this rule will moderate. Relates to a collection configured and found in the Collections page|
+|payload|`<JSON>`|No|The payload to be passed to the workflow|
```
markdboyd, line 85:
```suggestion
|payload|`<JSON Object>`|No|The payload to be passed to the workflow|
```

Just for consistency with the other entries in this table

example/spec/parallel/discoverGranules/DiscoverGranulesDuplicateHandlingSkipSpec.js
```diff
@@ -0,0 +1,206 @@
...
+        { timeout: 15 }
+      );
+
+      // Get the completed "DiscoverGranules" execution
+      finishedDiscoverGranulesExecution = await getCompletedExecution({
+        prefix,
+        arn: discoverGranulesExecutionArn
+      });
+    } catch (err) {
+      beforeAllFailed = true;
```
markdboyd, line 148: It would probably be preferable to set `beforeAllFailed = err` and then call `fail(beforeAllFailed)` so that we know why the test failed from the beforeAll

packages/integration-tests/package.json
```diff
@@ -12,13 +12,14 @@
...
   "repository": {
     "type": "git",
     "url": "https://github.com/nasa/cumulus"
   },
   "scripts": {
     "build": "rm -rf dist && mkdir dist && webpack",
+    "build-docs": "jsdoc2md --template README.hbs Collections.js Executions.js Granules.js Providers.js Rules.js > README.md",
```
markdboyd, line 22: Genuine question: What do you think about generating the docs to `docs/API.md`? I ask because I started doing this in other packages since we are linting the markdown in our package READMEs, but not in subdirectories of the packages. And I worry that jsdoc2md may produce invaild markdown

yjpa7145:
Replying to [this comment](https://github.com/nasa/cumulus/pull/1567#discussion_r407697552):

The problem with putting the error in the `fail()` call is that it is going to cause the error to be printed for every test. As written, it should only report the exception at the end of the test output.

packages/integration-tests/package.json
yjpa7145: The big advantage of having the API docs in the readme is that they also show up when you view the package on npm's website. I checked just now and confirmed that, at least so far, the docs being generated by `jsdoc2md` do pass our markdown linter.

packages/integration-tests/Granules.js
```diff
@@ -0,0 +1,68 @@
...
+ * @param {string} params.granuleId - the `granuleId` of the granule
+ * @param {Function} [params.callback=cumulusApiClient.invokeApifunction] - an async function to
+ * invoke the API Lambda that takes a prefix / user payload
+ * @param {integer} [params.timeout=30] - the number of seconds to wait for the
+ *   execution to reach a terminal state
+ * @returns {Promise<Object>} the granule as returned by the `GET /granules/<granule-id>` endpoint
+ *
+ * @alias module:Granules
+ */
+const getCompletedGranule = async (params = {}) =>
```
markdboyd, line 29: Should we refactor this to be generic for a completed or failed granule?

PR #1573, opened 2020-04-15
Adds documentation on serving Dashboard from CloudFront.

flamingbear:
----

#

Addresses [CUMULUS-1817: Dashboard served from Cloudfront only seems to work when SERVED_BY_CUMULUS_API is true](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1817)

## Changes

* Updates Documentation only.

### DIFF
--- a/docs/operator-docs/serve-dashboard-from-cloudfront.md
+++ b/docs/operator-docs/serve-dashboard-from-cloudfront.md
@@ -1,0 +1,26 @@
+---
+id: serve-dashboard-from-cloudfront
+title: Serve Dashboard with CloudFront
+hide_title: true
+---
+
+# Serve Cumulus Dashboard behind CloudFront
+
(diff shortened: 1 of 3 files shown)


docs/operator-docs/serve-dashboard-from-cloudfront.md
```diff
@@ -0,0 +1,26 @@
+---
+id: serve-dashboard-from-cloudfront
+title: Serve Dashboard with CloudFront
+hide_title: true
+---
+
+# Serve Cumulus Dashboard behind CloudFront
```
laurenfrederick, line 7: I think NASD/NGAP/Platform specific documentation should go into the onboarding documentation in the wiki, especially since we already link back to the wiki in a few places here. As part of the epic next PI I am going to have us move Platform specific docs to the internal wiki.

PR #1579, opened 2020-04-21
Make rule type error more descriptive

laurenfrederick:

Addresses [CUMULUS-1812](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1812)

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/packages/api/models/rules.js
+++ b/packages/api/models/rules.js
@@ -254,7 +254,7 @@ class Rule extends Manager {
       newRuleItem = await this.validateAndUpdateSqsRule(newRuleItem);
       break;
     default:
-      throw new Error(`Rule type ${newRuleItem.rule.type} not supported.`);
+      throw new Error(`Rule type \'${newRuleItem.rule.type}\' not supported.`);
     }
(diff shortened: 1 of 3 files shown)


packages/api/models/rules.js
```diff
@@ -254,7 +254,7 @@ class Rule extends Manager {
       newRuleItem = await this.validateAndUpdateSqsRule(newRuleItem);
       break;
     default:
-      throw new Error('Type not supported');
+      throw new Error(`Rule type ${newRuleItem.rule.type} not supported.`);
```
chuckwondo, line 257: My only suggestion would be to put the type quotes (single or double) within the message to avoid having it look like a regular word in the message. For example: `Rule type "foo" not supported`. This also covers the empty string, which is not noticeable without quotes.

laurenfrederick: @chuckwondo I made these updates.

PR #1580, opened 2020-04-21
Migrate logger package to Typescript

yjpa7145:
Part of CUMULUS-1883


package.json
```diff
@@ -13,11 +13,12 @@
...
     "lint": "eslint . && lerna run python-lint",
     "lint-md": "markdownlint docs/**/*.md docs/*.md tf-modules/**/*.md tf-modules/*.md",
-    "test": "nyc lerna run --concurrency 1 test",
-    "bootstrap": "lerna bootstrap --no-ci --force-local",
-    "bootstrap-no-build": "lerna bootstrap --no-ci --force-local --ignore-scripts",
-    "bootstrap-no-build-no-concurrency": "lerna bootstrap --no-ci --force-local --ignore-scripts --concurrency 1",
-    "bootstrap-no-build-quiet": "lerna bootstrap --loglevel=error --no-ci --force-local --ignore-scripts",
+    "test": "npm run tsc && nyc lerna run --concurrency 1 test",
+    "tsc": "lerna run tsc",
+    "bootstrap": "lerna bootstrap --no-ci --force-local --ignore-scripts && lerna run tsc && lerna run prepublish && lerna run prepare",
```
markdboyd, line 18: Why is bootstrap now doing `prepublish` and `prepare` as well? I thought `lerna bootstrap` ran those anyway

packages/logger/package.json
```diff
@@ -10,6 +10,11 @@
   "engines": {
     "node": ">=10.16.3"
   },
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.js",
```
markdboyd, line 14: Interesting. I thought the types file had to have a `.ts` extension.

packages/logger/src/index.ts
```diff
@@ -0,0 +1,177 @@
...
+  granules?: string,
+  parentArn?: string,
+  pretty?: boolean,
+  sender?: string,
+  stackName?: string,
+  version?: string
+}
+
+class Logger {
+  readonly #asyncOperationId: string | undefined;
```
markdboyd, line 19: I'm assuming this a convention for read-only properties?

packages/logger/tests/index.js
```diff
@@ -2,7 +2,7 @@
 
 const { Console } = require('console');
 const { Writable } = require('stream');
-const isString = require('lodash/isString');
+const isString = require('lodash.isstring');
 const moment = require('moment');
 const test = require('ava');
```
markdboyd, line 7: I spent some time looking just now but couldn't figure out if Ava supports running your TS code directly rather than relying on the compiled JS code for tests

https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md

markdboyd: Actually, does the line below this still work..?

`const Logger = require('..');`

Just curious since we're no longer compiling `index.js` in the root but instead `dist/index.js`

packages/logger/src/index.ts
```diff
@@ -0,0 +1,177 @@
...
+    this.#sender = options.sender || 'unknown';
+    this.#stackName = options.stackName;
+    this.#console = options.console || global.console;
+    this.#version = options.version;
+  }
+
+  /**
+   * Log a debug message
+   *
+   * @param {...any} [messageArgs] - the message to log
```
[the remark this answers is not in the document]
yjpa7145: It actually shouldn't be anything at all. `tsdoc` doesn't include the type in the doc string, since it's now part of the actual code. I'll fix that.

packages/logger/tests/index.js
```diff
@@ -2,7 +2,7 @@
 
 const { Console } = require('console');
 const { Writable } = require('stream');
-const isString = require('lodash/isString');
+const isString = require('lodash.isstring');
```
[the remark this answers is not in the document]
yjpa7145: I've refactored the test so that package is no longer used.

packages/logger/tests/index.js
yjpa7145: > I spent some time looking just now but couldn't figure out if Ava supports running your TS code directly rather than relying on the compiled JS code for tests

Ava runs against the compiled `.js` file that is generated. If we wanted better TS support in our tests, [Jest](https://jestjs.io/) seems to be what almost everyone recommends.

yjpa7145: > Actually, does the line below this still work..?
> 
> `const Logger = require('..');`
> 
> Just curious since we're no longer compiling `index.js` in the root but instead `dist/index.js`

Yeah, I was curious about that too. It does work, so `require()` must be respecting the `main` that we're setting in `package.json`.

packages/logger/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
```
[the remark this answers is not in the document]
yjpa7145: That option is being set here: https://github.com/nasa/cumulus/blob/CUMULUS-1883-migrate-logger-package/tsconfig.json#L6

We do want to publish that `.d.ts` file (and should with the current configuration). We don't want to commit it, though, since it's generated from the `.ts` files.

packages/logger/src/index.ts
Jkovarik: And re: #privateVariable .... neat.

Jkovarik: Adding a comment as I want to re-read the weakmap private variable oddness vs update  to make sure I'm not missing something here.

packages/logger/src/index.ts
```diff
@@ -0,0 +1,151 @@
...
+      );
+    } else {
+      this.writeLogEvent('error', messageArgs);
+    }
+  }
+
+  /**
+   * Log a fatal message
+   */
+  fatal(...messageArgs: any[]) {
```
Jkovarik, line 81: Why is this set as any?  The interfacing looks similar to the other non-error methods, including the map-through from common/log 🤔

packages/logger/package.json
```diff
@@ -19,8 +24,9 @@
     "url": "https://github.com/nasa/cumulus"
   },
   "scripts": {
-    "build": "true",
-    "test": "ava"
+    "tsc": "tsc",
+    "prepare": "npm run tsc",
+    "test": "npm run tsc && ava"
```
Jkovarik, line 29: I half wonder if we should follow a pattern of test explicitly running prepare.... except that's not great for more complicated packages. 

I suppose running tsc isn't *too* bad on test.... something like Jest to avoid it is probably not what we want to do now.

packages/logger/.npmignore
```diff
@@ -1 +0,0 @@
-/tests/
```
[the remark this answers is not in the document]
Jkovarik: @mh81  Presumably this is one of the packages we *do* still want to publish?

packages/api/package.json
```diff
@@ -22,7 +22,7 @@
     "serve-dist-remote": "cumulus-api serve-dist",
     "reset-tables": "LOCALSTACK_HOST=localhost NODE_ENV=test FAKE_AUTH=true cumulus-api reset-tables",
     "link-api-bin": "sh ./bin/link-api-bin.sh",
-    "prepare": "npm run build && npm run link-api-bin && npm run build-lambda-zips",
+    "webpack": "npm run build && npm run link-api-bin && npm run build-lambda-zips",
```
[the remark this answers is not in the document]
Jkovarik: Feel free to tell me I'm crazy and webpack is good.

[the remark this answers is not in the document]
yjpa7145: No, I think this really should be in a "publish" script, or something like that.

packages/logger/package.json
yjpa7145: I think it should probably be taken out, so that people can use nodemon or something similar to build their .js files as they make changes to .ts, rather than running tsc before every test run. I'll remove this.

packages/logger/src/index.ts
yjpa7145: I was thinking about it last night, and I think that I'm going to go with the `private varName: string` instead of `#varName: string`. Typescript compiles them both to the same Javascript, but "private" is more likely to pass the two beers test rather than the less obvious `#`.

packages/logger/src/index.ts
yjpa7145: They should all actually be `any` because [that is what `format` accepts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/util.d.ts#L8). I'll fix the inconsistency.

packages/api/package.json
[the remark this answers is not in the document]
markdboyd: What is the motivation for not leaving it as `prepare`?

packages/logger/.npmignore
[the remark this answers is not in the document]
yjpa7145: There are a couple of ways to specify what files should or should not be included in your npm package. One way is to use a `.npmignore` file, which will add everything to the npm package except what is in `.npmignore`. Another option is to use the files property in your `package.json` file ([documented here](https://docs.npmjs.com/files/package.json#files)). Rather than adding everything by default and not including the stuff in `.npmignore`, using the `files` property will include nothing _except_ for what's listed in `files`. That seemed like a much more reliable way for us to only include the stuff that we care about.

To answer @Jkovarik's question, yes, this package is still being published.

packages/logger/tsconfig.json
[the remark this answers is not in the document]
markdboyd: I don't have strong feelings - but why not commit it? Even if it is generated, it will only change when changes are made in this package

packages/logger/src/index.ts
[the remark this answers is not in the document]
markdboyd: Hmm, I'm thinking this isn't going to be compatible with our jsdoc-to-markdown approach

packages/api/package.json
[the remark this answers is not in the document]
yjpa7145: When one of the packages that uses Typescript is installed locally (when we run `lerna bootstrap`), we need to compile those files into `.js` and `.d.ts`. We also need to make sure that compilation is done in the right order, so that a package like `aws-client` is compiled after something it depends on, like `errors`. Using the `prepare` script was the right place to do that. (In addition, the webpages I've read about using Typescript to publish npm packages, [like this one](https://itnext.io/step-by-step-building-and-publishing-an-npm-typescript-package-44fe7164964c), all recommend that `prepare` is the place to compile your Typescript.

Up until now, we've been using the `prepare` script to run webpack in packages that use it. That isn't really something that we want to do in the `prepare` stage, though, but really something that we want to do before publishing or deploying. Moving webpack out of `prepare` makes sense, but I don't think we've found the right place to actually put it yet, though.

packages/logger/tests/index.js
markdboyd: Interesting. If I open it up in VScode and click on the `require('..')` to open the file, it opens the type definition file. This is how most published packages behave, so that's fine

markdboyd: However, while it is true that Ava runs the tests against the compiled JS files, that experience is not great. The compiled JS files aren't really human readable and putting breakpoints in there isn't really helpful because it's not the actual source code. It might be worth exploring this a little further

packages/api/package.json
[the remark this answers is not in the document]
markdboyd: Fair enough

packages/logger/tests/index.js
markdboyd: It turns out that being able to set breakpoints in the TS code is as easy as enabling sourcemaps for the Typescript compilation. But I assume we only want this for local development. 

Is there any clever way to do `tsc --sourceMap` only for local development?

Jkovarik: Possibly if we use an env variable or an alternate package.json devel script chain (e.g. bootstrap-devel) that enables it.   

🤔

yjpa7145: We can build the source maps using `"sourceMap": true` in the `tsconfig.json` file, then update the `files` section of `package.json` to:

```json
"files": [
  "./dist/**/*.js",
  "./dist/**/*.d.ts",
],
```

packages/logger/src/index.ts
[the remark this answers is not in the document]
yjpa7145: Yeah, I don't have a good answer for that. I've done a little research on parsing tsdoc into md, but haven't found a solution.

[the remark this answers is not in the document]
markdboyd: This seems promising?

https://typedoc.org/

package.json
```diff
@@ -9,15 +9,18 @@
...
     "install-locks": "lerna exec --no-sort -- npm i --package-lock-only --no-audit",
-    "lint": "eslint . && lerna run python-lint",
+    "lint": "npm run eslint && lerna run python-lint && npm run lint-md",
     "lint-md": "markdownlint docs/**/*.md docs/*.md tf-modules/**/*.md tf-modules/*.md",
-    "test": "nyc lerna run --concurrency 1 test",
-    "bootstrap": "lerna bootstrap --no-ci --force-local",
-    "bootstrap-no-build": "lerna bootstrap --no-ci --force-local --ignore-scripts",
-    "bootstrap-no-build-no-concurrency": "lerna bootstrap --no-ci --force-local --ignore-scripts --concurrency 1",
-    "bootstrap-no-build-quiet": "lerna bootstrap --loglevel=error --no-ci --force-local --ignore-scripts",
+    "test": "npm run tsc && nyc lerna run --concurrency 1 test",
```
yjpa7145, line 17: TODO Don't run tsc here

PR #1581, opened 2020-04-22
Validate rule before kicking off workflows or creating event source mappings

laurenfrederick:
Addresses [CUMULUS-1854: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1854)

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


packages/api/tests/models/test-rules-model.js
```diff
@@ -358,6 +362,23 @@ test.serial('Creating a kinesis rule where an event source mapping already exist
   }
 });
 
+test.serial('creating an invalid kinesis type rule does not add event mappings', async (t) => {
+  const { kinesisRule } = t.context;
+
+  const newKinesisRule = cloneDeep(kinesisRule);
```
markdboyd, line 368: Nit: you shouldn't need to `cloneDeep` because `t.context.kinesisRule` is reset in `beforeEach` for every test

packages/api/models/rules.js
```diff
@@ -228,6 +228,12 @@ class Rule extends Manager {
       newRuleItem.state = 'ENABLED';
     }
 
+    newRuleItem.createdAt = Date.now();
+    newRuleItem.updatedAt = Date.now();
+
+    // Validate rule before kicking off workflows or adding event source mappings
+    await this.constructor.recordIsValid(newRuleItem, this.schema, this.removeAdditional);
```
markdboyd, line 235: Super, super good change

packages/api/tests/models/test-rules-model.js
```diff
@@ -358,6 +362,23 @@ test.serial('Creating a kinesis rule where an event source mapping already exist
   }
 });
 
+test.serial('creating an invalid kinesis type rule does not add event mappings', async (t) => {
```
markdboyd, line 365: Do we care about having the same test to ensure SNS triggers aren't created for invalid rules? I could go either way

laurenfrederick: My thinking was if we tested one that was good, but what do you think? I mean I could go either way too, so took the easy way out.

packages/api/tests/models/test-rules-model.js
laurenfrederick: Oh yeah that is a good point. Habit I guess!

packages/api/tests/models/test-rules-model.js
markdboyd: I guess I incline towards being thorough but it does seem repetitive. Alternatively, if all the code in the rules model from lines 237 - 264 was in a separate (private) method, then we could just test using sinon that the method is never called when you try to create an invalid rule.

More importantly, we probably also want to explicitly check that the rule is valid in `rule.update` before updating rule triggers?

laurenfrederick: Yep, like to all of that. Took your suggestions.

packages/api/models/rules.js
```diff
@@ -134,50 +134,61 @@ class Rule extends Manager {
...
-      await this.addRule(updatedRuleItem, payload);
-      break;
-    }
-    case 'kinesis':
-      if (valueUpdated) {
-        await this.deleteKinesisEventSources(updatedRuleItem);
-        const updatedRuleItemArns = await this.addKinesisEventSources(updatedRuleItem);
-        updatedRuleItem = this.updateKinesisRuleArns(updatedRuleItem,
-          updatedRuleItemArns);
+      case 'scheduled': {
```
markdboyd, line 154: When I check this out locally, ESlint is complaining about the spaces, though I don't know why

packages/api/tests/models/test-rules-model.js
```diff
@@ -31,6 +31,18 @@ async function getKinesisEventMappings() {
...
 }
 
+async function deleteKinesisEventSourceMappings() {
+  const eventMappings = await getKinesisEventMappings();
+
+  if(!eventMappings) {
+    return Promise.resolve();
+  }
+
+  const allEventMappings = eventMappings[0].EventSourceMappings.concat(eventMappings[1].EventSourceMappings);
```
markdboyd, line 41: Eslint is complaining about the length of this line

packages/api/tests/models/test-rules-model.js
```diff
@@ -31,6 +31,18 @@ async function getKinesisEventMappings() {
...
+
+  return Promise.all(allEventMappings.map((e) => awsServices.lambda().deleteEventSourceMapping({ UUID: e.UUID }).promise()));
```
markdboyd, line 43: Eslint is complaining about the length of this line

packages/api/models/rules.js
```diff
@@ -228,39 +239,51 @@ class Rule extends Manager {
...
     // save
     return super.create(newRuleItem);
   }
 
+  async createRuleTrigger(ruleItem) {
+    let newRuleItem = cloneDeep(ruleItem);
+
+    const payload = await Rule.buildPayload(newRuleItem);
+    switch (newRuleItem.rule.type) {
+      case 'onetime': {
```
markdboyd, line 259: Weirdly, ESlint seems to think that case statements should not be indented more than the switch statement.

packages/api/tests/models/test-rules-model.js
```diff
@@ -149,10 +161,27 @@ test('Creating a rule with an invalid type throws an error', async (t) => {
...
   await t.throwsAsync(
     () => rulesModel.create(ruleItem),
-    { message: 'Rule type \'invalid\' not supported.' }
+    { name: 'SchemaValidationError' }
   );
 });
 
+test.serial('Creating an invalid rule does not create workflow triggers', async (t) => {
+  const { onetimeRule } = t.context;
+  const ruleItem = cloneDeep(onetimeRule);
```
markdboyd, line 170: Nit: You don't need to cloneDeep

packages/api/tests/models/test-rules-model.js
```diff
@@ -149,10 +161,27 @@ test('Creating a rule with an invalid type throws an error', async (t) => {
...
+
+  const createTriggerStub = sinon.stub(models.Rule.prototype, 'createRuleTrigger').resolves(ruleItem);
+
+  await t.throwsAsync(
+    () => rulesModel.create(ruleItem),
+    { name: 'SchemaValidationError' }
+  );
+
+  t.true(createTriggerStub.notCalled);
+  createTriggerStub.restore();
```
markdboyd, line 182:
```suggestion
  const createTriggerStub = sinon.stub(models.Rule.prototype, 'createRuleTrigger').resolves(ruleItem);

  try {
    await t.throwsAsync(
      () => rulesModel.create(ruleItem),
      { name: 'SchemaValidationError' }
    );
    t.true(createTriggerStub.notCalled);
  } finally {
    createTriggerStub.restore();
  }
```

We want to make sure that this stub gets restored even if the test fails on `throwsAsync`, so that the stub isn't still active when other tests run. We also want to do this for the `update` test

packages/api/tests/models/test-rules-model.js
```diff
@@ -311,6 +362,7 @@ test.serial('create a kinesis type rule, using existing event source mappings',
 
   await rulesModel.delete(rule);
   await rulesModel.delete(newRule);
+  await deleteKinesisEventSourceMappings();
```
markdboyd, line 365: I don't doubt that it's necessary, but why did you add these to most of the tests? To ensure test isolation?

markdboyd:
@laurenfrederick Looks great! Just a few small cleanup comments

packages/api/tests/models/test-rules-model.js
laurenfrederick: Yes, when I was writing the test the other way this was needed and I figured there's no reason to go delete it since it's actually a good change to isolate the tests and may help someone down the road.

packages/api/tests/models/test-rules-model.js
chuckwondo: I'm curious about this. Why is `cloneDeep` not necessary?

I see that ava does attempt to isolate data between tests (https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#test-context).

However, it's not clear that a deep clone is performed on the context. Specifically:

> Context created in .before() hooks is cloned before it is passed to .beforeEach() hooks and / or tests. The .after() and .after.always() hooks receive the original context value.
>
> For .beforeEach(), .afterEach() and .afterEach.always() hooks the context is not shared between different tests, allowing you to set up data such that it will not leak to other tests.

The first sentence links to lodash's shallow clone (see the ava docs at the link above), but the second sentence seems to _imply_ that some sort of deep clone is happening, but it is not clear as to what actually happens.

markdboyd: `t.context.onetimeRule` is created in `beforeEach`. And as the docs you posted say, context set up in beforeEach is not shared between tests. So there's no need to clone deep because `t.context.onetimeRule` will already be a new object in each test

chuckwondo: Oh, I see. That makes sense now. Thanks!

PR #1582, opened 2020-04-23
Migrate the errors package to Typescript

yjpa7145:


packages/errors/src/index.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export const UnparsableFileLocationError = createErrorType('UnparsableFileLocationError');
+
+export const RecordDoesNotExist = createErrorType('RecordDoesNotExist');
+
+export const InvalidRegexError = createErrorType('InvalidRegexError');
+
+export const UnmatchedRegexError = createErrorType('UnmatchedRegexError');
+
+export const ValidationError = createErrorType('ValidationError');
```
markdboyd, line 122: So it doesn't matter if we write our TS using export/import syntax because it compiles to JS using CommonJS style? I actually prefer export/import syntax, but is confusing to have different styles in different places?

packages/errors/src/index.ts
```diff
@@ -0,0 +1,122 @@
+/**
+ * Creates a new error type with the given name and parent class. Sets up
+ * boilerplate necessary to successfully subclass Error and preserve stack trace
+ * @param name - The name of the error type
+ * @param ParentType - The error that serves as the parent
+ * @returns The new type
+ */
+export const createErrorType = (
+  name: string,
+  ParentType: new (message: string) => Error = Error
```
markdboyd, line 10: Whoa. Now this I can't quite follow

packages/errors/src/index.ts
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Returns true if the error is a resource error.
+ */
+export const isWorkflowError = (error: Error) => error.name.includes('WorkflowError');
+
+/**
+ * Returns true if the error is a DynamoDB conditional check exception.
+ */
+export const isConditionalCheckException = (error: { code?: string }) =>
```
markdboyd, line 40: Use the `ErrorWithOptionalCode` interface here

packages/errors/src/index.ts
yjpa7145: It actually does matter, but in a good way. Typescript uses the export / import syntax only. It will then compile it to the module format that you specify in your `tsconfig.json` file.

packages/errors/src/index.ts
yjpa7145: Just wait, things are going to get weirder the higher up the stack we go.

yjpa7145: Here are the docs for what I'm doing here: https://www.typescriptlang.org/v2/docs/handbook/generics.html

Basically, I'm saying that the `ParentType` argument should be a type that is a constructor function (a class) that takes a single argument that's a string and returns an instance of an Error. The default is set to Error.

packages/errors/src/index.ts
yjpa7145: Fixed, good catch.

packages/errors/src/index.ts
yjpa7145: Incidentally, [Typescript Weekly](https://www.typescript-weekly.com/) just included an article about this: https://2ality.com/2020/04/classes-as-values-typescript.html

PR #1583, opened 2020-04-23
Migrate the checksum package to Typescript

yjpa7145:


packages/checksum/src/cksum.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module 'cksum' {
```
markdboyd, line 1: I presume we're doing this because there is no published typing for the cksum module?

packages/checksum/src/checksum.ts
```diff
@@ -0,0 +1,63 @@
...
+  options: TransformOptions
+): Promise<number | string> {
+  if (algorithm.toLowerCase() === 'cksum') {
+    return _getCksumFromStream(stream);
+  }
+
+  return _getChecksumFromStream(algorithm, stream, options);
+}
+
+export default generateChecksumFromStream;
```
markdboyd, line 63: So it is both exported as a function and the default export? Is this compatible with people importing it elsewhere as `const { generateChecksumFromStream } = require('...`

markdboyd: I guess it is since the tests are passing

packages/checksum/src/checksum.ts
```diff
@@ -0,0 +1,63 @@
+import * as cksum from 'cksum';
+import * as crypto from 'crypto';
+import { Readable, TransformOptions } from 'stream';
+
+/**
+ * Get file checksum (cksum) from readable stream
+ *
+ * @param stream - A readable file stream
+ *
+ * @returns The file checksum
```
markdboyd, line 10: Nit: It's a promise returning the file checksum if we want to be consistent

packages/checksum/src/checksum.ts
markdboyd: Hmm, but actually it seems like we're importing through the module index everywhere like so

`const { generateChecksumFromStream } = require('@cumulus/checksum');`

Is there a reason to export it both as a function and the default export?

packages/checksum/src/validate.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Validate expected checksum against calculated checksum
+ *
+ * @param algorithm - Checksum algorithm
+ * @param stream - A readable file stream
+ * @param expectedSum - expected checksum
+ * @param options - Checksum options
+ *
+ * @returns whether expectedSum === calculatedSum
```
markdboyd, line 12: Also a promise

packages/checksum/src/cksum.d.ts
yjpa7145: Correct.

packages/checksum/src/checksum.ts
yjpa7145: Not really. I'm inclined to remove the `default export`, since who knows, we could add something to that file later that's a more appropriate default.

.eslintrc.json
```diff
@@ -125,18 +126,19 @@
...
         "@typescript-eslint/no-throw-literal": "off",
         "comma-dangle": ["warn", "never"],
+        "import/first": "off",
+        "import/prefer-default-export": "off",
         "lines-between-class-members": "off",
+        "lodash/prefer-lodash-typecheck": "off",
         "node/no-unsupported-features/es-syntax": "off",
-        "valid-jsdoc": "off",
-        "tsdoc/syntax": "warn"
+        "unicorn/no-abusive-eslint-disable": "off"
```
markdboyd, line 141: What's the reason for this change?

.eslintrc.json
```diff
@@ -125,18 +126,19 @@
...
       ],
       "extends": ["airbnb-typescript/lib/shared"],
       "rules": {
         "@typescript-eslint/no-implied-eval": "off",
         "@typescript-eslint/no-throw-literal": "off",
         "comma-dangle": ["warn", "never"],
+        "import/first": "off",
+        "import/prefer-default-export": "off",
         "lines-between-class-members": "off",
+        "lodash/prefer-lodash-typecheck": "off",
```
markdboyd, line 139: If we're going to require lodash everywhere, don't we want to encourage its use?

.eslintrc.json
```diff
@@ -125,18 +126,19 @@
...
-        "@typescript-eslint",
-        "eslint-plugin-tsdoc"
+        "@typescript-eslint"
       ],
       "extends": ["airbnb-typescript/lib/shared"],
       "rules": {
         "@typescript-eslint/no-implied-eval": "off",
         "@typescript-eslint/no-throw-literal": "off",
         "comma-dangle": ["warn", "never"],
+        "import/first": "off",
```
markdboyd, line 136: This seems useful, why disable it?

packages/checksum/templates/README.md.hbs
```diff
@@ -0,0 +1,33 @@
+# @cumulus/checksum
```
markdboyd, line 1: I like having our API documented in the README, but I kinda dislike that the README contents get hidden in this file and it's less obvious for future package maintenance where to update the README

.eslintrc.json
yjpa7145: We aren't requiring `lodash`, but we are using it when it's available and when it makes sense. This rule was forcing us to do things like `isUndefined(x)` instead of `x === undefined`, which isn't necessary.

.eslintrc.json
yjpa7145: I believe it was because it was complaining about the order of `import x from 'x';` vs `import y = require('y');`, which shouldn't matter. I re-enabled the rule and didn't get any errors this time, so I must have refactored out the code that it was complaining about. I'll leave it re-enabled for the time being.

.eslintrc.json
yjpa7145: I forget what the issue was but, again, I seem to have refactored it out. I _believe_ it was returning a false positive if you disabled an eslint rule for a single line, and that line was the first line of a file.

packages/checksum/templates/README.md.hbs
yjpa7145: In addition to the "Generated automatically using `npm run build-docs`" comment at the bottom of the README.md, we could add "from `templates/README.md.hbs`?

markdboyd: I think this bears a larger discussion but I don't have a great answer and it's not worth holding up this PR on

PR #1584, opened 2020-04-23
Migrate aws-client package to TS

yjpa7145:


packages/aws-client/src/DynamoDb.ts
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * Call DynamoDb client get
+ *
+ * See [DocumentClient.get()](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/DynamoDB/DocumentClient.html#get-property)
+ * for descriptions of `params` and the return data.
+ *
+ * @param {Object} params
+ * @param {string} params.tableName - Table name to read
+ * @param {any} params.item - Key identifying object to get
```
yjpa7145, line 17: fix this one

.eslintignore
```diff
@@ -8,3 +8,5 @@ tmp/**
 docs/**
 _book/**
 website/**
+/packages/aws-client/*.js
```
markdboyd, line 11: Why do we need this for `aws-client` if we didn't for `errors`? If we do need it, can we have `aws-client/.eslintignore`?

packages/aws-client/src/CloudwatchEvents.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+import { cloudwatchevents } from './services';
+
+/**
+ * Create a CloudWatch Events rule
+ *
+ * @param {string} name - the rule name
+ * @param {string} schedule - a ScheduleExpression
+ * @param {string} state - the statue of the rule
```
markdboyd, line 12:
```suggestion
 * @param {string} state - the state of the rule
```

packages/aws-client/src/DynamoDb.ts
```diff
@@ -0,0 +1,170 @@
...
+ * @param {Object} params.client - Instance of a DynamoDb DocumentClient
```
markdboyd, line 18: The type should be `AWS.DynamoDB.DocumentClient`?

packages/aws-client/src/DynamoDbSearchQueue.ts
```diff
@@ -28,30 +35,24 @@ class DynamoDbSearchQueue {
...
-   * @returns {Promise<undefined>} - resolves when the queue has been updated
-   * @private
-   */
   async fetchItems() {
     let response;
     do {
       response = await this.dynamodbDocClient[this.searchType](this.params).promise(); // eslint-disable-line no-await-in-loop, max-len
       if (response.LastEvaluatedKey) this.params.ExclusiveStartKey = response.LastEvaluatedKey;
-    } while (response.Items.length === 0 && response.LastEvaluatedKey);
+    } while ((response.Items || []).length === 0 && response.LastEvaluatedKey);
```
markdboyd, line 50: I don't quite understand the refactor

packages/aws-client/src/KMS.ts
```diff
@@ -20,24 +23,26 @@ const createKey = (params = {}) => kms().createKey(params).promise();
...
-const decryptBase64String = (ciphertext) =>
-  kms().decrypt({
+export const decryptBase64String = async (ciphertext: string) => {
+  const { Plaintext } = await kms().decrypt({
     CiphertextBlob: Buffer.from(ciphertext, 'base64')
-  }).promise()
-    .then(({ Plaintext }) => Plaintext.toString());
+  }).promise();
+
+  if (Plaintext === undefined) return undefined;
```
markdboyd, line 45: So TS made you do this because it was warning you about calling `toString` on `undefined`?

packages/aws-client/src/Lambda.ts
```diff
@@ -1,26 +1,21 @@
...
-const awsServices = require('./services');
-const { inTestMode } = require('./test-utils');
+import Logger = require('@cumulus/logger');
+import { lambda } from './services';
+import { inTestMode } from './test-utils';
 
 const log = new Logger({ sender: 'aws-client/Lambda' });
 
-const invoke = async (name, payload, type = 'Event') => {
+export const invoke = async (name: string, payload: unknown, type = 'Event') => {
```
markdboyd, line 7: Why not `any` instead of `unknown`?

packages/aws-client/src/S3.ts
```diff
@@ -1,44 +1,49 @@
...
-}
+const S3_RATE_LIMIT = inTestMode() ? 1 : 20;
 
 /**
  * Join strings into an S3 key without a leading slash or double slashes
  *
  * @param {...string|Array<string>} args - the strings to join
  * @returns {string} the full S3 key
+ *
+ * @static
```
markdboyd, line 37: What does this mean outside of a class context?

packages/aws-client/src/S3.ts
```diff
@@ -1,44 +1,49 @@
...
  */
-function s3Join(...args) {
-  const tokens = Array.isArray(args[0]) ? args[0] : args;
+export function s3Join(...args: [string | string[], ...string[]]) {
```
markdboyd, line 39: `args` is a string, an array of strings, or ?

packages/aws-client/src/S3.ts
```diff
@@ -189,53 +224,27 @@ exports.downloadS3File = (s3Obj, filepath) => {
...
+  headObject(bucket, key, { retries: 3 })
     .then((response) => response.ContentLength);
 
 /**
 * Get object Tagging from S3
 *
 * @param {string} bucket - name of bucket
 * @param {string} key - key for object (filepath + filename)
-* @returns {Promise} - returns response from `S3.getObjectTagging` as a promise
+* @returns {Promise} returns response from `S3.getObjectTagging` as a promise
```
markdboyd, line 243: Does AWS provide a class for the response type inside the promise?

packages/aws-client/src/S3.ts
```diff
@@ -576,16 +627,17 @@ exports.calculateS3ObjectChecksum = async ({
  * @param {Object} [params.options] - crypto.createHash options
  *
  * @throws {InvalidChecksum} - Throws error if validation fails
- * @returns {boolean} - returns true for success
+ * @returns {boolean} returns true for success
```
markdboyd, line 630:
```suggestion
 * @returns {Promise<boolean>} returns true for success
```

packages/aws-client/src/S3.ts
```diff
@@ -553,15 +601,18 @@ exports.listS3ObjectsV2 = listS3ObjectsV2;
  * @param {string} params.key - S3 key
  * @param {Object} [params.options] - crypto.createHash options
  *
- * @returns {number|string} - calculated checksum
+ * @returns {number|string} calculated checksum
```
markdboyd, line 604:
```suggestion
 * @returns {Promise<number|string>} calculated checksum
```

packages/aws-client/src/client.ts
```diff
@@ -29,22 +24,32 @@ const memoize = (fn) => {
...
- * @param {Function} Service - an AWS service object constructor function
- * @param {string} version - the API version to use
- * @returns {Function} - a function which, when called, will return an AWS service object
+ * @param {AWS.Service} Service - an AWS service object constructor function
+ * @param {string} [version] - the API version to use
+ *
+ * @returns {Function} a function which, when called, will return an instance of an AWS service
+ * object
+ *
+ * @private
```
markdboyd, line 33: Why is this private if it's the main thing exported from this file?

packages/aws-client/src/test-utils.ts
```diff
@@ -92,38 +98,48 @@ function localStackAwsClient(Service, options) {
...
 
-  return {};
+  return <T>{};
 }
-exports.testAwsClient = testAwsClient;
 
-const throwThrottlingException = () => {
-  const throttlingException = new Error('ThrottlingException');
-  throttlingException.code = 'ThrottlingException';
+class ThrottlingException extends Error {
```
markdboyd, line 118: Is there any reason not to move this to the `errors` package?

tsconfig.json
```diff
@@ -3,9 +3,10 @@
     "target": "ES2018",
     "module": "commonjs",
     "lib": ["ES2018"],
+    "esModuleInterop": true,
     "declaration": true,
     "sourceMap": true,
-    "removeComments": true,
+    "removeComments": false,
```
markdboyd, line 9: Why the change?

.eslintrc.json
```diff
@@ -125,18 +126,19 @@
...
         "@typescript-eslint/no-throw-literal": "off",
         "comma-dangle": ["warn", "never"],
+        "import/first": "off",
+        "import/prefer-default-export": "off",
         "lines-between-class-members": "off",
+        "lodash/prefer-lodash-typecheck": "off",
         "node/no-unsupported-features/es-syntax": "off",
-        "valid-jsdoc": "off",
-        "tsdoc/syntax": "warn"
+        "unicorn/no-abusive-eslint-disable": "off"
```
markdboyd, line 141: You don't need to do this if you upgrade `eslint-plugin-unicorn` to the latest version. I just tested it. This issue proves that the plugin supports `@typescript` rules:

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/212

packages/aws-client/src/DynamoDbSearchQueue.ts
yjpa7145: Apparently `response.Items` can potentially be `undefined`. This catches that case and defaults to an empty array, which gives the behavior that we're looking for.

tsconfig.json
yjpa7145: `jsdoc2md` can't parse the `.ts` files, so it's configured to parse the generated `.js` files instead. When the comments were being stripped out, though, `jsdoc2md` didn't have anything to parse.

The `jsdoc2md` docs say that you _may_ be able to use `jsdoc2md` and babel to use babel to compile your `.ts` files and have `jsdoc2md` build docs off of those, but there are a number of documented cases where that doesn't work. I ran into enough of those cases that I decided to just build the docs off of the `.js` files that `tsc` compiled.

.eslintrc.json
yjpa7145: Fixed in 2a86d8408

Upgrading `eslint-plugin-unicorn` resulted 1,000+ new linter issues being reported. I disabled those rules in the config but have documented them so that we can discuss them in an architecture meeting and then file tickets to re-enable them.

.eslintignore
yjpa7145: This wasn't needed in the `errors` package because that package writes its compiled `.js` to a `dist` directory, which is covered by this top-level `.eslintignore` rule: `**/dist/**`.  In order to not break backward compatibility, we need the `aws-client` compiled Javascript files to appear at the root level of the `aws-client` package.

eslint does not support per-directory `.eslintignore` files. From the docs:

> When ESLint is run, it looks in the current working directory to find an .eslintignore file before determining which files to lint. If this file is found, then those preferences are applied when traversing directories. Only one .eslintignore file can be used at a time, so .eslintignore files other than the one in the current working directory will not be used.

I tried creating `aws-client/.eslintignore` and confirmed that it was not used. It looks like we have `.eslintignore` files in other places, too, which are also not going to have any effect. Well, they would only have an effect if you ran `eslint` from inside of that package directory.

yjpa7145: I have removed those extra `.eslintignore` files to prevent confusion in the future, and confirmed that removing them didn't change the behavior of `npm run eslint`.

packages/aws-client/src/DynamoDb.ts
yjpa7145: Fixed.

packages/aws-client/src/KMS.ts
yjpa7145: Correct, apparently `Plaintext` could potentially be undefined.

packages/aws-client/src/Lambda.ts
yjpa7145: This explains it better than I could: https://mariusschulz.com/blog/the-unknown-type-in-typescript

packages/aws-client/src/S3.ts
yjpa7145: This falls under the "I don't always understand how `jsdoc2md` works" category. This was an issue in JS too, it isn't new in TS.

If you export a function that was declared as `const listS3Objects = () => blah;`, the `jsdoc2md` will generate documentation as expected.

If you export a function that was declared as `function s3Join () {}`, the `jsdoc2md` documents it as this, for some reason:

```
    * _inner_
        * [~s3Join(...args)](#module_S3..s3Join) ⇒ <code>string</code>
```

I don't know why, but I know that adding `@static` to the jsdoc of those `function blah()` exports will fix it.

packages/aws-client/src/S3.ts
yjpa7145: Yeah, I would love to break the backward compatibility of this function. You can call `s3Join` with either an array of strings to join, or any number of strings to be joined (rather than an array). It would be a lot simpler if it behaved like `path.join` and just took a variable number of strings rather than that _or_ an array.

This is saying that the first argument to `s3Join` can be either a string or an array of strings. Any arguments _after_ that first one must be strings. Say the word and I'll be happy to make a breaking change here.

packages/aws-client/src/S3.ts
yjpa7145: Updated

packages/aws-client/src/client.ts
yjpa7145: To prevent it from being published in our API docs. I'd hate for anyone else to use this function.

packages/aws-client/src/test-utils.ts
yjpa7145: Moved.

packages/aws-client/src/S3.ts
markdboyd: Any reason to not just export it as a `const` instead?

packages/aws-client/src/S3.ts
markdboyd: I mean, it's extremely hard to follow even after that explanation, but if you've already handled it I don't feel extremely compelled to break it. Your call.

packages/errors/src/index.ts
```diff
@@ -25,6 +26,17 @@ export const createErrorType = (
     }
   );
 
+export class ThrottlingException extends Error {
+  readonly code: string;
+
+  constructor() {
+    super('ThrottlingException');
+    this.name = 'ThrottlingException';
+    this.code = 'ThrottlingException';
```
markdboyd, line 35: So this whole class is only necessary to set the `code` on the error object?

packages/aws-client/src/S3.ts
yjpa7145: Yeah, looks like that fixes it. I'll make that change.

packages/errors/src/index.ts
yjpa7145: Yes.

PR #1585, opened 2020-04-24
CUMULUS-1891

markdboyd:

Addresses [CUMULUS-1891: //replays endpoint should not return 202 if async operation fails to start](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1891)

## Changes

- Updated the following endpoints using async operations to return a 503 error if the ECS task  cannot be started and a 500 response for a non-specific error:
  - POST `/replays`
  - POST `/bulkDelete`
  - POST `/elasticsearch/index-from-database`
  - POST `/granules/bulk`
- Created an error handler middleware for async operation based endpoints, `asyncOperationEndpointErrorHandler`, and updated endpoints to use it
  - Added unit tests for middleware

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests


packages/api/endpoints/bulk-delete.js
```diff
@@ -17,25 +19,18 @@ async function startBulkDeleteAsyncOperation(req, res) {
...
-      description: 'Bulk Delete',
-      operationType: 'Bulk Delete',
-      payload: { granuleIds: req.body.granuleIds }
-    });
-  } catch (err) {
-    if (err.name !== 'EcsStartTaskError') throw err;
-
-    return res.boom.serverUnavailable(`Failed to run ECS task: ${err.message}`);
-  }
+  const asyncOperation = await asyncOperationModel.start({
```
markdboyd, line 22: There is no need to catch these errors anymore because if an error is thrown here then the `asyncOperationEndpointErrorHandler` middleware is automatically triggered and will return the appropriate response

packages/api/app/middleware.js
```diff
@@ -0,0 +1,26 @@
...
+// Catch and send the error message down (instead of just 500: internal server error)
+// Need all 4 params, because that's how express knows this is the error handler
+// eslint-disable-next-line no-unused-vars
+const defaultErrorHandler = (error, req, res, next) => {
+  logger.error(error);
+  return res.boom.badRequest(error.message, error);
+};
+
+// eslint-disable-next-line no-unused-vars
+const asyncOperationEndpointErrorHandler = async (err, req, res, next) => {
```
markdboyd, line 15: So as I was updating the error handling for the `/replays` endpoint, it occurred to me that we probably want to do the error handling consistently for all the endpoints which kick off async operations. I debated a few options but made the call that a custom error middleware for these endpoints might be best, because we can handle it in one place and just attach this middleware everywhere we define an endpoint for async operations. But interested in what others think

packages/api/tests/app/test-middleware.js
```diff
@@ -0,0 +1,43 @@
...
+
+app.post(
+  '/fake-async-endpoint2',
+  () => {
+    throw new EcsStartTaskError('failed to start');
+  },
+  asyncOperationEndpointErrorHandler
+);
+
+test('request to replays endpoint returns 500 if starting ECS task throws unexpected error', async (t) => {
```
markdboyd, line 30: Because the error handling for async operation endpoints is controlled by middleware, I also thought maybe we only need to test that middleware once (on fake endpoints), instead of writing separate tests each endpoint (/replays, bulk granules, etc) to see whether they handle the errors correctly. But is that a good assumption? 

All this proves is that the middleware will return the correct response if it's the only middleware in the chain, but the endpoint routes could be defined with other middleware

packages/api/app/routes.js
```diff
@@ -111,10 +111,6 @@ router.use('/ems', ensureAuthorized, ems);
 
 // Catch and send the error message down (instead of just 500: internal server error)
 // Need all 4 params, because that's how express knows this is the error handler
```
yjpa7145, line 113: This comment can probably be removed now.

packages/api/tests/models/test-async-operation-model.js
```diff
@@ -153,24 +155,35 @@ test.serial('The AsyncOperation.start() method starts an ECS task with the corre
   t.is(environmentOverrides.payloadUrl, `s3://${systemBucket}/${asyncOperationModel.stackName}/async-operation-payloads/${id}.json`);
 });
 
-test('The AsyncAdapter.start() method sets the output if it is unable to create an ECS task', async (t) => {
+test('The AsyncAdapter.start() method throws error and updates operation if it is unable to create an ECS task', async (t) => {
```
yjpa7145, line 158:
```suggestion
test('The AsyncOperation.start() method throws error and updates operation if it is unable to create an ECS task', async (t) => {
```

packages/api/app/middleware.js
```diff
@@ -0,0 +1,26 @@
...
+  logger.error('Failed to run ECS task', err);
+  if (err instanceof EcsStartTaskError) {
+    return res.boom.serverUnavailable(`Failed to run ECS task: ${err.message}`);
```
yjpa7145, line 18: I'm not sure we want to be exposing an implementation detail like "we're using ECS for async operations."

```suggestion
    return res.boom.serverUnavailable(`Failed to start async operation: ${err.message}`);
```

markdboyd: These errors would be obscured the user anyway because boom hides the message for any 5xx errors, but I can still make the change.

packages/api/models/async-operation.js
```diff
@@ -81,52 +118,28 @@ class AsyncOperation extends Manager {
...
-      }
-    }).promise();
+    const runTaskResponse = await this.startECSTask({
+      ...params,
+      id,
+      payloadBucket,
+      payloadKey
+    });
 
     // If creating the stack failed, update the database
```
yjpa7145, line 128: This comment no longer applies.

packages/api/tests/models/test-async-operation-model.js
```diff
@@ -153,24 +154,25 @@ test.serial('The AsyncOperation.start() method starts an ECS task with the corre
   t.is(environmentOverrides.payloadUrl, `s3://${systemBucket}/${asyncOperationModel.stackName}/async-operation-payloads/${id}.json`);
 });
 
-test('The AsyncAdapter.start() method sets the output if it is unable to create an ECS task', async (t) => {
+test('The AsyncOperation.start() method throws error and updates operation if it is unable to create an ECS task', async (t) => {
```
yjpa7145, line 157:
```suggestion
test('The AsyncOperation.start() method throws error if it is unable to create an ECS task', async (t) => {
```

packages/api/tests/models/test-async-operation-model.js
```diff
@@ -153,24 +154,25 @@ test.serial('The AsyncOperation.start() method starts an ECS task with the corre
...
   stubbedEcsRunTaskResult = {
     tasks: [],
     failures: [{ arn: randomString(), reason: 'out of cheese' }]
   };
 
-  const { id } = await asyncOperationModel.start({
+  const expectedMessage = 'Failed to start AsyncOperation: out of cheese';
```
yjpa7145, line 163: 🐜 🐜 🐜 This string can really just be moved down to line 174: `message: ...`

packages/api/tests/app/test-middleware.js
```diff
@@ -0,0 +1,43 @@
...
+  const response = await request(app)
+    .post('/fake-async-endpoint')
+    .expect(500);
+  t.is(response.status, 500);
```
yjpa7145, line 34: If this is anything other than `500`, the `.expect(500)` line above is going to cause an exception to be thrown, and this line will never be reached. I'd probably drop the `.expect(500)`. Same thing for the other test in this file.

packages/api/app/middleware.js
yjpa7145: Seems reasonable.

PR #1586, opened 2020-04-24
Log granules in Discover Granules task

laurenfrederick:

Addresses [CUMULUS-1859](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1859)

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


tasks/discover-granules/tests/index.js
```diff
@@ -422,3 +428,45 @@ test.serial('checkGranuleHasNoDuplicate throws an error on an unexpected API lam
...
+
+      const granules = JSON.parse(process.env.GRANULES);
+      t.deepEqual(granules, [
+        'granule-1',
+        'granule-2',
+        'granule-3'
+      ]);
+
+      // Check that the second log.info() call had the granules set
+      const loggerCall = loggerInfoSpy.getCall(1);
```
laurenfrederick, line 462: I put this in as a test that we actually set the granules before we did the logging. I'm back and forth though because it's really brittle. If you add another `log.info()` call before this one then this test would fail.

tasks/discover-granules/tests/index.js
```diff
@@ -422,3 +428,45 @@ test.serial('checkGranuleHasNoDuplicate throws an error on an unexpected API lam
...
+      t.deepEqual(loggerCall.thisValue.granules[
```
yjpa7145, line 463: It horrifies me that this is passing. There is a comma missing between `loggerCall.thisValue.granules` and `[`, which happens to all of us, but then ... why is it passing? Javascript is doing some awful interpretation of this line and coming up with ... I don't know what.

In any case, without that comma, it's definitely not testing what you think it's testing.

PR #1588, opened 2020-04-29
Use p-limit for concurrency in index-from-database

abarciauskas-bgse:
**Summary:** 

When running a large elasticsearch indexing operation, sending all indexing requests at once results in an error from Elasticsearch: `es_rejected_execution_exception`. I also noticed that making a large indexing request causes memory consumption to spike to 100% very quickly. This PR uses `p-limit` to enable setting a maximum number of concurrent requests when indexing items from each DynamoDB table being indexed.

It is also worth considering if we should use p-limit on the call to `Promise.all([indexModel..., indexModel..., indexModel` since right now I believe p-limit is being to limit the number of concurrent requests for each table separately, as opposed to items for all tables.

## Changes

* Add `ES_CONCURRENCY` / `es_concurrency` variable to index_from_database terraform module
* Use ES_CONCURRENCY or default in `index-from-database` lambda function call to index items in DynamoDB

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/packages/api/lambdas/index-from-database.js
+++ b/packages/api/lambdas/index-from-database.js
@@ -1,6 +1,8 @@
 'use strict';
 
 const log = require('@cumulus/common/log');
+const pLimit = require('p-limit');
+const limit = pLimit(process.env.ES_CONCURRENCY || 10);
 
 const DynamoDbSearchQueue = require('@cumulus/aws-client/DynamoDbSearchQueue');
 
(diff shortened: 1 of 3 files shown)


markdboyd:
Thanks @abarciauskas-bgse !

PR #1591, opened 2020-04-29
Add index database concurrency

markdboyd:
Integrates #1588 

- Additional improvements for logging
- Adds unit tests

### DIFF
--- a/packages/api/lambdas/index-from-database.js
+++ b/packages/api/lambdas/index-from-database.js
@@ -1,6 +1,8 @@
 'use strict';
 
 const log = require('@cumulus/common/log');
+const pLimit = require('p-limit');
+const limit = pLimit(process.env.ES_CONCURRENCY || 10);
 
 const DynamoDbSearchQueue = require('@cumulus/aws-client/DynamoDbSearchQueue');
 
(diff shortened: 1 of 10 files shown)


packages/api/lambdas/index-from-database.js
```diff
@@ -25,33 +40,113 @@ async function indexModel(esClient, tableName, esIndex, indexFn) {
...
 
   await Promise.all([
-    indexModel(esClient, tables.collectionsTable, esIndex, indexer.indexCollection),
-    indexModel(esClient, tables.executionsTable, esIndex, indexer.indexExecution),
-    indexModel(esClient, tables.asyncOperationsTable, esIndex, indexer.indexAsyncOperation),
-    indexModel(esClient, tables.granulesTable, esIndex, indexer.indexGranule),
-    indexModel(esClient, tables.pdrsTable, esIndex, indexer.indexPdr),
-    indexModel(esClient, tables.providersTable, esIndex, indexer.indexProvider),
-    indexModel(esClient, tables.rulesTable, esIndex, indexer.indexRule)
+    indexModel({
```
markdboyd, line 83: I thought about using `p-limit` to limit how many **models** the Lambda tries to do indexing for at once, but wondered if it was overkill? The existing changes already ensure that each model will only try to index so many records at a time, though I guess if that limit is set sufficiently high, it may also be wise to limit how many models are indexing at once?

packages/api/lambdas/index-from-database.js
```diff
@@ -25,33 +40,113 @@ async function indexModel(esClient, tableName, esIndex, indexFn) {
...
+      esClient,
+      tableName: tables.collectionsTable,
+      esIndex,
+      indexFn: indexer.indexCollection,
+      limitEsRequests
```
markdboyd, line 88: I really dislike passing the `p-limit` function as a parameter, but it seemed wasteful to pass the limit value into `indexModel` and re-initialize the concurrency function every time

PR #1596, opened 2020-05-02
CUMULUS-408: HTTPS Ingest

nemreid:

Addresses [CUMULUS-408: HTTPS Ingest Test](https://bugs.earthdata.nasa.gov/browse/CUMULUS-408)

## Changes

* Fix HTTPS Ingest
* Add new `certificateUri` field to providers to allow custom CA bundles.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


example/spec/helpers/ecsHelpers.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+const { ecs } = require('@cumulus/aws-client/services');
+
+async function findAsyncOperationTaskDefinitionForDeployment(stackName) {
```
nemreid, line 5: Pagination fix. All updates to Async* spec tests are to use this fix.

DATA_MODEL_CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- **CUMULUS-408**
```
laurenfrederick, line 12: Do we need to make dashboard updates for this?

example/spec/helpers/Providers.js
```diff
@@ -41,17 +41,28 @@ const buildFtpProvider = async (postfix = '') => {
   return provider;
 };
 
-const buildHttpProvider = async (postfix = '') => {
+const fakeProviderPortMap = {
+  http: process.env.PROVIDER_HTTP_PORT ? Number(process.env.PROVIDER_HTTP_PORT) : 3030,
+  https: process.env.PROVIDER_HTTPS_PORT ? Number(process.env.PROVIDER_HTTPS_PORT) : 4040,
+};
+
+const buildHttpOrHttpsProvider = async (postfix = '', protocol = 'http', systemBucket) => {
```
laurenfrederick, line 49: I'm confused here about why defaults are specified for the first 2 params if you have to specify the `systemBucket` param.

nemreid: Fair point, I can switch them around.

DATA_MODEL_CHANGELOG.md
nemreid: Looks like the providers page has fields instead of a JSON form, so yes.

nemreid: It's an optional field for edge case providers that use self-signed certificates, so it's not exactly top priority, but we should file a ticket.

nemreid: Filed CUMULUS-1992 for this.

PR #1605, opened 2020-05-07
Add index database concurrency lauren

laurenfrederick:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


example/cumulus-tf/main.tf
```diff
@@ -112,6 +112,8 @@ module "cumulus" {
 
   token_secret = var.token_secret
 
+  es_request_concurrency = var.es_request_concurrency
```
markdboyd, line 115: So we want to be able to set this specifically for the example deployments? Because the default value for the `cumulus` module is 10 already

packages/api/lambdas/index-from-database.js
```diff
@@ -71,7 +71,7 @@ async function indexModel({
 
 async function indexFromDatabase(event) {
   const {
-    index: esIndex,
+    indexName: esIndex,
```
markdboyd, line 74: Great catch. Epic fail on my part

markdboyd: Also something that TS would catch if we had a description for the typing of `event`, but I digress

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Changed
+
+Limit concurrent Elasticsearch functions when doing an index from database
```
markdboyd, line 11:
```suggestion
  - Added a limit for concurrent Elasticsearch requests when doing an index from database operation
```

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Changed
```
markdboyd, line 9: Should the header be "Added" instead?

example/cumulus-tf/main.tf
laurenfrederick: I was testing it so I used it

CHANGELOG.md
laurenfrederick: Yeah I could see that. I changed it.

example/cumulus-tf/main.tf
laurenfrederick: I removed it

PR #1607, opened 2020-05-08
Add @cumulus/aws-client/S3.multipartCopyObject()

yjpa7145:
This is an incremental step toward CUMULUS-1894, adding the ability to handle large S3 file copies.


packages/aws-client/lib/S3MultipartUploads.js
```diff
@@ -0,0 +1,115 @@
...
+// This byte range has a starting byte and an ending byte (inclusive) that makes
+// up the part. All parts must be larger than 5 MB, except that the last part
+// may be smaller than 5 MB. Parts cannot be larger than 5 GB.
+//
+// This function takes a file size and an optional maxSize. It returns an array
+// of objects, each containing a `start` and an `end` value. These will make up
+// the ranges of the multi-part copy.
+//
+// https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html
+const createMultipartChunks = (size, maxSize = 5 * GB) => {
```
chuckwondo, line 20: nit: Consider using more specific names, such as `objectSize` and `maxChunkSize` to make it more immediately clear as to what they mean. Otherwise, the existing name similarity forces a good read of the code to grok their purposes.

packages/aws-client/lib/S3MultipartUploads.js
```diff
@@ -0,0 +1,115 @@
...
+  const lastChunkSize = size % maxSize;
+
+  // Build the list of full-size chunks
+  const chunks = range(0, size - lastChunkSize, maxSize).map((start) => {
```
chuckwondo, line 24: nit: No need to subtract `lastChunkSize`

packages/aws-client/lib/S3MultipartUploads.js
```diff
@@ -0,0 +1,115 @@
...
+// up the part. All parts must be larger than 5 MB, except that the last part
+// may be smaller than 5 MB. Parts cannot be larger than 5 GB.
+//
+// This function takes a file size and an optional maxSize. It returns an array
+// of objects, each containing a `start` and an `end` value. These will make up
+// the ranges of the multi-part copy.
+//
+// https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html
+const createMultipartChunks = (size, maxSize = 5 * GB) => {
+  const lastChunkSize = size % maxSize;
```
chuckwondo, line 21: Don't we need to clamp `maxSize` (or `maxChunkSize` as suggested above) to the range [5 MB, 5 GB]? There's nothing here that would prevent specifying a size outside of this range.

packages/aws-client/S3.js
```diff
@@ -621,3 +622,70 @@ exports.getFileBucketAndKey = (pathParams) => {
...
+ * Copy an S3 object to another location in S3 using a multipart copy
+ *
+ * @param {Object} params
+ * @param {string} params.sourceBucket
+ * @param {string} params.sourceKey
+ * @param {string} params.destinationBucket
+ * @param {string} params.destinationKey
+ * @returns {Promise<undefined>}
+ */
+const multipartCopyObject = async (params) => {
```
chuckwondo, line 636: nit: Do we want to consider naming this `uploadCopy` to show its correspondence to both the `upload` and `uploadPartCopy` functions that are already part of the SDK? Just a thought on name consistency. I'm not bothered either way.

packages/aws-client/tests/lib/test-S3MultipartUploads.js
```diff
@@ -0,0 +1,128 @@
...
+  );
+
+  t.deepEqual(
+    createMultipartChunks(12, 10),
+    [
+      { start: 0, end: 9 },
+      { start: 10, end: 11 }
+    ]
+  );
+});
```
chuckwondo, line 45: I suggest splitting each of these `deepEqual` cases out into separate tests, as a general practice, as these represent parameterized tests and you want them _all_ to be executed, regardless of the failure of the others. In other words, if the first `deepEqual` fails, all of the remaining ones are skipped, and you won't see _all_ of the failures at once (if there would be multiple).

However, given this is pretty trivial, I'd be okay if you would prefer to leave things as they are, but want to at least plant the general idea in your mind.

packages/aws-client/lib/S3MultipartUploads.js
yjpa7145: 👍 Will fix.

packages/aws-client/lib/S3MultipartUploads.js
yjpa7145: 👍 Will fix.

yjpa7145: Removing that does actually cause test failures:

```
lib › S3MultipartUploads › createMultipartChunks(11, 10) returns the correct chunks

  tests/lib/test-S3MultipartUploads.js:36

   35: test('createMultipartChunks(11, 10) returns the correct chunks', (t) => {
   36:   t.deepEqual(                                                           
   37:     createMultipartChunks(11, 10),                                       

  Difference:

    [
      {
        end: 9,
        start: 0,
      },
  -   {
  -     end: 19,
  -     start: 10,
  -   },
      {
        end: 10,
        start: 10,
      },
    ]

  › test (tests/lib/test-S3MultipartUploads.js:36:5)
  › process._tickCallback (internal/process/next_tick.js:68:7)
```

packages/aws-client/tests/lib/test-S3MultipartUploads.js
yjpa7145: 👍

packages/aws-client/lib/S3MultipartUploads.js
chuckwondo: Yes, that's because the next line should be corrected from this:
```
const end = start + maxSize - 1;
```
to this:
```
const end = Math.min(start + maxSize, size) - 1;
```
which would then allow you to eliminate the block for creating the last chunk that follows.

packages/aws-client/S3.js
yjpa7145: I thought about that but decided that it would be more confusing. It's already pretty confusing that AWS makes you refer to the "S3 to S3" copy operations as "upload" operations. I'd rather insulate the users of this function from that confusion if possible.

packages/aws-client/lib/S3MultipartUploads.js
yjpa7145: I agree that would work. My concern is that this function is a little confusing already. I'd rather err on the side of being a little more verbose so that, when someone has to fix a bug in it on a Friday night after a couple 🍺 🍺 , it'll be easier to understand.

packages/aws-client/lib/S3MultipartUploads.js
yjpa7145: In an awesome twist, it turns out that you can use multi-part upload copies on files of any size. I was testing against the real AWS service and realized that I was able to upload a 10-byte file using multi-part. The documentation says that the only chunk that can be less than 5 MB is the last chunk. Turns out that, if you only send one chunk, it can be as small as you want.

I also did some performance testing against AWS and found that a chunk size of about 250 MB actually seems to give the best performance. Copying an 11 GB file using 5 GB chunks took about 133 seconds, while copying the same file with 250 MB chunks only took 10 seconds. Small sample size, but the smaller chunks did show an order of magnitude improvement.

To be on the safe side, I did update this function to throw an exception if someone tries to use a `maxSize` greater than 5GB.

packages/aws-client/src/lib/S3MultipartUploads.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  // If necessary build the last, not-full-size chunk
+  if (lastChunkSize !== 0) {
+    const start = objectSize - lastChunkSize;
+    const end = objectSize - 1;
+    chunks.push({ start, end });
+  }
+
+  return chunks;
+};
```
chuckwondo, line 50: This is a small amount of code, so I'm perfectly fine with the additional verbosity, so the rest of this comment is for optional perusal.

BTW, glad to see you converted this file to TypeScript, and also that you found a good default chunk size for good performance. Looks like AWS nicely parallelizes the part uploads.

I agree that a bit of verbosity can indeed be preferable when it greatly clarifies things.

I think it's debatable as to whether that's the case here. I arguably (arguing with myself) find the code below to be more clear, but that's perhaps because I wrote it. Notice that it also avoids checking `maxChunkSize`, leaving `uploadPartCopy` to do the complaining, if necessary.

```javascript
export const createMultipartChunks = (
  objectSize: number,
  maxChunkSize = 250 * MB
): Chunk[] =>
  range(0, objectSize, maxChunkSize).map((start) => ({
    start,
    end: Math.min(start + maxChunkSize, objectSize) - 1
  }));
```

packages/aws-client/S3.js
chuckwondo: Works for me.

packages/aws-client/lib/S3MultipartUploads.js
chuckwondo: Ah, yes, that makes perfect sense, and nice speedup! See my other comment about either also throwing when chunk size is too small, or not throwing in either case.

packages/aws-client/src/lib/S3MultipartUploads.ts
```diff
@@ -0,0 +1,150 @@
...
+// From anecdotal testing, a chunk size of 250 MB seems to perform fairly well.
+//
+// https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html
+export const createMultipartChunks = (
+  objectSize: number,
+  maxChunkSize = 250 * MB
+): Chunk[] => {
+  if (maxChunkSize > (5 * GB)) {
+    throw new Error('maxChunkSize can be no greater than 5368709120');
+  }
```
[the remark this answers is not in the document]
yjpa7145: > Alternatively, I suggest we don't bother checking for either case and simply let AWS throw an error when we pass the "illegal" chunks to uploadPartCopy.

I'm inclined to go with this approach. That way, if AWS adjusts their allowed sizes, we'll get that benefit automatically.

This function is also not part of the public API, it was extracted out to make `multipartCopyObject()` easier to understand. Since it will only be used internally, I'm comfortable with us not passing an invalid value here, or at least noticing when we do.

packages/aws-client/src/lib/S3MultipartUploads.ts
yjpa7145: Assuming the tests pass, I'll to it this way.

PR #1610, opened 2020-05-10
Fix granule reporting test timeout

yjpa7145:
This test was failing intermittently because it was timing out. The test creates a couple granules with a dummy file in each of them. Those files were not created in S3 so that, when the code under test tried to get the size of those files from S3, it was retrying for 7 seconds before failing.

By staging the dummy files to S3, the "get size" operation returns immediately and took the test time from 7+ seconds and intermittent failures to working consistently in milliseconds.

### DIFF
--- a/packages/api/tests/models/granules/test-granule-reporting.js
+++ b/packages/api/tests/models/granules/test-granule-reporting.js
@@ -1,6 +1,7 @@
 const test = require('ava');
 const sinon = require('sinon');
 
+const S3 = require('@cumulus/aws-client/S3');
 const StepFunctions = require('@cumulus/aws-client/StepFunctions');
 const cmrjs = require('@cumulus/cmrjs');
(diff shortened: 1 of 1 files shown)


packages/api/tests/models/granules/test-granule-reporting.js
```diff
@@ -235,40 +236,52 @@ test('_validateAndStoreGranuleRecord() does not throw an error for a failing rec
...
-        version: '001'
+  try {
+    const granule1 = fakeGranuleFactoryV2({
+      files: [fakeFileFactory({ bucket })]
+    });
+    const granule2 = fakeGranuleFactoryV2({
+      files: [fakeFileFactory({ bucket })]
+    });
+
+    await Promise.all([
```
markdboyd, line 250: Was there some mocking going on previously that we're superseding here? I can't see it, but if so, can it be removed?

PR #1613, opened 2020-05-10
Allow test files to rely on dev deps from the top-level package.json

yjpa7145:
⚠️ This PR depends on #1612

In order to speed up bootstrapping and CI build time, we are looking to move `devDependencies` out of individual packages and into the top-level `package.json`. One issue that we ran into was that the `node/no-extraneous-require` eslint rule, which tells you when you forgot to include a dependency in your `package.json` was complaining about those missing `devDependencies`.

This PR fetches the list of `devDependencies` that are defined in the top-level `package.json` and adds them to the `allowModules` config for the `node/no-extraneous-require` rule, but only for test files. By doing this, missing dependencies will still generate a warning but, if a test file contains a package that is listed in the top-level `package.json`, it will not generate an error.


.eslintrc.js
```diff
@@ -119,6 +136,9 @@ module.exports = {
 
     'node/no-missing-require': 'off',
 
+    'unicorn/filename-case': 'off',
+    'unicorn/no-process-exit': 'off',
+    'unicorn/prefer-spread': 'off',
```
markdboyd, line 141: These are already included below

PR #1625, opened 2020-05-12
Pull dev dependencies of packages/* up

yjpa7145:
This should improve bootstrap time.


packages/api/package.json
```diff
@@ -99,20 +99,6 @@
...
-    "fs-extra": "^7.0.0",
-    "nock": "^10.0.0",
-    "nyc": "^14.0.0",
-    "proxyquire": "^2.0.0",
-    "rewire": "^4.0.1",
-    "sinon": "^7.1.1",
-    "supertest": "^3.3.0",
-    "webpack": "~4.5.0",
-    "webpack-cli": "~2.0.14"
+    "@cumulus/test-data": "1.22.1"
```
chuckwondo, line 102: I notice that this dependency (and a couple of others in other places) are left behind in non-root `devDependencies`. Just curious as to why that is.

PR #1637, opened 2020-05-15
Add support for specifying the number of es index shards

laurenfrederick:

Addresses [CUMULUS-1995](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1995)

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


CHANGELOG.md
```diff
@@ -31,6 +31,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added a limit for concurrent Elasticsearch requests when doing an index from database operation
 - Added the `es_request_concurrency` parameter to the archive and cumulus Terraform modules
 
+- **CUMULUS-1995**
+  - Added the `es_index_shards` paramater to the archive and cumulus Terraform modules to configure the number of shards for the ES index
+    - If you have an existing ES index, you will need to [reindex](https://nasa.github.io/cumulus-api/#reindex) to take advantage of shard updates
```
markdboyd, line 36: So you need to reindex and then change-index, correct? That's what the API docs seem to say but not sure if we want to specify that here

packages/api/lambdas/bootstrap.js
```diff
@@ -88,7 +88,14 @@ async function bootstrapElasticSearch(host, index = 'cumulus', alias = defaultIn
     // add mapping
     await esClient.indices.create({
       index,
-      body: { mappings }
+      body: {
```
markdboyd, line 91: I think it would be nice to reuse `createIndex` here rather than repeating the logic

laurenfrederick: I started to make that change and what i didn't love about it is you do the exists check right above so you'd be doing it twice.  And then you could do `createIndex` and then handle the error and set `exists` that way, but that made it seem like you shoehorned `createIndex` in. And then it's like well you could pull the actual creation piece out to its own function which seemed like overengineering. I hope you enjoyed this recap of the conversation I had with myself.

markdboyd: I guess I don't think pulling the code just doing the index creation into its own function is overengineering and it avoids future drift between the code using this new env vars. The current `createIndex` helper is overloaded in that it checks for the index and creates the index. If those were two separate functions, then all of the calling code could implement them as desired (throw exceptions or not, etc). 

I don't feel super strongly, but I think it would be nice. But if you think this is out of scope we can nix it for now.

CHANGELOG.md
```diff
@@ -31,6 +31,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added a limit for concurrent Elasticsearch requests when doing an index from database operation
 - Added the `es_request_concurrency` parameter to the archive and cumulus Terraform modules
 
+- **CUMULUS-1995**
+  - Added the `es_index_shards` paramater to the archive and cumulus Terraform modules to configure the number of shards for the ES index
```
markdboyd, line 35:
```suggestion
  - Added the `es_index_shards` parameter to the archive and cumulus Terraform modules to configure the number of shards for the ES index
```

PR #1638, opened 2020-05-15
CUMULUS-1993: Respect event collection in SQS message consumer

markdboyd:

Addresses [CUMULUS-1993: Fix SQS message consumer to respect collection in queue message](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1993)

## Changes

* Update SQS message consumer to filter rules for processing based on the collection in the message (if any)
* Update SQS message consumer to do the primary iteration over queues instead of all rules, so queue messages are only read once per Lambda invocation rather than N times per rule

## PR Checklist

- [ ] Update CHANGELOG
- [x] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


packages/api/lambdas/sqs-message-consumer.js
```diff
@@ -18,66 +18,95 @@ const Rule = require('../models/rules');
  * messages from SQS queue are processed
  */
 async function processQueues(event, dispatchFn) {
-  const model = new Rule();
-  const rules = await model.getRulesByTypeAndState('sqs', 'ENABLED');
+  const rulesModel = new Rule();
+  const rules = await rulesModel.queryRules({
```
markdboyd, line 22: The only downside to this change is that `rulesModel.queryRules` throws an error if it finds no rules. But if this Lambda should do nothing if it finds no rules. And if it throws an error, it should be logged so I'm not sure if this is a problem

packages/api/lambdas/sqs-message-consumer.js
```diff
@@ -18,66 +18,95 @@ const Rule = require('../models/rules');
...
-    if (!(await sqsQueueExists(queueUrl))) return Promise.resolve();
+    if (!(await sqsQueueExists(queueUrl))) {
+      const ruleIds = rulesForQueue.map((rule) => rule.id);
+      log.info(`Could not find queue ${queueUrl}. Unable to process rules ${ruleIds}`);
+      return Promise.resolve();
+    }
+
+    // Use the max of the visibility timeouts for all the rules
+    // bound to this queue.
+    const visibilityTimeout = rulesHelpers.getMaxTimeoutForRules(rulesForQueue);
```
markdboyd, line 49: I discussed this with Menno - but the issue here is that we potentially have multiple rules for a given queue at this point, whereas we always had 1 in the previous implementation since we were iterating over rules and not queues. Since we have multiple rules, we still want to respect their `rule.meta.visibilityTimeout` value, but those values could be different, so taking the maximum of them seemed logical

packages/api/lib/rulesHelpers.js
```diff
@@ -6,6 +6,31 @@ const { removeNilProperties } = require('@cumulus/common/util');
...
+      ? get(rule, 'collection.name') === collection.name
+      : true;
+    const versionMatch = collection.version
+      ? get(rule, 'collection.version') === collection.version
+      : true;
+    return nameMatch && versionMatch;
+  }
+);
+
+const getMaxTimeoutForRules = (rules) => rules.reduce(
```
laurenfrederick, line 22: Very nitpicky, but something like this seems a bit simpler to me:
```
const values = rules.map((rule) => get(rule, 'meta.visibilityTimeout')).filter((t) => t);
values.reduce((a, b) => Math.max(a, b));
```

packages/api/lambdas/sqs-message-consumer.js
```diff
@@ -18,66 +18,95 @@ const Rule = require('../models/rules');
...
-    };
-    return sqs().changeMessageVisibility(params).promise();
-  }
+  const eventObject = JSON.parse(message.Body);
+  const eventCollection = rulesHelpers.lookupCollectionInEvent(eventObject);
 
-  if (messageReceiveCount !== 1) {
-    log.debug(`message ${message.MessageId} from queue ${queueUrl} is being processed ${messageReceiveCount} times`);
-  }
+  const rulesToSchedule = rulesHelpers.filterRulesbyCollection(rulesForQueue, eventCollection);
```
laurenfrederick, line 81: What if there is no collection in the event?

packages/api/lambdas/sqs-message-consumer.js
laurenfrederick: Can we just handle the error and write a log message like 'No SQS rules configured. Aborting.'?

packages/api/lambdas/sqs-message-consumer.js
```diff
@@ -18,66 +18,95 @@ const Rule = require('../models/rules');
...
+  const rules = await rulesModel.queryRules({
+    type: 'sqs',
+    state: 'ENABLED'
+  });
 
   const messageLimit = event.messageLimit || 1;
   const timeLimit = event.timeLimit || 240;
 
-  await Promise.all(rules.map(async (rule) => {
+  const rulesByQueueMap = rules.reduce((map, rule) => {
```
laurenfrederick, line 30: Super nitpicky, so don't change it, but I feel like this could be easier to understand as a separate function with separate tests.

packages/api/lambdas/sqs-message-consumer.js
markdboyd: For sure

packages/api/lib/rulesHelpers.js
markdboyd: The refactor isn't quite equivalent. Because of the `filter` operation, rules without a timeout get filtered out, so if there was only one rule without a timeout, then there would be an empty array for `.reduce` to operate on. Also, I *think* that we want to return `undefined` if none of the rules have a visibility timeout, because a value of `undefined` won't be used by the consumer when it goes to fetch messages whereas any numeric value would be.

The tests for `getMaxTimeoutForRules` fail with this change:

https://github.com/nasa/cumulus/pull/1638/files?file-filters%5B%5D=.js#diff-023361c061b8df15ce4e31d181c7e2adR108

packages/api/lambdas/sqs-message-consumer.js
markdboyd: If there is no collection in the event, then `lookupCollectionInEvent` returns an empty object:

https://github.com/nasa/cumulus/pull/1638/files?file-filters%5B%5D=.js#diff-023361c061b8df15ce4e31d181c7e2adR229

And if `filterRulesbyCollection` receives an empty object for `eventCollection`, it will simply return all the rules passed into, so effectively no filtering by collection will happen and workflows will queued for all rules. I added an explicit test for this:

https://github.com/nasa/cumulus/pull/1638/files?file-filters%5B%5D=.js#diff-82db426278e52d5ff0bfdbad86db20ebR282

packages/api/tests/models/rules/test-sns-rules.js
```diff
@@ -0,0 +1,316 @@
...
+    SubscriptionArn: rule.rule.arn
+  }));
+
+  t.teardown(() => {
+    snsStub.restore();
+    unsubscribeSpy.restore();
+  });
+});
+
+test.serial('multiple rules using same SNS topic can be created and deleted', async (t) => {
```
markdboyd, line 268: This is the test I added to confirm that we can create/delete multiple SNS rules using the same topic. The rest of the tests in here were moved from `test-rules-model.js` just to keep things more organized logically

PR #1639, opened 2020-05-15
add ability to disable deprecation warnings

markdboyd:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


example/cumulus-tf/main.tf
```diff
@@ -141,6 +141,8 @@ module "cumulus" {
   additional_log_groups_to_elk  = var.additional_log_groups_to_elk
 
   tags = local.tags
+
+  enable_deprecation_warnings = var.enable_deprecation_warnings
```
markdboyd, line 145: I only added a TF variable so that we could see these warnings in the task for our example Cumulus deployments, but upon further review that may not even be necessary

packages/common/util.js
```diff
@@ -31,11 +31,16 @@ const log = require('./log');
...
-  if (!('NO_DEPRECATION_WARNINGS' in process.env)) {
-    log.warn(message);
+  const key = `${name}-${version}`;
+  if (!warned.has(key)) {
+    warned.add(key);
+    let message = `${name} is deprecated after version ${version} and will be removed in a future release.`;
+    if (alternative) message += ` Use ${alternative} instead.`;
+    if (process.env.NO_DEPRECATION_WARNINGS !== 'true') {
+      log.warn(message);
+    }
```
chuckwondo, line 43: Thanks for this. This will massively reduce output even when warnings are not disabled.

This should also be updated in `packages/aws-client/src/StepFunctions.ts`, where `deprecate` is duplicated. I don't know if there are other places.

tf-modules/ingest/discover-granules-task.tf
```diff
@@ -15,7 +15,8 @@ resource "aws_lambda_function" "discover_granules_task" {
     variables = {
       stackName                     = var.prefix
       GranulesTable                 = var.dynamo_tables.granules.name
-      CUMULUS_MESSAGE_ADAPTER_DIR = "/opt/"
+      CUMULUS_MESSAGE_ADAPTER_DIR   = "/opt/"
+      NO_DEPRECATION_WARNINGS       = var.enable_deprecation_warnings
```
chuckwondo, line 19: Hmmm. This looks off, and the names should be consistent. Consider changing `enable_deprecation_warnings` to `no_deprecation_warnings` and set its default value to `false`.

Alternatively, change `NO_DEPRECATION_WARNINGS` to `ENABLE_DEPRECATION_WARNINGS` and set the default value of `enable_deprecation_warnings` to `true`.

Either way will keep default behavior consistent with current behavior.

packages/common/util.js
```diff
@@ -31,10 +31,13 @@ const log = require('./log');
  *
  * @alias module:util
  */
+const warned = new Set();
 exports.deprecate = (name, version, alternative) => {
-  let message = `${name} is deprecated after version ${version} and will be removed in a future release.`;
-  if (alternative) message += ` Use ${alternative} instead.`;
-  if (!('NO_DEPRECATION_WARNINGS' in process.env)) {
+  const key = `${name}-${version}`;
+  if (!warned.has(key)) {
```
yjpa7145, line 37: 🐜 🐜 🐜 

By returning quickly, you don't need the following if-block.

```suggestion
  if (warned.has(key)) return;
```

packages/common/util.js
```diff
@@ -31,10 +31,13 @@ const log = require('./log');
  *
  * @alias module:util
  */
+const warned = new Set();
```
yjpa7145, line 34: You might want to check the generated docs and make sure that the JSDoc here isn't now documenting the `const warned ...` line.

yjpa7145, line 34: I like the use of the `Set` object here.

yjpa7145, line 34: Given the fact that Node will [create a closure around everything in this module](https://nodejs.org/api/modules.html#modules_the_module_wrapper), creating the `warn` Set object here works and keeps it isolated from the rest of the program. It _is_ still accessible by the other functions in this module.

I don't think it's worth holding up the review for, but one thing to consider might be creating an explicit closure for just the `warn` object and the function that actually uses it. Doing that will make it very clear where that `warned` object is intended to be used.

```js
exports.deprecate = (() => {
  const warned = new Set();

  return (name, version, alternative) => {
    if (warned.has(name)) return;

    let message = `${name} is deprecated after version ${version} and will be removed in a future release.`;
    if (alternative) message += ` Use ${alternative} instead.`;

    log.warn(message);
  };
})();
```

packages/aws-client/src/StepFunctions.ts
```diff
@@ -21,13 +21,19 @@ export const doesExecutionExist = (describeExecutionPromise: Promise<unknown>) =
...
 
 // Copied here to avoid requiring @cumulus/common just for this function
-const deprecate = (name: string, version: string, alternative?: string) => {
-  let message = `${name} is deprecated after version ${version} and will be removed in a future release.`;
-  if (alternative) message += ` Use ${alternative} instead.`;
-  if (!('NO_DEPRECATION_WARNINGS' in process.env)) {
+const deprecate = (() => {
+  const warned = new Set();
+
+  return (name:string, version: string, alternative?:string) => {
```
yjpa7145, line 27:
```suggestion
  return (name: string, version: string, alternative?: string) => {
```

packages/common/docs/API.md
```diff
@@ -227,7 +227,7 @@ isNil(undefined); // => true
 ```
 
 * [util](#module_util)
-    * [exports.deprecate(name, version, [alternative])](#exp_module_util--exports.deprecate) ⏏
+    * [exports.deprecate](#exp_module_util--exports.deprecate) ⏏
```
yjpa7145, line 230: Ugh, stupid `jsdoc2md`. This method signature is going to need some love.

packages/common/docs/API.md
```diff
@@ -245,10 +245,13 @@ isNil(undefined); // => true
...
-### exports.deprecate(name, version, [alternative]) ⏏
-Mark a piece of code as deprecated
+### exports.deprecate ⏏
+Mark a piece of code as deprecated.
 
-**Kind**: Exported function  
+Each deprecation notice for a given name and version combination will
+only be printed once.
+
+**Kind**: Exported member
```
yjpa7145, line 254: Not sure why this is an "Exported member" now instead of an "Exported function". Yay `jsdoc2md`.

PR #1641, opened 2020-05-19
Prevent redeployment of unchanged Lambdas

yjpa7145:
Every time we build the zip files for our Lambda functions, the checksum of those Lambdas was changing, which was resulting in all of the Lambdas being redeployed. This was happening even if there were no actual changes to the Lambda function. This was being caused by the changing timestamps of the files being inserted into the zip files.

I have added `bin/zip.js`, which builds a zip file but hard-codes the timestamp of the files in the zip file. Since we are just using the zip to deploy a Lambda function, the timestamp really doesn't matter.

I updated all of the scripts that generate the Lambda zip files to use this script and verified that only Lambdas that have actually changed are being deployed as a result.

Incidentally, because we are not using `package-lock.json` files, we are still going to see Lambdas redeployed when we have not made any changes, because there could be patch releases of some package that the Lambda depends on.

### DIFF
--- a/bin/zip.js
+++ b/bin/zip.js
@@ -1,12 +1,21 @@
 'use strict';
 
+/**
+ * This script should be used to build zip files for deployment to Lambda. It
+ * hard-codes the timestamp of each file in the generatd zip so that identical
+ * files with different timestamps will result in an identical Lambda.
+ */
+
 const fs = require('fs');
 const archiver = require('archiver');
 
 const isDirectory = (x) => fs.statSync(x).isDirectory();
 
+// The first command line argument is the name of the zip file to generate. The
+// other arguments are the files to add to the zip.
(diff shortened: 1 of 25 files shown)


bin/zip.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+const fs = require('fs');
+const archiver = require('archiver');
+
+const isDirectory = (x) => fs.statSync(x).isDirectory();
+
+const [zipPath, ...files] = process.argv.slice(2);
+
+const date = new Date('2009-06-12');
```
laurenfrederick, line 10: Can you comment this line and maybe the file please? I can just see us in a year being like wtf.

yjpa7145: Done: 11162946d3c310f12cf1275d8b4b0e6b6d0dc4ac

PR #1643, opened 2020-05-19
Update DiscoverGranules to get provider_path from config, not config.collection

yjpa7145:
The `DiscoverGranules` task previously expected its `provider_path` to be specified in `event.config.collection.provider_path`. It is now expected in `event.config.provider_path`. This is a step toward [CUMULUS-1969](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1969): "Set provider_path on the rule instead of the collection."


example/cumulus-tf/browse_example.tf
```diff
@@ -19,6 +19,7 @@ module "discover_granules_browse_example_workflow" {
           "event.$": "$",
           "task_config": {
             "provider": "{$.meta.provider}",
+            "provider_path": "{$.meta.collection.provider_path}",
```
markdboyd, line 22: Do we want to also update the ingest browse data cookbook to specify that the rule should have the provider path?

https://nasa.github.io/cumulus/docs/data-cookbooks/browse-generation

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### BREAKING CHANGES
 
+- **CUMULUS-1969**
```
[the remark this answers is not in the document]
yjpa7145: Done: caa432feb

example/cumulus-tf/browse_example.tf
yjpa7145: Done: caa432feb

markdboyd: I'm wondering if it should also be updated here: https://nasa.github.io/cumulus/docs/data-cookbooks/run-tasks-in-lambda-or-docker

The purpose of that doc isn't necessarily to show a "working" discover granules workflow, but people might easily copy/paste from it and be confused when it doesn't work

yjpa7145: Fixed here: 1b5706565

PR #1644, opened 2020-05-20
Update discover_granules and browse_example workflows to use meta.provider_path

yjpa7145:
Rather than using the `provider_path` stored in the collection, our `discover_granules` and `browse_example` workflows should set and use `meta.provider_path` instead.


yjpa7145:
> Looks great! One unrelated question: when are these references to provider path on the collection going to get cleaned up?
> 
> https://nasa.github.io/cumulus/docs/operator-docs/discovery-filtering
> https://nasa.github.io/cumulus/docs/data-cookbooks/sips-workflow#collection

There will be another related PR that removes `provider_path` from collections once the other related PRs have been merged. I wanted to update the code to make `collection.provider_path` unused, make sure it all works, then remove `collection.provider_path`.

PR #1645, opened 2020-05-20
Update DiscoverPdrs to get provider_path from config, not config.collection

yjpa7145:
The `DiscoverPdrs` task previously expected its `provider_path` to be specified in `event.config.collection.provider_path`. It is now expected in `event.config.provider_path`.


CHANGELOG.md
```diff
@@ -8,6 +8,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### BREAKING CHANGES
 
+- **CUMULUS-1969**
+  - The `DiscoverPdrs` task now expects `provider_path` to be provided at
+    `event.config.provider_path`, not `event.config.collection.provider_path`
+  - `event.config.provider_path` is now a required parameter of the
+    `DiscoverGranules` task
```
laurenfrederick, line 15:
```suggestion
    `DiscoverPdrs` task
```

tasks/discover-pdrs/tests/index.js
```diff
@@ -192,20 +174,14 @@ test('test pdr discovery with SFTP assuming some PDRs are new', async (t) => {
...
-    } catch (e) {
-      if (e instanceof RemoteResourceError) {
-        t.pass('Ignoring this test. Test server seems to be down');
-      } else t.fail(e);
-    }
+    await validateOutput(t, output);
+
+    t.is(output.pdrs.length, 4);
+    const names = output.pdrs.map((p) => p.name);
+    newPdrs.forEach((pdr) => t.true(names.includes(pdr)));
```
laurenfrederick, line 184: We need to remember this one in case we start seeing these fails pop up. Guessing they were intermittent.

markdboyd:
I think we should update the example rule in this doc to show `rule.meta.provider_path`

https://nasa.github.io/cumulus/docs/data-cookbooks/sips-workflow

yjpa7145:
> I think we should update the example rule in this doc to show `rule.meta.provider_path`
> 
> [nasa.github.io/cumulus/docs/data-cookbooks/sips-workflow](https://nasa.github.io/cumulus/docs/data-cookbooks/sips-workflow)

For this particular PR there isn't a required change needed to that file, since the `provider_path` is still coming from the collection: https://github.com/nasa/cumulus/pull/1645/files#diff-0fdb64609fdea5b06c91ff7efc6e6780R25

When the `provider_path` is removed from the collection, though, we will need to update this file. I'll make a note.

PR #1646, opened 2020-05-20
Update discover_and_queue_pdrs workflow to use meta.provider_path

yjpa7145:
Rather than using the `provider_path` stored in the collection, our `discover_and_queue_pdrs` workflow should set and use `meta.provider_path` instead.


example/spec/parallel/ingest/ingestFromPdrSpec.js
```diff
@@ -159,9 +150,11 @@ describe('Ingesting from PDR', () => {
       workflowExecution = await buildAndExecuteWorkflow(
         config.stackName,
         config.bucket,
-        taskName,
-        collection,
-        provider
+        workflowName,
+        undefined,
```
markdboyd, line 154: So this doesn't workflow really doesn't need a collection because it's a PDR based workflow?

yjpa7145: Apparently not. The `discoverPdrs` task does not use collection, and the `queuePdrs` task takes a collection, but it's optional and is only passed along to the triggered workflow. The workflow triggered by `queuePdrs` is `parse_pdr_workflow`, which does not use a collection.

PR #1648, opened 2020-05-20
CUMULUS-1952: HTTP(S) ingest with EDL (Basic auth) support

nemreid:

Addresses [CUMULUS-1952: As a Cumulus Integrator, I would like to ingest files via HTTP/S from locations that require URS authentication](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1952)

## Changes

- Updated HTTP(S) provider to accept username/password for Basic authorization. This change adds support for Basic Authorization such as Earthdata login redirects to ingest (i.e. as implemented in SyncGranule), but not to discovery (i.e. as implemented in DiscoverGranules). Discovery still expects the provider's file system to be publicly accessible, but not the individual files and their contents.
- **NOTE**: Using this in combination with the HTTP protocol may expose usernames and passwords to intermediary network entities. HTTPS is highly recommended.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


nemreid:
I've realized I forgot to add username/password decryption support. Working on that now.

packages/ingest/util.js
```diff
@@ -1,6 +1,16 @@
 'use strict';
 
 const mime = require('mime-types');
+const KMS = require('@cumulus/aws-client/KMS');
+const { S3KeyPairProvider } = require('@cumulus/common/key-pair-provider');
+
+const decrypt = async (ciphertext) => {
+  try {
+    return await KMS.decryptBase64String(ciphertext);
```
markdboyd, line 9: So we try base 64 decryption otherwise try to decrypt with the S3 keypair that was uploaded during deployment? Does the S3 keypair ever come into play for basic auth?

CHANGELOG.md
```diff
@@ -70,6 +70,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - **CUMULUS-1912**
   - Updated reconciliationReports endpoints to query against Elasticsearch, delete report from both database and s3
   - Added `@cumulus/api-client/reconciliationReports`
+- **CUMULUS-1952**
+  - Updated HTTP(S) provider to accept username/password for Basic authorization. This change adds support for Basic Authorization such as Earthdata login redirects to ingest (i.e. as implemented in SyncGranule), but not to discovery (i.e. as implemented in DiscoverGranules). Discovery still expects the provider's file system to be publicly accessible, but not the individual files and their contents.
```
markdboyd, line 74: I suppose we didn't implement this for discovery given that it was outside the scope of the request?

packages/ingest/test/test-HttpsProviderClient.js
```diff
@@ -67,17 +102,72 @@ test('download() downloads a file', async (t) => {
...
+  const localPath = path.join(tmpdir(), randomString());
+  try {
+    await httpsProviderClient.download(protectedFile, localPath);
+    t.is(fs.existsSync(localPath), true);
+  } finally {
+    fs.unlinkSync(localPath);
+  }
+});
+
+test('HttpsProviderClient supports basic auth with redirects for sync', async (t) => {
```
markdboyd, line 155: Is there a test in here for using encrypted credentials with basic auth?

packages/ingest/HttpProviderClient.js
```diff
@@ -28,25 +29,45 @@ const validateHost = (host) => {
...
     this.endpoint = buildURL({
       protocol: this.protocol,
       host: this.host,
       port: this.port
     });
   }
 
+  async setUpGotOptions() {
+    if (this.encrypted === true) {
+      this.username = await decrypt(this.providerConfig.username);
```
markdboyd, line 50: I probably lack some fundamental understanding of basic auth: but how does this scenario work? A user configures encrypted (base64?) username/password for the provider, then we decrypt those credentials before putting them in the basic auth header? Is the point here just to avoid storing un-encrypted credentials on the provider and not having them exposed in the cumulus message?

packages/ingest/util.js
nemreid: S3 Keypair stuff is legacy support, shouldn't really come into play. The meat of this change is DRYing up this function between the HTTP/STP/FTP provider clients.

CHANGELOG.md
nemreid: Yes, outside of the scope. Also a completely separate implementation because we use a completely different library. But the ticket and user both indicate no need for discovery support.

packages/ingest/HttpProviderClient.js
nemreid: A user can provide the plaintext username/password when configuring the provider, but with `encrypted: true` that password will be encrypted when we create the provider record. So yes, this setting has us encrypt the credentials and store the encrypted copy to avoid exposing them in the record and message. Brings this ProviderClient in line with FTP/SFTP, which also support username/password.

packages/ingest/test/test-HttpsProviderClient.js
nemreid: I have yet to figure out how to adequately test this. Should I mock out `decrypt`, or go the end-to-end route with full KMS testing?

markdboyd: I would mock out `decrypt`. Proving that the decryption works is a separate unit test. So the test here is "end to end" in the sense that it proves that encrypted credentials are unencrypted before being used for auth and a separate test verifies that the decryption itself works.

PR #1651, opened 2020-05-21
Add a timeout on the parallel tests

laurenfrederick:
**Summary:** Add a timeout to parallel test command

## Changes

* Added a 50 minute timeout to the parallel test command so if the tests hang we can see the output after it times out. `ruleSpec` has been hanging recently and this will help debug since it hasn't been able to be reproduced locally.

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


example/scripts/tests-parallel.sh
```diff
@@ -14,7 +14,7 @@ testOutputDir=scripts/test_output
 rm -r -f $testOutputDir
 mkdir -p $testOutputDir
 
-echo "" | ../node_modules/.bin/parallel -j 0 sh scripts/run_test.sh  $testOutputDir ::: $TESTS
+echo "" | ../node_modules/.bin/parallel -j 0 --timeout 3000 sh scripts/run_test.sh  $testOutputDir ::: $TESTS
```
markdboyd, line 17: Does 3000 = 5 mins, not 50 minutes?

laurenfrederick: Yep, thank you. I had been going back and forth and playing around with different values.

PR #1653, opened 2020-05-21
Update mocking for CMR unit tests

markdboyd:
The unnecessary mocking was causing these tests to fail persistently for me when I have Localstack running. 

`nock.enableNetConnect('127.0.0.1')` means that only requests to 127.0.0.1 will be able to go to network and the rest will be blocked (unless they are mocked with `nock` separately). Since our commands to Localstack to go `localhost` and not 127.0.0.1, I was getting this error:

```
  getCmrPassword returns password from AWS secret when set
  Rejected promise returned by test. Reason:
  NetConnectNotAllowedError {
    code: 'NetworkingError',
    hostname: 'localhost',
    message: 'Nock: Disallowed net connect for "localhost:4584/"',
```

I was tempted to remove this mocking altogether - but I guess the point of it is to ensure that no requests to CMR ever go to network, because we're only intending to hit mocked URLs. So instead I updated the command to `nock.enableNetConnect(/(localhost|127.0.0.1)/);`, which will allow requests to `localhost` or `127.0.0.1` to get through


markdboyd:
@laurenfrederick Fixed `test-UmmUtils.js`: 9337e39

PR #1656, opened 2020-05-22
Relocate some workflow endpoint tests

yjpa7145:
Trying to slowly clean up an intermittently failing integration test.


packages/api/tests/endpoints/workflows.js
```diff
@@ -95,7 +95,6 @@ test('GET with no path parameters and an authorized user returns a list of workf
     .set('Authorization', `Bearer ${jwtAuthToken}`)
     .expect(200);
 
-  t.is(response.status, 200);
```
markdboyd, line 98: Why remove these?

yjpa7145: If the status is not `200`, the `.expect(200)` on line 96 will have already thrown an exception.

PR #1659, opened 2020-05-26
getExecutionWithStatus retries if the execution does not exist

yjpa7145:
The `getExecutionWithStatus` function was not retrying if the execution did not yet exist. Fixed it so that it will retry even if the execution does not exist.

### DIFF
--- a/packages/integration-tests/tests/test-Executions.js
+++ b/packages/integration-tests/tests/test-Executions.js
@@ -1,0 +1,44 @@
+'use strict';
+
+const test = require('ava');
+const cryptoRandomString = require('crypto-random-string');
+const { getExecutionWithStatus } = require('../Executions');
+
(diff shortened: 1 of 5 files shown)


packages/integration-tests/tests/test-Executions.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+const test = require('ava');
+const cryptoRandomString = require('crypto-random-string');
+const { getExecutionWithStatus } = require('../Executions');
+
+const randomId = (prefix, separator = '-') =>
+  `${prefix}${separator}${cryptoRandomString({ length: 6 })}`;
+
+test('getExecutionWithStatus() will retry if the execution does not exist', async (t) => {
```
markdboyd, line 10: :100:

PR #1660, opened 2020-05-26
Make granule deletion faster

yjpa7145:
I noticed that the test for deleting a granule with a file that does not exist in S3 was taking a long time (7 seconds). I tracked the problem down to the fact that, when you call `GranuleModel.get()` (which `del` was doing), it fetches each of the files in the granule to determine their size. The `getObjectSize` function retries getting a non-existent file three times, so that was the cause of the 7-second delay.

Because deleting a granule does not require knowing the size of the granule, I added a `getRecord` method to the `Granules` model which only returns the raw DB record. I then updated the granules endpoint's `del` function to use `getRecord`.

### DIFF
--- a/packages/api/endpoints/granules.js
+++ b/packages/api/endpoints/granules.js
@@ -12,6 +12,7 @@ const { asyncOperationEndpointErrorHandler } = require('../app/middleware');
 const Search = require('../es/search').Search;
 const indexer = require('../es/indexer');
 const models = require('../models');
+const FileUtils = require('../lib/FileUtils');
 const { deconstructCollectionId } = require('../lib/utils');
 
 /**
@@ -133,7 +134,7 @@ async function del(req, res) {
   log.info(`granules.del ${granuleId}`);
 
   const granuleModelClient = new models.Granule();
-  const granule = await granuleModelClient.get({ granuleId });
+  const granule = await granuleModelClient.getRecord({ granuleId });
 
   if (granule.detail) {
     return res.boom.badRequest(granule);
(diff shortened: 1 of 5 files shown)


packages/api/endpoints/granules.js
```diff
@@ -133,7 +134,7 @@ async function del(req, res) {
   log.info(`granules.del ${granuleId}`);
 
   const granuleModelClient = new models.Granule();
-  const granule = await granuleModelClient.get({ granuleId });
+  const granule = await granuleModelClient.getRecord({ granuleId });
```
markdboyd, line 137: Do we still need to support the `translateGranule` logic? If so, is it because we never provided a migration from old granule schema to new granule schema? I don't love having two separate methods for getting a granule

packages/api/tests/models/granules/test-granules-model.js
```diff
@@ -391,6 +391,23 @@ test('get() will correctly return a granule file stored using the new schema', a
   );
 });
 
+test('getRecord() returns a granule record from the database', async (t) => {
+  const granule = fakeGranuleFactoryV2();
+
+  await awsServices.dynamodbDocClient().put({
```
markdboyd, line 397: Curious: why not `granule.create` here?

yjpa7145: When I first wrote these tests, I wanted to make sure that matching errors in `create` and `get` didn't result in tests passing that shouldn't. I don't have a particularly strong opinion on it now, but I was matching the way that the other tests in the file are written.

packages/api/endpoints/granules.js
yjpa7145: > is it because we never provided a migration from old granule schema to new granule schema?

Correct.

> I don't love having two separate methods for getting a granule

I agree. Ideally, I think that `get` should return the raw record (like `getRecord` is doing now), and whoever called `get` could then have the option of calling `translateGranule` on the result if they wanted to. I'll make a note and see about coming back to that in an IP sprint.

markdboyd: If the only benefit here is a speedup in one of our tests, I'd prefer to not add more code as a temporary solution (since it has a way of hanging around) and instead just wait until we can possibly make the entire fix at once. And in fact, since my work now focuses on granule deletion, I may be able to do this as part of my ticket

markdboyd:
I have included these changes in #1668 so I think this can be closed

PR #1665, opened 2020-05-28
MoveGranule waits for input objects to exist

yjpa7145:
There have been cases where the SyncGranule task is copying a file to S3 but, when the MoveGranules task runs later in the workflow, the object cannot be found. This is likely being caused by S3's eventual consistency model and the fact that we are calling `headObject` on the object to be created in the SyncGranule task.

This PR updates the MoveGranules task to wait 30 seconds for the object to exist in S3 before trying to move it.


packages/aws-client/src/S3.ts
```diff
@@ -152,6 +153,35 @@ export const s3ObjectExists = (params: { Bucket: string, Key: string }) =>
...
+/**
+ * Wait for an object to exist in S3
+ *
+ * @param {Object} params
+ * @param {string} params.bucket
+ * @param {string} params.key
+ * @param {number} [params.interval=1000]
+ * @param {number} [params.timeout=30000]
+ * @returns {Promise<undefined>}
+ */
```
chuckwondo, line 165: Do we want to add some descriptions to the params and return?

yjpa7145: Done: 66cc746b9

PR #1669, opened 2020-05-29
RuleSpec intermittent test fail fix

laurenfrederick:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


example/spec/parallel/testAPI/ruleSpec.js
```diff
@@ -226,13 +257,16 @@ describe('When I create a one-time rule via the Cumulus API', () => {
...
         async () => {
-          const listRulesResponse = await rulesApi.listRules({ prefix });
+          let pageNumber = 1;
+          let ruleResults = await queryForRule(helloWorldRule.name, prefix, pageNumber);
 
-          const rules = JSON.parse(listRulesResponse.body).results;
-
-          const ruleNames = rules.map((x) => x.name);
+          // If the rule list spans multiple pages, query the other pages
+          while (!ruleResults.ruleFound && ruleResults.count > (LIST_RULES_LIMIT * pageNumber)) {
```
markdboyd, line 264: Rather than paging through all the "list rules" results, can't we use the list rules endpoint to query directly for the rule that we want?

https://github.com/nasa/cumulus/commit/adc0877791ca106ebe58e0f0c2f224bcd17a8830#diff-e3907763a932ff41fd1df5a2e3169932R242

I say that while being aware that I'm not sure the test was stable as previously written, but it's confusing to me as to why it wouldn't be

markdboyd: Also, the intent of this test was changed along the way:

https://github.com/nasa/cumulus/commit/1bd8c669e4850d90bd8d64df5124a03e29ebc8de#diff-e3907763a932ff41fd1df5a2e3169932R225

As previously written it was trying to confirm that the rule existed in the list with the `updatedCheck` value set in a previous test. But now we're just checking to be sure that the a rule with the right name exists. I'm not really saying which version of the test is more correct, but flagging it

packages/integration-tests/Executions.js
```diff
@@ -30,11 +32,31 @@ const findExecutionArn = async (prefix, matcher, options = { timeout: 0 }) =>
...
+          // eslint-disable-next-line no-await-in-loop
+          const response = await executionsApi.getExecutions({
+            prefix,
+            query: {
+              limit: EXECUTION_LIST_LIMIT,
+              page: pageNumber += 1
+            }
+          });
+          executions = JSON.parse(response.body);
+          execution = executions.results.find(matcher);
```
markdboyd, line 58: Similarly, I wonder here if we couldn't query ES directly with the properties we're expecting in the `matcher`. I suppose if this is working it's not essential though

example/spec/parallel/testAPI/ruleSpec.js
laurenfrederick: It does get checked for the updates above, so I'll leave that. I guess we could just check that we can search it in the list. I think the test is trying to test the list functionality called by the dashboard, but I guess it's testing the same thing.

packages/integration-tests/Executions.js
laurenfrederick: Yeah so this one gets a bit more complicated. I looked into it and there are some instances where we're trying to match on `originalPayload.testExecutionId` which I think would be impossible through the field queries without a more advanced ES query.

PR #1670, opened 2020-05-29
Remove "provider_path" from "collection"

yjpa7145:
- Remove the "provider_path" field from the allowed properties on collection items
- Update the `@cumulus/api-client/collections.create` function to remove `provider_path` before creating a collection. This was much simpler than updating all of the `.json` files that contain our collection fixtures used in our tests.
- Update the `Collections` model to not return deprecated fields (like `provider_path`) from `.get` calls. Because `Collection` inherited from `Manager`, ensuring that `provider_path` was never returned would have required overriding every method from the parent class that returned a collection. There was also the risk that a method could be added to the parent class in the future that would have returned the colleciton without `provider_path` removed. To handle that, I refactored the `Collection` class to no longer inerhit from `Manager` and [use delegation](https://media.pragprog.com/titles/tpp20/inheritance-tax.pdf) to a `Manager` instance instead.
- Fixed a bug in the `Collection.create` method so that the `create` method now properly supports either a single item, or an array of items.
- Updated the `@cumulus/aws-client/SNS.publishSnsMessage` function to suppress debug messages in test mode.


CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     `DiscoverPdrs` task
   - `event.config.collection` is no longer a parameter to the `DiscoverPdrs`
     task
+  - Collections no longer support the `provider_path` property. The tasks that
```
yjpa7145, line 18: Update data model changelog as well

packages/api-client/collections.js
```diff
@@ -21,7 +24,7 @@ const createCollection = async ({ prefix, collection, callback = invokeApi }) =>
     resource: '/{proxy+}',
     headers: { 'Content-Type': 'application/json' },
     path: '/collections',
-    body: JSON.stringify(collection)
+    body: JSON.stringify(omit(collection, deprecatedFields))
```
yjpa7145, line 27: Remove provider_path from all the fixtures.

packages/aws-client/src/SNS.ts
```diff
@@ -25,6 +25,8 @@ export const publishSnsMessage = async (
 ) =>
   pRetry(
     async () => {
+      if (!snsTopicArn && process.env.NODE_ENV === 'test') return;
```
yjpa7145, line 28: Revert this.

PR #1677, opened 2020-06-02
Improve globalConnectionLimit debug message

yjpa7145:


yjpa7145:
@amarouane-ABDELHAK Is this error message better? Could it still be improved?

amarouane-ABDELHAK:
@yjpa7145 Yes, thank you that error message is descriptive enough for us to know we need to increase the provider connection limit

PR #1687, opened 2020-06-05
Update collection reporting to match rest of reporting

markdboyd:
This refactors collection reporting to work the same as execution, PDR, and granule reporting:

- Record is written to Dynamo
- Dynamo event stream triggers Lambda that publishes record to SNS

https://nasa.github.io/cumulus/docs/data-cookbooks/ingest-notifications

Besides consistency, a nice side effect of this change is that we should no longer see annoying warning messages in Ava tests about how a collection could not be published to the topic.

Since the publishing to SNS no longer happens directly in the model and relies on asynchronous triggers, I had to add an integration test to prove this is working, but the test runs in < 5 seconds, so it seemed okay.


packages/api/lambdas/publish-collections.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
```
yjpa7145, line 1: This Lambda needs tests.

chuckwondo: Good call. Will do.

chuckwondo: Tests added. Please review.

PR #1700, opened 2020-06-10
Update get object size

yjpa7145:
The `@cumulus/aws-client/S3.getObjectSize` function was retrying three times if the object did not exist, which introduced a 7-second delay. This retry value was not configurable.

Adding a retry parameter was an option but, because the function used positional parameters, adding an optional `retries` parameter would make it more difficult to add additional parameters (like `version`) in the future.

To address those issues, this PR changes the function signature to take named parameters. It also removed the retry logic in favor of a simpler, more targeted function. In cases where the object may not exist yet, but will shortly, `waitForObjectToExist()` should be used before fetching the object size.

Lastly, in order to make testing more flexible, the function now takes an `s3` parameter. This should be an `AWS.S3` instance. Previously, the function had been using a singleton `AWS.S3` client, and did not support a way to specify a different client.


packages/api/lib/FileUtils.js
```diff
@@ -99,15 +106,15 @@ const filterDatabaseProperties = (file) =>
...
+const buildDatabaseFile = (s3, providerURL, file) =>
   flow([
     setBucket,
     setKey,
     setChecksum,
     setFileName,
     partial(setSource, providerURL),
     parseSource,
-    setS3FileSize // This one is last because it returns a Promise
+    partial(setS3FileSize, s3) // This one is last because it returns a Promise
```
chuckwondo, line 117: While I do like to see your use of some "functional" functions, I'm not crazy about the use of `flow` with all of the setter functions that each create a new object, but I'm not going to hold up approval because of it.

Just for future reference, take a look at Ramda's `applySpec` function. Here's how I might construct things with it:

```javascript
import { applySpec } from 'ramda';

setS3FileSize(s3, parseSource(applySpec({
  bucket: getBucket,
  key: getKey,
  checksum: getChecksum,
  fileName: getFileName,
  source: partial(buildFileSourceURL, providerURL)
})(file)));
```

PR #1709, opened 2020-06-12
Migrate common/log to TS

yjpa7145:
Beginning an incremental migration of `@cumulus/common` to Typescript. This migrates the `log` module.

This also deprecates the `convertLogLevel` function, which is only used in the `api` package, and moves that function to the `api` package.


packages/common/src/log.ts
```diff
@@ -86,7 +84,10 @@ function trace(...args) {
  * @param {string/number} level - log level in string or number
  * @returns {number/string} - level in number or string
  */
-function convertLogLevel(level) {
+// @ts-ignore
+export function convertLogLevel(level) {
+  warn('@cumulus/common/log.convertLogLevel() is deprecated after version 1.23.2 and will be removed in a future release.');
```
chuckwondo, line 89: Is this directly using `warn` in order to avoid a potential infinite loop via the `deprecate` function?

yjpa7145: Correct. The `util` module (which contains `deprecate`) includes `log`, so I didn't want `log` to then include `util`.

PR #1710, opened 2020-06-12
Migrate common/util to TS

yjpa7145:


packages/common/src/util.ts
```diff
@@ -191,9 +195,13 @@ exports.setErrorStack = (error, newStack) => {
  * @param {Object} obj - object to update
  * @returns {Object} a shallow clone of the object with updated property name
  *
+ * @deprecated
+ *
  * @alias module:util
  */
-exports.renameProperty = (from, to, obj) => {
+// @ts-ignore
+export const renameProperty = (from, to, obj) => {
```
chuckwondo, line 203: Why are you using `@ts-ignore` on some of the deprecated functions and not on others?

packages/common/src/util.ts
```diff
@@ -124,63 +125,66 @@ exports.omit = (objectIn, keys) => {
...
  * const isEven = (x) => x % 2 === 0;
  * const isOdd = negate(isEven);
  *
  * isOdd(2); // => false
  * isOdd(3); // => true
  */
-exports.negate = (predicate) => (...args) => !predicate.apply(this, args);
+// @ts-ignore
+export const negate = (predicate) => {
+  deprecate('@cumulus/common/util.omit()', '1.23.2', 'lodash/omit');
```
chuckwondo, line 139: Copy-paste error on the `deprecate` args?

packages/common/src/util.ts
```diff
@@ -69,33 +68,31 @@ exports.sleep = (waitPeriodMs) => {
...
  * @param {string} name - A base name for the temp dir, to be uniquified for the final name
  * @returns {string} The absolute path to the created dir
+ *
  * @private
+ *
+ * @deprecated
  */
-exports.mkdtempSync = (name) => {
+export const mkdtempSync = (name: string) => {
+  deprecate('@cumulus/common/util.noop()', '1.23.2');
```
chuckwondo, line 77: Copy-paste error?

example/spec/helpers/granuleUtils.js
```diff
@@ -154,14 +153,17 @@ const loadFileWithUpdatedGranuleIdPathAndCollection = (
...
-  JSON.parse
-);
+) => {
+  const fileContents = fs.readFileSync(filename, 'utf8');
+
+  return flow([
+    replace(new RegExp('replace-me-granuleId', 'g'), newGranuleId),
+    replace(new RegExp('replace-me-path', 'g'), newPath),
+    replace(new RegExp('replace-me-collectionId', 'g'), newCollectionId),
+    replace(new RegExp('replace-me-stackId', 'g'), stackId),
```
chuckwondo, line 163: Curious. Why the choice of the `RegExp` constructor over literals?

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     - **bucket**
     - **key**
   - The `getObjectSize` function will no longer retry if the object does not exist
+- **CUMULUS-1930**
+  - The `@cumulus/common/util.uuid()` function has been removed
```
chuckwondo, line 25: Why isn't `setErrorStack` also listed here?

CHANGELOG.md
```diff
@@ -182,6 +198,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - The deprecated `@cumulus/common/string.replace` functon has been removed
 - The deprecated `@cumulus/common/string.toLower` functon has been removed
 - The deprecated `@cumulus/common/string.toUpper` functon has been removed
+- The deprecated `@cumulus/common/util.setErrorStack` function has been removed
```
chuckwondo, line 201: Why isn't `uuid` also listed here?

CHANGELOG.md
yjpa7145: `setErrorStack` had been previously deprecated, so removing it isn't considered a breaking change.

CHANGELOG.md
yjpa7145: I hadn't listed it here since it was already listed under "Breaking Changes", but it probably should be. I'll fix that.

example/spec/helpers/granuleUtils.js
yjpa7145: Not sure, I was maintaining what was there before, but I agree that it could be cleaned up a bit. I'll do that.

packages/common/src/util.ts
yjpa7145: Thanks, fixed.

packages/common/src/util.ts
yjpa7145: Thanks.

packages/common/src/util.ts
yjpa7145: It was purely a "this looks hard to migrate and we're gonna get rid of it anyway ..." decision.

packages/sftp-client/index.js
```diff
@@ -119,7 +119,8 @@ class SftpClient {
       Bucket: bucket,
       Key: key,
       Body: pass,
-      ContentType: lookupMimeType(key)
+      // eslint-disable-next-line unicorn/no-null
+      ContentType: mime.lookup(key) || null
```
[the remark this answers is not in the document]
yjpa7145: Not sure why that was done. It's `undefined` now. Incidentally, I'm explicitly setting a value because `mime.lookup()` returns `false` if a type could not be found, which is not what we want.

PR #1731, opened 2020-06-23
Fix edge cases in parseS3Uri

yjpa7145:


packages/aws-client/tests/S3/test-parseS3Uri.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+const test = require('ava');
+const { parseS3Uri } = require('../../S3');
+
+[
+  ['s3://bucket', { Bucket: 'bucket', Key: undefined }],
```
nemreid, line 7: Is an undefined `Key` a useful return value?

yjpa7145: I think so, since we need a way to differentiate between `s3://bucket` and `s3://bucket/`. In the first case you would get `Key === undefined`, and in the second `Key === ''`.

nemreid: Right, I guess what I'm asking is, if we pass that over to an S3 API call, it will throw a `Key is undefined`-type error, won't it? Is it a useful thing for any downstream callers?

yjpa7145: Given the wide range of situations where `parseS3Uri` is called, I think that's a decision that needs to be made by the caller. They will have more context as to whether they want to handle `s3://bucket` and `s3://bucket/` as the same or different.

nemreid: OK. I think I would ideally like to see it log a warning of some sort if there is no discernible key in the URL. Otherwise, approved.

yjpa7145: I've changed this PR to maintain backward compatibility.

PR #1732, opened 2020-06-23
Build the api package using Typescript

yjpa7145:
This doesn't migrate anything _to_ TS yet, but it will give us better code completion and detect some common errors. It will also make it easier to incrementally migrate things to TS.


packages/api/src/es/search.js
```diff
@@ -96,7 +96,8 @@ const esConfig = async (host, metrics = false) => {
   return config;
 };
 
-class BaseSearch {
+// eslint-disable-next-line import/prefer-default-export, node/no-unsupported-features/es-syntax
+export class BaseSearch {
```
chuckwondo, line 100: Out of curiosity, why the need to ignore `node/no-unsupported-features/es-syntax`?

yjpa7145: The `ignore` is required because I added the `export`.

The `export` is required because, when `tsc` tried to build the `.d.ts` files for that file, it failed saying that `BaseSearch` was private. I don't entirely understand why, TS compiling JS seems a little magical. In any case, when that file and the files that use it are actually migrated over to `.ts`, that will go away.

yjpa7145:
> Looks good, but curious as to why the need to manually read the json files?

The `require('../blah.json')`s were using relative paths but, because the location of those JSON files relative to the `.ts` files and their location relative to the generated `.js` files was different, the build was failing. (There are also problems related to caching of those JSON files when you use `require`, but that's only bitten us when we do it in tests so far.)

By determining the actual path of the JSON and then reading them in, we can work around that.

yjpa7145:
We need to sort out the `require('xxx.json')` issues before we can move forward with this.

PR #1746, opened 2020-06-27
CUMULUS-1955

chuckwondo:
[CUMULUS-1955](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1955): Validate CMR file checksum

Add ETag output to sync methods for the various provider clients.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/packages/aws-client/src/S3.ts
+++ b/packages/aws-client/src/S3.ts
@@ -895,7 +895,7 @@ const uploadPartCopy = async (
  * @param {string} params.destinationKey
  * @param {string} [params.ACL] - an [S3 Canned ACL](https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl)
  * @param {boolean} [params.copyTags=false]
- * @returns {Promise<undefined>}
+ * @returns {Promise} promise of result returned by [S3.completeMultipartUpload](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#completeMultipartUpload-property)
  */
 export const multipartCopyObject = async (
   params: {
@@ -958,7 +958,7 @@ export const multipartCopyObject = async (
     );
 
     // Let S3 know that the multi-part upload (copy) is completed
-    await S3MultipartUploads.completeMultipartUpload({
+    return await S3MultipartUploads.completeMultipartUpload({
       UploadId: uploadId,
       Bucket: destinationBucket,
       Key: destinationKey,
--- a/packages/aws-client/src/S3.ts
+++ b/packages/aws-client/src/S3.ts
@@ -895,7 +895,8 @@ const uploadPartCopy = async (
  * @param {string} params.destinationKey
  * @param {string} [params.ACL] - an [S3 Canned ACL](https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl)
  * @param {boolean} [params.copyTags=false]
- * @returns {Promise} promise of result returned by [S3.completeMultipartUpload](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#completeMultipartUpload-property)
+ * @returns {Promise.<{ etag: string }>} object containing the ETag of the
+ *    destination object
  */
 export const multipartCopyObject = async (
   params: {
@@ -958,7 +959,7 @@ export const multipartCopyObject = async (
     );
 
     // Let S3 know that the multi-part upload (copy) is completed
-    return await S3MultipartUploads.completeMultipartUpload({
+    const { ETag: etag } = await S3MultipartUploads.completeMultipartUpload({
       UploadId: uploadId,
       Bucket: destinationBucket,
       Key: destinationKey,
@@ -966,6 +967,8 @@ export const multipartCopyObject = async (
         Parts: uploadPartCopyResponses
       }
     });
+
+    return { etag };
   } catch (error) {
     // If anything went wrong, make sure that the multi-part upload (copy)
     // is aborted.
--- a/packages/aws-client/tests/S3/test-multipartCopyObject.js
+++ b/packages/aws-client/tests/S3/test-multipartCopyObject.js
@@ -73,7 +73,7 @@ test('multipartCopyObject() copies a file between buckets', async (t) => {
     Key: sourceKey
   });
 
-  await multipartCopyObject({
+  const { ETag: etag } = await multipartCopyObject({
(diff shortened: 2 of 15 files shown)


packages/aws-client/src/S3.ts
```diff
@@ -958,7 +958,7 @@ export const multipartCopyObject = async (
     );
 
     // Let S3 know that the multi-part upload (copy) is completed
-    await S3MultipartUploads.completeMultipartUpload({
+    return await S3MultipartUploads.completeMultipartUpload({
```
yjpa7145, line 961: I think that we should probably abstract away how we are implementing the `multipartCopyObject` function. If we return the output of `completeMultipartUpload` like we do here, then it's exposing some details that could make it harder to change later.

Rather than returning the result of `completeMultipartUpload`, consider returning `{ etag: response.ETag }`.

packages/aws-client/tests/S3/test-multipartCopyObject.js
```diff
@@ -73,7 +73,7 @@ test('multipartCopyObject() copies a file between buckets', async (t) => {
     Key: sourceKey
   });
 
-  await multipartCopyObject({
+  const { ETag: etag } = await multipartCopyObject({
```
yjpa7145, line 76: 🐜🐜🐜 Nit-picking question, but what is the benefit to renaming this variable in the test?

packages/sftp-client/index.js
```diff
@@ -105,7 +105,8 @@ class SftpClient {
    * @param {string} remotePath - the full path to the remote file to be fetched
    * @param {string} bucket - destination s3 bucket of the file
    * @param {string} key - destination s3 key of the file
-   * @returns {Promise} s3 uri of destination file
+   * @returns {Promise.<{ s3uri, etag }>} promise resolving to an object
```
yjpa7145, line 108: I don't know jsdoc well enough, is there a way to indicate that s3uri and etag are strings?

packages/sftp-client/index.js
```diff
@@ -125,7 +126,7 @@ class SftpClient {
 
     const result = await S3.promiseS3Upload(params);
     log.info('Downloading to s3 is complete(sftp)', result);
-    return s3uri;
+    return { s3uri, etag: result.ETag };
```
yjpa7145, line 129: 👍

packages/aws-client/src/S3.ts
chuckwondo: I did wrestle with that thought as well, but went the route of returning the full result since that's what the rest of the functions here do.  However, since this is not a simple wrapper like the other functions are, I can see how it would make sense here to reduce the exposure.

However, for consistency with what other functions return, which are AWS response objects, should we keep the property name casing consistent by returning `{ ETag }`?

packages/aws-client/tests/S3/test-multipartCopyObject.js
chuckwondo: No benefit. That's just me being pedantic with JS variable name casing. No other reason. (BTW, it bugs me how the AWS Node SDK doesn't use camel case conventions for object property names.)

packages/ingest/FtpProviderClient.js
```diff
@@ -184,11 +185,11 @@ class FtpProviderClient {
       Body: pass,
       ContentType: lookupMimeType(key)
     };
-    await S3.promiseS3Upload(params);
+    const { ETag: etag } = await S3.promiseS3Upload(params);
     log.info('Uploading to s3 is complete(ftp)', s3uri);
 
     client.destroy();
-    return s3uri;
+    return { s3uri, etag };
```
chuckwondo, line 192: @yjpa7145 Given my other comment about `multipartUploadCopy` perhaps returning `{ ETag }` rather than `{ etag }` (for consistency with the other functions in that module), do we want to do the same here (also for consistency)?

Although I moaned about AWS not using camel case, I think consistency might trump convention here. It might be confusing to change this function (and the others part of this PR) to return on object with a property named `etag`, but everywhere else where we deal with AWS objects, the property is named `ETag`.

packages/sftp-client/index.js
chuckwondo: I noticed that I failed to be consistent because there is one place where I do this, but not sure if it is JSDoc-friendly:
```
@returns {Promise.<{ s3uri: string, etag: string }>}
```

yjpa7145:
These changes look good.

packages/ingest/S3ProviderClient.js
```diff
@@ -51,10 +51,10 @@ class S3ProviderClient {
     return objects.map(({ Key, Size, LastModified }) => ({
       name: basename(Key),
       // If the object is at the top level of the bucket, path.dirname is going
-      // to return "." as the dirname.  It should instead be null.
-      path: dirname(Key) === '.' ? null : dirname(Key),
+      // to return "." as the dirname.  It should instead be undefined.
+      path: dirname(Key) === '.' ? undefined : dirname(Key),
       size: Size,
-      time: (new Date(LastModified)).valueOf()
+      time: LastModified.valueOf()
```
yjpa7145, line 57: 👍

packages/ingest/S3ProviderClient.js
```diff
@@ -51,10 +51,10 @@ class S3ProviderClient {
     return objects.map(({ Key, Size, LastModified }) => ({
       name: basename(Key),
       // If the object is at the top level of the bucket, path.dirname is going
-      // to return "." as the dirname.  It should instead be null.
-      path: dirname(Key) === '.' ? null : dirname(Key),
+      // to return "." as the dirname.  It should instead be undefined.
+      path: dirname(Key) === '.' ? undefined : dirname(Key),
```
yjpa7145, line 55: This is a breaking change, and needs to be noted as such in the Changelog. (A good change, but still a breaking one.) If the calling code is checking for `null`, it will break.

packages/aws-client/tests/S3/test-multipartCopyObject.js
```diff
@@ -86,6 +86,7 @@ test('multipartCopyObject() copies a file between buckets', async (t) => {
   });
 
   t.is(sourceChecksum, destinationChecksum, 'Source and destination checksums do not match');
+  t.truthy(etag, 'Missing ETag in copy response');
```
yjpa7145, line 89:
```suggestion
  t.truthy(etag, 'Missing etag in copy response');
```

PR #1752, opened 2020-07-02
Mhs/cumulus 1961/refactor

flamingbear:
**Summary:** Adds ESScrollSearch inbetween ESFileQueue and Search.

### DIFF
--- a/packages/api/es/search.js
+++ b/packages/api/es/search.js
@@ -42,7 +42,10 @@ const getLocalEsHost = () => {
 
 const esTestConfig = () => ({
   node: getLocalEsHost(),
-  requestTimeout: 5000
+  requestTimeout: 5000,
+  ssl: {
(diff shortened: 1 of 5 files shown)


packages/api/es/search.js
```diff
@@ -42,7 +42,10 @@ const getLocalEsHost = () => {
 
 const esTestConfig = () => ({
   node: getLocalEsHost(),
-  requestTimeout: 5000
+  requestTimeout: 5000,
+  ssl: {
```
laurenfrederick, line 46: Do we want this committed? I _think_ it's fine since it's just the test config?

flamingbear: I want it committed and it is only the test config, but if there's real reason not to I can stash it and just apply when I want to use it.  I think it's super handy to work against the AWS Cluster locally through a tunnel.

PR #1754, opened 2020-07-06
CUMULUS-1991: Update granule file descriptions in CMR metadata

markdboyd:

Addresses [CUMULUS-1991: As a user, I would like URL Description in granule metadata to have a more useful description](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1991)

## Changes

* Updated code generating CMR metdata for ECHO10/UMM-G to use "Download file.hdf" as the file metadata description instead of "File to download"
* Changed code for updating CMR metadata to include any changed metadata descriptions (previously only updated URLs were included)
* Updated test fixtures to reflect new file metadata description expectations

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


packages/cmrjs/src/cmr-utils.js
```diff
@@ -317,10 +325,11 @@ async function constructOnlineAccessUrl({
   if (distributionApiBuckets.includes(bucketType)) {
     const fileUrl = await generateFileUrl({ file, distEndpoint, cmrGranuleUrlType, distributionBucketMap });
     if (fileUrl) {
+      const filename = getFilename(file);
```
markdboyd, line 328: The only logic changes are here in this file

packages/cmrjs/src/cmr-utils.js
```diff
@@ -44,6 +44,14 @@ function getS3UrlOfFile(file) {
   throw new Error(`Unable to determine location of file: ${JSON.stringify(file)}`);
 }
 
+function getFilename(file) {
+  if (file.name) return file.name;
+  if (file.filename) return path.basename(file.filename);
+  if (file.filepath) return path.basename(file.filepath);
+  if (file.key) return path.basename(file.key);
+  throw new Error(`Unable to determine name of file: ${JSON.stringify(file)}`);
```
markdboyd, line 52: Should we throw an error here? I'm not sure if is beneficial and may only cause unexpected failures

packages/cmrjs/tests/cmr-utils/test-cmr-utils.js
```diff
@@ -74,7 +75,7 @@ test.before(async (t) => {
   t.context.launchpadStub = sinon.stub(launchpad, 'getLaunchpadToken')
     .callsFake((config) => Promise.resolve(`${config.passphrase}-${config.api}-${config.certificate}`));
 
-  const bucketsJson = await readJsonFixture('./tests/fixtures/buckets.json');
+  const bucketsJson = await readJsonFixture(path.join(__dirname, '../fixtures/buckets.json'));
```
markdboyd, line 78: Only reason for this change is to allow me to debug these tests using the VScode debugger. Otherwise the debugger would fail trying to load these relative paths

packages/cmrjs/src/cmr-utils.js
yjpa7145: I don't have a problem with it returning `undefined` in this case, and let the calling code handle that.

packages/cmrjs/src/cmr-utils.js
```diff
@@ -44,6 +44,27 @@ function getS3UrlOfFile(file) {
   throw new Error(`Unable to determine location of file: ${JSON.stringify(file)}`);
 }
 
+function getFilename(file) {
```
yjpa7145, line 47: I hate this so much. It's not your fault, but having to check in five places to figure out a filename on what should be a uniform type? That's awful. Maybe worth an architecture meeting discussion.

packages/cmrjs/src/cmr-utils.js
```diff
@@ -44,6 +44,27 @@ function getS3UrlOfFile(file) {
...
+  if (file.fileName) return file.fileName;
+  if (file.name) return file.name;
+  if (file.filename) return path.basename(file.filename);
+  if (file.filepath) return path.basename(file.filepath);
+  if (file.key) return path.basename(file.key);
+  throw new Error(`Unable to determine name of file: ${JSON.stringify(file)}`);
+}
+
+function getFileDescription(file) {
```
yjpa7145, line 56: If `getFilename` returns `undefined` in the failure case then this can just be:

```js
  const filename = getFilename(file);

  return filename ? `Download ${filename}` : 'File to download';
```

packages/cmrjs/src/cmr-utils.js
markdboyd: Yeah, as far as I know we can't predict at least if we'll have `fileName` or `name`. And I think the same with filename/filepath/key, because the schema was changed at some point but there could still be old records with or without those properties?

PR #1769, opened 2020-07-14
Better handle closing of sftp connections

yjpa7145:


packages/ingest/SftpProviderClient.js
```diff
@@ -19,12 +19,41 @@ class SftpProviderClient {
...
+    await this.privateSftpClient.connect(this.clientOptions);
+
+    this.connected = true;
   }
 
+  async end() {
+    if (this.connected) {
+      this.connected = false;
+
+      await this.privateSftpClient.end();
```
markdboyd, line 47: Should we do the await before setting this.connected = false in case .end throws?

CHANGELOG.md
```diff
@@ -7,6 +7,20 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
+  result in a "Client not connected" exception being thrown.
+- Instances of `@cumulus/ingest/SftpProviderClient` no longer implicitly
+  disconnect from the SFTP server when `list` is called.
+- Instances of `@cumulus/sftp-client/SftpClient` must now be expclicitly closed
+  by calling `.end()`
+- Instances of `@cumulus/sftp-client/SftpClient` no longer implicitly connect to
+  the server when `download`, `unlink`, `syncToS3`, `syncFromS3`, and `list` are
+  called. You must explicitly call `connect` before calling one of those
+  methods.
+- Instances of `@cumulus/sftp-client/SftpClient` must now be explicitly closed
```
markdboyd, line 22: I think this note is in here twice

packages/ingest/S3ProviderClient.js
```diff
@@ -12,6 +12,14 @@ class S3ProviderClient {
     this.bucket = bucket;
   }
 
+  async connect() {
```
markdboyd, line 15: Hmm, I don't having to add these methods to providers where they have no contextual meaning, but I assume we have no choice given how our provider clients are used?

yjpa7145: Not really. I can pull those empty functions out into a mix-in if you'd prefer.

CHANGELOG.md
yjpa7145: c2f282ae60832876d4a69ed689d976b9c4926da3

packages/ingest/SftpProviderClient.js
yjpa7145: b423db8aa66411a8b8512f7ab02dd6ccb45e1ae9

packages/ingest/SftpProviderClient.js
```diff
@@ -67,17 +98,13 @@ class SftpProviderClient {
...
-      this.sftpClientInstance = new SftpClient({
-        host: this.host,
-        port: get(this.providerConfig, 'port', 22),
-        username: await this.getUsername(),
-        password: await this.getPassword(),
-        privateKey: await this.getPrivateKey()
-      });
+  /* @private */
+  getSftpClient() {
+    if (!this.connected) {
```
markdboyd, line 103: What's the advantage of this pattern and the changes throughout here? Why throw an error in a method that's just returning the client if it's not connected?

markdboyd: Because it forces you to explicitly connect before calling something like `download`?

packages/ingest/test/test-SftpProviderClient.js
```diff
@@ -67,7 +73,14 @@ test('SftpProviderClient supports plaintext usernames and passwords', async (t)
     encrypted: false
   });
 
-  const files = await sftpProviderClient.list('/');
+  let files;
```
markdboyd, line 76: Can we not use the provider client set up in the context?

packages/ingest/test/test-SftpProviderClient.js
```diff
@@ -67,7 +73,14 @@ test('SftpProviderClient supports plaintext usernames and passwords', async (t)
...
+  try {
+    await sftpProviderClient.connect();
+    files = await sftpProviderClient.list('/');
+  } finally {
```
markdboyd, line 80: No love for `t.teardown`?

packages/sftp-client/index.js
```diff
@@ -21,35 +20,33 @@ class SftpClient {
...
+  }
 
-      this.client.on('error', reject);
+  async end() {
+    if (this.connected) {
+      this.connected = false;
 
-      this.client.connect(this.clientOptions);
-    });
+      await this.sftpClient.end();
```
markdboyd, line 39: We probably want to do the await before setting `this.connected` = false

packages/sftp-client/index.js
```diff
@@ -136,44 +121,15 @@ class SftpClient {
...
-
-    return new Promise((resolve, reject) => {
-      const readStream = this.sftp.createReadStream(remotePath);
-      readStream.on('error', reject);
-      this.client.on('error', reject);
-      return resolve(readStream);
-    });
+    const remoteFiles = await this.sftp.list(remotePath);
+
+    return remoteFiles.map((remoteFile) => ({
```
markdboyd, line 126: I'm just assuming that the properties returned here are equivalent to how they were being determined before?

packages/sftp-client/tests/test-index.js
```diff
@@ -38,14 +40,18 @@ test('SftpClient supports password authentication', async (t) => {
     password: 'password'
   });
 
+  let files;
   try {
-    const files = await sftpClient.list('/');
-    const fileNames = files.map((f) => f.name);
+    await sftpClient.connect();
```
markdboyd, line 45: Same question: why not use the client from `t.context`?

tasks/discover-pdrs/index.js
```diff
@@ -10,18 +10,33 @@ const { runCumulusTask } = require('@cumulus/cumulus-message-adapter-js');
...
- *                                             before timing out
- * @param {*}      path                      - the provider path to search
- * @returns {Array<Object>}                  - a list of discovered file objects
+ * @param {Object} params.providerConfig - the connection config for the
+ *   provider
+ * @param {bool} params.useList - flag to tell ftp server to use 'LIST'
+ *   instead of 'STAT'
+ * @param {number} [params.httpRequestTimeout=300] - seconds for http provider
+ * to wait before timing out
+ * @param {*} params.path - the provider path to search
```
markdboyd, line 19: We don't know the type of `params.path`?

packages/ingest/S3ProviderClient.js
markdboyd: That or we could hide them in a base provider client that we extend, but it's 6 one way and a half-dozen the other

packages/ingest/test/test-SftpProviderClient.js
yjpa7145: No, because we're testing different types of authentication in this test and a few others, so we need different provider clients rather than using the default one.

packages/ingest/SftpProviderClient.js
yjpa7145: > Because it forces you to explicitly connect before calling something like `download`?

Correct. Since users need to explicitly end the connection, it is more clear when a connection needs to be closed if they are also explicitly connecting as well.

packages/sftp-client/index.js
yjpa7145: Correct.

packages/sftp-client/tests/test-index.js
yjpa7145: It is testing a different auth scheme, which requires a new client.

packages/ingest/S3ProviderClient.js
yjpa7145: 0cfb14418

packages/ingest/test/test-SftpProviderClient.js
yjpa7145: 0cfb14418

packages/sftp-client/index.js
yjpa7145: 0cfb14418

tasks/discover-pdrs/index.js
yjpa7145: 0cfb14418

PR #1780, opened 2020-07-21
Fix granules API client list querying

laurenfrederick:
**Summary:** This was causing intermittent test failures

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/packages/api-client/src/executions.ts
+++ b/packages/api-client/src/executions.ts
@@ -47,7 +47,7 @@ export const getExecutions = async (params: {
   query?: { [key: string]: string },
   callback?: InvokeApiFunction
 }): Promise<ApiGatewayLambdaHttpProxyResponse> => {
-  const { prefix, callback = invokeApi } = params;
+  const { prefix, query, callback = invokeApi } = params;
 
   return callback({
     prefix,
--- a/packages/api-client/src/executions.ts
+++ b/packages/api-client/src/executions.ts
@@ -44,6 +44,7 @@ export const getExecution = async (params: {
  */
 export const getExecutions = async (params: {
   prefix: string,
+  query?: { [key: string]: string },
   callback?: InvokeApiFunction
 }): Promise<ApiGatewayLambdaHttpProxyResponse> => {
   const { prefix, callback = invokeApi } = params;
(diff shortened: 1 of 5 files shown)


packages/api-client/src/executions.ts
```diff
@@ -44,16 +44,18 @@ export const getExecution = async (params: {
  */
 export const getExecutions = async (params: {
   prefix: string,
+  query?: { [key: string]: string },
```
markdboyd, line 47: Typescriptin like a boss

packages/api-client/tests/test-executions.js
```diff
@@ -37,7 +37,8 @@ test('getExecutions calls the callback with the expected object', async (t) => {
     payload: {
       httpMethod: 'GET',
       resource: '/{proxy+}',
-      path: '/executions'
+      path: '/executions',
+      queryStringParameters: undefined
```
markdboyd, line 41: Why can't we just exclude `queryStringParameters` if we're wanting it to be `undefined`?

packages/api-client/tests/test-granules.js
```diff
@@ -228,14 +228,14 @@ test('moveGranule calls the callback with the expected object', async (t) => {
...
-  const query = 'testQuery';
+  const query = { limit: 50 };
   const expected = {
     prefix: t.context.testPrefix,
     payload: {
       httpMethod: 'GET',
       resource: '/{proxy+}',
       path: '/granules',
-      body: JSON.stringify({ query })
+      queryStringParameters: { limit: 50 }
```
markdboyd, line 238:
```suggestion
      queryStringParameters: query
```

packages/api-client/tests/test-granules.js
```diff
@@ -258,7 +258,7 @@ test('listGranules calls the callback with the expected object if there is no qu
       httpMethod: 'GET',
       resource: '/{proxy+}',
       path: '/granules',
-      body: undefined
+      queryStringParameters: undefined
```
markdboyd, line 261: Same comment as above: why include this property if we don't want it to defined?

packages/api-client/tests/test-executions.js
laurenfrederick: The test will holler at you, since that is the expected value

PR #1787, opened 2020-07-23
Cumulus 1955 cmr checksum

chuckwondo:
After moving granules, the CMR metadata files are updated with new URLs. Due to AWS's eventual consistency, it is possible that the pre-updated CMR files are published to CMR.  This work uses etags to guarantee that the updated CMR files are published as expected, avoiding publication of outdated CMR files.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [ ] Integration tests


packages/cmrjs/src/cmr-utils.js
```diff
@@ -203,43 +211,55 @@ async function getXMLMetadataAsString(xmlFilePath) {
...
   if (isECHO10File(cmrFilename)) {
-    return metadataObjectFromCMRXMLFile(cmrFilename);
+    return metadataObjectFromCMRXMLFile(cmrFilename, etag);
   }
   if (isUMMGFile(cmrFilename)) {
-    return metadataObjectFromCMRJSONFile(cmrFilename);
+    return metadataObjectFromCMRJSONFile(cmrFilename, etag);
   }
-  throw new Error(`cannot return metdata from invalid cmrFilename: ${cmrFilename}`);
+  throw new Error(`Invalid CMR filename: ${cmrFilename}`);
```
markdboyd, line 262: I personally think the previous error message was a little more descriptive, but it's not a huge deal

packages/aws-client/src/S3.ts
```diff
@@ -342,34 +353,185 @@ export const s3PutObjectTagging = improveStackTrace(
...
+ * @param {string|pRetry.Options} keyOrRetryOptions - key for object
+ *    (filepath + filename), or retry options (see `retryOptions` parameter)
+ * @param {Object} [retryOptions={ maxTimeout: 10_000, retries: 0 }] - options
+ *    to control retry behavior when an object does not exist (by default,
+ *    retries will not be performed)
+ * @returns {Promise} response from `S3.getObject` as a promise
+ * @see {@link https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#getObject-property|S3.getObject}
+ * @see {@link https://github.com/tim-kos/node-retry#retryoperationoptions|retry options}
+ */
+export const getS3Object: S3GetObject = (
```
markdboyd, line 523: So while I understand the intent here was to avoid a larger refactor and preserve the existing call signature of `getS3Object`, ultimately I feel like the maintainability burden of these changes is a bit too high and we've sacrificed some clarity/readability with them. 

I'd probably prefer to just add a `getS3ObjectWithEtag` function and leave `getS3Object` as it was. And remove all the other code in here that was added to support the refactor

packages/aws-client/src/S3.ts
```diff
@@ -342,34 +353,185 @@ export const s3PutObjectTagging = improveStackTrace(
...
+ * @example s3().getObject(params).then(checkMatchPreconditions(params))
+ *
+ * @function
+ * @param {S3MatchParams} params - object optionally containing
+ * @returns {function} a function that compares the resolved result of an S3
+ *    request with the specified match parameters, and either returns the result
+ *    if pre-conditions are met (or unspecified), or throws an error (with
+ *    `statusCode` set to `412`) indicating a pre-condition failed
+ */
+const checkMatchPreconditions = (params: S3MatchParams) =>
```
markdboyd, line 432: If this is here to just support testing via Localstack and we add a `getS3ObjectWithEtag` function, it should be easy to mock this behavior for the tests rather than needing it in this S3 client library itself?

package.json
```diff
@@ -4,6 +4,11 @@
   "description": "Cumulus Framework for ingesting and processing NASA Earth data streams",
   "scripts": {
     "audit": "audit-ci --config ./audit-ci.json && lerna exec -- audit-ci --config $(pwd)/audit-ci.json",
+    "deploy": "npm run deploy-cumulus",
+    "deploy-all": "npm run deploy-data && npm run deploy-cumulus",
+    "predeploy-cumulus": "npm install && npm run bootstrap",
+    "deploy-cumulus": "cd example/cumulus-tf && terraform init && terraform apply -auto-approve",
```
markdboyd, line 10: I personally don't love having a script that does `terraform apply -auto-approve` in the repo, because it's easy to deploy changes you didn't intend to, but think I'll be outvoted on that one

packages/cmrjs/src/cmr-utils.js
```diff
@@ -576,8 +601,8 @@ async function getCmrSettings(cmrConfig = {}) {
   };
 
   if (oauthProvider === 'launchpad') {
-    const launchpadPassphraseSecretName = cmrConfig.passphraseSecretName ||
-      process.env.launchpad_passphrase_secret_name;
+    const launchpadPassphraseSecretName = cmrConfig.passphraseSecretName
```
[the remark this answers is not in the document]
chuckwondo: Yes, this fixes an ESLint warning.

tasks/move-granules/index.js
```diff
@@ -360,3 +384,4 @@ async function handler(event, context) {
 }
 
 exports.handler = handler;
+exports.moveGranules = moveGranules;
```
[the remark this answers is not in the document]
chuckwondo: Not added, just moved it down to the bottom so that it was obvious (clearly it wasn't :-)

tasks/move-granules/tests/test-index.js
```diff
@@ -32,7 +32,7 @@ async function uploadFiles(files, bucket) {
   await Promise.all(files.map((file) => promiseS3Upload({
     Bucket: bucket,
     Key: parseS3Uri(file).Key,
-    Body: file.endsWith('.cmr.xml')
+    Body: isCMRFilename(file)
```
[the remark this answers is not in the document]
chuckwondo: Ah yes, my bad. I should have used `isECHO10File`. If that is already exported, I'll use that, otherwise I'll just revert this change.

tasks/post-to-cmr/index.js
```diff
@@ -142,3 +142,4 @@ async function handler(event, context) {
 }
 
 exports.handler = handler;
+exports.postToCMR = postToCMR;
```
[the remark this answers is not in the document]
chuckwondo: Not added, just moved it down to the bottom so that it's obvious.

tasks/post-to-cmr/tests/cmr-test.js
```diff
@@ -103,17 +102,49 @@ test.serial('postToCMR succeeds with correct payload', async (t) => {
...
+    'meta.xml'), 'utf8');
+  // "Minify" the XML simply to make it differ from the original XML so that S3
+  // treats them as different versions (i.e., generates different ETags)
+  const updatedCmrXml = cmrXml.split('\n').join('');
+
   try {
-    await promiseS3Upload({
+    // Upload updated XML to obtain the updated ETag so we can add it to the
+    // input CMR file.
+    const { ETag: newEtag } = await promiseS3Upload({
```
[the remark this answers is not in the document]
chuckwondo: Makes sense. Will do.

tasks/post-to-cmr/tests/cmr-test.js
```diff
@@ -1,22 +1,21 @@
...
 const sinon = require('sinon');
-const { promisify } = require('util');
 
 const cmrClient = require('@cumulus/cmr-client');
 const awsServices = require('@cumulus/aws-client/services');
 const { promiseS3Upload, recursivelyDeleteS3Bucket } = require('@cumulus/aws-client/S3');
 const { randomString } = require('@cumulus/common/test-utils');
 const { CMRMetaFileNotFound } = require('@cumulus/errors');
 const launchpad = require('@cumulus/launchpad-auth');
+const { isCMRFile } = require('@cumulus/cmrjs');
```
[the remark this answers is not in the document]
chuckwondo: Yep, I think I missed that because I added a test for failure to the S3 tests.

packages/aws-client/src/S3.ts
chuckwondo: Yes, I'm not thrilled with the added complexity to the `getS3Object` function, but I'm also not thrilled with adding another function that does the _same_ thing, differing only in the expected parameters.

I did consider adding another function, but preferred not doing so, so if we do add a function, I want to make sure it can fully handle all parameters that we could possibly send to `S3.getObject`. If we add `getS3ObjectWithEtag`, and then later we want to pass yet another parameter through to S3, then we'll have to add yet another nearly duplicate function.

I was considering adding a `getObject` function that takes a `params` object to mirror the `S3.getObject` function that we're wrapping, but was concerned that having `getS3Object` _and_ `getObject` would be confusing (i.e., why 2 functions that do the same thing, but with different parameters?), unless we _also_ deprecate `getS3Object`.

What do you think about that?

packages/aws-client/src/S3.ts
chuckwondo: Good point. I hadn't considered that, but I'll see how that could be done.

package.json
chuckwondo: I'm open to suggestions for improvement. I could drop the `-auto-approve` flag, but support adding it to the command line. For example `npm run deploy -- -auto-approve` (which would actually allow you to add _any_ valid `terraform` option).  What do you think about that?

chuckwondo:
> A few comments, mostly that I think we need to simplify the changes we're making to our S3 library.
> 
> Beyond that, I'm wondering if our integration tests for ingest & publish workflows are going to pass. You've updated `move-granules` to include etags in the output and `post-to-cmr` to use that etag - but there is a `hyrax-metadata-updates` task that usually runs in between those two and which also updates the CMR file. So I would think that after that task has run, the actual etag of the file on S3 has changed relative to what was output by `move-granules` and post-to-cmr should fail trying to use the old etag. If that is an issue, the only way I see to address it is by updating the hyrax task to include the updated etags in its output. If it's not an issue, I'm a bit surprised

Aha! That would explain why post-to-cmr is failing in CI.  It is failing with a PreconditionFailed error, which exactly what it should do when the ETag is not correct.  (At least that proves that the unit test I need to add for this case should succeed.)

Looks like I have work to do on the hyrax task. Actually, does this mean I can remove the changes I made to move-granules, or are those necessary in case there is a different workflow where there is no hyrax task in between?

package.json
markdboyd: There are other much more important changes in this PR so let's just not worry about this one. It's just my opinion anyway

packages/aws-client/src/S3.ts
yjpa7145: My vote would be to deprecate `getS3Object` and add something like this:

```ts
export const getObject = (
  params: AWS.S3.GetObjectRequest & { s3: AWS.S3 }
): Promise<AWS.S3.GetObjectOutput> =>
  params.s3.getObject(params).promise();
```

package.json
```diff
@@ -4,6 +4,11 @@
   "description": "Cumulus Framework for ingesting and processing NASA Earth data streams",
   "scripts": {
     "audit": "audit-ci --config ./audit-ci.json && lerna exec -- audit-ci --config $(pwd)/audit-ci.json",
+    "deploy": "npm run deploy-cumulus",
```
yjpa7145, line 7: My leaning is to not include these additional scripts. If we do, though, then it seems like they belong in `example` rather than at the top level.

CHANGELOG.md
```diff
@@ -14,6 +14,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 - Fixed bug with request headers in `@cumulus/launchpad-auth` causing Launchpad token requests to fail
 
+- **CUMULUS-1955**
```
yjpa7145, line 17: If there are changes being made to `aws-client/S3`, those should be noted here as well.

.gitignore
```diff
@@ -1,8 +1,9 @@
+.env
+.envrc
 package-lock.json
 .git
 yarn.lock
 yarn-error.log
-node_modules
```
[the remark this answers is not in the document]
chuckwondo: It was a duplicate entry in the file.

tasks/post-to-cmr/index.js
```diff
@@ -39,19 +39,20 @@ function buildOutput(results, granules) {
...
-async function addMetadataObjects(cmrFiles) {
-  const updatedCMRFiles = [];
-  const objectPromises = cmrFiles.map(async (cmrFile) => {
-    const metadataObject = await metadataObjectFromCMRFile(cmrFile.filename);
-    const updatedFile = { ...cmrFile, metadataObject };
-    updatedCMRFiles.push(updatedFile);
-  });
-  await Promise.all(objectPromises);
-  return updatedCMRFiles;
+function addMetadataObjects(cmrFiles) {
```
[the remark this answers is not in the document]
chuckwondo: Sure.

package.json
chuckwondo: I considered putting them in `example`, but that would then require the "precumulus-deploy" to cd to the parent directory: `"cd .. && npm install && npm run bootstrap"`. This isn't terrible, but I felt that it was mildly more odd than this putting it at the top level.

However, if the consensus is to put them in `example`, that's fine.

Of course, if there is a stronger preference to not include them at all, then my question would be why not? These don't appear to be any less appropriate than some of the existing ones, unless, of course, everybody prefers slightly different steps, in which case they aren't generally usable.

CHANGELOG.md
chuckwondo: Will do.

chuckwondo:
@yjpa7145 what about adding retry options to this?

```javascript
export const getObject = (
  params: AWS.S3.GetObjectRequest & { s3: AWS.S3 }
): Promise<AWS.S3.GetObjectOutput> =>
  params.s3.getObject(params).promise();
```

yjpa7145:
> @yjpa7145 what about adding retry options to this? (I don't know why github didn't let me directly reply to your comment above.)
> 
> ```js
> export const getObject = (
>   params: AWS.S3.GetObjectRequest & { s3: AWS.S3 }
> ): Promise<AWS.S3.GetObjectOutput> =>
>   params.s3.getObject(params).promise();
> ```

I suspect that there are going to be plenty of places that we would want to use this function but would not care about retries. There are also going to be places where we may want to retry for differing reasons. I would prefer to see this function stay small, providing the ability to invoke `getObject` using the passed-in S3 instance, and then returning a promised response. There will certainly be places where we need to retry, but I think those should be build on top of a smaller function. Maybe something like `waitForObjectWithEtag()` or something like that, in this case?

chuckwondo:
> > @yjpa7145 what about adding retry options to this? (I don't know why github didn't let me directly reply to your comment above.)
> > ```js
> > export const getObject = (
> >   params: AWS.S3.GetObjectRequest & { s3: AWS.S3 }
> > ): Promise<AWS.S3.GetObjectOutput> =>
> >   params.s3.getObject(params).promise();
> > ```
> 
> I suspect that there are going to be plenty of places that we would want to use this function but would not care about retries. There are also going to be places where we may want to retry for differing reasons. I would prefer to see this function stay small, providing the ability to invoke `getObject` using the passed-in S3 instance, and then returning a promised response. There will certainly be places where we need to retry, but I think those should be build on top of a smaller function. Maybe something like `waitForObjectWithEtag()` or something like that, in this case?

Good. I like the idea of keeping functions small and focused, but easily composed.

chuckwondo:
> > @yjpa7145 what about adding retry options to this? (I don't know why github didn't let me directly reply to your comment above.)
> > ```js
> > export const getObject = (
> >   params: AWS.S3.GetObjectRequest & { s3: AWS.S3 }
> > ): Promise<AWS.S3.GetObjectOutput> =>
> >   params.s3.getObject(params).promise();
> > ```
> 
> I suspect that there are going to be plenty of places that we would want to use this function but would not care about retries. There are also going to be places where we may want to retry for differing reasons. I would prefer to see this function stay small, providing the ability to invoke `getObject` using the passed-in S3 instance, and then returning a promised response. There will certainly be places where we need to retry, but I think those should be build on top of a smaller function. Maybe something like `waitForObjectWithEtag()` or something like that, in this case?

Does this signature also mean that the caller will be _required_ to pass and S3 instance? For example:

```javascript
const object = await getObject({ s3: s3(), Bucket: 'b', Key: 'k' });
```

packages/aws-client/src/S3.ts
```diff
@@ -348,32 +348,58 @@ export const s3PutObjectTagging = improveStackTrace(
...
+ * @example
+ * const obj = await getObject(s3(), { Bucket: 'b', Key: 'k' })
+ *
+ * @param {AWS.S3} s3 - an `AWS.S3` instance
+ * @param {AWS.S3.GetObjectRequest} params - parameters object to pass through
+ *    to `AWS.S3.getObject()`
+ * @returns {Promise} response from `AWS.S3.getObject()` as a Promise
+ */
+// eslint-disable-next-line no-shadow
+export const getObject = (s3: AWS.S3, params: AWS.S3.GetObjectRequest) =>
```
markdboyd, line 362: Is it not going to be a problem that we're deprecating `getS3Object` in favor of `getObject` when `getObject` has no retries? I suppose eventually the retries might not be as essential once we've updated everything to use etags when making requests, but there might be some interim where updating to use this function without retries could break things due to eventual consistency

packages/aws-client/src/utils.ts
```diff
@@ -68,3 +68,34 @@ export const retryOnThrottlingException = <T, U extends unknown[]>(
...
+ * @param {Function} fn - function (that returns a Promise) to retry on error
+ * @param {Object} options - retry options
+ * @returns {Function} a function that will retry the specified function when
+ *    it throws a "not found" or "pre-condition failed" error (i.e., the
+ *    `statusCode` property of the error is either 404 or 412)
+ * @see https://github.com/sindresorhus/p-retry#options
+ * @see https://github.com/tim-kos/node-retry#retryoperationoptions
+ * @see https://github.com/tim-kos/node-retry#retrytimeoutsoptions
+ */
+export const retryOnMissingObjectError = <T, U extends unknown[]>(
```
markdboyd, line 93: Is this actually used? Do we need it? I'd rather not add it if we're not using it, especially given that the signature is more than a little complex to understand

packages/aws-client/tests/test-S3.js
```diff
@@ -169,6 +163,35 @@ test('getS3Object() will retry if the requested key does not exist', async (t) =
...
+
+  t.is(response.Body.toString(), 'asdf');
+});
+
+test('getObject() throws an exception if the requested bucket does not exist', async (t) => {
+  const Bucket = randomString();
+
+  const promisedObject = getObject(awsServices.s3(), { Bucket, Key: 'asdf' });
+
+  const error = await t.throwsAsync(pTimeout(promisedObject, 5000));
```
markdboyd, line 180: Is there any reason to make this wait 5 seconds? Why not 500 milliseconds to make the tests run a little faster

packages/aws-client/src/utils.ts
markdboyd: I see that it is used later. Given the complexity of this function I would vastly prefer putting the `p-retry` directly inside `getObject`

packages/aws-client/tests/test-S3.js
```diff
@@ -169,6 +163,35 @@ test('getS3Object() will retry if the requested key does not exist', async (t) =
...
+
+  t.is(error.code, 'NoSuchBucket');
+});
+
+test('getObject() throws an exception if the requested key does not exist', async (t) => {
+  const { Bucket } = t.context;
+
+  const promisedObject = getObject(awsServices.s3(), { Bucket, Key: 'asdf' });
+
+  const err = await t.throwsAsync(pTimeout(promisedObject, 5000));
```
markdboyd, line 190: Same question about timeout

packages/aws-client/src/S3.ts
chuckwondo: Yes, @yjpa7145 and I covered this. The idea is to separate concerns here so that the S3 functions are not concerned with retries. Callers will need to wrap retries around the functions themselves, if that's what's desired.

However, I do find use of `pRetries` a bit clumsy, and I had originally added a convenience function around the use of `pRetries` to make it simpler and more convenient to wrap retry behavior around any desired function, but pulled it out (before pushing to this PR) because I felt it would have raised further discussions that would have further delayed this PR.

Such a topic is perhaps worthy of discussion at the next Arch. Mtg.

packages/aws-client/src/utils.ts
chuckwondo: Yes, I added it because I use it on other code in this PR. Regarding the signature, it is identical to the `retryOnThrottlingException` that is already part of this file.

However, this does tie back into my other comment about `pRetry` being a bit clumsy and this was my compromise in favor of the more general convenience wrapper that I decided against adding to avoid a long discussion on this PR and save it for an Arch. Mtg.

packages/aws-client/tests/test-S3.js
chuckwondo: Could do, I suppose, but I was just going off of existing test code.

packages/aws-client/tests/test-S3.js
chuckwondo: Again, this is based on existing test code that I mimicked, but would be happy to dial it down, if you like.

packages/cmrjs/tests/cmr-utils/test-cmr-utils.js
```diff
@@ -661,3 +676,40 @@ test('getFileDescription returns fallback if file name cannot be determined', (t
...
+};
+
+test(
+  'metadataObjectFromCMRFile throws PreconditionFailed when ETag does not match CMR XML metadata file',
+  testMetadataObjectFromCMRFile('s3://bucket/fake.cmr.xml')
+);
+
+test(
+  'metadataObjectFromCMRFile throws PreconditionFailed when ETag does not match CMR UMMG JSON metadata file',
+  testMetadataObjectFromCMRFile('s3://bucket/fake.cmr.json')
```
markdboyd, line 714: Given the test mocking - is there any difference between this and the previous test? It's not a huge deal, just wondering

tasks/hyrax-metadata-updates/index.js
```diff
@@ -271,30 +290,31 @@ async function updateSingleGranule(config, granuleObject) {
...
  * @param {Object} event - input from the message adapter
  * @returns {Object} the granules
  */
-async function hyraxMetadataUpdate(event) {
-  const granulesInput = event.input.granules;
-
-  await Promise.all(
-    granulesInput.map((granuleObject) => updateSingleGranule(event.config, granuleObject))
-  );
-  // We don't create anything, we just update existing metadata. So we return what we got.
```
markdboyd, line 295: One: refactoring this function wasn't really necessary. But more importantly, I think the code comment was perhaps useful context, so even if we were going to refactor, I wouldn't have removed it

tasks/hyrax-metadata-updates/tests/test-integration-index.js
```diff
@@ -152,45 +179,252 @@ const event = {
...
+
+  const e = {
+    config: event.config,
+    input: t.context.payload.input
+  };
+
+  try {
+    const metadataFile = t.context.payload.input.granules[0].files.find((f) =>
+      f.type === 'metadata');
+    metadataFile.etag = randomString();
```
markdboyd, line 335: This is relying on mutation. Not going to hold the PR up on it, but it's a bit brittle

packages/aws-client/tests/test-S3.js
markdboyd: Let's not worry about it changing it for now

packages/aws-client/src/S3.ts
markdboyd: OK. I'm going to defer to @yjpa7145 on the refactored code

markdboyd:
Functionally, the updates to the hyrax task look like they should do what we need and are well covered by the unit tests. I assume the integration tests should pass now. Going to defer to @yjpa7145 on the code refactoring

packages/cmrjs/tests/cmr-utils/test-cmr-utils.js
chuckwondo: They take 2 different paths to finally get to the call to `getObject`, so this tests both paths.

tasks/hyrax-metadata-updates/index.js
chuckwondo: The refactoring was due to my adding the `etag` property to the metadata file of each input granule, and as part of that I avoided using mutation in the `updateSingleGranule` function, which meant changing that function to return the cloned/updated granule object instead of not returning anything at all.

Before, since no changes were made to the granule object (and the function didn't return anything) the comment made sense.  It no longer makes sense because now we _do_ create something (the `etag` properties).

markdboyd: Fair enough, that makes sense 👍

tasks/hyrax-metadata-updates/tests/test-integration-index.js
chuckwondo: Yes, but these tests are fully isolated in that sense. I would have to either had to modify the fixture files (which I don't want to do since they are used by other tests as well) or rebuild the whole payload, which seems overkill.

Perhaps I'm missing something though. What would you suggest?

packages/aws-client/src/utils.ts
chuckwondo: Can you elaborate on what you find complex about it? Keep in mind, I copied the existing `retryOnThrottlingException` function in this file, so there is precedence for it.

For comparison, a bare call to `getObject` would look like so:
```
const obj = getObject(s3(), { Bucket, Key });
```
Adding retries using the function in question might look like so (again, similar to usage of `retryOnThrottlingException`):
```
const retryGetObject = retryOnMissingObjectError(getObject, { retries: 5 });
const obj = retryGetObject(s3(), { Bucket, Key });
```

More generally, regardless of whether you like this particular function or not (I'm actually not a fan of it myself -- it's a compromise I made -- which ties into my desire to discuss this topic in an Arch. Mtg.), decoupling the retry capabilities from `getObject` (as well as the other S3 functions) is important because we may not always want to retry on the same set of exceptions.

If we bake the retry capabilities into the `getObject` function, we cannot change which exceptions to retry on, unless we add _further_ complexity to the function signature, beyond the addition of the retry options object.

For example, in one case, we might want to retry _only_ on pre-condition failures, and _not_ on "no such bucket/key" failures, or vice versa, or even for something else, like "not modified."

tasks/hyrax-metadata-updates/tests/test-integration-index.js
markdboyd: Given that the payload is coming from the test context, you're right, this is fine as is

yjpa7145:
I think that the things I would like to see addressed will be better illustrated by a PR against this branch, so I'm working on that now.

chuckwondo:
Ok, I pushed the changes that drop the use of `t.like`, so hopefully that will get past that build error. All integration tests pass locally, so unless the build hits an intermittent test failure, I expect the build to pass

yjpa7145:
You'll need to run `npm run build-docs` in aws-client to update the README there.

packages/aws-client/tests/test-S3.js
```diff
@@ -169,6 +163,35 @@ test('getS3Object() will retry if the requested key does not exist', async (t) =
   t.is(response.Body.toString(), 'asdf');
 });
 
+test('getObject() returns an existing S3 object', async (t) => {
```
yjpa7145, line 166: Consider moving these tests to `tests/S3/test-getObject.js`

tasks/move-granules/tests/test-index.js
```diff
@@ -128,7 +125,7 @@ test.beforeEach(async (t) => {
   );
 
   const payloadPath = path.join(__dirname, 'data', 'payload.json');
-  const rawPayload = await readFile(payloadPath, 'utf8');
+  const rawPayload = fs.readFileSync(payloadPath, 'utf8');
```
yjpa7145, line 128: I'm not sure I understand what this change is fixing.

packages/cmrjs/src/cmr-utils.js
```diff
@@ -74,17 +75,21 @@ function isCMRFile(fileobject) {
...
+ * @returns {Array<Object>} an array of CMR file objects, each with properties
+ *    `granuleId`, `filename`, and possibly `etag` (if present on input)
  */
 function granuleToCmrFileObject(granule) {
   return granule.files
     .filter(isCMRFile)
-    .map((f) => ({ // handle both new-style and old-style files model
-      filename: f.key ? buildS3Uri(f.bucket, f.key) : f.filename, granuleId: granule.granuleId
+    .map((file) => ({
+      ...pick(file, 'etag'),
```
yjpa7145, line 89: Is this functionally different than `etag: file.etag,`?

packages/aws-client/tests/test-S3.js
chuckwondo: Will do.

tasks/move-granules/tests/test-index.js
chuckwondo: Just a minor cleanup. Everywhere else is using `fs.readFileSync` rather than importing `promisify` and using it with `fs.readFile`, so I did it for consistency to and to eliminate the `promisify` import.

packages/cmrjs/src/cmr-utils.js
chuckwondo: Slightly. This _avoids_ inserting `etag: undefined` when `file.etag` is `undefined`. This is perhaps a cryptic way to avoid an explicit check:
```
    .map((file) => ({
      ...pick(file, 'etag'),
      // handle both new-style and old-style files model
      filename: file.key ? buildS3Uri(file.bucket, file.key) : file.filename,
      granuleId: granule.granuleId
    }));
```
Equivalently, this would be:
```
    .map((file) => {
      const result = {
        // handle both new-style and old-style files model
        filename: file.key ? buildS3Uri(file.bucket, file.key) : file.filename,
        granuleId: granule.granuleId
      };
      if (file.etag) result.etag = file.etag;
      return result;
    }));
```
If you'd prefer the latter (or similar), I'd be happy to change it.

chuckwondo: In essence, this expression:
```
...pick(file, 'etag')
```
is equivalent to this:
```
...(file.etag ? { etag: file.etag } : {})
```
which I suppose is another alternative to the `if` statement.

markdboyd: Ultimately this is pretty minor in the scope of this PR so going to ignore it for now

markdboyd:
You need to re-run `npm run build-docs` in the `aws-client` package. And move the tests as suggested by @marc above. Let's also advertise the added `waitForObject` function in the CHANGELOG. Then I'm good to approve

PR #1788, opened 2020-07-23
CUMULUS-1886: Add multiple sort keys support to api

jennyhliu:

Addresses [CUMULUS-1886: Sort dashboard tables on more than 1 column](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1886)

## Changes

* Add `multiple sort keys` support to es query builder `@cumulus/api/es/es/queries`
* Updated ApiEndpoint lambda to support multiValueQueryStringParameters

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


jennyhliu:
> For example, for this search `/dev/collections?version=006&sort_key[]=+name` I get `{"statusCode":400,"error":"Bad Request","message":"Cannot read property 'filter' of undefined"}`
> and I have gotten it when sorting on something like `process`.
We need to escape `+` with `%2B`,
`curl "https://apiendpoint:8000/dev/collections?version=006&sort_key[]=%2Bname"`

jennyhliu:
Lauren, I have added a fix, the es.collection will keep the error response from es if there is no result.  The unescaped + is a space, so the error for `dev/collections?version=006&sort_key[]=+name` would be:
{"name":"ResponseError","meta":{"body":{"error":{"root_cause":[{"type":"query_shard_exception","reason":"No mapping found for [ name] in order to sort on","index_uuid":"Z7xd6knFTk-_Jlb2OpxRJg","index":"cumulus"}],...

PR #1794, opened 2020-07-28
Migrate ingest/util to TS

yjpa7145:
This is a small incremental step toward migrating the entire ingest package.


.eslintrc.js
```diff
@@ -46,6 +46,7 @@ module.exports = {
     'plugin:eslint-comments/recommended',
     'plugin:import/errors',
     'plugin:import/warnings',
+    'plugin:import/typescript',
```
yjpa7145, line 49: This was needed to solve the problem where `.js` files were reporting that they couldn't find their dependencies that had been moved to `.ts`.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### BREAKING CHANGES
+
+- The `@cumulus/ingest/util.lookupMimeType` function now returns `undefined`
```
markdboyd, line 11: Opinion: this is a breaking change for sure, but I'm wondering if we could group all these code ones that probably 95% of users don't care about under a bullet like "Code changes"

packages/ingest/src/util.ts
```diff
@@ -0,0 +1,23 @@
...
+  } catch (_) {
+    return S3KeyPairProvider.decrypt(ciphertext);
+  }
+};
+
+/**
+ * Return mime-type based on input url or filename
+ *
+ * @param {string} key
+ * @returns {string} mimeType or null
```
markdboyd, line 19:
```suggestion
 * @returns {string} mimeType or undefined
```

packages/ingest/package.json
```diff
@@ -23,10 +27,11 @@
     "CUMULUS",
     "NASA"
   ],
-  "homepage": "https://github.com/nasa/cumulus/tree/master/packages/ingest",
+  "homepage": "https://github.com/nasa/cumulus/tree/master/packages/ingest#readme",
   "repository": {
     "type": "git",
-    "url": "https://github.com/nasa/cumulus"
+    "url": "https://github.com/nasa/cumulus",
+    "directory": "packages/ingest"
```
markdboyd, line 34: What does this do - take you directly to this folder on the repo when you run `npm repo` or something?

yjpa7145: That's what I would expect from [the docs](https://docs.npmjs.com/files/package.json#repository), but I haven't actually seen it have any effect 🤷‍♂️

PR #1795, opened 2020-07-28
Migrate ingest/lock to TS

yjpa7145:
Blocked by #1794.


packages/ingest/src/lock.ts
```diff
@@ -31,30 +43,54 @@ async function checkOldLocks(bucket, locks = []) {
...
+): Promise<number> {
   const s3Objects = await listS3ObjectsV2({
     Bucket: bucket,
     Prefix: `${lockPrefix}/${providerName}`
   });
-  return checkOldLocks(bucket, s3Objects);
+
+  if (s3Objects === undefined) return 0;
+
+  const x = <Lock[]>s3Objects.filter(({ Key }) => Key !== undefined);
```
markdboyd, line 57: The list operation returns objects where key is undefined?

Also, let's change `x` to `oldLocks`

yjpa7145: > The list operation returns objects where key is undefined?

Apparently? AWS created the TS definitions, and it says it can be undefined, so I guess so. I've never seen it in the wild, but I guess it could happen if they say so.

packages/ingest/src/lock.ts
```diff
@@ -31,30 +43,54 @@ async function checkOldLocks(bucket, locks = []) {
...
 }
 
-function addLock(bucket, providerName, filename) {
-  return s3PutObject({
+async function addLock(
+  bucket: string,
+  providerName: string,
+  filename: string
+): Promise<void> {
+  await s3PutObject({
```
markdboyd, line 67: Any reason for `await` over `return`?

yjpa7145: The return value from `addLock` was never being used, so I took the opportunity to hide how it was implemented by not returning anything.

packages/ingest/src/lock.ts
```diff
@@ -31,30 +43,54 @@ async function checkOldLocks(bucket, locks = []) {
...
 
-async function proceed(bucket, provider, filename, counter = 0) {
+export async function proceed(
+  bucket: string,
+  provider: {
+    id: string,
+    globalConnectionLimit?: number
+  },
+  filename: string,
+  counter = 0
```
markdboyd, line 92: Is this mixing type definition and default assignment?

yjpa7145: Yes. TS will determine the type based on the default.

markdboyd: I guess so, but it's allowed when you're not destructuring a params object?

yjpa7145: Yup.

packages/ingest/src/lock.ts
```diff
@@ -17,10 +21,18 @@ const lockPrefix = 'lock';
 * @param {Array} locks - The list of locks in the bucket
 * @returns {integer} - Number of locks remaining in bucket
 **/
-async function checkOldLocks(bucket, locks = []) {
+export async function checkOldLocks(
```
markdboyd, line 24: `checkOldLocks` is horribly named, since it returns a list of locks that are **not** expired, which is the opposite of what I expect based on the name. But I digress

yjpa7145: Yeah, this entire file needs to go and be replaced by the Dynamo semaphores.

packages/ingest/src/lock.ts
```diff
@@ -31,30 +43,54 @@ async function checkOldLocks(bucket, locks = []) {
...
 }
 
-function removeLock(bucket, providerName, filename) {
-  return deleteS3Object(
+export async function removeLock(
+  bucket: string,
+  providerName: string,
+  filename: string
+): Promise<void> {
+  await deleteS3Object(
```
markdboyd, line 79: Oof - I don't love relying on the global state of`lockPrefix` in these methods. Do you mind at least making that a function returning the same value that is used in these functions?

yjpa7145: `lockPrefix` isn’t global, it’s a constant declared in this module.

markdboyd: Yeah I meant global to the scope of the module. I still think it'd be preferable to change for the reasons globals are generally discouraged, but I won't hold up the PR on it

PR #1798, opened 2020-07-29
Migrate ingest/granule and ingest/recursion to TS

yjpa7145:


packages/ingest/src/granule.ts
```diff
@@ -1,71 +1,109 @@
...
+}
 
-const encodeurl = require('encodeurl');
-const errors = require('@cumulus/errors');
-const get = require('lodash/get');
-const moment = require('moment');
-const S3 = require('@cumulus/aws-client/S3');
-const log = require('@cumulus/common/log');
-const { deprecate } = require('@cumulus/common/util');
+export interface File {
```
markdboyd, line 21: Given the likelihood of using this multiple places (the API package code?), should it go into the types package?

packages/ingest/src/granule.ts
```diff
@@ -1,71 +1,109 @@
...
+  * rename s3 file with timestamp
+  *
+  * @param {string} bucket - bucket of the file
+  * @param {string} key - s3 key of the file
+  * @returns {Promise} promise that resolves when file is renamed
+  */
+export async function renameS3FileWithTimestamp(
+  bucket: string,
+  key: string
+): Promise<void> {
```
markdboyd, line 51: So you made the same change in the other PR and said the point of doing `await` instead of `return` was to hide the implementation? Do you mean by not exposing the actual return of `S3.moveObject`? Is there some reason that is preferable?

packages/ingest/src/granule.ts
```diff
@@ -1,71 +1,109 @@
...
+  let renamedKey = `${key}.v${moment.utc(timestamp).format(formatString)}`;
 
-  return S3.s3CopyObject(params)
-    .catch((error) => {
-      log.error(`Failed to copy s3://${CopySource} to s3://${target.Bucket}/${target.Key}: ${error.message}`);
-      throw error;
-    });
+  // if the renamed file already exists, get a new name
+  // eslint-disable-next-line no-await-in-loop
+  while (await S3.s3ObjectExists({ Bucket: bucket, Key: renamedKey })) {
```
markdboyd, line 63: Why is this a `while` insted of a basic `if`?

packages/ingest/src/granule.ts
```diff
@@ -1,71 +1,109 @@
...
-  await copyGranuleFile(source, target, options);
-  return S3.deleteS3Object(source.Bucket, source.Key);
+  if (s3list === undefined) {
+    return [];
+  }
+
+  return s3list
+    .filter((o) => o.Key !== undefined && o.Size !== undefined)
+    .map((c) => (
+      <{ Bucket: string, Key: string, size: number }>{
```
markdboyd, line 101: Casting with a custom type is a bit hard to read. Plus the same type is used as the return of this function, so we should define this interface in the file here at least

packages/ingest/src/granule.ts
```diff
@@ -1,71 +1,109 @@
...
+  });
 
-  await copyGranuleFile(source, target, options);
-  return S3.deleteS3Object(source.Bucket, source.Key);
+  if (s3list === undefined) {
+    return [];
+  }
+
+  return s3list
+    .filter((o) => o.Key !== undefined && o.Size !== undefined)
```
markdboyd, line 99: Would it be better if `listS3ObjectsV2` itself handled these conditions, so it always returns at least an empty array with any objects with undefined bucket/key removed? Seems like it might be be better than repeating this logic everywhere we call the function

packages/ingest/src/granule.ts
```diff
@@ -137,14 +185,23 @@ async function moveGranuleFileWithVersioning(source, target, sourceChecksumObjec
...
-  ACL
-}) {
+export async function handleDuplicateFile(params: {
+  source: { Bucket: string, Key: string },
+  target: { Bucket: string, Key: string },
+  duplicateHandling: DuplicateHandling,
+  checksumFunction?: (bucket: string, key: string) => Promise<[string, string]>,
+  syncFileFunction?: (bucket: string, key: string) => Promise<void>,
+  ACL?: string
+}): Promise<{ Bucket: string, Key: string, size: number }[]> {
```
markdboyd, line 195: Should use the same interface that we extract from the above function

packages/ingest/src/granule.ts
```diff
@@ -247,15 +319,26 @@ function generateMoveFileParams(sourceFiles, destinations) {
...
+  destinations: {
+    regex: string,
+    bucket: string,
+    filepath: string
+  }[]
+): Promise<{ bucket: string, key: string, name?: string }[]> {
   const moveFileParams = generateMoveFileParams(sourceFiles, destinations);
 
-  const processedFiles = [];
+  const processedFiles: { bucket: string, key: string, name?: string }[] = [];
```
markdboyd, line 336: Same here: can we move this type into a definition so it's shared by the function return type?

packages/ingest/src/recursion.ts
```diff
@@ -1,7 +1,12 @@
...
+import * as log from '@cumulus/common/log';
 
-const path = require('path');
-const log = require('@cumulus/common/log');
+export interface RecursionFile {
+  name: string,
+  type: string
+}
+
+export type RecursionFunction = (x: string) => RecursionFile[];
```
markdboyd, line 9: What is `x`? Can we give it a better name?

packages/ingest/src/granule.ts
```diff
@@ -200,13 +258,23 @@ const getNameOfFile = (file) => file.fileName || file.name;
...
   return sourceFiles.map((file) => {
     const fileName = getNameOfFile(file);
+
+    if (fileName === undefined) return { file };
+
     const destination = destinations.find((dest) => fileName.match(dest.regex));
 
     // if there's no match, we skip the file
-    if (!destination) return { source: null, target: null, file };
+    if (!destination) return { file };
```
markdboyd, line 277: Is this a breaking change?

packages/ingest/src/granule.ts
yjpa7145: Probably, but I want to finish migrating the rest of the ingest package first to really flesh out what this type really is. I suspect it may end up getting broken into other types before this is all said and done. Leaving it here for now makes it easier to change.

packages/ingest/src/granule.ts
yjpa7145: Correct. That prevents people from expecting to get that exact return value back. If we need to change the way that we're implementing `renameS3FileWithTimestamp` to use some other S3 function, we won't break anybody if we aren't returning that underlying response.

packages/ingest/src/granule.ts
yjpa7145: It looks like this will handle cases where there are already multiple versions of a file. It will continue increasing the number until it finds one that doesn't exist yet. I agree that it should be fixed, but that refactor felt beyond the scope of this PR.

packages/ingest/src/granule.ts
yjpa7145: It is, noted here: https://github.com/nasa/cumulus/pull/1798/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR16

packages/ingest/src/granule.ts
yjpa7145: I think I've cleaned that up with these changes: [https://github.com/nasa/cumulus/pull/1798/commits/08ceba70690ea61f11cecfef5c6a71d1e1c13c47 ](https://github.com/nasa/cumulus/pull/1798/commits/08ceba70690ea61f11cecfef5c6a71d1e1c13c47)#diff-69e435f10b5478031e5d601d73639322

packages/ingest/src/granule.ts
yjpa7145: See [https://github.com/nasa/cumulus/pull/1798/commits/08ceba70690ea61f11cecfef5c6a71d1e1c13c47 ](https://github.com/nasa/cumulus/pull/1798/commits/08ceba70690ea61f11cecfef5c6a71d1e1c13c47)#diff-69e435f10b5478031e5d601d73639322

packages/ingest/src/granule.ts
yjpa7145: https://github.com/nasa/cumulus/pull/1798/commits/08ceba70690ea61f11cecfef5c6a71d1e1c13c47#diff-8b192e80bd09d39c934ec14baef99981R47

packages/ingest/src/granule.ts
yjpa7145: https://github.com/nasa/cumulus/pull/1798/commits/08ceba70690ea61f11cecfef5c6a71d1e1c13c47#diff-8b192e80bd09d39c934ec14baef99981R338

packages/ingest/src/recursion.ts
yjpa7145: https://github.com/nasa/cumulus/pull/1798/commits/08ceba70690ea61f11cecfef5c6a71d1e1c13c47#diff-fc77b0b42462304bc512aca0c3380f33R9

CHANGELOG.md
```diff
@@ -22,8 +32,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - **[PR 1790](https://github.com/nasa/cumulus/pull/1790)**
-
-  - Fixed bug with request headers in `@cumulus/launchpad-auth` causing Launchpad token requests to fail
+- Fixed bug with request headers in `@cumulus/launchpad-auth` causing Launchpad token requests to fail
```
markdboyd, line 35: I think this is a bad merge: this bullet should be indented inside the above

packages/aws-client/src/S3.ts
```diff
@@ -673,22 +681,28 @@ export const listS3Objects = async (
...
       // Update the params with a Continuation Token
       {
 
         ...params,
         ContinuationToken: listObjectsResponse.NextContinuationToken
       }
-    ).promise();
-    discoveredObjects = (discoveredObjects || []).concat(listObjectsResponse.Contents || []);
+    ).promise());
+    discoveredObjects = discoveredObjects.concat(listObjectsResponse.Contents);
```
markdboyd, line 705: So we have confidence based on the SDK docs that the return here should never be undefined?

CHANGELOG.md
```diff
@@ -11,6 +11,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - The `@cumulus/ingest/util.lookupMimeType` function now returns `undefined`
   rather than `null` if the mime type could not be found.
 - The `@cumulus/ingest/lock.removeLock` function now returns `undefined`
+- The `@cumulus/ingest/granule.generateMoveFileParams` function now returns
+  `source = undefined` and `target = undefined` if either could not be
```
markdboyd, line 15:
```suggestion
  `source: undefined` and `target :undefined` on the response object if either could not be
```

packages/ingest/src/granule.ts
```diff
@@ -296,48 +381,7 @@ async function moveGranuleFiles(sourceFiles, destinations) {
...
-}
-
-/**
-  * rename s3 file with timestamp
-  *
-  * @param {string} bucket - bucket of the file
-  * @param {string} key - s3 key of the file
-  * @returns {Promise} promise that resolves when file is renamed
-  */
-async function renameS3FileWithTimestamp(bucket, key) {
```
markdboyd, line 309: Do we need to flag this deletion since it was exported? And the function below?

packages/aws-client/src/S3.ts
yjpa7145: I think so. It's the assumption that we've been working off of for the past few years.

packages/ingest/src/granule.ts
yjpa7145: This function is still there and exported, it just had to be moved higher in the file so that it was declared before it was used. I'll update the Changelog to note that the `@cumulus/ingest/granule.getRenamedS3File` function has been renamed to
  `listVersionedObjects`.

PR #1804, opened 2020-07-31
Add vscode debugger config

markdboyd:


README.md
```diff
@@ -119,6 +119,10 @@ Run the test commands next
 
 For more information please [read this](docs/development/integration-tests.md).
 
+### Running tests via VScode debugger
```
chuckwondo, line 122: nit: VScode should be VS Code (same on following line too)

README.md
```diff
@@ -119,6 +119,10 @@ Run the test commands next
...
+
+Copy `launch.json.example` to `.vscode/launch.json`. Refer to the [VScode documentation on how to use the debugger](https://code.visualstudio.com/docs/editor/debugging).
```
chuckwondo, line 124: How about creating the dir `.vscode.example/` (to avoid cluttering the root dir with non-hidden files) and putting the example file in there and just naming the file `launch.json` (to get the benefit of JSON editing capabilities in the editor)

PR #1813, opened 2020-08-13
Fixes for rec reports integration tests

laurenfrederick:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/packages/api/es/esScrollSearch.js
+++ b/packages/api/es/esScrollSearch.js
@@ -2,7 +2,7 @@
 
 const { Search } = require('./search');
 const defaultESScrollSize = 1000;
-const defaultESScrollDuration = '30s';
+const defaultESScrollDuration = '2m';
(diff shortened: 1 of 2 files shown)


packages/api/es/esScrollSearch.js
```diff
@@ -2,7 +2,7 @@
 
 const { Search } = require('./search');
 const defaultESScrollSize = 1000;
-const defaultESScrollDuration = '30s';
+const defaultESScrollDuration = '2m';
```
laurenfrederick, line 5: This was the main issue is that the search context was dying. I did some reading and increased it. I exposed it an env var (to just us for now) in case we need to test with increasing it or decreasing it on the fly.

PR #1830, opened 2020-08-26
Remove some deprecated code

yjpa7145:


packages/cmrjs/src/index.js
```diff
@@ -1,13 +1,9 @@
 'use strict';
 
-const got = require('got');
```
markdboyd, line 3: I think we no longer need `got` as a dependency in this package

PR #1831, opened 2020-08-26
[CUMULUS-2113] Granule Not Found Report

bt6453:

Addresses [CUMULUS-2113: Modify create-reconciliation-report lambda to capture each file's associated granuleID](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2113)

## Changes

* Add okCountyByGranuleId to inventory report
* With this info we _should_ have all the info we need to display a Granule Not Found report in the dashboard

## PR Checklist

- [x] Update CHANGELOG
- [X] Unit tests
- [X] Adhoc testing
- [X] Integration tests


bt6453:
I just added to the inventory report, based on previous discussions. We could also change the report so it's saved as "Granule Not Found," but I kept it the same for now. We could also take away any sections that would be irrelevant for a GNF report.

dopeters:
Just my two cents, but I think we do want to set the report type to Granule Not Found so that we can easily distinguish between types on the dashboard side

PR #1855, opened 2020-09-09
Increase timeouts

laurenfrederick:
I have seen test fails due to not waiting long enough for the workflow to kick off (I went and actually searched for it based on the testExecutionId and found it, compared the timestamps), so this is increasing the timeout.

### DIFF
--- a/packages/integration-tests/Executions.js
+++ b/packages/integration-tests/Executions.js
@@ -69,6 +69,7 @@ const findExecutionArn = async (prefix, matcher, queryParameters = { }, options 
     },
     {
       retries: options.timeout,
+      minTimeout: 1000,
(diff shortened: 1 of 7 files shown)


packages/integration-tests/Executions.js
```diff
@@ -69,6 +69,7 @@ const findExecutionArn = async (prefix, matcher, queryParameters = { }, options
     },
     {
       retries: options.timeout,
+      minTimeout: 1000,
```
laurenfrederick, line 72: Line 25 was misleading, because that was not true. Timeout is the number of times, but this makes it true. I changed this because I didn't see a real reason we need to hit the executions more than once per second.

markdboyd (approved):

PR #1869, opened 2020-09-16
Mhs/cumulus 1963/granule id parameters to reconcilation reports

flamingbear:
**Summary:** Adds granuleId input parameter to the reconciliationReport endpoint.

Addresses [CUMULUS-1963: Update reconciliation-report endpoint to take parameter: granule ID](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1963)

## Changes

* splits oneWay Report testing into one way for granules and one way for collections. When granuleIds are passed to reconcilationReport handler, collection comparisons become one way.
* CMR is searched for granuleIds using array of granuleIds each one passed to CMR on a `readable_granule_name[]` key
* Non-'Internal' reports now fail when both collectionId and granuleId are passed to the lambda.
* Adds a number of tests to describe the desired behavior of normalizeEvent
* Adds two tests for granuleId input to the lambda.
* updates CMRSearchConceptQueue to add provider_short_name to an input plain Object or an input URLSearchParams.
* updates cmr-client/searchConcept internally to use URLSearchParams rather than plan object so that multiple keyword params can be included to search CMR.

PR for API doc change https://github.com/nasa/cumulus-api/pull/289

## PR Checklist

- [X] Update CHANGELOG
- [X] Unit tests
- [X] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/packages/api/es/search.js
+++ b/packages/api/es/search.js
@@ -35,10 +35,11 @@ const getCredentials = () =>
  * @returns {string} elasticsearch local address
  */
 const getLocalEsHost = () => {
-  const prot = (process.env.LOCAL_ES_HOST_PROTOCOL) ? process.env.LOCAL_ES_HOST_PROTOCOL : 'http';
-  if (process.env.LOCAL_ES_HOST) return `${prot}://${process.env.LOCAL_ES_HOST}:9200`;
-  if (process.env.LOCALSTACK_HOST) return `${prot}://${process.env.LOCALSTACK_HOST}:4571`;
-  return `${prot}://localhost:9200`;
+  const port = process.env.LOCAL_ES_HOST_PORT || 9200;
+  const protocol = (process.env.LOCAL_ES_HOST_PROTOCOL) ? process.env.LOCAL_ES_HOST_PROTOCOL : 'http';
+  if (process.env.LOCAL_ES_HOST) return `${protocol}://${process.env.LOCAL_ES_HOST}:${port}`;
+  if (process.env.LOCALSTACK_HOST) return `${protocol}://${process.env.LOCALSTACK_HOST}:4571`;
+  return `${protocol}://localhost:9200`;
 };
 
 const esTestConfig = () => ({
--- a/packages/api/lambdas/create-reconciliation-report.js
+++ b/packages/api/lambdas/create-reconciliation-report.js
@@ -115,12 +115,9 @@ async function fetchESCollections(recReportParams) {
   let esCollectionIds;
   // [MHS, 09/02/2020] We are doing these two because we can't use
   // aggregations on scrolls yet until we update elasticsearch version.
-  log.debug(`esGranuleSearchParams ${JSON.stringify(esGranuleSearchParams)}`);
-  log.debug(`esCollectionSearchParams ${JSON.stringify(esCollectionSearchParams)}`);
   if (shouldAggregateGranulesForCollections(esGranuleSearchParams)) {
     // Build an ESCollection and call the aggregateGranuleCollections to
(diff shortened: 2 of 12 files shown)


packages/api/es/search.js
```diff
@@ -35,10 +35,11 @@ const getCredentials = () =>
  * @returns {string} elasticsearch local address
  */
 const getLocalEsHost = () => {
-  const prot = (process.env.LOCAL_ES_HOST_PROTOCOL) ? process.env.LOCAL_ES_HOST_PROTOCOL : 'http';
-  if (process.env.LOCAL_ES_HOST) return `${prot}://${process.env.LOCAL_ES_HOST}:9200`;
-  if (process.env.LOCALSTACK_HOST) return `${prot}://${process.env.LOCALSTACK_HOST}:4571`;
-  return `${prot}://localhost:9200`;
+  const port = process.env.LOCAL_ES_HOST_PORT || 9200;
```
flamingbear, line 38: Just a new local-only ENV var for choosing a port to find the elasticsearch when testing locally.

packages/api/lambdas/create-reconciliation-report.js
```diff
@@ -712,6 +734,58 @@ function isoTimestamp(dateable) {
   return undefined;
 }
 
+/**
+ * Transforms input collectionId into correct parameters for
```
jennyhliu, line 738: doc doesn't reflect the functionality

packages/api/lib/reconciliationReport.js
```diff
@@ -113,18 +129,24 @@ function initialReportHeader(recReportParams) {
...
+    granuleId,
+    collectionIds,
     collectionId,
   } = recReportParams;
 
   return {
-    reportType,
-    createStartTime: createStartTime.toISOString(),
+    collectionId,
+    collectionIds,
```
jennyhliu, line 140: I think we only need `collectionId` to match the event input

jennyhliu (commented):
It looks good to me.  I haven't tested it yet.  Are you going to add `granuleId` parameter to integration test spec?

packages/api/lib/reconciliationReport.js
flamingbear: > It looks good to me. I haven't tested it yet. Are you going to add `granuleId` parameter to integration test spec?

I was not going to because it would require another complete run of the reconcilation report as we can't have both collectionId and granuleId on the same request.

flamingbear: That is true when your work gets finished, but right now, the normalizeEvent code, for non 'Internal' types moves the input string or array<string> from collectionId to collectionIds. (same for granuleId) (or it should.)

flamingbear: I think your PR that handles multiple collectionId and granuleId should stop that behavior and just transform the input string or array into an array on the singular version... eg.
if the input is collectionId = "collectionId" => collectionId = ["collectionId"] (rather than moving it to "collectionIds")

packages/api/lambdas/create-reconciliation-report.js
flamingbear: Oops. good catch.

flamingbear: 1477306bb

packages/api/lib/reconciliationReport.js
flamingbear: I am open to something else.  I tried also just adding the entire recReportParams key, but it had duplicates, It might be best to do that because we will keep adding and removing parameters and want to know exactly how the lambda was called in our report.

jennyhliu (approved):
👍

PR #1878, opened 2020-09-22
Fix local db cleanup

yjpa7145:
**Summary:** Remove localstack DB tables in the correct order

Addresses [CUMULUS-2174: ./bin/db-local-drop.js failing](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2174)

## Changes

- Use migration rollbacks to clean up local db rather than just dropping tables
- Remove `./bin/db-local-drop.js`
- Add `./bin/db-local-rollback.js`
- Remove unused `example/lambdas/db-drop-tables` package
- Remove unused `@cumulus/db.dropAllTables` function

## PR Checklist

- [ ] Update CHANGELOG
- [ ] ~Unit tests~
- [ ] Adhoc testing
- [ ] ~Integration tests~

PR #1873, opened 2020-09-17
Update all granule/file database writes to write to both Dynamo and RDS

yjpa7145:
**Summary:** Summary of changes

Addresses [CUMULUS-2128: Update all granule/file database writes to write to both Dynamo and RDS](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2128)

## Changes

* Detailed list or prose of changes
* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

PR #1879, opened 2020-09-22
CUMULUS-2157: Dynamo providers migration

markdboyd:

Addresses [CUMULUS-2157: Dynamo -> RDS providers migration](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2157)

## Changes

* Added Dynamo -> RDS providers migration to data-migration1
* Code reorganization
* Added `provider_kms_key_id` variable to data-migration1 module

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing


CHANGELOG.md
```diff
@@ -152,8 +133,34 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - **CUMULUS-2113**
   - Added Granule not found report to reports endpoint
   - Update reports to return breakdown by Granule of files both in DynamoDB and S3
+- **CUMULUS-2123**
```
markdboyd, line 136: A lot of these changes are reorganizing the CHANGELOG to be sorted by ticket number ascending

example/data-migration1-tf/README.md
```diff
@@ -0,0 +1,7 @@
+# Data migration deployment module
+
+## How to get value for `provider_kms_key_id` Terraform variable
```
markdboyd, line 3: I don't really like this option, but there's no other way to easily get the provider KMS key ID to use for this module. This module has to deploy **before** the cumulus module, so it can't be retrieved as an output from the cumulus module

lambdas/data-migration1/src/providers.ts
```diff
@@ -0,0 +1,157 @@
...
+
+import { RecordAlreadyMigrated } from './errors';
+import { MigrationSummary } from './types';
+
+const Manager = require('@cumulus/api/models/base');
+const schemas = require('@cumulus/api/models/schemas');
+
+const logger = new Logger({ sender: '@cumulus/data-migration/providers' });
+
+export interface RDSProviderRecord {
```
yjpa7145, line 17: I guess it's fine that this lives here for now but, as soon as we write code that's reading or writing to the DB, we should move this somewhere more widely accessible.

lambdas/data-migration1/src/providers.ts
```diff
@@ -0,0 +1,157 @@
...
+ *   if record was already migrated
+ */
+export const migrateProviderRecord = async (
+  dynamoRecord: AWS.DynamoDB.DocumentClient.AttributeMap,
+  knex: Knex
+): Promise<number> => {
+  // Use API model schema to validate record before processing
+  Manager.recordIsValid(dynamoRecord, schemas.provider);
+
+  const [existingRecord] = await knex('providers')
```
yjpa7145, line 76: 🐜 Consider `knex('providers').first()`

lambdas/data-migration1/src/providers.ts
```diff
@@ -0,0 +1,157 @@
...
+    .where('name', dynamoRecord.id);
+  // Throw error if it was already migrated.
+  if (existingRecord) {
+    throw new RecordAlreadyMigrated(`Provider name ${dynamoRecord.id} was already migrated, skipping`);
+  }
+
+  let { username, password, encrypted } = dynamoRecord;
+  if (username || password) {
+    username = await encryptProviderCredential(username, encrypted);
```
yjpa7145, line 85: If these are returning `undefined` I'm pretty sure that will cause a failure when you write to the DB.

lambdas/data-migration1/src/providers.ts
```diff
@@ -0,0 +1,157 @@
...
+    password = await encryptProviderCredential(password, encrypted);
+    encrypted = true;
```
yjpa7145, line 87: After this migration, I don't know that `encrypted` has any use as it's always "true".

lambdas/data-migration1/src/providers.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  // Map old record to new schema.
+  const updatedRecord: RDSProviderRecord = {
```
yjpa7145, line 91: For all of these fields, please make sure that `undefined` works.

yjpa7145, line 91: We may not end up merging my PR for granule / file migrations, but I think that these functions will be helpful for you:

https://github.com/nasa/cumulus/pull/1873/files#diff-a53ed033e4dedc280ea96e597da29392R1

lambdas/data-migration1/src/providers.ts
```diff
@@ -0,0 +1,157 @@
...
+ *   Source record from DynamoDB
+ * @param {Knex} knex - Knex client for writing to RDS database
+ * @returns {Promise<number>} - Cumulus ID for record
+ * @throws {RecordAlreadyMigrated}
+ *   if record was already migrated
+ */
+export const migrateProviderRecord = async (
+  dynamoRecord: AWS.DynamoDB.DocumentClient.AttributeMap,
+  knex: Knex
+): Promise<number> => {
```
yjpa7145, line 72: The return value is never used.

```suggestion
): Promise<void> => {
```

lambdas/data-migration1/src/providers.ts
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * @param {AWS.DynamoDB.DocumentClient.AttributeMap} dynamoRecord
+ *   Source record from DynamoDB
+ * @param {Knex} knex - Knex client for writing to RDS database
+ * @returns {Promise<number>} - Cumulus ID for record
+ * @throws {RecordAlreadyMigrated}
+ *   if record was already migrated
+ */
+export const migrateProviderRecord = async (
+  dynamoRecord: AWS.DynamoDB.DocumentClient.AttributeMap,
```
yjpa7145, line 70: The answer may be "no", but is it worth creating a more provider-specific type for this record, rather than just accepting anything?

lambdas/data-migration1/src/providers.ts
```diff
@@ -0,0 +1,157 @@
...
+  updated_at?: Date
+}
+
+const encryptProviderCredential = async (
+  value: string,
+  encrypted?: boolean
+) => {
+  if (isNil(value)) return undefined;
+
+  const providerKmsKeyId = envUtils.getRequiredEnvVar('provider_kms_key_id');
```
yjpa7145, line 39: Consider passing in `provider_kms_key_id` as an argument rather than pulling it from the environment here.

lambdas/data-migration1/src/providers.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  if (encrypted) {
+    return keyPairProvider.S3KeyPairProvider
+      .decrypt(value)
+      .then(
+        (decryptedValue) => KMS.encrypt(providerKmsKeyId, decryptedValue),
+        async (error) => {
```
yjpa7145, line 46: I'd rather see `try / catch` here rather than the less-common `then(resolve, reject)` syntax.

lambdas/data-migration1/src/providers.ts
```diff
@@ -0,0 +1,157 @@
...
+const encryptProviderCredential = async (
+  value: string,
+  encrypted?: boolean
+) => {
+  if (isNil(value)) return undefined;
+
+  const providerKmsKeyId = envUtils.getRequiredEnvVar('provider_kms_key_id');
+
+  if (encrypted) {
+    return keyPairProvider.S3KeyPairProvider
```
yjpa7145, line 42: It might be more clear what's going on in this function if there were a separate function called `decrypt` or something similar that takes an encrypted value and returns the plaintext.

lambdas/data-migration1/tests/test-providers.js
```diff
@@ -0,0 +1,363 @@
...
+  cmKeyId: 'key-id',
+  certificateUri: 'uri',
+  ...params,
+});
+
+let providersModel;
+let rulesModel;
+
+test.before(async (t) => {
+  t.context.knexAdmin = await getKnexClient({
```
yjpa7145, line 48: This might be helpful: https://github.com/nasa/cumulus/pull/1878/files#diff-69d8e7315b4a49aed756eedad7781a18R7

lambdas/data-migration1/tests/test-providers.js
```diff
@@ -0,0 +1,363 @@
...
+  const createKeyResponse = await KMS.createKey();
+  process.env.provider_kms_key_id = createKeyResponse.KeyMetadata.KeyId;
+
+  providersModel = new Provider();
+  await providersModel.createTable();
+
+  rulesModel = new Rule();
+  await rulesModel.createTable();
+
+  await t.context.knexAdmin.raw(`create database "${testDbName}";`);
```
yjpa7145, line 83: We shouldn't have to create the database and run migrations in the test, it should be handled by `npm run db:local:migrate`.

yjpa7145 (changes requested):

lambdas/data-migration1/src/collections.ts
```diff
@@ -0,0 +1,131 @@
...
+ *   Source record from DynamoDB
+ * @param {Knex} knex - Knex client for writing to RDS database
+ * @returns {Promise<number>} - Cumulus ID for record
+ * @throws {RecordAlreadyMigrated}
+ *   if record was already migrated
+ */
+export const migrateCollectionRecord = async (
+  dynamoRecord: AWS.DynamoDB.DocumentClient.AttributeMap,
+  knex: Knex
+): Promise<number> => {
```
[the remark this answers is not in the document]
markdboyd: It is used by the tests. Is there a better way?

https://github.com/nasa/cumulus/blob/master/lambdas/data-migration1/tests/test-collections.js#L138

lambdas/db-migration/src/migrations/20200911163145_create_providers_table.ts
```diff
@@ -0,0 +1,54 @@
...
+      .integer('port')
+      .comment('Port name for acessing the provider');
+    table
+      .text('username')
+      .comment('Username for acessing the provider');
+    table
+      .text('password')
+      .comment('password for acessing the provider');
+    table
+      .boolean('encrypted')
```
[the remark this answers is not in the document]
markdboyd: It is so long as we have code like this using it:

https://github.com/nasa/cumulus/blob/master/packages/ingest/src/HttpProviderClient.js

I'm a little hesitant to change it because that goes beyond a simple "switch to reading to RDS instead of dymamo". It seems like something we could deprecate in the future, but I don't know if I want to add it to this work

lambdas/data-migration1/src/collections.ts
```diff
@@ -0,0 +1,131 @@
...
+    granuleIdValidationRegex: dynamoRecord.granuleId,
+    granuleIdExtractionRegex: dynamoRecord.granuleIdExtraction,
+    // have to stringify on an array of values
+    files: JSON.stringify(dynamoRecord.files),
+    reportToEms: dynamoRecord.reportToEms,
+    sampleFileName: dynamoRecord.sampleFileName,
+    ignoreFilesConfigForDiscovery: dynamoRecord.ignoreFilesConfigForDiscovery,
+    meta: dynamoRecord.meta ? dynamoRecord.meta : undefined,
+    // have to stringify on an array of values
+    tags: dynamoRecord.tags ? JSON.stringify(dynamoRecord.tags) : undefined,
```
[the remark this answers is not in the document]
markdboyd: The default for the collection tests is to migrate with tags:

https://github.com/nasa/cumulus/blob/master/lambdas/data-migration1/tests/test-collections.js#L37

Do you mean adding a test for where `tags` is undefined?

lambdas/data-migration1/src/providers.ts
markdboyd: See above comment about keeping this value

lambdas/data-migration1/src/providers.ts
markdboyd: Since we're immediately turning around and validating the dynamo record against the provider schema, does it matter?

lambdas/data-migration1/tests/test-providers.js
markdboyd: This goes back to a pattern from @Jkovarik's previous PR. I believe that his argument was that each test should be using a separate database and thus needs the migrations applied to that database.

lambdas/data-migration1/src/providers.ts
markdboyd: I thought previously we had discussed that we only need to verify that the fields which can be null/undefined in the Dynamo record need to be tested explicitly. For fields that aren't optional, the migration will fail on the schema validation if they're missing

markdboyd: And the test for the nullable fields is here:

https://github.com/nasa/cumulus/pull/1879/files?file-filters%5B%5D=.js&hide-deleted-files=true#diff-a8b8713199bb6242c4074966671aaeedR241

example/data-migration1-tf/README.md
yjpa7145: I'm fine with this solution.

lambdas/data-migration1/src/collections.ts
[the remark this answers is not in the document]
yjpa7145: Sorry, I mistyped. I meant a test like this:

```js
test.serial('migrateCollectionRecord correctly migrates collection record without tags', async (t) => {
  const fakeCollection = generateFakeCollection();
  delete fakeCollection.tags;

  const cumulusId = await migrateCollectionRecord(fakeCollection, t.context.knex);

  const [createdRecord] = await t.context.knex.queryBuilder()
    .select()
    .table('collections')
    .where('cumulusId', cumulusId);

  t.deepEqual(
    omit(createdRecord, ['cumulusId']),
    omit(
      {
        ...fakeCollection,
        granuleIdValidationRegex: fakeCollection.granuleId,
        granuleIdExtractionRegex: fakeCollection.granuleIdExtraction,
        created_at: new Date(fakeCollection.createdAt),
        updated_at: new Date(fakeCollection.updatedAt),
        tags: null, // eslint-disable-line unicorn/no-null
      },
      ['granuleId', 'granuleIdExtraction', 'createdAt', 'updatedAt']
    )
  );
});
```

That being said, the test passes, so I guess doing an insert with some values set to `undefined` does work. I'm not sure what the difference was when I got failures in that case, but 🤷‍♂️

lambdas/data-migration1/src/collections.ts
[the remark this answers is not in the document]
yjpa7145: In the tests, rather than:

```js
const [createdRecord] = await t.context.knex.queryBuilder()
    .select()
    .table('collections')
    .where('cumulusId', cumulusId);
```

you could do something like:

```js
const [createdRecord] = await t.context.knex.queryBuilder()
    .select()
    .table('collections')
    .where({ name: fakeCollection.name, version: fakeCollection.version });
```

lambdas/db-migration/src/migrations/20200911163145_create_providers_table.ts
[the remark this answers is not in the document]
yjpa7145: Until we update code to read from the DB, that provider code isn't reading from the DB. In more complex cases I agree that we should just lift and shift from Dynamo to RDS. This case seems trivial enough that creating that column and deleting it later without it ever being used seems like more work. I'll defer to @nemreid and @Jkovarik, though.

lambdas/data-migration1/src/collections.ts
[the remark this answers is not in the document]
markdboyd: 🤦 Indeed

lambdas/data-migration1/src/providers.ts
yjpa7145: Probably not in this case.

lambdas/data-migration1/tests/test-providers.js
yjpa7145: I'm not sure why each test would need its own database, though, if what we are testing is a data migration and not a schema migration.

lambdas/data-migration1/src/providers.ts
yjpa7145: 👍

lambdas/data-migration1/src/collections.ts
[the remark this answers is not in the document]
markdboyd: I think that case is also handled by this test?

https://github.com/nasa/cumulus/blob/master/lambdas/data-migration1/tests/test-collections.js#L136

lambdas/data-migration1/src/providers.ts
markdboyd: I'm not sure I grasp what they would be used for?

lambdas/data-migration1/src/collections.ts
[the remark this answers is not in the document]
yjpa7145: 👍

lambdas/data-migration1/src/providers.ts
yjpa7145: They were useful in the case where setting a field to `undefined` was failing, but that does not seem to be an issue here.

lambdas/data-migration1/src/providers.ts
markdboyd: This is a bit complex. If the decryption fails and the KMS decryption succeeds, then we want to return the initial passed in value from this function, but otherwise we want to encrypt the decrypted value. This felt cleaner than other options that I tried, but I can revisit.

lambdas/data-migration1/src/providers.ts
markdboyd: 9c20c460168bf02cf59826e5f2bf64106bf8f23c

lambdas/data-migration1/src/collections.ts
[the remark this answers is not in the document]
markdboyd: 9c20c460168bf02cf59826e5f2bf64106bf8f23c

lambdas/data-migration1/src/collections.ts
```diff
@@ -0,0 +1,131 @@
...
+ *   if record was already migrated
+ */
+export const migrateCollectionRecord = async (
+  dynamoRecord: AWS.DynamoDB.DocumentClient.AttributeMap,
+  knex: Knex
+): Promise<number> => {
+  // Use API model schema to validate record before processing
+  Manager.recordIsValid(dynamoRecord, schemas.collection);
+
+  const [existingRecord] = await knex('collections')
```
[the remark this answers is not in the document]
markdboyd: 9c20c460168bf02cf59826e5f2bf64106bf8f23c

lambdas/data-migration1/src/collections.ts
```diff
@@ -0,0 +1,131 @@
...
+    .where('name', dynamoRecord.name)
```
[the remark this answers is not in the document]
markdboyd: 9c20c460168bf02cf59826e5f2bf64106bf8f23c

lambdas/data-migration1/src/providers.ts
markdboyd: 9c20c460168bf02cf59826e5f2bf64106bf8f23c

lambdas/data-migration1/src/providers.ts
markdboyd: e8cd413a857e96829e08e38277ddc74ad3a93f44

lambdas/data-migration1/src/providers.ts
markdboyd: I've tried to refactor this a couple times now and I'm not seeing a great way without the `then(resolve,reject)` syntax. The problem is that we want to attempt verifying that it is a KMS encrypted value **only if** the S3 decryption failed, not if the `resolve` function itself throws an error. And we only need to encrypt the value if the S3 decryption succeeded. So far, I'm not finding an elegant way to achieve this in try/catch syntax. This gets close, but the catch would be triggered for any error on KMS.encrypt as well, which is not what we want.

```javascript
    let decryptedValue;
    try {
      decryptedValue = await decryptProviderCredential(value);
      return KMS.encrypt(providerKmsKeyId, decryptedValue);
    } catch (error) {
      // If we already have a KMS encrypted value, return it.
      if ((await KMS.decryptBase64String(value))) {
        return value;
      }
      throw error;
    }
```

markdboyd: OK, got this: 159c7e27f6386873b8b6c14dd34ef4adddf1145a

I'm not sure that I really like it better though

lambdas/data-migration1/tests/test-providers.js
markdboyd: 63da80250b13951b2e6723f175eb2b536bcc7bb5

lambdas/data-migration1/src/collections.ts
```diff
@@ -0,0 +1,131 @@
...
+    sampleFileName: dynamoRecord.sampleFileName,
+    ignoreFilesConfigForDiscovery: dynamoRecord.ignoreFilesConfigForDiscovery,
+    meta: dynamoRecord.meta ? dynamoRecord.meta : undefined,
+    // have to stringify on an array of values
+    tags: dynamoRecord.tags ? JSON.stringify(dynamoRecord.tags) : undefined,
+    created_at: new Date(dynamoRecord.createdAt),
+    updated_at: new Date(dynamoRecord.updatedAt),
+  };
+
+  return knex('collections').insert(updatedRecord);
```
yjpa7145, line 84: This is returning a value despite the return type of `Promise<void>`

```suggestion
  await knex('collections').insert(updatedRecord);
```

lambdas/data-migration1/src/providers.ts
yjpa7145: Here is an option: https://github.com/nasa/cumulus/pull/1882/files

lambdas/db-migration/src/migrations/20200911163145_create_providers_table.ts
yjpa7145, line 30: Should this be `notNullable`?

markdboyd: Hmm. It wasn't in the original schema, but I suppose if we're enforcing in the migration, it could be? Though we would also need to enforce it with the parallel writes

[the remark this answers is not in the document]
markdboyd: 39a7cfe091e64390e9196b171af342e5b08e8314
24936aa28b3c18cf2b0fbbe99b35d7fc7ce1adbf

lambdas/data-migration1/src/collections.ts
markdboyd: ae2047d5a0498188fdf241e81466c2ee176f0158

yjpa7145 (approved):

PR #1881, opened 2020-09-23
Mhs/cumulus 2162/granule csv fails with large numbers of granules

flamingbear:
**Summary:** moves the granules-csv endpoint into an asyncOperation as a new 'Granule Inventory' report on the '/reconciliationReport' endpoint.

Addresses [CUMULUS-2162: granule-csv fails with large numbers of granules](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2162)

## Changes

* deprecate `@cumulus/api/endpoints/granule-csv/list`
* adds that functionality to the create-reconciliation-report lambda

## PR Checklist

- [X] Update CHANGELOG
- [X] Unit tests
- [x] Adhoc testing
- [X] Integration tests

~Working~ on adhoc testing of large numbers of granules.  Was able to create report 8.1Mb large (too large for direct lambda response) and download the file via the signed url.


packages/api/lambdas/reports/granule-inventory-report.js
```diff
@@ -0,0 +1,65 @@
...
+    'granuleUr',
+    'collectionId',
+    'createdAt',
+    'startDateTime',
+    'endDateTime',
+    'status',
+    'updatedAt',
+    'published',
+  ];
+  const parser = new Parser({ fields });
```
jennyhliu, line 52: Should we use `Transform` here as you did for packages/api/endpoints/granule-csv.js? As `parser` is for small amount of data.

packages/api/lambdas/reports/granule-inventory-report.js
```diff
@@ -0,0 +1,65 @@
...
+  return s3()
+    .putObject({
+      Bucket: systemBucket,
+      Key: reportKey,
+      Body: csv,
+    })
+    .promise();
+}
+
+exports.createGranuleInventoryReport = createGranuleInventoryReport;
```
jennyhliu, line 65: Unit test for this function?

jennyhliu:
I wasn't able to download the `Granule Inventory` report from dashboard.

Screen Shot 2020-09-25 at 1 27 41 PM

packages/api/endpoints/reconciliation-reports.js
```diff
@@ -61,6 +74,7 @@ async function getReport(req, res) {
     }
     throw error;
   }
+  return res.boom.badImplementation('Should never get here');
```
laurenfrederick, line 77: This should have a more user friendly message just in case.

jennyhliu:
> I wasn't able to download the `Granule Inventory` report from dashboard.
> 
After I googled '303', I believe it will be dashboard side change to issue a separate GET message for the directed url to get the data.

packages/api/lambdas/reports/granule-inventory-report.js
flamingbear: Yes I should.

packages/api/lambdas/reports/granule-inventory-report.js
flamingbear: I thought I had done those, but looks like not.  I will add some.

packages/api/endpoints/reconciliation-reports.js
flamingbear: Well, it is a 500 so the user will only ever see {message: "Internal Error"}, but I can try to think of something else to put in there.

flamingbear:
> > I wasn't able to download the `Granule Inventory` report from dashboard.
> 
> After I googled '303', I believe it will be dashboard side change to issue a separate GET message for the directed url to get the data.

Yes, I think we will have to do something on the dashboard side, but I didn't know what that was yet.

jennyhliu (approved):
I was able to create 'Granule Inventory' report via reconciliationReports endpoint and download the file with the signed URL.
Nice work!👍

PR #1882, opened 2020-09-24
Refactor provider migration

yjpa7145:


lambdas/data-migration1/src/providers.ts
```diff
@@ -87,12 +71,21 @@ export const migrateProviderRecord = async (
...
+    const plaintext = encrypted ? await decrypt(username) : username;
+    username = await KMS.encrypt(providerKmsKeyId, plaintext);
   }
 
+  if (password) {
+    const plaintext = encrypted ? await decrypt(password) : password;
+    password = await KMS.encrypt(providerKmsKeyId, plaintext);
+  }
+
+  encrypted = typeof encrypted === 'boolean'
```
yjpa7145, line 85: I don't particularly like this behavior of `encrypt` possibly being null, it makes it ambiguous if the value is encrypted or not. I added this to make sure that my refactor does not alter the behavior that is being verified by the tests, though.

PR #1892, opened 2020-09-28
CUMULUS-2181: update bamboo script for RDS feature branch

markdboyd:


Jkovarik (approved):
👍    This should *hopefully* work, although I'm not convinced the status will be reported.   You should check that's happening in CI before merging.

PR #1894, opened 2020-09-29
CUMULUS-2201: Fix Reconciliation Report integration test failures

jennyhliu:

Addresses [CUMULUS-2201: Investigate Reconciliation Report integration test failures](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2201)

## Changes

* Increase timeout value for checking if a granule is deleted from elasticsearch

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/example/spec/helpers/granuleUtils.js
+++ b/example/spec/helpers/granuleUtils.js
@@ -179,8 +179,8 @@ const waitForGranuleRecordInOrNotInList = async (stackName, granuleId, granuleIs
     return granuleIsIncluded ? ids.includes(granuleId) : !ids.includes(granuleId);
   },
   {
-    interval: 3000,
-    timeout: 240 * 1000,
+    interval: 10000,
+    timeout: 480 * 1000,
   }
 );
 


example/spec/parallel/createReconciliationReport/CreateReconciliationReportSpec.js
```diff
@@ -311,6 +330,13 @@ describe('When there are granule differences and granule reconciliation is run',
         activeCollectionPromise,
       ]);
 
+      console.log('XXXXX Waiting for collections in list');
```
kkelly51, line 333: Are we keeping this in?

example/spec/parallel/createReconciliationReport/CreateReconciliationReportSpec.js
```diff
@@ -239,6 +241,23 @@ async function updateGranuleFile(granuleId, granuleFiles, regex, replacement) {
...
+    const ids = JSON.parse(collsResp.body).results.map((c) => constructCollectionId(c.name, c.version));
+    return isEqual(ids.sort(), collectionIds.sort());
+  },
+  {
+    interval: 10000,
+    timeout: 360 * 1000,
+  }
+);
+
+
```
kkelly51, line 260: Extra line here

example/spec/parallel/createReconciliationReport/CreateReconciliationReportSpec.js
jennyhliu: Yeah.  We leave it for debugging purpose

example/spec/helpers/granuleUtils.js
```diff
@@ -179,8 +179,8 @@ const waitForGranuleRecordInOrNotInList = async (stackName, granuleId, granuleIs
     return granuleIsIncluded ? ids.includes(granuleId) : !ids.includes(granuleId);
   },
   {
-    interval: 3000,
-    timeout: 240 * 1000,
+    interval: 10000,
+    timeout: 480 * 1000,
```
markdboyd, line 183: Is there any context to these specific numbers? I'm a bit concerned that by just upping the timeouts, we may make the failures less frequent, but not have identified/resolved the cause of the issue

jennyhliu: The reconciliation report generation queries collection/granule records from ElasticSearch.  Many failures are due to the timeout when waiting granules to be inserted/deleted into ES.  In this case, wait for the granule added to the ES, then deleted from ES.
https://github.com/nasa/cumulus/blob/99ed6f1137d3191a5def4ddefa0fc93fbbda4820/example/spec/parallel/createReconciliationReport/CreateReconciliationReportSpec.js#L219-L222

Do you know if there is configuration to manage how soon the db records get propagated  into ES. @markdboyd

markdboyd: The DB->ES syncing happens via a lambda triggered on updates to Dynamo, so no, there is no way to predict its timing that I know of. I assume updating these values has made the test more stable in local testing?

jennyhliu: Yeah, it should improve the stability.  We want to make sure the granules/collections are present/absent in ES before creating reconciliation reports.

markdboyd (approved):

PR #1899, opened 2020-09-30
Migrate some ingest files to TypeScript

yjpa7145:


packages/ingest/test/test-consumer.js
```diff
@@ -1,23 +1,30 @@
...
 
 async function stubReceiveSQSMessages(_url, { numOfMessages }) {
   await delay(timeToReceiveMessages);
-  return Array.apply(null, { length: numOfMessages }).map(() => 'i am a message'); // eslint-disable-line prefer-spread
+  return Array.apply(undefined, { length: numOfMessages }).map(() => 'i am a message'); // eslint-disable-line prefer-spread
 }
-consumer.__set__('receiveSQSMessages', stubReceiveSQSMessages);
-consumer.__set__('deleteSQSMessage', async () => true);
+
+const consumer = proxyquire(
```
markdboyd, line 17: More of a preference thing, but could we achieve the same mocking effect by just doing:

```
const SQS = require('@cumulus/aws-client/SQS');
sinon.stub(SQS, 'deleteSQSMessage').resolves(true)
```

If so, is there any benefit to one approach vs the other?

packages/ingest/src/consumer.ts
```diff
@@ -43,20 +63,20 @@ class Consumer {
       log.info(`processing ${messages.length} messages`);
       const processes = messages.map((message) => this.processMessage(message, fn));
       const results = await Promise.all(processes);
-      counter = results.reduce((s, v) => s + v, 0);
+      counter = results.reduce((s: number, v) => s + v, 0);
```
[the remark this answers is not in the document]
yjpa7145: 6ae1955a8

packages/ingest/test/test-consumer.js
yjpa7145: For consistency, 001345cae

packages/ingest/src/types.ts
```diff
@@ -0,0 +1,26 @@
+export interface ProviderClientListItem {
```
markdboyd, line 1: What does this do..?

yjpa7145: It's making sure that the providers all provide the same expected interface (once they're each migrated to TS).

yjpa7145: Each of the providers' `list` method is returning objects with different fields. This is meant to specify the minimum fields that all of the providers provide. It'll get fleshed out more as additional providers are migrated to TS.

PR #1905, opened 2020-10-01
Migrate FtpProviderClient to TS

yjpa7145:


packages/ingest/src/FtpProviderClient.ts
```diff
@@ -99,37 +132,38 @@ class FtpProviderClient {
...
+      client.ls(path, (err: Error | ListError, data) => {
         if (err) {
-          const message = err.message || err.text;
+          const message = isJSFtpError(err) ? err.text : err.message;
           if (message && message.includes('Timed out') && counter < 3) {
             log.error(`Connection timed out while listing ${path}. Retrying...`);
-            counter += 1;
-            return this._list(path, counter).then((r) => {
-              log.info(`${counter} retry succeeded`);
+            return this._list(path, counter + 1).then((r) => {
```
markdboyd, line 148: Does this really behave the same without the `counter += 1`? I guess we don't need to update the value of `counter` in this scope, just pass the incremented value into the recursion?

packages/ingest/src/FtpProviderClient.ts
```diff
@@ -139,15 +173,22 @@ class FtpProviderClient {
...
     // should not be part of the returned result.
-    return files.map((file) => omit(file, 'type'));
+    return files.map((file) => ({
+      name: file.name,
+      path: file.path,
+      size: file.size,
+      time: file.time,
+    }));
+
+    // return files.map((file) => omit(file, 'type'));
```
markdboyd, line 191: Remove this

packages/ingest/src/FtpProviderClient.ts
```diff
@@ -139,15 +173,22 @@ class FtpProviderClient {
...
+  async list(path: string): Promise<ProviderClientListItem[]> {
     const listFn = this._list.bind(this);
     const files = await recursion(listFn, path);
 
     log.info(`${files.length} files were found on ${this.host}`);
 
     // Type 'type' field is required to support recursive file listing, but
     // should not be part of the returned result.
-    return files.map((file) => omit(file, 'type'));
+    return files.map((file) => ({
```
markdboyd, line 184: Why do we only want these file properties?

packages/ingest/src/FtpProviderClient.ts
```diff
@@ -1,63 +1,96 @@
...
+interface FtpProviderClientConstructorParams {
+  host: string;
+  port?: number;
+  useList?: boolean;
+  username?: string;
+  password?: string;
+  encrypted?: boolean;
+}
+
+function isJSFtpError(error: Error | ListError): error is ListError {
```
markdboyd, line 20: What does the `error is ListError` part of this do? Specify that is what the return should be? Because we're obviously testing to determine if this error is a ListError?

packages/ingest/src/jsftp.d.ts
```diff
@@ -0,0 +1,91 @@
+// Type definitions for jsftp 2.1
+// Project: https://github.com/sergi/jsftp
+// Definitions by: Konrad Księski <https://github.com/xyleen>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+/// <reference types="node" />
+
+declare module 'jsftp' {
```
markdboyd, line 8: Man, this is sad to have to maintain this

packages/ingest/src/recursion.ts
```diff
@@ -72,18 +70,18 @@ async function recurOnDirectory(
...
+): Promise<T[]> {
   const normalizedPath = path.normalize(configuredPath);
   const isAbsolutePath = path.isAbsolute(normalizedPath);
   try {
     const segments = normalizedPath
       .split('/') // split on divider
       .filter((segment) => segment.trim() !== ''); // filter out empty strings from split
     const startingPath = isAbsolutePath ? '/' : '.';
-    return await recurOnDirectory(fn, startingPath, segments, -1);
+    return recurOnDirectory<T>(fn, startingPath, segments, -1);
```
markdboyd, line 84: `return await` is required for any thrown error from the recursion to be caught, isn't it?

packages/ingest/src/recursion.ts
```diff
@@ -2,12 +2,10 @@ import path from 'path';
 import * as log from '@cumulus/common/log';
 
 export interface RecursionFile {
-  name: string,
-  type: string
+  type: number,
```
markdboyd, line 5: How could this type change so significantly?

packages/ingest/src/FtpProviderClient.ts
yjpa7145: It's a [User-Defined Type Guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards), which tells TypeScript what type an object is.

packages/ingest/src/FtpProviderClient.ts
yjpa7145: Correct.

packages/ingest/src/FtpProviderClient.ts
yjpa7145: 03396bf58

packages/ingest/src/FtpProviderClient.ts
yjpa7145: Those were the only properties that were being returned before, this just made them more explicit, and easier to define in TS.

packages/ingest/src/jsftp.d.ts
yjpa7145: It is, we should probably issue a PR for the `@types/jsftp` package. That being said, the `jsftp` package hasn't been updated in three years, so the odds of that changing and breaking something is pretty small.

packages/ingest/src/recursion.ts
yjpa7145: 🤷‍♂️

packages/ingest/src/recursion.ts
yjpa7145: 89df42190

yjpa7145 (commented):
@markdboyd I think I've addressed your feedback, this is ready for re-review.

packages/ingest/src/recursion.ts
markdboyd: I'll rephrase: why did we have change it?

packages/ingest/src/jsftp.d.ts
markdboyd: Yeah, it still makes me pretty nervous. 

Alternatively: is there a more maintained package to use?

packages/ingest/src/recursion.ts
yjpa7145: Before this PR, nothing that was actually using `recursion.ts` had been migrated to TypeScript. This PR starts using that recursion code, and it flagged the fact that `type` is actually a number, not a string.

markdboyd (approved):

PR #1906, opened 2020-10-02
Upgrade the got package

yjpa7145:
The newer version of `got` supports TypeScript, which will make those migrations easier.


packages/api/ecs/async-operation/index.js
```diff
@@ -113,14 +115,10 @@ async function fetchLambdaFunction(codeUrl) {
...
   await pRetry(
-    () => new Promise((resolve, reject) => {
-      const file = fs.createWriteStream('/home/task/fn.zip')
-        .on('finish', resolve)
-        .on('error', reject);
-      got.stream(codeUrl)
-        .on('error', reject)
-        .pipe(file);
-    }),
+    () => promisify(pipeline)(
```
markdboyd, line 118: Based on my reading of `pipeline`, if we promisifying [`pipeline`](https://nodejs.org/api/stream.html#stream_stream_pipeline_source_transforms_destination_callback), then we don't have the callback style error handling, so any error encountered here will just be thrown?

packages/common/src/http.ts
```diff
@@ -1,34 +0,0 @@
...
-/**
- * Download a file to disk
- *
- * @param {string} uri - the URI to request
- * @param {string} destination - Where to store file locally
- * @param {Object} [options] - additional download options
- * @param {Object} [options.headers] - headers to include in the request
- * @returns {Promise<undefined>} - resolves when the download is complete
- */
-export const download = (
```
markdboyd, line 18: I love deleting code, but this wasn't deprecated? I'm not necessarily opposed to removing it regardless, but what led you to do so?

packages/ingest/src/HttpProviderClient.js
```diff
@@ -59,13 +60,11 @@ class HttpProviderClient {
...
-    const auth = `${this.username}:${this.password}`;
-    this.gotOptions = {
-      ...this.gotOptions,
-      auth,
-      cookieJar,
-    };
+
+    this.gotOptions.cookieJar = new CookieJar();
+
+    if (this.username) this.gotOptions.username = this.username;
```
markdboyd, line 66: 👍

markdboyd: Though I'm not sure if `gotOptions.username` would achieve anything without `password`, so should we test for both before assigning them to gotOptions?

yjpa7145: Maybe, but that seems like something that should be handled in the `providers` endpoint or model, not here.

packages/common/src/http.ts
yjpa7145: I took it out because it was only used in one place, and the replacement code (which I copied from the `got` documentation) was even simpler than this was.

packages/api/ecs/async-operation/index.js
yjpa7145: That's correct. It's actually a bit safer, since there were potentially events that we _weren't_ handling before. This approach is what is recommended by the `got` documentation.

markdboyd (approved):

PR #1908, opened 2020-10-02
update typescript compilation to be incremental, speeding up build times

markdboyd:


markdboyd:
This seems to have made the build times slightly worse, because the first typescript compilation takes slightly longer to generate the build info that it uses for the subsequent incremental builds. So this may not help in CI, but does help locally

Results from local running a full `npm run bootsrap` with this enabled:

1st run

`npm run bootstrap  863.01s user 104.78s system 211% cpu 7:37.80 total`

2nd run

`npm run bootstrap  493.57s user 55.73s system 205% cpu 4:26.93 total`

PR #1920, opened 2020-10-07
Mhs/cumulus 1/add dopeters to api users

flamingbear:
**Summary:** Add user to default API users in example stack.

PR #1922, opened 2020-10-08
Returns a signedS3Url to get the Granule Inventory report.

flamingbear:
**Summary:** Instead of redirecting with a 303, return 200 and an object that holds the
signedS3Url.

Addresses [CUMULUS-2200:](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2200)

## Changes

* Changes the return from reconcilationReport/reportname.csv from 303 redirect to an object {url: signedS3URL}

## PR Checklist

- [x ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [x ] Integration tests

### DIFF
--- a/example/spec/parallel/createReconciliationReport/CreateReconciliationReportSpec.js
+++ b/example/spec/parallel/createReconciliationReport/CreateReconciliationReportSpec.js
@@ -625,22 +625,22 @@ describe('When there are granule differences and granule reconciliation is run',
       reportRecord = JSON.parse(asyncOperation.output);
     });
 
-    it('Fetches a signed URL to the Granule Inventory report through the Cumulus API', async () => {
+    it('Fetches a object with a signedURL to the Granule Inventory report through the Cumulus API', async () => {
       redirectResponse = await reconciliationReportsApi.getReconciliationReport({
         prefix: config.stackName,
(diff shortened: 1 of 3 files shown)


example/spec/parallel/createReconciliationReport/CreateReconciliationReportSpec.js
```diff
@@ -625,22 +625,22 @@ describe('When there are granule differences and granule reconciliation is run',
       reportRecord = JSON.parse(asyncOperation.output);
     });
 
-    it('Fetches a signed URL to the Granule Inventory report through the Cumulus API', async () => {
+    it('Fetches a object with a signedURL to the Granule Inventory report through the Cumulus API', async () => {
```
dopeters, line 628: Small nitpick, but this should say "an" object

dopeters (approved):
Thank you for making this change! Looks good and does what we need

example/spec/parallel/createReconciliationReport/CreateReconciliationReportSpec.js
flamingbear: I will sneak this into #1920

PR #1930, opened 2020-10-14
Improve example variable definitions file

botanical:

- Comments out optional variables so that, when a user copies and pastes to create their own `terraform.tfvars`, they don't need to manually comment out variables that they aren't using.
## Changes

* Updates `terraform.tfvars.example` in `cumulus-tf` module to comment out optional vars
* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


Jkovarik (approved):
:100: good idea!

PR #1932, opened 2020-10-14
CUMULUS-2230: Add Rules schema migrations & rest of migration skeletons

markdboyd:


lambdas/db-migration/src/migrations/20201014105024_create_rules_table.ts
```diff
@@ -0,0 +1,66 @@
...
+    table
+      .text('name')
+      .comment('Rule name')
+      .notNullable();
+    table
+      .text('workflow')
+      .comment('Workflow name to invoke for this rule')
+      .notNullable();
+    table
+      .integer('collectionCumulusId')
```
nemreid, line 17: I'm not particularly in favor of one way or another, but on your diagram this is called `cumulusCollectionId`. We should try to keep things consistent. If you think this is the better name, then we should update the diagram.

lambdas/db-migration/src/migrations/20201014105024_create_rules_table.ts
```diff
@@ -0,0 +1,66 @@
...
+      .references('cumulusId')
+      .inTable('collections')
+      .notNullable();
+    table
+      .integer('providerCumulusId')
```
nemreid, line 22: Same note about diagram naming this `cumulusProviderId`

lambdas/db-migration/src/migrations/20201014105024_create_rules_table.ts
```diff
@@ -0,0 +1,66 @@
...
+      .jsonb('meta')
+      .comment('Optional metadata for the rule. Contents will be automatically added to $.meta on invoked workflows.');
+    table
+      .jsonb('tags')
+      .comment('Optional tags for the rule');
+    table
+      .text('queueUrl')
+      .comment('Optional SQS queue URL used to schedule executions for this rule');
+    table
+      .timestamps(false, true);
```
nemreid, line 61: Can you explain the preference for datetime over timestamp for my understanding?

nemreid (commented):
Looks good overall! Just some questions.

lambdas/db-migration/src/migrations/20201014105024_create_rules_table.ts
markdboyd: Yeah, I think I prefer this naming convention. I'll update the diagram

lambdas/db-migration/src/migrations/20201014105024_create_rules_table.ts
markdboyd: For Knex, I'm not sure if it makes a difference:

http://knexjs.org/#Schema-datetime
http://knexjs.org/#Schema-timestamp

However, using `timestamps` is nice because it creates the `created_at` and `updated_at` columns automatically:

http://knexjs.org/#Schema-timestamps

PostgreSQL docs on date/time column types:

https://www.postgresql.org/docs/11/datatype-datetime.html

nemreid (approved):

lambdas/db-migration/src/migrations/20201014105024_create_rules_table.ts
```diff
@@ -0,0 +1,66 @@
...
+      .integer('collectionCumulusId')
+      .references('cumulusId')
+      .inTable('collections')
+      .notNullable();
+    table
+      .integer('providerCumulusId')
+      .references('cumulusId')
+      .inTable('providers')
+      .notNullable();
+    table
```
markdboyd, line 26: It's probably worth calling out that this used to be nested under the `rule` object on the Dynamo schema, so `rule.type`. But given that part of the point of going to RDS is to facilitate easier querying on any record field, it seemed worth pulling all the `rule.*` properties to the top level (`type`, `value`, `arn`, `logEventArn`)

nemreid: I think that's fine. We'll need to keep it in mind when adding rule parallel writes to transform, but that'll shake out.

PR #1936, opened 2020-10-15
Update logging docs

markdboyd:
Fixing some docs that didn't get properly updated for the move to deploying TEA as a separate module


nemreid (approved):
LGTM

docs/deployment/README.md
```diff
@@ -214,12 +214,6 @@ $ aws dynamodb create-table \
     --region us-east-1
 ```
 
-### Create a secret for signing Thin Egress App JWTs
```
laurenfrederick, line 217: In the template deploy we still have `thin_egress_app` in the cumulus-tf deployment, which needs this. I think it should either be removed from there and the docs updated on how to do a distribution deployment or this should stay.

laurenfrederick: I am thinking have this stay and update the deployment docs to call out the configuration. I am making deployment doc updates so can do that. What do you think?

laurenfrederick (changes requested):

laurenfrederick (approved):

PR #1944, opened 2020-10-19
[CUMULUS-2170] Add FilterExpression to granule scan

dopeters:
**Summary:** Makes Granule Inventory Reports filterable

Addresses [CUMULUS-2170: Update reconciliationReport with Granule Inventory type endpoint to take filter parameters](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2170)

## Changes

* Pass searchParams to `granuleAttributeScan` when creating a granule inventory report
* Use searchParams to create FilterExpression for DynamoDbSearchQueue
* Made a separate function to handle creating the params to pass to the search queue so that it could be reused.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [ ] Integration tests


packages/api/lambdas/reports/granule-inventory-report.js
```diff
@@ -38,6 +30,24 @@ async function createGranuleInventoryReport(recReportParams) {
     'updatedAt',
     'published',
   ];
+
+  const { reportKey, systemBucket, ...params } = recReportParams;
+  const searchParams = {};
+  fields.forEach((field) => {
+    let searchField = field;
+    if (field === 'granuleUr') searchField = 'granuleId';
```
kkelly51, line 38: Are they not named the same thing?

dopeters: The csv column is called `granuleUr`, while the API accepts `granuleId`. Didn't want to change the output, but wanted to use the fields as an easy way of getting relevant search params, so just had it replace the name here.

packages/api/lib/reconciliationReport.js
```diff
@@ -117,6 +117,30 @@ function convertToDBGranuleSearchParams(params) {
...
 
+/**
+ *
+ * @param {Object} params - request params to convert to database params
+ * @returns {Object} object of desired parameters formated for database granule search
+ */
+function convertToDBScanGranuleSearchParams(params) {
+  const {
+    collectionId,
+    granuleId,
```
jennyhliu, line 128: Is the `granuleId` from the `search` box in dashboard granules page?  The search field is actually `infix`, matches part of the granuleId.

jennyhliu: nvm.  I see `contains` is used for the db search.

packages/api/models/granules.js
```diff
@@ -446,16 +420,72 @@ class Granule extends Manager {
...
       }
     });
 
+    return {
+      attributeNames,
+      attributeValues,
+      filterArray,
+      filterExpression: (filterArray.length > 0) ? filterArray.join(' AND ') : undefined,
+      keyConditionArray,
+      keyExpression: keyConditionArray.join(' AND '),
```
jennyhliu, line 443: `keyExpression` is not used

packages/api/models/granules.js
```diff
@@ -446,16 +420,72 @@ class Granule extends Manager {
...
+    // Filter Expression can't contain key attributes
+    if (searchParams.collectionId) {
+      throw new CumulusModelError('Could not search granule records for collection, do not specify collectionId in searchParams');
+    }
+
+    const {
+      attributeNames,
+      attributeValues,
+      filterExpression,
+      keyConditionArray,
```
jennyhliu, line 484: use `keyExpression` instead?

jennyhliu (changes requested):
Looks good!
We can also add search parameters to https://github.com/nasa/cumulus/blob/master/example/spec/parallel/createReconciliationReport/CreateReconciliationReportSpec.js#L608
and verify that the selected granules/collection are reported.

jennyhliu:
https://github.com/nasa/cumulus/blob/master/packages/api/endpoints/reconciliation-reports.js#L122
You probably need to update the function which validates the input parameters.
Api doc needs to be updated.  https://github.com/nasa/cumulus-api

jennyhliu (approved):
Nice job! LGTM

jennyhliu (changes requested):
If I change the granuleId parameter to granuleId: 'MYD13Q1', in the integration test, the report doesn't have anything

packages/api/lib/reconciliationReport.js
```diff
@@ -117,6 +117,36 @@ function convertToDBGranuleSearchParams(params) {
...
 
+/**
+ *
+ * @param {Object} params - request params to convert to database params
+ * @returns {Object} object of desired parameters formated for database granule search
+ */
+function convertToDBScanGranuleSearchParams(params) {
+  const {
+    collectionId,
+    granuleId: granuleIdParam,
```
jennyhliu, line 128: There won't be `granuleId` parameter after normalizeEvent, only `granuleIds`

dopeters: Ahhh yeah I see that now. Thanks, will adjust

jennyhliu: Can you add a test case for the lambda `packages/api/tests/lambdas/test-create-reconciliation-report.js`?

dopeters: Absolutely!

dopeters:
> I think it should be `granuleId`.
> Can you also add `status` here and lambda test?
> Don't forget api doc.

Yep, you're right. Fixed.
Added status as a param.
Opened a PR for the docs here: https://github.com/nasa/cumulus-api/pull/295

dopeters:
Dashboard side, we were passing `granuleIds` instead of `granuleId` as the param, so opened up a PR to fix that https://github.com/nasa/cumulus-dashboard/pull/866

packages/api/models/granules.js
dopeters: Stopped `keyExpression` from being returned in `getDynamoDbSearchParams`. Want to use `keyConditionArray` here because we end up adding the collectionId to that array later in this function

`keyConditionArray.push('#collectionId = :collectionId');`

jennyhliu (approved):

PR #1945, opened 2020-10-19
CUMULUS-2239 - Add new aws_vpc_endpoint resource for deploying outside of NGAP

botanical:

- Adds infrastructure object code to create a VPC Endpoint for the service `com.amazonaws.${var.region}.execute-api` automatically if `deploy_to_ngap` is set to false.
## Changes

* Updates various terraform variables file
* Updates `tf-modules/tea-map-cache/main.tf` to include new resource definition

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


botanical:
Ran `terraform apply` locally with `deploy_to_ngap` set to `false` and verified in AWS Console that VPC Endpoint was created
Screen Shot 2020-10-19 at 2 22 55 PM

tf-modules/cumulus/variables.tf
```diff
@@ -526,3 +526,8 @@ variable "ems_deploy" {
   type        = bool
   default     = false
 }
+
+variable "region" {
+  type    = string
+  default = "us-east-1"
```
Jkovarik, line 532: see comment above re: the variable itself, but if this winds up being needed, having a default here should be noted as a breaking change as user configuration defaults would be directly impacted.    We should also consider us-west-(1,2?) as that where most of our customers are deploying.

tf-modules/tea-map-cache/main.tf
```diff
@@ -3,6 +3,18 @@ terraform {
     aws = ">= 2.31.0"
   }
 }
+
+resource "aws_vpc_endpoint" "config" {
```
Jkovarik, line 7: What happens here if VPC/lambda subnet IDs is null?     

Looking at the other existing code, it appears that it's (possibly  illegitimately) assuming no VPC subnets === null VPC, so that's not great (and not related to this PR!), but in a situation where VPC is null/no subnets, this would still break.      

This should probably be conditional on vpc_id.

Jkovarik (changes requested):
This looks good! Just one concern re: the region variable, the others are nits/minor.

CHANGELOG.md
```diff
@@ -5,8 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
-
 ### Added
+- Add resource declaration to create a VPC endpoint in tea-map-cache module if `deploy_to_ngap` is false.
```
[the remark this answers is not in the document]
botanical: Thanks for pointing this out! I don't believe this has a ticket. I can create one and then update the changelog to include it

tf-modules/tea-map-cache/main.tf
botanical: Good point!

tf-modules/cumulus/distribution.tf
```diff
@@ -23,6 +23,8 @@ module "distribution" {
   urs_client_password                            = var.urs_client_password
   urs_url                                        = var.urs_url
   vpc_id                                         = var.vpc_id
+  deploy_to_ngap                                 = var.deploy_to_ngap
+  region                                         = var.region
```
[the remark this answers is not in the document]
botanical: Based on your suggestions, I've included region as a data source

tf-modules/cumulus/variables.tf
Jkovarik: This should be removed after 01fa54f2a5ef5942c98dd60ca27051c9a39c2bf0

Jkovarik (changes requested):
Looks good - just one small removal and this should be good to go.

tf-modules/cumulus/variables.tf
botanical: Whoops! Not sure how that got back in. Let me update.

tf-modules/tea-map-cache/main.tf
```diff
@@ -3,6 +3,20 @@ terraform {
     aws = ">= 2.31.0"
   }
 }
+
+data "aws_region" "current" {}
+
+resource "aws_vpc_endpoint" "config" {
+  count = var.deploy_to_ngap && (var.vpc_id == null || var.lambda_subnet_ids == null) ? 0 : 1
```
Jkovarik, line 10: I'm sorry I missed this before (and with the caveat I don't have a non-NGAP account to really test it on): 
 

var.lambda_subnet_ids is a list of strings, with a default of `[]`, this should be checking it's length.  

Also - If deploy_to_ngap is true, *but* a user doesn't have vpc/subnets defined, this will evaluate to a count of 1 and attempt to deploy a vpc endpoint into NGAP.     I think we need to invert this so that in any situation `deploy_to_ngap` is true, it never tries to add the endpoint.  

How about 
```suggestion
  count = !var.deploy_to_ngap && (var.vpc_id == null || length(var.lambda_subnet_ids) != 0) ? 1 : 0
```

That gives you 

!T & (F) = 0
!T & (T) = 0
!F & (F) = 0
!F & (T) = 1

Jkovarik: Which.....removes the entire reason for having this conditional.   🤔

botanical: Ah, that's true, `lambda_subnet_ids` is a list of strings. My question is, when deploying to NGAP, are the VPC endpoints already configured? In any case, I do think we need the inversion as you suggested because we would _only_ want the VPC endpoint to be created in the case that the user is deploying outside of NGAP and that they have vpc IDs defined, ie:
```
deploy_to_ngap = false
vpc_id is _not_ null
and length(lambda_subnet_ids) != 0

!false & (true || true) 
which evaluates to true & true = 1

or
deploy_to_ngap = false
vpc_id is _not_ null
and length(lambda_subnet_ids) == 0

!false & (true || false)
which evaluates to true & true = 1
```
Technically, we can create a vpc_endpoint without specifying subnet IDs but I think the deployment would fail anyway if there were no subnet IDs specified by the user so I am inclined to rewrite it as follows:
```
count = !var.deploy_to_ngap && (var.vpc_id == null && length(var.lambda_subnet_ids) != 0) ? 1 : 0
```

botanical:
Verified with new logic where `deploy_to_ngap` was set to `false` and both VPC and subnet IDs were defined.

Screen Shot 2020-10-20 at 7 27 30 PM

tf-modules/tea-map-cache/main.tf
Jkovarik: 💯  that looks good.

Jkovarik (approved):

PR #1950, opened 2020-10-21
Mhs/cumulus 2216/make statistics inclusion optional on the collections endpoint

flamingbear:
**Summary:**

Splits es/Collection into BaseCollection and Collection and leaves the granule statistics aggregation in Collection, while moving the base functionality into BaseCollection.

/collection and /collection/active endpoints now return collections without aggregate statistics by default. To get the original behavior a query param of `includeStats=true` must be present on the request

Addresses [CUMULUS-2226: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2216)

## Changes
* Collection endpoint looks for an `includeStats` param and chooses which type of collection to instantiate before performing the endpoint's search.

## PR Checklist

- [X] Update CHANGELOG
- [X] Unit tests
- [X] Adhoc testing
- [ ] Integration tests


jennyhliu:
The `Example request` for `List collections` and `List collections with active granules` doesn't have parameter `includeStats=true`, but  the `Example Response` includes `stats`, this could cause confusion

packages/api/endpoints/collections.js
```diff
@@ -27,14 +28,16 @@ const log = new Logger({ sender: '@cumulus/api/collections' });
  * @returns {Promise<Object>} the promise of express response object
  */
 async function list(req, res) {
-  const { getMMT, ...queryStringParameters } = req.query;
-  const collection = new Collection(
+  const { getMMT, includeStats, ...queryStringParameters } = req.query;
+  const CollectionClass = includeStats === 'true' ? Collection : BaseCollection;
```
jennyhliu, line 32: Can we pass `includeStats` to `Collection` class, and let the class decide if the stats should be included or not?

flamingbear: I could do that. I actually thought this was cleaner. But I'm fine changing it if you think that is a better way to do it.

jennyhliu: I prefer letting `Collection` class worry about the `includeStats` flag.

flamingbear: Ok, I'm putting that up already.  Thanks.

flamingbear:
closing in favor of #1954

PR #1954, opened 2020-10-22
Mhs/cumulus 2216/2 make statistics inclusion optional with parameter

flamingbear:
**Summary:**

Adds new parameter to es/Collection to toggle the appending of aggregate granule statistics to the returns.

/collection and /collection/active endpoints now OMIT aggregate statistics by default. To get the original behavior a query param of `includeStats=true` must be present on the request.

Addresses [CUMULUS-2226: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2216)

## Changes
* Collection endpoint looks for an `includeStats` query paramameter and passes that to the Collection constructor separately.

## PR Checklist

- [X] Update CHANGELOG
- [X] Unit tests
- [X] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/packages/api/es/collections.js
+++ b/packages/api/es/collections.js
@@ -8,9 +8,9 @@ const { BaseSearch } = require('./search');
 const ES_MAX_AGG = 2147483647;
 
 class Collection extends BaseSearch {
-  constructor(event, type, index) {
+  constructor(event, type, index, includeStats = 'true') {
     super(event, type || 'collection', index);
-
+    this.includeStats = includeStats === 'true';
     // decrease the limit to 50
     this.size = this.size > 50 ? 50 : this.size;
   }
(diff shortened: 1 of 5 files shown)


packages/api/es/collections.js
```diff
@@ -8,9 +8,9 @@ const { BaseSearch } = require('./search');
 const ES_MAX_AGG = 2147483647;
 
 class Collection extends BaseSearch {
-  constructor(event, type, index) {
+  constructor(event, type, index, includeStats = true) {
```
jennyhliu, line 11: Should it default to false?

flamingbear: I didn't think so, because I didn't look to see how often it was being used before.  I did make the endpoints default to false.  I'll take a look.

flamingbear: Looks like it's only used in the create-reconciliation-report lambda and some tests.  So I can make it default to false

flamingbear: ef9077e
(and we don't actually need the stats in reconciliation report, so I left it alone.)

jennyhliu (approved):
LGTG.  Thanks!

PR #1957, opened 2020-10-22
CMA module output = cumulus module expected input

### LINKED ISSUE #1361: It's confusing whether I should specify Arn with version or not for CMA.
CMA deployment reports two things.
```
    "LayerArn": "arn:aws:lambda:us-east-1:12345:layer:tf-CMA-layer",
    "LayerVersionArn": "arn:aws:lambda:us-east-1:12345:tf-CMA-layer:1"
```

The following TF variable name seems to match ```LayerArn``` but example matches ```LayerVersionArn```.

```
# Replace 12345 with your actual AWS account ID
cumulus_message_adapter_lambda_layer_arn = "arn:aws:lambda:us-east-1:12345:layer:Cumulus_Message_Adapter:4"
```

### LINKED ISSUE #1956: Add cumulus-message-adapter-tf module output to match cumulus-tf input
https://github.com/nasa/cumulus/blob/1e9119edb099a2670740e6012712c1375d576f7d/tf-modules/cumulus-message-adapter/outputs.tf#L2

Currently the cumulus-message-adapter module outputs only the layer_arn, which does not include the version. (See [terraform documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_layer_version#arn))
However, the cumulus-tf module is expecting the arn *with* the version as an input variable, or it fails.

By allowing these to match values, one can chain these modules together, for example:

```
cumulus_message_adapter_lambda_layer_arn = module.cumulus-message-adapter.outputs.cumulus_message_adapter_layer_arn
```

elkirby:
Modifies cumulus-message-adapter module output to match expected cumulus module input. Fixes #1956, fixes #1361 

Addresses Issue #1956: Add cumulus-message-adapter module output to match cumulus-tf input
## Changes

* Changed existing cumulus_message_adapter_layer_arn output to the expected value of the cumulus-tf variable with the same name
* Added a new output, cumulus_message_adapter_layer_layer_arn to cover the original output value. (Follows naming convention of the [Terraform attributes for aws_lambda_layer_version](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_layer_version#arn))

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/tf-modules/cumulus-message-adapter/outputs.tf
+++ b/tf-modules/cumulus-message-adapter/outputs.tf
@@ -1,3 +1,7 @@
 output "cumulus_message_adapter_layer_arn" {
+  value = aws_lambda_layer_version.cumulus_message_adapter.arn
+}
+
+output "cumulus_message_adapter_layer_layer_arn" {
   value = aws_lambda_layer_version.cumulus_message_adapter.layer_arn
 }
--- a/tf-modules/cumulus-message-adapter/outputs.tf
+++ b/tf-modules/cumulus-message-adapter/outputs.tf
@@ -1,3 +1,3 @@
-output "cumulus_message_adapter_layer_arn" {
+output "cumulus_message_adapter_lambda_layer_version_arn" {
   value = aws_lambda_layer_version.cumulus_message_adapter.arn
 }
--- a/tf-modules/cumulus-message-adapter/outputs.tf
+++ b/tf-modules/cumulus-message-adapter/outputs.tf
@@ -1,7 +1,3 @@
 output "cumulus_message_adapter_layer_arn" {
   value = aws_lambda_layer_version.cumulus_message_adapter.arn
 }
-
-output "cumulus_message_adapter_layer_layer_arn" {
-  value = aws_lambda_layer_version.cumulus_message_adapter.layer_arn
-}
--- a/bamboo/bootstrap-tf-deployment.sh
+++ b/bamboo/bootstrap-tf-deployment.sh
@@ -76,7 +76,7 @@ echo "Deploying Cumulus example to $DEPLOYMENT"
   -input=false \
   -var-file="../deployments/cumulus/$BASE_VAR_FILE" \
   -var-file="../deployments/cumulus/$DEPLOYMENT.tfvars" \
-  -var "cumulus_message_adapter_lambda_layer_arn=arn:aws:lambda:us-east-1:$AWS_ACCOUNT_ID:layer:Cumulus_Message_Adapter:$CMA_LAYER_VERSION" \
+  -var "cumulus_message_adapter_lambda_layer_version_arn=arn:aws:lambda:us-east-1:$AWS_ACCOUNT_ID:layer:Cumulus_Message_Adapter:$CMA_LAYER_VERSION" \
   -var "cmr_username=$CMR_USERNAME" \
   -var "cmr_password=$CMR_PASSWORD" \
   -var "cmr_client_id=cumulus-core-$DEPLOYMENT" \
--- a/docs/workflows/input_output.md
+++ b/docs/workflows/input_output.md
@@ -25,7 +25,7 @@ In order to make use of this configuration, a Lambda layer must be uploaded to y
 
 Once you've deployed the layer, integrate the CMA layer with your Lambdas:
 
-- If using the `cumulus` module, set the `cumulus_message_adapter_lambda_layer_arn` in your `.tfvars` file to integrate the CMA layer with all core Cumulus lambdas.
+- If using the `cumulus` module, set the `cumulus_message_adapter_lambda_layer_version_arn` in your `.tfvars` file to integrate the CMA layer with all core Cumulus lambdas.
 - If including your own Lambda or ECS task Terraform modules, specify the CMA layer ARN in the Terraform resource definitions.  Also, make sure to set the `CUMULUS_MESSAGE_ADAPTER_DIR` environment variable for the task to `/opt` for the CMA integration to work properly.
 
 In the future if you wish to update/change the CMA version you will need to update the deployed CMA, and update the layer configuration for the impacted Lambdas as needed.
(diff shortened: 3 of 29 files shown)


elkirby:
I haven't done the PR Checks yet because
* I'm not sure if this change is actually ... desired
* also unsure on what the final verbiage should be for a CHANGELOG
* Am fully open to suggestion on what the actual outputs should be named, since I know `layer_layer_arn` is a bit repetitive.

tf-modules/cumulus-message-adapter/outputs.tf
```diff
@@ -1,3 +1,7 @@
 output "cumulus_message_adapter_layer_arn" {
```
markdboyd, line 1: Thanks so much for this contribution @elkirby !

I know that we'd be breaking with the Terraform conventions, but maybe we call this output `cumulus_message_adapter_layer_version_arn` and the one below it `cumulus_message_adapter_layer_arn`. Avoids the repetitive names and is maybe clear about which output contains the version.

Alternately, do we even need an output for the ARN without the version? It seems like anything using this module needs the layer ARN with the version in it for compatibility with the Cumulus module.

Thoughts?

elkirby: No problem -- happy to contribute! I'm new to the whole project and am totally fine with whatever people decide makes the most sense. 

I personally like the naming that includes the word version, and will admit I didn't see anywhere that the version wasn't also required. 

----
Finally, and this is a bit beyond the scope of this PR, but, since we're talking naming, I thought I'd still bring it up for discussion... 😅 

Before writing this issue, I found #1361, which talks about not knowing whether to specify version or not for `cumulus_message_adapter_lambda_layer_arn` in the cumulus module. Your `cumulus_message_adapter_layer_version_arn` suggestion got me wondering if maybe applying this naming in both modules would solve that issue, in addition to keeping the names consistent and making it obvious that they're the same value. I'll admit that I too got tripped up originally by the issue outlined in #1361, and even when writing the linked issue for this PR, the description had to be edited because I originally got their names wrong.

markdboyd: I like the `cumulus_message_adapter_layer_version_arn` naming, and it's nice that it would also address #1361. It would be a breaking change for deployment configuration, albeit a very minor one. Let's see what the team thinks.

@Jkovarik @laurenfrederick @nemreid - thoughts?

markdboyd: @elkirby We discussed this as a team and we'd like to move forward with this approach:

- Change the output for the layer version ARN to `cumulus_message_adapter_lambda_layer_version_arn` in `tf-modules/cumulus_message_adapter/outputs.tf`. Remove the non-layer version output.
- Change the variable name for the layer from `cumulus_message_adapter_lambda_layer_arn` to `cumulus_message_adapter_lambda_layer_version_arn` in both `tf-modules/cumulus/variables.tf` and `tf-modules/ingest/variables.tf`

Would you be up for making those changes?

elkirby: @markdboyd, 
I would absolutely be up to it. 

Restating my understanding, just to be sure:
1. in cumulus-message-adapter/outputs.tf, there should be **one** output: a variable called `cumulus_message_adapter_lambda_layer_version_arn`, whose output should be `aws_lambda_layer_version.cumulus_message_adapter.arn` (that is, the ARN with the version appended)
2. Change ***all*** references of `cumulus_message_adapter_lambda_layer_arn` to `cumulus_message_adapter_lambda_layer_version_arn` in the following modules:
2a. `cumulus`
2b. `ingest`

Am I on the same page?

Does this cause downstream changes to other repositories that I should be aware of? I know cumulus-template-deploy [does not currently have a direct hook-up](https://github.com/nasa/cumulus-template-deploy/blob/master/cumulus-tf/main.tf#L43) between `cumulus-message-adapter` and `cumulus` child modules (which I have some ideas about how that could be done ...for another time 😅), but are there others I should be aware of that I could/should submit a contribution to?

markdboyd: @elkirby Yep, your understanding is correct. Would you also mind updating `example/cumulus-tf/variables.tf` to replace `cumulus_message_adapter_lambda_layer_arn` with `cumulus_message_adapter_lambda_layer_version_arn`? This file is only used internally by the Cumulus core team, but it would be helpful to us 😄 

In terms of other repos, yeah the only other one would be cumulus-template-deploy, where we would want to update all instances of `cumulus_message_adapter_lambda_layer_arn` to `cumulus_message_adapter_lambda_layer_version_arn`. Thanks so much!

elkirby: ✅ Done!
Everything is still split up into very small commits to make it clear what I did, and to make it easy to drop anything that y'all didn't want. I saw a couple other references to the variable, so once I validated that it made sense to do so, I added commits for each of those changes too. I will annotate them. Feel free to request other revisions, additions, or redactions.

markdboyd:
Fixes #1361

tf-modules/cumulus-message-adapter/outputs.tf
markdboyd: This is great @elkirby! Thanks for catching all those references that I didn't even mention. Going to merge this now. 💯 awesome contribution!

markdboyd (approved):

tf-modules/cumulus-message-adapter/outputs.tf
markdboyd: Also, just to shed some light on our PR merge process from forks: we run our test suite on all PRs before merge as a blocking step, but we don't run any tests PRs from forks for security reasons, so I have to merge your work to an intermediary branch off of `master` and then merge that branch to `master`. And that PR is open now:

https://github.com/nasa/cumulus/pull/1977

PR #1969, opened 2020-10-23
[CUMULUS-2138] Add update-granules-metadata task

botanical:

Addresses [CUMULUS-2138: As a user I would like the move-granules task to update the metadata files when moveStagedFiles is false](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2138)

## Changes
* Created `update-granules-metadata` task
  * Decouples updating CMR metadata and adding CMR file etags from `move-granules` task and puts them into a new `update-granules-metadata tasks`
  * Updates unit tests on said functionality to account for `.cmr.json` files in addition to `.cmr.xml` files.
* Updates `example/cumulus-tf/ingest_and_publish_granule_workflow.asl.json` to incorporate new `UpdateGranulesMetadata` step and call to `HyraxMetadataUpdatesTask`
* Updates README to point to correct HelloWorld spec
* Updates FileUtils to log when there is an error
* Updates `config.yml` to include my internal bucket
* Updates `example/spec/parallel/ingestGranule/IngestGranuleSuccessSpec.js` to account for added step in workflow

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Adhoc testing
- [x] Integration tests


botanical:
Tested locally against `spec/parallel/ingestGranule/IngestGranuleSuccessSpec.js` and `spec/parallel/ingestGranule/IngestUMMGSuccessSpec.js` and both integration tests were successful.

tasks/update-granules-metadata/package.json
```diff
@@ -0,0 +1,47 @@
...
+  },
+  "author": "Cumulus Authors",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@cumulus/aws-client": "2.0.1",
+    "@cumulus/cmrjs": "2.0.1",
+    "@cumulus/common": "2.0.1",
+    "@cumulus/cumulus-message-adapter-js": "1.3.1",
+    "@cumulus/errors": "2.0.1",
+    "@cumulus/ingest": "2.0.1",
```
markdboyd, line 44: I don't think either of these packages are used in the code for this task that I can tell?

tasks/update-granules-metadata/schemas/config.json
```diff
@@ -0,0 +1,72 @@
+{
+  "title": "UpdateGranulesMetadataConfig",
+  "description": "Describes the config used by the update-granules-metadata task",
+  "type": "object",
+  "required": [
+    "bucket",
+    "distribution_endpoint",
+    "collection",
```
markdboyd, line 8: I'm not seeing where config.collection is used in the task, so I think that can be removed from the config schema?

tasks/update-granules-metadata/schemas/config.json
```diff
@@ -0,0 +1,72 @@
+{
+  "title": "UpdateGranulesMetadataConfig",
+  "description": "Describes the config used by the update-granules-metadata task",
+  "type": "object",
+  "required": [
+    "bucket",
```
markdboyd, line 6: I also don't think config.bucket is used in the task code?

example/cumulus-tf/ingest_and_publish_granule_workflow.asl.json
```diff
@@ -237,6 +237,34 @@
...
     },
+    "UpdateGranulesMetadataStep": {
+      "Parameters": {
+        "cma": {
+          "event.$": "$",
+          "task_config": {
+            "bucket": "{$.meta.buckets.internal.name}",
+            "buckets": "{$.meta.buckets}",
+            "distribution_endpoint": "{$.meta.distribution_endpoint}",
+            "collection": "{$.meta.collection}"
```
markdboyd, line 248: Based on the comments below, I think we might not need `bucket` or `collection` task config for this task

tasks/update-granules-metadata/tests/test-index.js
```diff
@@ -0,0 +1,122 @@
...
+});
+
+test.afterEach.always(async (t) => {
+  await recursivelyDeleteS3Bucket(t.context.publicBucket);
+  await recursivelyDeleteS3Bucket(t.context.stagingBucket);
+  await recursivelyDeleteS3Bucket(t.context.protectedBucket);
+  await recursivelyDeleteS3Bucket(t.context.systemBucket);
+});
+
+test.serial('Should add etag to each CMR metadata file by checking that etag is one of more characters not whitespace', async (t) => {
```
markdboyd, line 99: I think it might be helpful to have another test where the incoming payload already has an etag on the CMR file and verify that the etag changes after running the task. This scenario would not occur for any of current workflows, but it could if people use this task later in a workflow chain. What do you think?

tasks/update-granules-metadata/tests/data/payload.json
```diff
@@ -0,0 +1,121 @@
...
+          "name": "cumulus-protected",
+          "type": "protected"
+        },
+        "public": {
+          "name": "cumulus-public",
+          "type": "public"
+        }
+      },
+      "distribution_endpoint": "https://something.api.us-east-1.amazonaws.com/",
+      "collection": {
```
markdboyd, line 23: I think you might not need this for your task, in which case it'd good to simplify the fixture

tasks/update-granules-metadata/tests/test-index.js
```diff
@@ -0,0 +1,122 @@
...
+  const newPayload = buildPayload(t);
+  const filesToUpload = cloneDeep(t.context.filesToUpload);
+  await uploadFiles(filesToUpload, t.context.stagingBucket);
+
+  const output = await updateGranulesMetadata(newPayload);
+  output.granules[0].files
```
markdboyd, line 105: Since your input payload has two granules, each with a CMR file, why not do this test against the CMR file for both granules?

markdboyd (changes requested):
A few requested changes and comments

tasks/update-granules-metadata/tests/test-index.js
botanical: Good point, I forgot to add the second test!

tasks/update-granules-metadata/tests/test-index.js
botanical: @markdboyd 
Yes, I think that is useful to test. I think I need more context on what that workflow could look like. Would a workflow chain in which that scenario might occur look like this:

```
      sync_granule_task_arn: module.cumulus.sync_granule_task.task_arn,
      add_missing_file_checksums_task_arn: 
      module.cumulus.add_missing_file_checksums_task.task_arn,
      fake_processing_task_arn: module.cumulus.fake_processing_task.task_arn,
      files_to_granules_task_arn: module.cumulus.files_to_granules_task.task_arn,
      move_granules_task_arn: module.cumulus.move_granules_task.task_arn,
      hyrax_metadata_updates_task_arn: module.cumulus.hyrax_metadata_updates_task.task_arn,
      post_to_cmr_task_arn: module.cumulus.post_to_cmr_task.task_arn,
      update_granules_metadata_task_arn: module.cumulus.update_granules_metadata_task.task_arn
```
where it updates the granules metadata last? Correct me if I am wrong, but etags are updated on any s3 PUT event, right? So, in this case, hyrax metadata update will write to s3 and it will consequently update the etag. Let me know if there is a fault in my understanding/reasoning!

tasks/update-granules-metadata/schemas/config.json
botanical: Yeah, that's true. I'll remove it

tasks/update-granules-metadata/tests/test-index.js
markdboyd: I would just update this one test to run the same assertion for the CMR file in both granules

tasks/update-granules-metadata/tests/test-index.js
markdboyd: > etags are updated on any s3 PUT event

Yes, that is correct. 

That workflow you outlined would trigger the behavior - but I don't think it would ever occur in reality. Once `post_to_cmr ` is run, the metadata on S3 has already been posted to CMR , so there's no need to update the metadata filed stored on S3.

Really, I can't actually think of a real life workflow where update_granules_metadata might run and receive etags as part of the input, but keep in mind that users can assemble our tasks into workflows however they want including with their own custom tasks. So my motivation for testing this is because it is **theoretically** possible for the task to receive input with etags on the files and it seems easy enough to test.

tasks/update-granules-metadata/tests/test-index.js
botanical: Yeah, that's what I ended up doing

tasks/update-granules-metadata/schemas/config.json
botanical: It's not but I was running into a strange error when I removed it.

example/cumulus-tf/ingest_and_publish_granule_workflow.asl.json
botanical:
```"CumulusMessageAdapterExecutionError: warning setting command to loadAndUpdateRemoteEvent",
    "warning setting command to loadNestedEvent",
    "Unexpected Error <class 'jsonschema.exceptions.ValidationError'>. config schema: 'collection' is a required property",

```
and
```
"errorType": "CumulusMessageAdapterExecutionError",
  "errorMessage": "warning setting command to loadAndUpdateRemoteEvent\nwarning setting command to loadNestedEvent\nUnexpected Error <class 'jsonschema.exceptions.ValidationError'>. config schema: 'collection' is a required property\n\nFailed validating 'required' in schema:\n    {'description': 'Describes the config used by the '\n                    'update-granules-metadata task',\n     'properties': {'bucket': {'description': 'the bucket the has the '\n                                              'private/public key needed '\n                                              'for decrypting cmr password',\n                               'type': 'string'},\n              
```

markdboyd: Yeah, so these errors occurred because you removed them from the `task_config`, which sets the `config.collection` value in the task, but you hadn't removed them from the task config schema. So the value of `config.collection` was empty, but the task config schema was still requiring `config.collection` to be present.

So if you remove `bucket` and `collection` from tasks/update-granules-metadata/schemas/config.json and then also from this `task_config`, I would expect those errors to go away.

tasks/update-granules-metadata/tests/data/payload.json
```diff
@@ -0,0 +1,86 @@
...
+             {
+              "name": "MOD11A1.A2017200.h19v04.006.2017201090723.cmr.json",
+              "bucket": "cumulus-internal",
+              "filename": "s3://cumulus-internal/file-staging/subdir/MOD11A1.A2017200.h19v04.006.2017201090723.cmr.json",
+              "type": "metadata",
+              "fileStagingDir": "file-staging/subdir"
+            }
+          ]
+        },
+        {
```
markdboyd, line 71: I'm not opposed, but what's the rationale for having three granules in the test fixture as opposed to two? Two is the sufficient minimum case to prove that multiple granules get updated.

tasks/update-granules-metadata/tests/test-index.js
```diff
@@ -0,0 +1,135 @@
...
+  delete newPayload.config.distribution_endpoint;
+
+  const filesToUpload = cloneDeep(t.context.filesToUpload);
+  await uploadFiles(filesToUpload, t.context.stagingBucket);
+
+  const error = await t.throwsAsync(
+    () => updateGranulesMetadata(newPayload)
+  );
+
+  t.is(error.message, 'cmrGranuleUrlType is distribution, but no distribution endpoint is configured.');
```
markdboyd, line 134: I know that this is how the test is already written, but it can be refactored to now that we're on a later version of `ava`:

```
  const error = await t.throwsAsync(
    () => updateGranulesMetadata(newPayload),
    { message: 'cmrGranuleUrlType is distribution, but no distribution endpoint is configured.' } 
  );
```

tasks/update-granules-metadata/tests/test-index.js
```diff
@@ -0,0 +1,135 @@
...
+    .forEach(({ etag = '' }) => t.regex(etag, /"\S+"/)));
+});
+
+test.serial('Should update existing etag on CMR metadata file', async (t) => {
+  const newPayload = buildPayload(t);
+  const filesToUpload = cloneDeep(t.context.filesToUpload);
+  await uploadFiles(filesToUpload, t.context.stagingBucket);
+
+  const output = await updateGranulesMetadata(newPayload);
+  const granule = output.granules[output.granules.findIndex((g) => g.granuleId === 'MOD11A1.A2017200.h19v04.006.2017201090722')];
```
markdboyd, line 118: Maybe this is personal preference, but even though we're working with a fixture and have a predictable granuleId, I'd prefer that this test find the granules to test on by looking for which ones have an etag. It de-couples the test from knowledge of the fixture.

tasks/update-granules-metadata/tests/test-index.js
```diff
@@ -0,0 +1,135 @@
...
+  const fileWithPrevEtag = granule.files.filter(isCMRFile);
+  t.true(fileWithPrevEtag !== '13f2bb38e22496fe9d42e761c42a0e67');
```
markdboyd, line 120: I could be misreading this, but it looks like we're comparing here whether a `file` object found in `granule.files` is not equal to a specific Etag string, so this test will always pass even if the etag isn't being updated.

tasks/update-granules-metadata/tests/data/payload.json
markdboyd: I see. I think you added this just for your etag tests? That's fine in that case

botanical: Yeah, I added the third granule for my etag test

tasks/update-granules-metadata/tests/test-index.js
markdboyd: I think the logic that I would prefer to see for this test is: iterate over granules in input (`newPayload`) and if those granules have files with `etags`, then find the corresponding granule and file in `output` and compare whether those two etags are different

example/cumulus-tf/ingest_and_publish_granule_workflow.asl.json
botanical: I removed them from both `task_config` in this file and `schemas/config.json` and still ran into the error, but I realized it's because I didn't remove `bucket` and `collections` from the `required` list in `config.json`... oops 😬

tasks/update-granules-metadata/tests/test-index.js
```diff
@@ -0,0 +1,136 @@
...
+  const newPayload = buildPayload(t);
+  const filesToUpload = cloneDeep(t.context.filesToUpload);
+  await uploadFiles(filesToUpload, t.context.stagingBucket);
+  const inputGranule = newPayload.input.granules[newPayload.input.granules.findIndex((g) => g.granuleId === 'MOD11A1.A2017200.h19v04.006.2017201090722')];
+  const previousEtag = inputGranule.files.filter(isCMRFile)[0].etag;
+
+  const output = await updateGranulesMetadata(newPayload);
+  const updatedGranule = output.granules[output.granules.findIndex((g) => g.granuleId === 'MOD11A1.A2017200.h19v04.006.2017201090722')];
+  const newEtag = updatedGranule.files.filter(isCMRFile)[0].etag;
+  t.true(newEtag !== previousEtag);
```
markdboyd, line 122: Let's use the `t.not` assertion here: https://github.com/avajs/ava/blob/master/docs/03-assertions.md#notvalue-expected-message

tasks/update-granules-metadata/tests/test-index.js
```diff
@@ -0,0 +1,136 @@
...
+  output.granules.forEach((g) => g.files
+    .filter(isCMRFile)
+    .forEach(({ etag = '' }) => t.regex(etag, /"\S+"/)));
+});
+
+test.serial('Should update existing etag on CMR metadata file', async (t) => {
+  const newPayload = buildPayload(t);
+  const filesToUpload = cloneDeep(t.context.filesToUpload);
+  await uploadFiles(filesToUpload, t.context.stagingBucket);
+  const inputGranule = newPayload.input.granules[newPayload.input.granules.findIndex((g) => g.granuleId === 'MOD11A1.A2017200.h19v04.006.2017201090722')];
```
markdboyd, line 116: So the test is still depending on knowing the granuleId from the fixture. I think it could be refactored like this to avoid that dependency:

```
const inputGranulesWithEtags = newPayload.input.granules.reduce(
  (acc, granule) => {
    const cmrFile = granule.files.find(isCMRFile);
    if (!cmrFile) return acc;
    acc.push([granule.granuleId, cmrFile]);
  },
  []
);

inputGranulesWithEtags.forEach(([inputGranuleId, inputCmrFile]) => {
  const outputGranule = output.granules.find((outputGranule) => outputGranule.granuleId === inputGranuleId);
  if (outputGranule) {
    const outputCmrFile = outputGranule.files.find(isCMRFile);
    t.not(inputCmrFile.etag, outputCmrFile.etag)
  }
})
```

What do you think? Does that seem preferable or like unnecessary optimization? Also I didn't test the above code so it could need some tweaking

example/cumulus-tf/ingest_and_publish_granule_workflow.asl.json
markdboyd: Ah, makes sense. Glad you figured it out!

tasks/update-granules-metadata/tests/test-index.js
botanical: Hm, I did it a little differently. https://github.com/nasa/cumulus/pull/1969/files#diff-41cb5ec44ef17d2ccd2ba9142b73b29aee1d0c497c17766f8877a8d58648aa57R112-R125

Let me know what you think!

tasks/update-granules-metadata/tests/test-index.js
```diff
@@ -0,0 +1,136 @@
...
+    .filter(isCMRFile)
+    .forEach(({ etag = '' }) => t.regex(etag, /"\S+"/)));
+});
+
+test.serial('Should update existing etag on CMR metadata file', async (t) => {
+  const newPayload = buildPayload(t);
+  const filesToUpload = cloneDeep(t.context.filesToUpload);
+  const inputGranules = newPayload.input.granules;
+  const granuleWithEtag = inputGranules.find((g) => g.files.some((f) => f.etag));
+  const previousEtag = granuleWithEtag.files[0].etag;
```
markdboyd, line 116: My only issue is that we're assuming that the first file is the one that has the etag, which is true for the fixture but could change. I think adding the `.filter(isCMRFile)` like you have on line 121 would solve that nicely 💯

markdboyd: Also this solution is much more readable than mine, so 💯

botanical: Good point! I'll update that

markdboyd (approved):
This is approved from my perspective, but want to get another team member's input before merging

nemreid (changes requested):
This is a good PR! 💯 I only have some nitpicks, and most importantly a request for more and detailed changelog entries.

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     throwing consistent errors under load.
 
 ### Added
+- **CUMULUS-2138**
+  - Adds a new task `update-granules-metadata` to Ingest and Publish Granule workflow. This new task updates CMR Metadata and adds CMR file etags. It decouples said functionality from the `move-granules` task.
```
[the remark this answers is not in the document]
nemreid: It occurs to me that the files are what's being updated, with URLs, and there is no pushing of CMR metadata to CMR as in PostToCmr, so something like `update-granules-cmr-metadata-file-urls` or `-links`  would be even more specific.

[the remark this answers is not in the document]
nemreid: Naming is hard. 😅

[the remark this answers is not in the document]
botanical: I definitely agree with everything you said. I will consider what we can name this task. I will also add a note under `BREAKING CHANGES` as you have pointed out. Thanks for your feedback!

tasks/update-granules-metadata/tests/data/ummg-meta.json
```diff
@@ -0,0 +1,61 @@
...
+    },
+    "TemporalExtent": {
+      "RangeDateTime": {
+        "BeginningDateTime": "2016-01-09T11:40:45.032Z",
+        "EndingDateTime": "2016-01-09T11:41:12.027Z"
+      }
+    },
+    "GranuleUR": "L0A_RAD_RAW_product_0017-of-0020"
+  }
+
```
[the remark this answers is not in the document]
botanical: This has a newline, doesn't it? 😮

CHANGELOG.md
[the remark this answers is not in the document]
markdboyd: We should also look in the docs and see if we need to update any docs where `MoveGranules`/`move-granules` is mentioned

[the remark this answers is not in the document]
botanical: I've gone with `update-granules-cmr-metadata-file-links` though I do need to acknowledge it is a pretty long name. I find that it most accurately describes what the task does, so I am torn. I think, in general, it's better to name it exactly to describe what it does so I've opted for a long task name despite my hesitations.

CHANGELOG.md
```diff
@@ -33,7 +32,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
   - Fixes ingestPdrWithNodeNameSpec parsePdr provider error
 
 ### BREAKING CHANGES
-
+- **CUMULUS-2138**
```
Jkovarik, line 35: Awesome!

packages/api/lib/FileUtils.js
```diff
@@ -72,6 +72,7 @@ const setS3FileSize = async (s3, file) => {
 
     return { ...file, size };
   } catch (error) {
+    console.log(`File is ${file}, Error is ${error}`);
```
Jkovarik, line 75: Was this intentionally left in?   If so, should we use the logger package, and does it need to be JSON.stringified to be useful given how nested Core file objects are?

tasks/update-granules-cmr-metadata-file-links/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "directories": {
+    "test": "tests"
+  },
+  "homepage": "https://github.com/nasa/cumulus/tree/master/tasks/update-granules-cmr-metadata-file-links",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/nasa/cumulus"
+  },
+  "publishConfig": {
+    "access": "public"
```
Jkovarik, line 15: I'm not sure we want to publish new tasks, given they're being pulled into TF via our release artifacts.   @nemreid  @markdboyd ?

tf-modules/ingest/update-granules-cmr-metadata-file-links-task.tf
```diff
@@ -0,0 +1,33 @@
+resource "aws_lambda_function" "update_granules_cmr_metadata_file_links_task" {
+  function_name    = "${var.prefix}-UpdateGranulesCmrMetadataFileLinks"
+  filename         = "${path.module}/../../tasks/update-granules-cmr-metadata-file-links/dist/lambda.zip"
+  source_code_hash = filebase64sha256("${path.module}/../../tasks/update-granules-cmr-metadata-file-links/dist/lambda.zip")
+  handler          = "index.handler"
+  role             = var.lambda_processing_role_arn
+  runtime          = "nodejs12.x"
+  timeout          = 300
+  memory_size      = 1024
```
Jkovarik, line 9: We should probably consider tuning/eval of the timeout here, depending on what load/timing projections looks like for a reasonable amount of granules.   

 Definitely shouldn't block the PR, but move granules had the same setup and was S3 timing bound, system performance/etc will benefit from reducing the timeout for failure cases if we're well above what users would reasonably want.

Jkovarik (commented):
This looks great - as it's already approved, I'll defer to the original reviewers, just wanted to note I left a comment for @nemreid  @markdboyd  to review.

tasks/update-granules-cmr-metadata-file-links/package.json
markdboyd: I think I would defer to publishing until we make a judgment call / change about all the tasks as a whole

packages/api/lib/FileUtils.js
markdboyd: When we were debugging failing integration tests for Jennifer, it was impossible to see what was really failing with logging some sort of error here. But 👍 to using the logger package

tf-modules/ingest/update-granules-cmr-metadata-file-links-task.tf
botanical: Good point! If I understand what you mean by `S3 timing bound`, this task similarly relies on S3 to retrieve an object. That would lead me to think that the timeout for this task is okay, but I am open to reducing the timeout for failure cases like you mentioned. @Jkovarik

packages/api/lib/FileUtils.js
botanical: Using logger package now - thanks for the tip

docs/data-cookbooks/browse-generation.md
```diff
@@ -534,7 +534,7 @@ The processing lambda you construct will need to do the following:
 
 If you do not already have a CMR file in the granules list, you will need to generate one for valid export. This example's processing script generates and adds it to the `FilesToGranules` file list via the payload but it can be present in the InputGranules from the DiscoverGranules task as well if you'd prefer to pre-generate it.
 
-Both downstream tasks `MoveGranules` and `PostToCmr` expect a valid CMR file to be available if you want to export to CMR.
+Both downstream tasks `MoveGranules`, `UpdateGranulesCmrMetadataFileLinks`, and `PostToCmr` expect a valid CMR file to be available if you want to export to CMR.
```
nemreid, line 537:
```suggestion
Both downstream tasks `UpdateGranulesCmrMetadataFileLinks`, and `PostToCmr` expect a valid CMR file to be available if you want to export to CMR.
```

Correct me if I'm wrong, but MoveGranules no longer does anything with CMR files with this PR?

tasks/update-granules-cmr-metadata-file-links/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@cumulus/update-granules-cmr-metadata-file-links",
+  "version": "2.0.1",
```
nemreid, line 3: guessing this version number needs updating now?

tasks/update-granules-cmr-metadata-file-links/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "serial": true,
+    "verbose": true,
+    "timeout": "15m"
+  },
+  "author": "Cumulus Authors",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@cumulus/aws-client": "2.0.1",
+    "@cumulus/cmrjs": "2.0.1",
+    "@cumulus/common": "2.0.1",
```
nemreid, line 41: Also version number updates. I'd expect bootstrap to break when current with master otherwise.

tasks/update-granules-cmr-metadata-file-links/tests/data/ummg-meta.json
```diff
@@ -0,0 +1,61 @@
...
+    },
+    "TemporalExtent": {
+      "RangeDateTime": {
+        "BeginningDateTime": "2016-01-09T11:40:45.032Z",
+        "EndingDateTime": "2016-01-09T11:41:12.027Z"
+      }
+    },
+    "GranuleUR": "L0A_RAD_RAW_product_0017-of-0020"
+  }
+
```
nemreid, line 61: missing newline

nemreid (commented):
Appreciate you addressing my previous comments! I think the version number thing is the only thing really of note here since master has a different major version now.

docs/data-cookbooks/browse-generation.md
botanical: Hm, I don't think that is entirely true. MoveGranules still deals with CMR files in `updateGranulesMetadata()` https://github.com/nasa/cumulus/blob/master/tasks/move-granules/index.js#L79

tasks/update-granules-cmr-metadata-file-links/tests/data/ummg-meta.json
botanical: Doesn't it already have a newline? Or does it need another? I will just add another but it seems to have one already.

tasks/update-granules-cmr-metadata-file-links/package.json
botanical: yeah, good call out - I've updated

tasks/update-granules-cmr-metadata-file-links/package.json
Jkovarik: I'm 'ok' with that, but let's get it on the architecture meeting for a discussion.

tf-modules/ingest/update-granules-cmr-metadata-file-links-task.tf
Jkovarik: Sorry - what I was (poorly) getting at was generally we probably want timeouts to be trimmed as much as reasonably possible with padding so that config errors/bugs/etc that might cause a timeout don't eat invocation charges at scale -- given what the lambda is doing compared with what move granules was doing, I was wondering if in terms of "wildest reasonable expectations" if 5 minutes makes sense.

If you feel this is reasonable given task testing, no objection at all.  

Again, not a PR stopper at all, it's super minor given the scaling/budgets probably involved.

docs/data-cookbooks/browse-generation.md
nemreid: OK, then my nitpick is the the use of 'Both' for 3 tasks is slightly odd use of language, but not really a blocking issue.

tasks/update-granules-cmr-metadata-file-links/tests/data/ummg-meta.json
nemreid: Screen Shot 2020-11-05 at 9 38 24 AM

I'm seeing Github report a missing newline, and I think it might actually be a JSON formatting issue. Try removing the whitespace before the terminating bracket on the last line?

nemreid (approved):

botanical:
Going to close because commits have already been merged to branch `merge-cumulus-2138-update-granules-cmr-task`

PR #1970, opened 2020-10-26
CUMULUS-2201-4: Reconciliation Report integration test failures

jennyhliu:

Addresses [CUMULUS-2201: CUMULUS-2201-3: Reconciliation Report integration test failures](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2201)

## Changes

* Update `dbIndexer` lambda to process requests in serial
* Fixes ingestPdrWithNodeNameSpec parsePdr provider error (no provider found or multiple providers with same host found)

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [x] Integration tests


example/deployments/cumulus/sandbox.tfvars
```diff
@@ -46,4 +46,6 @@ ems_retention_in_days = 30
 ems_submit_report     = true
 ems_username          = "cumulus"
 
+es_request_concurrency = 1
```
markdboyd, line 49: So the whole point of this is to make the dbIndexer run its requests in serial, rather than parallel, by setting the concurrency to 1? If so, I'd rather us just explicitly remove the parallel processing via Promise.all from dbIndexer. Basically, if it's the concurrency causing an issue that our tests are revealing, I don't think it should even be possible to enable concurrency in that lambda at least.

botanical:
hi - just wanted to jump in with a question I have! If this test can't be run in parallel, should we move it to the `serial` directory?

jennyhliu:
> hi - just wanted to jump in with a question I have! If this test can't be run in parallel, should we move it to the `serial` directory?
The concurrency in dbindexer caused the records saved to the ElasticSearch not in the order they came in, so we have to remove the concurrency.  The Reconcilation Report test itself can run in parallel with other tests.

example/deployments/cumulus/sandbox.tfvars
jennyhliu: Ok.  I will remove the concurrency code from dbIndexer.  We can support concurrency for indexing records from different types. But that would be separate ticket.
15a9dca[https://github.com/nasa/cumulus/pull/1970/commits/15a9dca3b889a14c80b2a2ee7deceb225c589114]

packages/api/lambdas/db-indexer.js
```diff
@@ -199,11 +199,7 @@ async function indexRecords(records) {
   return pMap(
     records,
     (record) => indexRecord(esClient, record).catch(log.error),
-    {
-      concurrency: process.env.ES_CONCURRENCY
-        ? Number.parseInt(process.env.ES_CONCURRENCY, 10)
-        : 3,
-    }
+    { concurrency: 1 }
```
markdboyd, line 202: Sorry, I'm not trying to nitpick here, but can we just remove the use of `pMap` altogether? I worry that with using a library that supports concurrency, it will be easy for someone to accidentally re-enable it and re-introduce the bug.

Maybe use https://github.com/sindresorhus/p-each-series instead?

jennyhliu: yeah.  It's more clear.
[a26302c](https://github.com/nasa/cumulus/pull/1970/commits/a26302c059d2fbf46db3b6cf228576f9262b1032)

example/cumulus-tf/main.tf
```diff
@@ -89,6 +89,8 @@ module "cumulus" {
   ems_submit_report     = var.ems_submit_report
   ems_username          = var.ems_username
 
+  es_request_concurrency = var.es_request_concurrency
```
markdboyd, line 92: Is this variable still used in the cumulus module anywhere if it's not used for dbIndexer?

markdboyd (approved):

example/cumulus-tf/main.tf
markdboyd: Ah yes, I see that it is

PR #1987, opened 2020-11-06
CUMULUS-2228: Write granules to RDS

markdboyd:

Addresses [CUMULUS-2228: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2228)

## Changes

* Update sfEventSqsToDbRecords Lambda to write granules to Dynamo/RDS
  * Required refactor of granule model methods so that each granule could be handled in a separate transaction

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


packages/api/lambdas/sf-event-sqs-to-db-records.js
```diff
@@ -77,30 +83,39 @@ const hasNoAsyncOpOrExists = async (cumulusMessage, knex) => {
...
-    return await knex(tableNames.providers).where({
+    const searchParams = {
       name: getMessageProviderId(cumulusMessage),
-    }).first();
+    };
+    const provider = await knex(tableNames.providers).where(searchParams).first();
+    if (!isRecordDefined(provider)) {
+      throw new Error(`Could not find provider with params ${JSON.stringify(searchParams)}`);
+    }
+    return provider.cumulusId;
```
markdboyd, line 118: Since all I really want from these helpers is the cumulusId of the objects, if they exist, this refactor seemed cleaner to me

packages/api/models/granules.js
```diff
@@ -576,60 +579,43 @@ class Granule extends Manager {
...
+      updateParams.ConditionExpression = '#execution <> :execution';
+    }
 
-    let executionDescription;
     try {
-      executionDescription = await StepFunctions.describeExecution({ executionArn });
+      return await this.dynamodbDocClient.update(updateParams).promise();
     } catch (error) {
-      log.error(`Could not describe execution ${executionArn}`, error);
+      if (isConditionalCheckException(error)) {
```
markdboyd, line 604: I thought this made sense when I wrote it, but looking closer, the previous code in `_validateAndStoreGranuleRecord` treated all errors the same and caught them, where they were a conditional check exception or not. So these conditional check exceptions should probably be thrown, since this method is used in the RDS transaction to write to Dynamo and we want to fail both writes if either write fails.

packages/api/lambdas/sf-event-sqs-to-db-records.js
```diff
@@ -135,67 +150,136 @@ const writeExecutionViaTransaction = async ({ cumulusMessage, trx }) =>
...
+}) => {
+  // If there are no granules in the message, then there's nothing to do here, which is fine
+  if (!messageHasGranules(cumulusMessage)) {
+    return true;
   }
+  if (!collectionCumulusId) {
+    throw new Error('Collection reference is required for granules');
+  }
+  // Process each granule in a separate transaction via Promise.allSettled
+  // so that they can succeed/fail indepdently
```
markdboyd, line 251: Spelling

packages/api/lambdas/sf-event-sqs-to-db-records.js
```diff
@@ -234,21 +318,29 @@ const writeRecords = async (cumulusMessage, knex) => {
...
     const pdrCumulusId = await writePdr({
       cumulusMessage,
-      collection,
-      provider,
+      collectionCumulusId,
+      providerCumulusId,
+      knex,
+      executionCumulusId,
+    });
+    return await writeGranules({
```
markdboyd, line 336: Re-evalute logic based on whether there is a PDR in the message. Granule is not dependent on PDR if there isn't one

packages/api/lambdas/sf-event-sqs-to-db-records.js
botanical: That makes sense to me

packages/api/lambdas/sf-event-sqs-to-db-records.js
```diff
@@ -135,67 +150,136 @@ const writeExecutionViaTransaction = async ({ cumulusMessage, trx }) =>
       arn: getMessageExecutionArn(cumulusMessage),
       cumulus_version: getMessageCumulusVersion(cumulusMessage),
       status: getMetaStatus(cumulusMessage),
-    });
+    })
+    .returning('cumulusId');
```
botanical, line 154: What is `returning`? I couldn't find any meaningful search results when I looked

packages/message/src/Granules.ts
```diff
@@ -8,18 +8,35 @@
...
+    granules?: object[]
+  }
+}
+
 /**
- * Get granules from execution message.
+ * Get granules from a workflow message.
  *
- * @param {Message.CumulusMessage} message - An execution message
+ * @param {MessageWithGranules} message - A workflow message
```
botanical, line 22: Are workflow and execution interchangeable?

packages/api/models/granules.js
```diff
@@ -576,60 +579,43 @@ class Granule extends Manager {
...
+      mutableFieldNames,
+    });
 
-    const executionArn = getMessageExecutionArn(cumulusMessage);
-    const executionUrl = StepFunctionUtils.getExecutionUrl(executionArn);
+    // Only allow "running" granule to replace completed/failed
+    // granule if the execution has changed
+    if (granuleRecord.status === 'running') {
+      updateParams.ConditionExpression = '#execution <> :execution';
+    }
```
botanical, line 599: What is a running granule?

packages/api/lambdas/sf-event-sqs-to-db-records.js
botanical: Do you mind elaborating what a PDR/ polling with delivery record mechanism is? Why was this re-written from `writePdr` to `writeGranules`?

packages/message/src/Granules.ts
markdboyd: Sort of. "Workflow" is the term we use generically in Cumulus for a Step Function composed of tasks. And an execution is an instance of running that workflow. I suppose that an "execution" message is a more accurate description

packages/api/models/granules.js
markdboyd: The sfEventSqsToDbRecords Lambda runs at the beginning and end of every Step Function, triggered by a Cloudwatch event. At the beginning of the workflow, its status is "running", so it saves the granules from the workflow as the database as being "running". This allows operators on the dashboard to see that the ingest of the granules is in process. Then when the workflow completes, another event fires to trigger the Lambda with a status of "completed", so the granules are re-saved as "completed".

packages/api/lambdas/sf-event-sqs-to-db-records.js
markdboyd: This comes from the Knex query builder. It allows you to return a field when you are inserting a record:

http://knexjs.org/#Builder-returning

packages/api/lambdas/sf-event-sqs-to-db-records.js
markdboyd: I think this is two separate questions. I can't fully elaborate on what a PDR is or how it works - but it's some kind of file format that is published by a provider that can be used to notify a DAAC that data is ready for ingest. So we have DiscoverPdrs workflows that discover PDRs on the provider, parse them, and ingest the referenced data. Is that correct @Jkovarik @nemreid ?

Also, this function wasn't rewritten from `writePdr` to `writeGranules`. `writePdr` still exists right above this function call. I did rename `saveGranulesToDb` to `writeGranules` to be consistent in the function naming with how the rest of the code was refactored

packages/api/models/granules.js
markdboyd: This document explains how data is "reported" from our workflows, though it only shows the data being saved to Dynamo:

https://nasa.github.io/cumulus/docs/data-cookbooks/ingest-notifications

packages/api/lambdas/sf-event-sqs-to-db-records.js
markdboyd: Also, I took another look at this and I think the logic here is fine. The key thing is that `writeGranules` is not attempted if `writePdr` throws an error, which would only happen if **there is a PDR on the message and writing it somehow fails**. If there is no PDR on the message, `writePdr` just returns `undefined` and `writeGranules` will continue as normal without a reference to a PDR.

packages/api/models/granules.js
markdboyd: Addressed here: bce1e9370a57e4bff67de02126e348a4e7adc6b9

packages/api/lambdas/sf-event-sqs-to-db-records.js
botanical: > Also, this function wasn't rewritten from writePdr to writeGranules. writePdr still exists right above this function call.

You're right, sorry I missed that. And yeah, the logic is solid

packages/message/src/Granules.ts
botanical: I see. Even though `execution` message is technically more accurate, I think I do prefer `workflow` as you have it because it ties back to the cumulus vocabulary and provides more context.

lambdas/db-migration/src/migrations/20201014105058_create_granules_table.ts
```diff
@@ -6,6 +6,14 @@ export const up = async (knex: Knex): Promise<void> =>
...
       .comment('Internal Cumulus ID for a granule')
       .primary();
+    table
+      .text('granuleId')
+      .comment('Granule ID')
+      .notNullable();
+    table
+      .enum('status', ['running', 'completed', 'failed'])
+      .comment('Ingest status of the granule')
+      .notNullable();
```
botanical, line 16: Is there not a reason to use snake case here and convert the rest of the table names to use snake case as well?

packages/api/models/granules.js
```diff
@@ -639,31 +607,26 @@ class Granule extends Manager {
...
-      });
-
-      // Only allow "running" granule to replace completed/failed
-      // granule if the execution has changed
-      if (granuleRecord.status === 'running') {
-        updateParams.ConditionExpression = '#execution <> :execution';
-      }
+    // TODO: Refactor this all to use model.update() to avoid having to manually call
+    // schema validation and the actual client.update() method.
+    await this.constructor.recordIsValid(granuleRecord, this.schema, this.removeAdditional);
```
Jkovarik, line 612: 👍  nice refactor

packages/api/models/granules.js
```diff
@@ -639,31 +607,26 @@ class Granule extends Manager {
...
+  }
 
-      await this.dynamodbDocClient.update(updateParams).promise();
-    } catch (error) {
-      log.error(
-        'Could not store granule record: ', granuleRecord,
-        error
-      );
-    }
+  async storeGranuleFromCumulusMessage({
```
Jkovarik, line 616: Docstring please!

packages/api/models/granules.js
```diff
@@ -673,9 +636,31 @@ class Granule extends Manager {
    * @returns {Promise}
    */
   async storeGranulesFromCumulusMessage(cumulusMessage) {
-    const granuleRecords = await this.constructor
-      ._getGranuleRecordsFromCumulusMessage(cumulusMessage);
-    return Promise.all(granuleRecords.map(this._validateAndStoreGranuleRecord, this));
+    const granules = getMessageGranules(cumulusMessage);
```
Jkovarik, line 639: See comments on getMessageGranules from sf-even-sqs-to-db-records

packages/api/models/granules.js
```diff
@@ -673,9 +636,31 @@ class Granule extends Manager {
...
+    const executionUrl = getExecutionUrlFromArn(executionArn);
+
+    let executionDescription;
+    try {
+      executionDescription = await StepFunctions.describeExecution({ executionArn });
+    } catch (error) {
+      log.error(`Could not describe execution ${executionArn}`, error);
+    }
+
+    return Promise.all(granules.map(
```
Jkovarik, line 655: Do we want this to abort on some of the granules if one granule fails validation in storeGranule?   

More importantly given this is async, if we *do* want that to occur, are we ok with a random set of them succeeding? 

Consider changing this to an allSettled with evaluation to throw if a failure occurred.

lambdas/db-migration/src/migrations/20201014105058_create_granules_table.ts
markdboyd: No, I plan on doing that once @nemreid's PR is merged, since he already renamed all of these fields

packages/message/src/Granules.ts
```diff
@@ -8,18 +8,35 @@
...
  * @alias module:Granules
  */
 export const getMessageGranules = (
-  message: Message.CumulusMessage
-): unknown[] | undefined => get(message, 'payload.granules');
+  message: MessageWithGranules
+): unknown[] | undefined => message.payload?.granules;
+
+/**
+ * Determine whether workflow message has granules.
```
Jkovarik, line 33: Nit: as this is a double lookup, can we add a comment here noting we're talking about payload?. granules?

Jkovarik (changes requested):
This looks solid, but there are a few comments.  I'm still looking at the PR, but wanted to add comments for your review sooner than later.

packages/api/tests/models/granules/test-granule-reporting.js
```diff
@@ -23,214 +23,303 @@ test.before(async (t) => {
...
-  try {
-    await granuleModel._validateAndStoreGranuleRecord(granule);
-    t.pass();
-  } catch (error) {
-    t.fail(`Expected error not to be thrown, caught: ${error}`);
-  }
+  await t.throwsAsync(granuleModel._validateAndStoreGranuleRecord(granule));
+});
+
+test('_validateAndStoreGranuleRecord() throws error for a conditional check exception', async (t) => {
```
Jkovarik, line 231: It's not clear from a casual read what this is actually doing - at first glance it looks like the test is checking that you can't update a failed record to running.   ALso if we want to be extra certain, we could check the error type. 

Something more like: 

```suggestion
test('_validateAndStoreGranuleRecord() throws an error if replacing granule without updated execution', async (t) => {
```

packages/api/lambdas/sf-event-sqs-to-db-records.js
```diff
@@ -77,30 +83,39 @@ const hasNoAsyncOpOrExists = async (cumulusMessage, knex) => {
   }, knex, tableNames.asyncOperations);
 };
 
-const getMessageCollection = async (cumulusMessage, knex) => {
+const getMessageCollectionCumulusId = async (cumulusMessage, knex) => {
```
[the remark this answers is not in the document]
markdboyd: Yeah, the `message` package is just helpers for parsing info out of a Cumulus message. I could put this in `db`, though I wasn't sure it would get used outside of this Lambda

packages/api/lambdas/sf-event-sqs-to-db-records.js
```diff
@@ -135,67 +150,136 @@ const writeExecutionViaTransaction = async ({ cumulusMessage, trx }) =>
...
+  collectionCumulusId,
+  providerCumulusId,
+  executionCumulusId,
+  pdrCumulusId,
+  trx,
+}) =>
+  trx(tableNames.granules)
+    .insert({
+      granuleId: granule.granuleId,
+      status: getMetaStatus(cumulusMessage) || granule.status,
```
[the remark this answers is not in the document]
markdboyd: Not particularly. I can add that.

packages/api/lambdas/sf-event-sqs-to-db-records.js
```diff
@@ -135,67 +150,136 @@ const writeExecutionViaTransaction = async ({ cumulusMessage, trx }) =>
...
+        await writeGranuleViaTransaction({
+          cumulusMessage,
+          granule,
+          collectionCumulusId,
+          providerCumulusId,
+          executionCumulusId,
+          pdrCumulusId,
+          trx,
+        });
+        return granuleModel.storeGranuleFromCumulusMessage({
```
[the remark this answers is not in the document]
markdboyd: It's hard to see from the diff, but `storeGranulesFromCumulusMessage` processed each granule in the message and stored it, but all errors for each granule were caught in `_validateAndStoreGranuleRecord`, so each granule succeeded/failed independently.

https://github.com/nasa/cumulus/blob/master/packages/api/models/granules.js#L708

Note that this behavior for writing granules to just Dynamo has been preserved.  See line 209 in this file

lambdas/db-migration/src/migrations/20201014105058_create_granules_table.ts
```diff
@@ -38,19 +42,17 @@ export const up = async (knex: Knex): Promise<void> =>
...
-      .comment('Granule ID')
-      .notNullable();
+      .integer('executionCumulusId')
+      .references('cumulusId')
+      .inTable('executions');
     table
-      .text('pdrName')
-      .comment('PDR associated with the granule');
+      .integer('pdrCumulusId')
+      .references('cumulusId')
```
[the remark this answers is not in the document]
markdboyd: Yeah, I think that was the plan

packages/api/tests/models/granules/test-granule-reporting.js
markdboyd: 8fa4866781a2269eaf70379c0d59cd30822e196d

packages/api/lambdas/sf-event-sqs-to-db-records.js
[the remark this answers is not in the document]
Jkovarik: I'm fine with YAGNI as a judgement call here, the concern really is maintainability situations where we don't realize this is here for whatever reason and re-implementing it.

packages/api/lambdas/sf-event-sqs-to-db-records.js
```diff
@@ -135,67 +150,136 @@ const writeExecutionViaTransaction = async ({ cumulusMessage, trx }) =>
...
+  cumulusMessage,
+  collectionCumulusId,
+  providerCumulusId,
+  executionCumulusId,
+  pdrCumulusId,
+  knex,
+  granuleModel = new Granule(),
+}) => {
+  // If there are no granules in the message, then there's nothing to do here, which is fine
+  if (!messageHasGranules(cumulusMessage)) {
```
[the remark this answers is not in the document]
markdboyd: Yeah that's very worth doing. Good call

packages/api/lambdas/sf-event-sqs-to-db-records.js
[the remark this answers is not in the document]
markdboyd: 5cfe3af6c3ade178abd2db816b366fe262446e94

packages/api/lambdas/sf-event-sqs-to-db-records.js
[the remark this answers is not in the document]
markdboyd: 6eca6687413705144fef9969af5b627df8722700
c7f09a9e3a1d583ff3b54aebdd90cbc197ec07c7
0fdc08fbe7c304be001b7374a8a24b5fe445eaab

packages/api/tests/models/granules/test-granule-reporting.js
```diff
@@ -23,214 +23,303 @@ test.before(async (t) => {
...
+
+  const updatedGranule = {
+    ...granule,
+    status: 'running',
+  };
+
+  await t.throwsAsync(granuleModel._validateAndStoreGranuleRecord(updatedGranule));
+});
+
+test('storeGranuleFromCumulusMessage() does throw an error for a failing record', async (t) => {
```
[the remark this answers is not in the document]
markdboyd: c253d6570ccd94942cfb6bb96dae59f5caa850b6

packages/message/src/Granules.ts
markdboyd: `messageHasGranules` was removed, but added to `getMessageGranules`:

3f60ec0d700e653789b08cd3be3c913c2e287d90

packages/api/models/granules.js
markdboyd: 0fdc08fbe7c304be001b7374a8a24b5fe445eaab

packages/api/models/granules.js
markdboyd: 96e39d78b372af42bb4558a7f913b26019218183

packages/api/lambdas/sf-event-sqs-to-db-records.js
```diff
@@ -135,67 +150,136 @@ const writeExecutionViaTransaction = async ({ cumulusMessage, trx }) =>
...
+      pdrCumulusId,
+    });
 
-  try {
-    await granuleModel.storeGranulesFromCumulusMessage(cumulusMessage);
-  } catch (error) {
-    const executionArn = getMessageExecutionArn(cumulusMessage);
-    log.fatal(`Failed to create/update granule records for execution ${executionArn}: ${error.message}`);
-    throw error;
+const writeGranules = async ({
```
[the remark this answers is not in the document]
markdboyd: d5dafe6d7d45b75d9cf34ec7f5e4c573a80d5f83

packages/api/lambdas/sf-event-sqs-to-db-records.js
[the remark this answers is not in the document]
markdboyd: `messageHasGranules` was removed, but `getMessageGranules` was updated

3f60ec0d700e653789b08cd3be3c913c2e287d90

Jkovarik:
Also - there's not a CHANGELOG update, is one warranted given prior PRs on this work?

lambdas/db-migration/src/migrations/20201014105058_create_granules_table.ts
```diff
@@ -6,6 +6,14 @@ export const up = async (knex: Knex): Promise<void> =>
...
+      .text('granuleId')
+      .comment('Granule ID')
+      .notNullable();
+    table
+      .enum('status', ['running', 'completed', 'failed'])
+      .comment('Ingest status of the granule')
+      .notNullable();
     table
       .integer('collectionCumulusId')
       .references('cumulusId')
```
botanical, line 19: I know this isn't really a part of your PR but I wanted to get some clarification. Is this setting `cumulusId` as a foreign key?

packages/api/lambdas/sf-event-sqs-to-db-records.js
```diff
@@ -135,67 +149,200 @@ const writeExecutionViaTransaction = async ({ cumulusMessage, trx }) =>
...
+ * @param {string} params.collectionCumulusId
+ *   Cumulus ID for collection referenced in workflow message, if any
+ * @param {string} params.executionCumulusId
+ *   Cumulus ID for execution referenced in workflow message, if any
+ * @param {Knex} params.knex - Client to interact with Postgres database
+ * @param {string} [params.providerCumulusId]
+ *   Cumulus ID for provider referenced in workflow message, if any
+ * @param {string} [params.pdrCumulusId]
+ *   Cumulus ID for PDR referenced in workflow message, if any
+ * @param {Object} [params.granuleModel]
```
botanical, line 248: Why did you put `[]` around these?
```
[params.providerCumulusId]
[params.pdrCumulusId]
[params.granuleModel]
```

lambdas/db-migration/src/migrations/20201014105058_create_granules_table.ts
botanical: Also, where would this `collectionCumulusId` value come from if `Collection` is a name and version?

Jkovarik: collectionCumlusId is a foriegn key referring to the cumulusId on the Collections table - it's an (incrementing) ID that's generated on insertion and is seperate from the name/version identifier.

packages/api/lambdas/sf-event-sqs-to-db-records.js
markdboyd: Because they're optional parameters: https://jsdoc.app/tags-param.html#optional-parameters-and-default-values

lambdas/db-migration/src/migrations/20201014105058_create_granules_table.ts
markdboyd: > where would this collectionCumulusId value come from if Collection is a name and version?

That's where `getMessageCollectionCumulusId` comes in. It looks up the collection name/version from a workflow message in RDS, returning its `cumulusId`. Then, we use that `cumulusId` to set `collectionCumulusId` on granules, PDRs, and executions.

https://github.com/nasa/cumulus/pull/1987/files#diff-bc96e8e9a9c08e28b506ef0387539a0fc75bb657b28dd974c4419818a5e72607R85

packages/api/tests/lambdas/test-sf-event-sqs-to-db-records.js
```diff
@@ -876,7 +991,7 @@ test.serial('Lambda sends message to DLQ when RDS_DEPLOYMENT_CUMULUS_VERSION env
   t.is(handlerResponse[0][1].body, sqsEvent.Records[0].body);
 });
 
-test('Lambda sends message to DLQ when any write to database fails', async (t) => {
+test('Lambda sends message to DLQ when an execution/PDR write to the database fails', async (t) => {
```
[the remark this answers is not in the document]
markdboyd: Going to push this off to an architecture discussion and will file a separate ticket/follow-up as necessary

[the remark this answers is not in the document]
markdboyd: 7ec6c044bfc27cb3bbb4df4bec784b9058527874 to address the initial concern of writing to DLQ for any failing granule write

The other concerns raised in the arch meeting will be handled in separate tickets

packages/api/lambdas/sf-event-sqs-to-db-records.js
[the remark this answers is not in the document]
Jkovarik: Nice 👍

Jkovarik (approved):
I took a look at the updates/comment responses and this all looks good.  Great work!

PR #1916, opened 2020-10-06
CI speed improvement: cache compiled TS code

markdboyd:
Successful build with cached bootstrap (only failed on integration tests) in **23 minutes**: https://ci.earthdata.nasa.gov/browse/CUM-CCRR0-46
Successful build without cached bootstrap (only failed on integration tests) in **29 minutes**: https://ci.earthdata.nasa.gov/browse/CUM-CCRR0-50

PR #1988, opened 2020-11-09
Fix issues with bootstrapping & TSC compilation

markdboyd:
Based on how the bootstrapping scripts were refactored in #1916, a bug was introduced where TSC compilation may fail locally because it runs in a non-deterministic order, so TS-compiled packages may not have been compiled before the packages that need them. Moving the TS compilation back to run as part of a `prepare` script which is executed by Lerna as part of bootstrapping, and which respects package dependency order, fixes the issue.


Jkovarik (approved):
Approved - we'll need to issue a separate PR re: updated packages in the feature branch as well.

PR #1989, opened 2020-11-09
[CUMULUS-2251] distribution module bug

botanical:
This PR solves an error users were running into:
```
The "count" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the
resources that the count depends on.
```
Basically, at the time terraform reaches
` count           = var.tea_api_egress_log_group != null ? 1 : 0`
the value here hasn't been allocated yet
`tea_api_egress_log_group = module.thin_egress_app.egress_log_group`
so we run into the count error we've seen. So the solution is to change count to depend on a variable it will know when we run terraform plan. The fix uses `log_api_gatway_to_cloudwatch` to set count.

Addresses [CUMULUS-2251: Fix issue with distribution module code](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2251)

## Changes

* Move `egress_api_gateway_log_subscription_filter` to `cumulus/main.tf`
* Remove unneeded variable definitions

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [ ] Integration tests


example/cumulus-tf/main.tf
```diff
@@ -216,6 +215,15 @@ module "thin_egress_app" {
   vpc_subnet_ids             = var.lambda_subnet_ids
 }
 
+resource "aws_cloudwatch_log_subscription_filter" "egress_api_gateway_log_subscription_filter" {
+  count           = (var.log_api_gateway_to_cloudwatch != null && var.log_destination_arn != null) ? 1 : 0
```
botanical, line 219: I modified count to include a check for `log_destination_arn` but I am not sure if there is a more elegant way to have this resource check on its dependent variables  (`log_api_gateway_to_cloudwatch` and `log_destination_arn`).

markdboyd: No, I think this is fine

markdboyd: However, since the default for `log_api_gateway_to_cloudwatch` is `false` and it has a type of `boolean`, the first check can just be`var.log_api_gateway_to_cloudwatch && ...`

example/cumulus-tf/main.tf
```diff
@@ -216,6 +215,15 @@ module "thin_egress_app" {
   vpc_subnet_ids             = var.lambda_subnet_ids
 }
 
+resource "aws_cloudwatch_log_subscription_filter" "egress_api_gateway_log_subscription_filter" {
```
markdboyd, line 218: This isn't required at all, but just a thought I had, what do you think about moving all the TEA related resources to a separate `tea.tf` file? It might be better organized than cluttering up this file.

This would include all the resources defined from line 173 to 226 in this file

CHANGELOG.md
```diff
@@ -31,6 +31,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - **CUMULUS-2201**
   - Update `dbIndexer` lambda to process requests in serial
   - Fixes ingestPdrWithNodeNameSpec parsePdr provider error
+- **CUMULUS-2251**
+  - Move Egress Api Gateway Log Group Filter from `distribution/main.tf` to `/cumulus-tf/main.tf` and remove `tea_api_egress_log_group` variable from `distribution/variables.tf` and `cumulus/variables.tf` and add `log_api_gateway_to_cloudwatch` variable to `cumulus-tf/variables.tf`
```
markdboyd, line 35: It would be good to split these notes about what was changed, removed, and added into separate notes under the different headers for Changed, Added, etc.

CHANGELOG.md
```diff
@@ -31,6 +31,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
+  - This fixes the count error users ran into when they had `log_api_gateway_to_cloudwatch` set to true in their `thin_egress_app` module.
```
markdboyd, line 36: We also usually add a separate header for "Fixed" that you could put this under. I would also maybe adjust the language to "fixed a deployment error caused by depending on a module output for a resource count" since "count error" is a bit ambiguous without more context

CHANGELOG.md
markdboyd: We also need to add a migration note here - since people will be expected to add the new resource definition to their cumulus-tf/main.tf file. Also, we need to update the [template-deploy repo](https://github.com/nasa/cumulus-template-deploy/pulls) that we maintain as a starter/example for users with these changes.

markdboyd (changes requested):
Looks great overall, just a few requested changes

example/cumulus-tf/main.tf
botanical: For posterity, we discussed on slack that it is probably better to keep resources and module in `main.tf`

botanical:
Also updated `template-deploy-repo` https://github.com/nasa/cumulus-template-deploy/pull/54

CHANGELOG.md
```diff
@@ -31,6 +35,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - **CUMULUS-2201**
   - Update `dbIndexer` lambda to process requests in serial
   - Fixes ingestPdrWithNodeNameSpec parsePdr provider error
+- **CUMULUS-2251**
+  - Moves Egress Api Gateway Log Group Filter from `distribution/main.tf` to `/cumulus-tf/main.tf` and remove `tea_api_egress_log_group` variable from `distribution/variables.tf` and `cumulus/variables.tf`.
```
markdboyd, line 39: Nitpicking for sure, but let's move the "removed.." statements under a "Removed" header

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Migration notes
 - Update the name of your `cumulus_message_adapter_lambda_layer_arn` variable for the `cumulus` module to `cumulus_message_adapter_lambda_layer_version_arn`. The value of the variable should remain the same (a layer version ARN of a Lambda layer for the [`cumulus-message-adapter`](https://github.com/nasa/cumulus-message-adapter/).
 - **CUMULUS-2138** - Update all workflows using the `MoveGranules` step to add `UpdateGranulesCmrMetadataFileLinksStep`that runs after it. See the example [`IngestAndPublishWorkflow`](https://github.com/nasa/cumulus/blob/master/example/cumulus-tf/ingest_and_publish_granule_workflow.asl.json) for reference.
+- **CUMULUS-2251**
+  - Add new resource defintion `egress_api_gateway_log_subscription_filter` to `cumulus-tf/main.tf`.
```
markdboyd, line 13:
```suggestion
  - Because it has been removed from the `cumulus` module, a new resource definition for `egress_api_gateway_log_subscription_filter` must be added to `cumulus-tf/main.tf`. For reference on how to define this resource, see [`example/cumulus-tf/main.tf`](https://github.com/nasa/cumulus/blob/master/example/cumulus-tf/main.tf) .
```

CHANGELOG.md
```diff
@@ -31,6 +35,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - **CUMULUS-2201**
   - Update `dbIndexer` lambda to process requests in serial
   - Fixes ingestPdrWithNodeNameSpec parsePdr provider error
+- **CUMULUS-2251**
+  - Moves Egress Api Gateway Log Group Filter from `distribution/main.tf` to `/cumulus-tf/main.tf`
```
markdboyd, line 39:
```suggestion
  - Moves Egress Api Gateway Log Group Filter from `tf-modules/distribution/main.tf` to `example/cumulus-tf/main.tf`
```

CHANGELOG.md
```diff
@@ -22,6 +24,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
   - Adds ability to filter granule inventory reports
 - **CUMULUS-2211**
   - Adds `granules/bulkReingest` endpoint to `@cumulus/api`
+- **CUMULUS-2251**
+  - Adds `log_api_gateway_to_cloudwatch` variable to `cumulus-tf/variables.tf`
```
markdboyd, line 28:
```suggestion
  - Adds `log_api_gateway_to_cloudwatch` variable to `example/cumulus-tf/variables.tf`
```

CHANGELOG.md
```diff
@@ -31,6 +35,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
+- **CUMULUS-2251**
+  - Moves Egress Api Gateway Log Group Filter from `distribution/main.tf` to `/cumulus-tf/main.tf`
+
+### Fixed
+- **CUMULUS-2251**
+  - This fixes a deployment error caused by depending on the `thin_egress_app` module output for a resource count.
+
+### Removed
+- **CUMULUS-2251**
+  -  Removes `tea_api_egress_log_group` variable from `distribution/variables.tf` and `cumulus/variables.tf`.
```
```suggestion
  -  Removes `tea_api_egress_log_group` variable from `tf-modules/distribution/variables.tf` and `tf-modules/cumulus/variables.tf`.
```

markdboyd (approved):
Approved with a few minor nits - I like to reference the full path to files from the repo root in the CHANGELOG, but it's a personal preference. Great work 💯 🚀 !

botanical:
> Approved with a few minor nits - I like to reference the full path to files from the repo root in the CHANGELOG, but it's a personal preference. Great work 💯 🚀 !

That makes sense to me. I've committed the suggestions, thanks! @markdboyd

PR #1991, opened 2020-11-10
Fix CI bootstrap script to not run TS compilation

markdboyd:
The `ci:bootstrap-no-scripts-quiet` script did not have the `--ignore-scripts` flag for `lerna bootstrap` set, so it was needlessly running `prepare` scripts and the TS compilation, slowing down CI times. I noticed that our unit tests were back up to ~16 minutes, but with this change they should go back to ~10 minutes.


botanical (approved):
lgtm

PR #1997, opened 2020-11-12
[CUMULUS-2251] Add log_api_gateway_to_cloudwatch var to thin_egress_app module definition

botanical:

Addresses [CUMULUS-2251: Distribution module bug](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2251)

## Changes

* Added `log_api_gateway_to_cloudwatch` var to `thin_egress_app` module defintion so that it could return a non-null value for `module.thin_egress_app.egress_log_group`
https://github.com/asfadmin/thin-egress-app/blob/devel/terraform/outputs.tf#L20-L22
since `log_api_gateway_to_cloudwatch` defaults to false https://github.com/asfadmin/thin-egress-app/blob/devel/terraform/variables.tf#L101-L105
* Update changelog

## PR Checklist

- [X] Update CHANGELOG
- [ ] Unit tests
- [X] Adhoc testing
- [ ] Integration tests


flamingbear (approved):
LGTM. thanks!

PR #2000, opened 2020-11-20
[CUMULUS-2198] Write Dynamo to RDS Migration for Rules

botanical:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [x] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


lambdas/data-migration1/src/rules.ts
```diff
@@ -0,0 +1,162 @@
...
+    .from<CollectionRecord>('collections')
+    .where({ name: name, version: version })
+    .first();
+
+  if (record === undefined) {
+    throw new RecordDoesNotExist(`Collection with name ${name} and version ${version} does not exist.`);
+  }
+  return record.collectionCumulusId;
+};
+
```
botanical, line 66: I was considering having a generic `getCumulusId` function instead of `getCollectionCumulusId` and `getProviderCumulusId`

botanical:
This is a draft because I still need to rebase the merged changes to `feature-rds` and convert everything to snake case.

lambdas/data-migration1/tests/test-rules.js
```diff
@@ -0,0 +1,394 @@
...
+        enabled: true,
+        log_event_arn: null,
+        execution_name_prefix: null,
+        payload: null,
+        queue_url: null,
+        tags: null,
+        created_at: new Date(fakeRule.createdAt),
+        updated_at: new Date(fakeRule.updatedAt),
+      },
+      ['createdAt', 'updatedAt', 'state', 'provider', 'collection', 'rule']
```
botanical, line 175: I realize I could create a `const` for this but it's only used in two places and it seems better to have this easily readable to anyone trying to interpret the test as opposed to referring to a `const` at the top of the file. Open to changing this though

lambdas/data-migration1/src/rules.ts
```diff
@@ -0,0 +1,194 @@
...
+    .first();
+
+  if (record === undefined) {
+    throw new RecordDoesNotExist(`Collection with name ${name} and version ${version} does not exist.`);
+  }
+  return record.collection_cumulus_id;
+};
+*/
+
+/**
```
botanical, line 66: Below is the generalized `getCumulusId` function I wrote to contrast the two non generalized functions above. My main issue is that I'm not sure how to pass in a type so that I could actually query the tables like:
```
const record = await knex.select('cumulus_id')
    .from<TYPE>(table)
    .where(whereClause)
    .first();
```
Any input on how to do that in TypeScript would be much appreciated (here's a cookie 🍪 )

lambdas/data-migration1/src/rules.ts
```diff
@@ -0,0 +1,194 @@
...
+    .where({ name: dynamoRecord.name })
+    .first();
+
+  // Throw error if it was already migrated.
+  if (existingRecord) {
+    throw new RecordAlreadyMigrated(`Rule name ${dynamoRecord.name} was already migrated, skipping`);
+  }
+
+  // const collectionCumulusId = await getCollectionCumulusId(dynamoRecord.collection.name, dynamoRecord.collection.version, knex);
+  // const providerCumulusId = await getProviderCumulusId(dynamoRecord.provider, knex);
```
botanical, line 118: I am aware that these are commented out

packages/db/src/types.ts
```diff
@@ -1,52 +1,52 @@
 import { AsyncOperationStatus, AsyncOperationType } from '@cumulus/types/api/async_operations';
```
botanical, line 1: I saw inconsistency in this file and opted to add commas, but I think we need to discuss if we want commas or nothing.

lambdas/data-migration1/src/rules.ts
```diff
@@ -0,0 +1,194 @@
...
+
+import { RecordDoesNotExist } from '@cumulus/errors';
+import { RecordAlreadyMigrated } from './errors';
+import { MigrationSummary } from './types';
+
+const logger = new Logger({ sender: '@cumulus/data-migration/rules' });
+const Manager = require('@cumulus/api/models/base');
+const schemas = require('@cumulus/api/models/schemas');
+
+/**
```
botanical, line 16: I am aware that two functions are commented out, see below ✨

lambdas/data-migration1/tests/test-rules.js
markdboyd: I don't have strong feelings, but there's always the argument of only having to update something in one place as opposed to two

lambdas/data-migration1/tests/test-rules.js
```diff
@@ -0,0 +1,394 @@
...
+        ...fakeRule,
+        arn: fakeRule.rule.arn ? fakeRule.rule.arn : null,
+        type: fakeRule.rule.type,
+        value: fakeRule.rule.value ? fakeRule.rule.value : null,
+        enabled: true,
+        log_event_arn: null,
+        execution_name_prefix: null,
+        payload: null,
+        queue_url: null,
+        tags: null,
```
markdboyd, line 171: I think I'd prefer to have the test validating the successful migration record including values for all possible fields. Then leave the nullable testing for that separate test.

I'd also like this test to verify that the correct provider/collection are attached to the rule. You could create those records in RDS before doing the migration.

lambdas/data-migration1/tests/test-rules.js
```diff
@@ -0,0 +1,394 @@
...
+  const { knex, providerKmsKeyId } = t.context;
+  const fakeRule = generateFakeRule(fakeCollection.name, fakeCollection.version, fakeProvider.id);
+
+  delete fakeRule.rule.value;
+  delete fakeRule.rule.arn;
+  delete fakeRule.payload;
+  delete fakeRule.logEventArn;
+  delete fakeRule.queueUrl;
+  delete fakeRule.meta;
+  delete fakeRule.tags;
```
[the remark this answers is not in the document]
markdboyd: And `collection` and `provider`

lambdas/data-migration1/tests/test-rules.js
```diff
@@ -0,0 +1,394 @@
...
+    encrypted: false,
+    privateKey: 'key',
+    cmKeyId: 'key-id',
+    certificateUri: 'uri',
+    createdAt: Date.now(),
+    updatedAt: Date.now(),
+  });
+  const fakeRule1 = generateFakeRule(fakeCollection.name, fakeCollection.version, fakeProvider.id);
+  const fakeRule2 = generateFakeRule(anotherFakeCollection.name, anotherFakeCollection.version, anotherFakeProvider.id);
+  const queueUrls1 = await createSqsQueues(randomString());
```
markdboyd, line 320: I don't think we need to use this helper, which actually creates an SQS queue in Localstack. We can just use random strings for the queue URLs, since that's all the field validation requires.

packages/db/src/types.ts
```diff
@@ -83,3 +83,25 @@ export interface PostgresProviderRecord extends PostgresProvider {
   updated_at: number,
   cumulusId: number,
 }
+
+export interface PostgresRule {
+  name: string,
+  workflow: string,
+  collection_cumulus_id: number,
```
markdboyd, line 90: Collection/provider aren't required in Dynamo, so I don't think it's intentional/correct to make them required in RDS

packages/db/src/types.ts
```diff
@@ -83,3 +83,25 @@ export interface PostgresProviderRecord extends PostgresProvider {
...
+  provider_cumulus_id: number,
+  execution_name_prefix?: string,
+  type: string,
+  enabled: boolean,
+  value: string,
+  arn: string,
+  log_event_arn: string,
```
markdboyd, line 97: I think these should all be optional based on the RDS schema

markdboyd (changes requested):
Looks good! I have a few comments, mostly in the tests

lambdas/data-migration1/src/rules.ts
```diff
@@ -0,0 +1,194 @@
...
+  table: 'providers' | 'collections',
+  knex: Knex
+): Promise<number> => {
+  const record = await knex.select('cumulus_id')
+    .from(table)
+    .where(whereClause)
+    .first();
+
+  if (record === undefined) {
+    throw new RecordDoesNotExist(`${table.toUpperCase()} with identifiers ${whereClause} does not exist.`);
```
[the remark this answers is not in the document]
botanical: Yeah, that makes sense. I wanted to uppercase it because I'm accustomed to starting my statements with an uppercased word. I'll change it!

lambdas/data-migration1/src/rules.ts
```diff
@@ -0,0 +1,194 @@
...
+    provider_cumulus_id: providerCumulusId,
+    collection_cumulus_id: collectionCumulusId,
+    enabled: dynamoRecord.state === 'ENABLED',
+    type: dynamoRecord.rule.type,
+    value: dynamoRecord.rule.value,
+    arn: dynamoRecord.rule.arn,
+    log_event_arn: dynamoRecord.rule.logEventArn,
+    execution_name_prefix: prefix,
+    payload: dynamoRecord.payload,
+    meta: dynamoRecord.meta ? JSON.stringify(dynamoRecord.meta) : undefined,
```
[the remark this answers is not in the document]
botanical: Oh! Okay, I will update that.

botanical:
@markdboyd 

> I'm almost wondering if we should move this to packages/db/database.ts. I don't want to create helpers for very low level stuff, but getting the cumulus ID for some type of record is probably something we'll do a lot. And a reusable helper could replace this type of code for me:
> https://github.com/nasa/cumulus/blob/feature-rds/packages/api/lambdas/sf-event-sqs-to-db-records.js#L92-L98
> What do you think?

I think that would make a lot of sense. I hadn't realized you were doing something similar in `sf-event-sqs-to-db-records` but since you are, we should probably have this code be defined once and imported for use when needed.

lambdas/data-migration1/src/rules.ts
```diff
@@ -0,0 +1,194 @@
...
+  if (record === undefined) {
+    throw new RecordDoesNotExist(`Collection with name ${name} and version ${version} does not exist.`);
+  }
+  return record.collection_cumulus_id;
+};
+*/
+
+/**
+ *
+ * Retrieve cumulus_id from table using name and version.
```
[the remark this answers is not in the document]
botanical: Ah, okay, will update. Thanks!

packages/db/src/types.ts
botanical: I see. I thought it was required in RDS because it's `nonNullable`
```
    table
      .integer('collection_cumulus_id')
      .references('cumulus_id')
      .inTable('collections')
      .notNullable();
    table
      .integer('provider_cumulus_id')
      .references('cumulus_id')
      .inTable('providers')
      .notNullable()
```
https://github.com/nasa/cumulus/blob/feature-rds/lambdas/db-migration/src/migrations/20201014105024_create_rules_table.ts#L17-L25

packages/db/src/types.ts
botanical: Yes, you are correct

lambdas/data-migration1/src/rules.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  // Throw error if it was already migrated.
+  if (existingRecord) {
+    throw new RecordAlreadyMigrated(`Rule name ${dynamoRecord.name} was already migrated, skipping`);
+  }
+
+  // const collectionCumulusId = await getCollectionCumulusId(dynamoRecord.collection.name, dynamoRecord.collection.version, knex);
+  // const providerCumulusId = await getProviderCumulusId(dynamoRecord.provider, knex);
+  const prefix = dynamoRecord.executionNamePrefix ? dynamoRecord.executionNamePrefix : undefined;
+  const collectionCumulusId = await getCumulusId(
```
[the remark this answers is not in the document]
botanical: Ah, that is true, I assumed that collection and provider are required references for a rule. I still think that is true though, please see https://github.com/nasa/cumulus/pull/2000#discussion_r529198031 . Open to hearing that I am wrong about this 👂

lambdas/data-migration1/tests/test-rules.js
```diff
@@ -0,0 +1,394 @@
...
+        tags: null,
+        created_at: new Date(fakeRule.createdAt),
+        updated_at: new Date(fakeRule.updatedAt),
+      },
+      ['createdAt', 'updatedAt', 'state', 'provider', 'collection', 'rule']
+    )
+  );
+});
+
+test.serial('migrateRuleRecord ignores extraneous fields from Dynamo', async (t) => {
```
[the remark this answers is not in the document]
botanical: Does `rule` count? Because we have in the DynamoDB schema 
```
rule: {
      title: 'Ingest Rule',
      type: 'object',
      properties: {
        type: {
          type: 'string',
          enum: ['onetime', 'scheduled', 'sns', 'kinesis', 'sqs'],
        },
        // Value is multi-use.   For a kinesis rule this is the target stream arn, for
        // a scheduled event it's the schedule pattern (e.g. cron), for a one-time rule.
        value: {
          type: 'string',
        },
        // Kinesis scheduled event arn
        arn: {
          type: 'string',
          readonly: true,
        },
        // Kinesis scheduled log event arn
        logEventArn: {
          type: 'string',
          readonly: true,
        },
      },
      required: ['type'],
    },
```

but we don't maintain this structure

packages/db/src/types.ts
markdboyd: Yeah, I think that table schema was defined erroneously. And fun fact: I'm the one who did it! See #1932 

But providers/collections are not set as required in the Dynamo schema, so I don't know why they suddenly would be for RDS.

lambdas/data-migration1/src/rules.ts
```diff
@@ -0,0 +1,141 @@
...
+): Promise<any> => {
+  const record = await knex.select('cumulus_id')
+    .from<T>(table)
+    .where(whereClause)
+    .first();
+
+  if (record === undefined) {
+    throw new RecordDoesNotExist(`Record in ${table} with identifiers ${whereClause} does not exist.`);
+  }
+  return record;
```
markdboyd, line 39: I don't understand why we changed this function to return the record instead of record.cumulus_id

lambdas/data-migration1/tests/test-rules.js
```diff
@@ -0,0 +1,399 @@
+/* eslint-disable max-len */
```
markdboyd, line 1: I'd rather not disable this rule and fix the cases where we have lines that are too long

lambdas/data-migration1/tests/test-rules.js
```diff
@@ -0,0 +1,399 @@
...
+    .first();
+
+  t.deepEqual(
+    omit(createdRecord, ['cumulus_id', 'collection_cumulus_id', 'provider_cumulus_id' ]),
+    omit(
+      {
+        name: fakeRule.name,
+        workflow: fakeRule.workflow,
+        meta: fakeRule.meta,
+        arn: fakeRule.rule.arn ? fakeRule.rule.arn : null,
```
markdboyd, line 167: In this test, I think the ternary expressions, `?`, are unnecessary, because we're expecting all of these properties to have values

lambdas/data-migration1/src/rules.ts
botanical: I was running into this error when I tried to return `record.cumulus_id`
```
Property 'cumulus_id' does not exist on type 'UnknownToAny<(AnyToUnknown<T> extends {} ? Boxed<"cumulus_id"> extends Boxed<keyof T> ? SafePick<T, "cumulus_id" & keyof T> : any : unknown) | undefined>'.
```

I attempted to define the function to have `<T extends { cumulus_id: number }>` and also extending an interface I wrote that has the same property to no avail. 

Do you have any insight? I would also prefer to just return `record.cumulus_id`

lambdas/data-migration1/tests/test-rules.js
botanical: That is true

lambdas/data-migration1/tests/test-rules.js
```diff
@@ -0,0 +1,399 @@
...
+      {
+        name: fakeRule.name,
+        workflow: fakeRule.workflow,
+        meta: fakeRule.meta,
+        arn: fakeRule.rule.arn,
+        type: fakeRule.rule.type,
+        value: fakeRule.rule.value,
+        enabled: true,
+        log_event_arn: fakeRule.rule.logEventArn,
+        execution_name_prefix: null,
```
markdboyd, line 172: Can we add a value for this to the source record so we can ensure this value is properly migrated?

lambdas/data-migration1/tests/test-rules.js
```diff
@@ -0,0 +1,399 @@
...
+    .where({ name: fakeRule.name })
+    .first();
+
+  t.deepEqual(
+    omit(createdRecord, ['cumulus_id', 'collection_cumulus_id', 'provider_cumulus_id']),
+    omit(
+      {
+        ...fakeRule,
+        arn: fakeRule.rule.arn ? fakeRule.rule.arn : null,
+        value: fakeRule.rule.value ? fakeRule.rule.value : null,
```
markdboyd, line 239: We know that these values should be null based on the setup in this test

lambdas/data-migration1/tests/test-rules.js
```diff
@@ -0,0 +1,399 @@
...
+
+  fakeCollection = fakeCollectionFactory();
+  fakeProvider = fakeProviderFactory({
+    encrypted: false,
+    privateKey: 'key',
+    cmKeyId: 'key-id',
+    certificateUri: 'uri',
+    createdAt: Date.now(),
+    updatedAt: Date.now(),
+  });
```
markdboyd, line 84: It seems like we need a fakeCollection/fakeProvider for most of the tests, so I'd move this to a `beforeEach` handler for the tests and attach them to `t.context`, like so:

```
test.beforeEach((t) => {
  t.context.fakeCollection = fakeCollectionFactory();
  t.context.fakeProvider = fakeProviderFactory({
    encrypted: false,
    privateKey: 'key',
    cmKeyId: 'key-id',
    certificateUri: 'uri',
    createdAt: Date.now(),
    updatedAt: Date.now(),
  });
});
```

Then, in each test, you'll have a new version of these objects available that you can access like so:

`const { fakeCollection, fakeProvider } = t.context`

lambdas/data-migration1/tests/test-rules.js
```diff
@@ -0,0 +1,399 @@
...
+  const fakeRule = generateFakeRule(fakeCollection.name, fakeCollection.version, fakeProvider.id);
+
+  await migrateCollectionRecord(fakeCollection, knex);
+  await migrateProviderRecord(fakeProvider, providerKmsKeyId, knex);
```
markdboyd, line 151: I'm kind of wary of using the collection/provider migration logic in the tests for the rule migration. With this setup, the rule migration tests could start failing due to bugs in the collection/provider migration logic, so the tests aren't very well isolated. I think I would prefer us manually creating the records in Postgres using `knex.insert`

botanical: That makes sense to me. I definitely agree. I'll update accordingly

lambdas/data-migration1/tests/test-rules.js
```diff
@@ -0,0 +1,399 @@
...
+    certificateUri: 'uri',
+    createdAt: Date.now(),
+    updatedAt: Date.now(),
+  });
+  const fakeRule = generateFakeRule(fakeCollection.name, fakeCollection.version, fakeProvider.id);
+
+  // make source record invalid
+  delete fakeRule.workflow;
+
+  await t.throwsAsync(migrateRuleRecord(fakeRule, t.context.knex));
```
markdboyd, line 199: Let's check the error type here to ensure it's the right type

```suggestion
  await t.throwsAsync(
    migrateRuleRecord(fakeRule, t.context.knex),
    { name: 'SchemaValidationError' }
  );
```

lambdas/data-migration1/tests/test-rules.js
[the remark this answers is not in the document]
markdboyd: Good point. Though I think this case is sufficiently covered by this test?

https://github.com/nasa/cumulus/pull/2000/files#diff-d7b0058e0252e8083289a57a228dc6cec19ad3877e20ff8c22626a419ae27affR160

If there was a `rule` property returned on the `createdRecord` in that test, I think it would fail. But feel free to verify that

If that is true, then this test isn't necessary

lambdas/data-migration1/src/rules.ts
markdboyd: Yeah, it comes down to the function not being aware of `cumulus_id` existing on the record now that we've introduced a generic type.  Let me take a look at it

markdboyd: What you tried is supposed to work: https://www.typescriptlang.org/docs/handbook/generics.html#generic-constraints

So the reason it doesn't may have more to do with Knex. I'll look into it.

packages/db/src/types.ts
botanical: Haha okay, I will update it so they are no longer required

lambdas/data-migration1/src/rules.ts
markdboyd: This is very tricky. The problem is that we're still applying the generic type to Knex here:

`.from<T>(table)`

And `T` is a generic type, so it doesn't specify that `cumulus_id` must be a part of it, hence the error. 

For now, I'd just drop the generic and do this:

```
interface CumulusRecord {
  cumulus_id: number
}

export const getCumulusId = async (
  whereClause : object,
  table: 'providers' | 'collections',
  knex: Knex
): Promise<number> => {
  const record = await knex.select('cumulus_id')
    .from<CumulusRecord>(table)
    .where(whereClause)
    .first();

  if (record === undefined) {
    throw new RecordDoesNotExist(`Record in ${table} with identifiers ${whereClause} does not exist.`);
  }
  return record.cumulus_id;
};
```

However, this is less than ideal, because someone could in theory run this function against a table where `cumulus_id` does not exist, and it wouldn't fail TS compilation, but would fail at runtime. You can try to look into a solution if you want, but I would timebox it to 30 minutes.

lambdas/data-migration1/tests/test-rules.js
botanical: [`2b0d9cf` (#2000)](https://github.com/nasa/cumulus/pull/2000/commits/2b0d9cf497d9b982d04544957b3a85a4059e4c7e#diff-d7b0058e0252e8083289a57a228dc6cec19ad3877e20ff8c22626a419ae27affR44-R55)

lambdas/data-migration1/src/rules.ts
botanical: Okay, I will look into this but timebox to 30mins. Thanks for looking into this further! I think it is not great that it would fail at runtime so I'm going to try to see a way to prevent that.

lambdas/data-migration1/tests/test-rules.js
```diff
@@ -0,0 +1,390 @@
...
+  );
+  await t.throwsAsync(result, { name: 'ColumnDoesNotExist' });
+});
+
+test.serial('migrateRuleRecord correctly migrates rule record', async (t) => {
+  const { knex } = t.context;
+  const fakeRule = generateFakeRule(fakeCollection.name, fakeCollection.version, fakeProvider.id);
+
+  await migrateFakeCollectionRecord(fakeCollection, knex);
+  await migrateFakeProviderRecord(fakeProvider, knex);
```
markdboyd, line 160: I'd like to remove the use of these migration functions for collections/providers in favor of creating the records directly via Knex. Otherwise, this test is now inadvertently also testing the migration logic for collections/providers, so it's not isolated from bugs in that logic.

lambdas/data-migration1/src/rules.ts
markdboyd: See #2006 for a proposed refactor

lambdas/data-migration1/tests/test-rules.js
```diff
@@ -0,0 +1,390 @@
...
+  await createBucket(process.env.system_bucket);
+
+  fakeCollection = fakeCollectionFactory();
+  fakeProvider = fakeProviderFactory({
+    encrypted: true,
+    privateKey: 'key',
+    cmKeyId: 'key-id',
+    certificateUri: 'uri',
+    createdAt: new Date(2020, 11, 17),
+  });
```
markdboyd, line 89: It seems like you're doing this in several of the tests, so I would move these calls to a `beforeEach` hook and attach them to `t.context` (e.g. `t.context.fakeCollection`). Then, in any given test, you can access `t.context.fakeCollection` rather than having to create it in the test setup.

lambdas/data-migration1/tests/test-rules.js
botanical: Yes, I agree with you. I've written the functions `migrateFakeCollectionRecord` and `migrateFakeProviderRecord` which will call knex directly https://github.com/nasa/cumulus/pull/2000/files#diff-d7b0058e0252e8083289a57a228dc6cec19ad3877e20ff8c22626a419ae27affR44-R54

```
const migrateFakeCollectionRecord = async (record, knex) => {
  const updatedRecord = translateApiCollectionToPostgresCollection(record);
  await knex('collections').insert(updatedRecord);
};

const fakeEncryptFunction = async () => 'fakeEncryptedString';

const migrateFakeProviderRecord = async (record, knex) => {
  const updatedRecord = await translateApiProviderToPostgresProvider(record, fakeEncryptFunction);
  await knex('providers').insert(updatedRecord);
};
```
Sorry if I didn't make that more clear!

markdboyd: Awesome, well done!

lambdas/data-migration1/tests/test-rules.js
```diff
@@ -0,0 +1,395 @@
...
+
+const migrateFakeCollectionRecord = async (record, knex) => {
+  const updatedRecord = translateApiCollectionToPostgresCollection(record);
+  await knex('collections').insert(updatedRecord);
+};
+
+const fakeEncryptFunction = async () => 'fakeEncryptedString';
+
+const migrateFakeProviderRecord = async (record, knex) => {
+  const updatedRecord = await translateApiProviderToPostgresProvider(record, fakeEncryptFunction);
```
markdboyd, line 52: Well done using the fake encrypt function

lambdas/data-migration1/tests/test-rules.js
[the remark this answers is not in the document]
markdboyd: Yeah, after looking at it more, I'm pretty sure that this test is unnecessary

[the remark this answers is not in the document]
botanical: I see, that's a good point. I agree and will remove the test!

lambdas/data-migration1/src/errors.ts
```diff
@@ -1,3 +1,4 @@
 import { createErrorType } from '@cumulus/errors';
 
 export const RecordAlreadyMigrated = createErrorType('RecordAlreadyMigrated');
+export const ColumnDoesNotExist = createErrorType('ColumnDoesNotExist');
```
markdboyd, line 4: I think this error type is no longer necessary/used

lambdas/data-migration1/src/rules.ts
```diff
@@ -0,0 +1,128 @@
...
+      knex
+    )
+    : undefined;
+  const providerCumulusId = dynamoRecord.provider
+    ? await getRecordCumulusId<PostgresProviderRecord>(
+      { name: dynamoRecord.provider },
+      tableNames.providers,
+      knex
+    )
+    : undefined;
```
markdboyd, line 59: I know that I'm the one who most recently changed these lines, but looking again, I think we need to surround these in a try/catch or Promise.allSettled, because `getRecordCumulusId` right now throws an error if the record is not found.

Or, `getRecordCumulusId` itself could just return `undefined` instead of throwing an error if the record is not found. That would simplify the code but I'm not sure if it's better or not. What do you think?

lambdas/data-migration1/src/errors.ts
botanical: Yes, that's correct, I'll remove it

markdboyd (approved):

PR #2006, opened 2020-11-25
CUMULUS-2198: getCumulusId refactor

markdboyd:
Changes:

- Move `getCumulusId` to `getRecordCumulusId` in the `db` package
- Update getRecordCumulusId to use type generics to enforce that the record type **must have a `cumulus_id` property**
- Changed the tableNames export to be an enum instead of just an object map. An enum is helpful for referencing table names in TS, but has no impact on the compiled output for JS
- Adds a `__name` property to the record types referring to the table name for that record

The refactor of `getRecordCumulusId` prevents us from using this function to do a query against a table that does not have a `cumulus_id` column, **so as long as our types describing the record types are accurate**. It also enforces that the specified table name must match the table for the record type, which requires an extra `__name` property on the types that I don't like, but this is the only implementation that seemed workable.

See this code sandbox for a basic demo of the implementation (hat tip to @alukach for assistance with this idea):

https://codesandbox.io/s/adoring-kepler-wohkr


packages/db/src/types.ts
```diff
@@ -35,6 +37,8 @@ export interface PostgresCollection {
 }
 
 export interface PostgresCollectionRecord extends PostgresCollection {
+  __name: tableNames.collections,
```
markdboyd, line 40: This is by far the part of this approach that I like the least, but it's the only workable solution that I could find to enforce that a separate table name argument to a function matches the record type specified. And using enums to references the table names makes me feel slightly better about it.

packages/db/src/database.ts
```diff
@@ -1,9 +1,37 @@
...
   params: Partial<T>,
   knex: Knex,
   tableName: string
 ): Promise<boolean> => isRecordDefined(await knex<T>(tableName).where(params).first());
+
+/**
+ * Retrieve cumulus_id for a record from the specified table.
+ *
+ * @param {Object} whereClause - where clause for query
+ * @param {string} table - Name of table
```
botanical, line 19: This is a nit but should the comment be `@param {tableNames} table - Name of table`? 
```suggestion
 * @param {tableNames} table - Name of table
```

packages/db/src/tables.ts
```diff
@@ -1,9 +1,9 @@
...
-};
+export enum tableNames {
+  asyncOperations = 'async_operations',
+  collections = 'collections',
+  executions = 'executions',
+  granules = 'granules',
+  pdrs = 'pdrs',
+  providers = 'providers',
+  rules = 'rules',
+}
```
botanical, line 9: Enums are awesome. Nice change!

packages/db/src/types.ts
```diff
@@ -81,7 +83,7 @@ export interface PostgresProvider {
 export interface PostgresProviderRecord extends PostgresProvider {
   created_at: number,
   updated_at: number,
-  cumulusId: number,
+  cumulus_id: number
```
botanical, line 86: This is a nit too - any reason why you removed the comma? 
```suggestion
  cumulus_id: number,
```
I don't mind either way but I'd like to be consistent.

packages/db/src/database.ts
markdboyd: Agreed and changed

botanical (approved):
🎉  Thanks for looking into the TS stuff around partials and generics. Really insightful and useful.

PR #2012, opened 2020-12-04
[CUMULUS-2199] Update all rule database writes to write to both Dynamo and RDS

botanical:

Addresses [CUMULUS-2199: Update all rule database writes to write to both Dynamo and RDS](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2199)

## Changes

* Add logic to rules endpoint to write to RDS 
* Add translation function to convert dynamodb record to RDS record for rules table

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


packages/api/lambdas/sf-event-sqs-to-db-records/index.js
```diff
@@ -87,14 +94,20 @@ const writeRecords = async (cumulusMessage, knex) => {
...
-    return await writeGranules({
+    await writeGranules({
       cumulusMessage,
       collectionCumulusId,
       providerCumulusId,
       executionCumulusId,
       pdrCumulusId,
       knex,
     });
+    return await writeRules({
```
markdboyd, line 105: I know that this is a WIP - but I just glanced at this and wanted to call out that rules are not one of the data types that is written to the database by the lambda. So the logic for saving them wouldn't go here, nor is there ever (that I'm aware of) rules data in `payload.rules`. Rules are created only via the API in `packages/api/endpoints/rules.js`, so we need to update that code to save to both DynamoDb and Postgres. We have already done for this collections if you want to look at that API endpoint code.

botanical: Thanks, I opened it as a draft _because_ I wasn't really sure I was going down the right path. I will take a look at the collections api endpoint code. I for some reason assumed that rules would be in the payload of a cumulus message but I can't remember how I came to that conclusion.

botanical:
This is _mostly_ ready for review apart from the one test I've marked as `skipped` since I am having trouble stubbing knex. https://github.com/nasa/cumulus/pull/2012/files#diff-49139a123843cc0118f2bc2b3275724314181f1de87be662e46c5125aca6c844R559

packages/db/src/rules.ts
```diff
@@ -0,0 +1,62 @@
...
+    provider_cumulus_id: record.provider ? providerCumulusId : undefined,
+    collection_cumulus_id: record.collection ? collectionCumulusId : undefined,
+    meta: (record.meta ? JSON.stringify(record.meta) : undefined),
+    payload: record.payload as any,
+    queue_url: record.queueUrl,
+    arn: record.rule.arn,
+    type: record.rule.type,
+    value: record.rule.value,
+    log_event_arn: record.rule.logEventArn,
+    enabled: record.state ? record.state === 'ENABLED' : true,
```
markdboyd, line 54:
```suggestion
    enabled: record.state === 'ENABLED' ,
```

The ternary expression is unnecessary, I believe. If `record.state` is undefined, then `record.state === 'ENABLED` will evaluate to false.

packages/db/src/rules.ts
```diff
@@ -0,0 +1,62 @@
...
+export const translateApiRuleToPostgresRule = async (
+  record: RuleRecord,
+  dbClient: Knex
+): Promise<PostgresRule> => {
+  const collectionCumulusId = await getCollectionCumulusId(record, dbClient);
+  const providerCumulusId = await getProviderCumulusId(record, dbClient);
+  const ruleRecord: PostgresRule = {
+    name: record.name,
+    workflow: record.workflow,
+    provider_cumulus_id: record.provider ? providerCumulusId : undefined,
```
markdboyd, line 45: Since we already have a ternary expression here, we can simplify the code to:

```suggestion
    provider_cumulus_id: record.provider ? await getRecordCumulusId<PostgresProviderRecord>(
      { name: record.provider },
      tableNames.providers,
      dbClient
    ) : undefined,
```

And remove the `getProviderCumulusId` helper. And same thing for collections

packages/db/tests/test-rules.js
```diff
@@ -0,0 +1,44 @@
+const test = require('ava');
+const omit = require('lodash/omit');
+const { translateApiRuleToPostgresRule } = require('../dist/rules');
+
+test('translateApiRuleToPostgresRule converts API rule to Postgres', async (t) => {
+  const record = {
+    name: 'name',
+    workflow: 'workflow_name',
+    provider: undefined,
```
markdboyd, line 9: You can just omit properties rather than setting them to `undefined` manuallly

packages/types/api/rules.d.ts
```diff
@@ -36,6 +36,9 @@ export interface NewRuleRecord extends PartialRuleRecord {
 }
 
 export interface RuleRecord extends NewRuleRecord {
+  executionNamePrefix?: string,
+  queueUrl?: string,
+  tags?: string[],
```
markdboyd, line 41: I think these properties belong on the `PartialRuleRecord`, because that type describes a rule record before it is saved to the database, which may include these properties. This type describes a rule record should look like when retrieved from the database, at which popint `createdAt` and `updatedAt` have been set and thus are required

markdboyd (changes requested):

packages/db/src/rules.ts
botanical: Yes, you are correct. Thanks!

packages/api/endpoints/rules.js
```diff
@@ -5,13 +5,19 @@ const { inTestMode } = require('@cumulus/common/test-utils');
...
 
+const { getKnexClient, translateApiRuleToPostgresRule, tableNames } = require('@cumulus/db');
 const { isBadRequestError } = require('../lib/errors');
 const models = require('../models');
 const { Search } = require('../es/search');
 const { addToLocalES, indexRule } = require('../es/indexer');
 
 const log = new Logger({ sender: '@cumulus/api/rules' });
 
+const dynamoRecordToDbRecord = (
```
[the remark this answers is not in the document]
botanical: There is no benefit

packages/db/src/rules.ts
```diff
@@ -0,0 +1,62 @@
...
+    collection_cumulus_id: record.collection ? collectionCumulusId : undefined,
+    meta: (record.meta ? JSON.stringify(record.meta) : undefined),
```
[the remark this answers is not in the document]
botanical: Ooh, noted.

packages/api/endpoints/rules.js
```diff
@@ -60,26 +66,33 @@ async function get(req, res) {
...
     } catch (error) {
-      if (error instanceof RecordDoesNotExist) {
-        const record = await model.create(data);
-
-        if (inTestMode()) {
-          await addToLocalES(record, indexRule);
-        }
-        return res.send({ message: 'Record saved', record });
-      }
+      await model.delete({ name });
```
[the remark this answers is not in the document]
botanical: That makes sense to me to use `knex.transaction`. Thanks for bringing that to my attention!

packages/api/endpoints/rules.js
```diff
@@ -111,19 +124,34 @@ async function put({ params: { name }, body }, res) {
...
 
     // if rule type is onetime no change is allowed unless it is a rerun
     if (body.action === 'rerun') {
-      return models.Rule.invoke(oldRule).then(() => res.send(oldRule));
+      return models.Rule.invoke(oldRule)
+        .then(async () => {
+          await dbClient.transaction(async (trx) => {
+            await trx(tableNames.rules).where({ name: name }).del();
+            await trx(tableNames.rules).insert(postgresRecord);
+          });
```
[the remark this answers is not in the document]
botanical: You're right. I thought it was updating the rule.

packages/db/tests/test-rules.js
```diff
@@ -0,0 +1,44 @@
...
+    payload: { result: { key: 'value' } },
+    tags: ['tag1', 'tag2'],
+    createdAt: Date.now(),
+    updatedAt: Date.now(),
+  };
+
+  const expectedPostgresRule = {
+    name: record.name,
+    workflow: record.workflow,
+    meta: (record.meta ? JSON.stringify(record.meta) : undefined),
```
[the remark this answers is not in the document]
botanical: Oh I see, I for some reason thought meta was an array but it's an object.

packages/db/src/rules.ts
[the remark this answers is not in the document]
botanical: @markdboyd Taking a closer look at this, record.meta looks like:

```
  meta?: {
    retries?: number,
    visibility?: number,
    [key: string]: unknown
  },
```
but meta expects 
```
export interface PostgresRule {
  name: string,
  workflow: string,
  type: string,
  enabled: boolean,
  collection_cumulus_id?: number,
  provider_cumulus_id?: number,
  execution_name_prefix?: string,
  value?: string,
  arn?: string,
  log_event_arn?: string,
  payload?: object,
  meta?: string,
  tags?: string,
  queue_url?: string,
  created_at: Date,
  updated_at: Date,
}
```
Should I have the Postgres `meta` match the object definition I shared?

packages/types/api/rules.d.ts
botanical: Noted

packages/db/src/rules.ts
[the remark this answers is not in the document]
markdboyd: `meta` should definitely at least be an `object` type. But in reality there's nothing in the postgres database schema that will enforce it to match a specific structure - it would accept any valid JSON. So I think `object` is fine

packages/api/endpoints/rules.js
```diff
@@ -5,13 +5,19 @@ const { inTestMode } = require('@cumulus/common/test-utils');
 const { RecordDoesNotExist } = require('@cumulus/errors');
 const Logger = require('@cumulus/logger');
 
+const { getKnexClient, translateApiRuleToPostgresRule, tableNames } = require('@cumulus/db');
```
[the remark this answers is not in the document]
markdboyd: It looks like this is still unaddressed?

packages/api/endpoints/rules.js
```diff
@@ -121,9 +130,17 @@ async function put({ params: { name }, body }, res) {
     // since body is expected to be a replacement rule, not a partial rule
     const fieldsToDelete = Object.keys(oldRule).filter((key) => !(key in body));
     const newRule = await model.update(oldRule, body, fieldsToDelete);
+    const newPostgresRecord = await translateApiRuleToPostgresRule(newRule, dbClient);
 
     if (inTestMode()) await addToLocalES(newRule, indexRule);
 
+    await dbClient.transaction(async (trx) => {
```
markdboyd, line 137: Using the transaction is only really useful if we're going to do multiple updates and we want to ensure that the Postgres update only happens if everything inside the transaction callback is successful. Thus, you need to move the `model.update` in this endpoint handler inside this transaction after the Postgres write. Then, if anything fails in the transaction, none of the writes happen so we have a guarantee that the update either happens in both Dynamo and Postgres or neither.

packages/api/endpoints/rules.js
```diff
@@ -60,26 +61,33 @@ async function get(req, res) {
...
     } catch (error) {
-      if (error instanceof RecordDoesNotExist) {
-        const record = await model.create(data);
-
-        if (inTestMode()) {
-          await addToLocalES(record, indexRule);
-        }
-        return res.send({ message: 'Record saved', record });
-      }
+      await model.delete({ name });
```
markdboyd, line 82: We should also use a transaction here instead of the try/catch with a model.delete

packages/db/src/rules.ts
markdboyd: This was resolved but I don't see the changes?

packages/api/tests/endpoints/test-rules.js
```diff
@@ -213,6 +291,14 @@ test('When calling the API endpoint to delete an existing rule it does not retur
 
   t.is(response.body.message, 'Record saved');
 
+  const fetchedPostgresRecord = await t.context.dbClient.queryBuilder()
+    .select()
+    .table(tableNames.rules)
+    .where({ name: newRule.name })
+    .first();
+
+  t.not(fetchedPostgresRecord, undefined);
```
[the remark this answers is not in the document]
markdboyd: It looks like none of the comments here in the test file were addressed

packages/db/src/rules.ts
botanical: Hmm odd, I am not sure why the changes are there. Let me add them again.

packages/api/tests/endpoints/test-rules.js
```diff
@@ -239,6 +325,14 @@ test('403 error when calling the API endpoint to delete an existing rule without
   newRule.createdAt = record.createdAt;
   newRule.updatedAt = record.updatedAt;
 
+  const fetchedPostgresRecord = await t.context.dbClient.queryBuilder()
+    .select()
+    .table(tableNames.rules)
+    .where({ name: newRule.name })
+    .first();
+
+  t.not(fetchedPostgresRecord, undefined);
```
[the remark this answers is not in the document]
botanical: Definitely agree, I didn't see your comment earlier about this but I also had the same thought and updated the test here: https://github.com/nasa/cumulus/pull/2012/commits/0589d8dbd1062090872cc349f76bbc16aa0daef7

packages/api/tests/endpoints/test-rules.js
```diff
@@ -287,11 +426,18 @@ test('POST creates a rule that is enabled by default', async (t) => {
     .send(newRule)
     .expect(200);
 
+  const fetchedPostgresRecord = await t.context.dbClient.queryBuilder()
+    .select()
+    .table(tableNames.rules)
+    .where({ name: newRule.name })
+    .first();
+
+  t.not(fetchedPostgresRecord, undefined);
```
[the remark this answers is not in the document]
botanical: That's correct, thanks for catching that

packages/api/tests/endpoints/test-rules.js
```diff
@@ -257,9 +351,19 @@ test('403 error when calling the API endpoint to delete an existing rule without
...
+  const { newRule, dbClient, fakeCollection, fakeProvider } = t.context;
+  newRule.provider = fakeProvider.id;
+  newRule.collection = {
+    name: fakeCollection.name,
+    version: fakeCollection.version,
+  };
+
+  const fakeEncryptFn = async () => 'fakeEncryptedString';
+  const collectionRecord = translateApiCollectionToPostgresCollection(fakeCollection);
+  const providerRecord = await translateApiProviderToPostgresProvider(fakeProvider, fakeEncryptFn);
```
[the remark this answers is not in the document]
botanical: That's a good point and I think a good practice overall to minimize the places the test could fail for reasons outside of what we are trying to actually test. I will update this logic.

packages/api/tests/endpoints/test-rules.js
```diff
@@ -390,16 +531,75 @@ test.serial('POST returns a 500 response if record creation throws unexpected er
...
+
+test('POST does not write to DynamoDB if writing to RDS fails', async (t) => {
+  const { newRule, dbClient } = t.context;
+
+  const failingRulesModel = {
+    exists: () => false,
+    create: () => {
+      throw new Error('Rule error');
+    },
+  };
```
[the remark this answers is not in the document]
botanical: Yes, that is true. I've updated

packages/api/tests/endpoints/test-rules.js
```diff
@@ -449,8 +656,62 @@ test('PUT returns 400 for name mismatch between params and payload',
     t.falsy(record);
   });
 
+test('PUT replaces onetime rule for rerun', async (t) => {
```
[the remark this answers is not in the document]
botanical: I added the test because the following line didn't seem to ever be explicitly tested: 
https://github.com/nasa/cumulus/blob/master/packages/api/endpoints/rules.js#L116

I could be wrong but I didn't find a test for that.

packages/api/endpoints/rules.js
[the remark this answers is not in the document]
botanical: https://github.com/nasa/cumulus/pull/2012/commits/ea9c7784b4d0e9841c8ebee053f05bbdcce89ce7

packages/api/endpoints/rules.js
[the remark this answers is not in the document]
botanical: https://github.com/nasa/cumulus/pull/2012/commits/bab31bb6c153ce370babbfd113b5bac6055ab578

packages/api/tests/endpoints/test-rules.js
```diff
@@ -33,17 +48,36 @@ const assertions = require('../../lib/assertions');
 const { app } = require('../../app');
 
 const esIndex = randomString();
-const testRule = fakeRuleFactoryV2();
+const workflow = randomId('workflow-');
+const testRule = {
```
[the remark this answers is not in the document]
botanical: I stopped using it because it was setting values for provider and collection and then causing the tests to fail when it tried looking up the provider and collection for the rule. I have since removed the beforeEach newRule logic and used `fakeRuleFactoryV2` for the tests that need it and then setting `provider` and `collection` to `undefined`. Let me know what you think:

https://github.com/nasa/cumulus/pull/2012/commits/29332bdf9a575edf5381cb1a905ba9c105fb24e4

packages/api/tests/endpoints/test-rules.js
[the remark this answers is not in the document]
botanical: https://github.com/nasa/cumulus/pull/2012/commits/29332bdf9a575edf5381cb1a905ba9c105fb24e4

Hopefully I addressed them all

packages/api/tests/endpoints/test-rules.js
```diff
@@ -268,16 +372,51 @@ test('POST creates a rule', async (t) => {
...
+    .where({ name: newRule.name })
+    .first();
+
   t.is(message, 'Record saved');
 
   newRule.createdAt = record.createdAt;
   newRule.updatedAt = record.updatedAt;
 
-  t.deepEqual(record, newRule);
+  t.not(fetchedPostgresRecord, undefined);
```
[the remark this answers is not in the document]
botanical: https://github.com/nasa/cumulus/pull/2012/commits/29332bdf9a575edf5381cb1a905ba9c105fb24e4#diff-49139a123843cc0118f2bc2b3275724314181f1de87be662e46c5125aca6c844L379

packages/api/tests/endpoints/test-rules.js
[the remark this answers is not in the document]
botanical: [`29332bd` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/29332bdf9a575edf5381cb1a905ba9c105fb24e4#diff-49139a123843cc0118f2bc2b3275724314181f1de87be662e46c5125aca6c844R427)

packages/api/tests/endpoints/test-rules.js
```diff
@@ -257,9 +351,19 @@ test('403 error when calling the API endpoint to delete an existing rule without
...
+  newRule.collection = {
+    name: fakeCollection.name,
+    version: fakeCollection.version,
+  };
+
+  const fakeEncryptFn = async () => 'fakeEncryptedString';
+  const collectionRecord = translateApiCollectionToPostgresCollection(fakeCollection);
+  const providerRecord = await translateApiProviderToPostgresProvider(fakeProvider, fakeEncryptFn);
+  await dbClient(tableNames.collections).insert(collectionRecord);
+  await dbClient(tableNames.providers).insert(providerRecord);
```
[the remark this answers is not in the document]
botanical: [`29332bd` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/29332bdf9a575edf5381cb1a905ba9c105fb24e4#diff-49139a123843cc0118f2bc2b3275724314181f1de87be662e46c5125aca6c844R353)

packages/api/tests/endpoints/test-rules.js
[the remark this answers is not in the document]
botanical: https://github.com/nasa/cumulus/pull/2012/commits/29332bdf9a575edf5381cb1a905ba9c105fb24e4

packages/api/tests/endpoints/test-rules.js
```diff
@@ -390,16 +531,75 @@ test.serial('POST returns a 500 response if record creation throws unexpected er
   }
 });
 
+test.serial('POST does not write to RDS if writing to DynamoDB fails', async (t) => {
```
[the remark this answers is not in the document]
botanical: [`29332bd` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/29332bdf9a575edf5381cb1a905ba9c105fb24e4#diff-49139a123843cc0118f2bc2b3275724314181f1de87be662e46c5125aca6c844R539)

packages/api/tests/endpoints/test-rules.js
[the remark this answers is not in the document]
botanical: [`29332bd` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/29332bdf9a575edf5381cb1a905ba9c105fb24e4#diff-49139a123843cc0118f2bc2b3275724314181f1de87be662e46c5125aca6c844R539)

packages/api/tests/endpoints/test-rules.js
```diff
@@ -390,16 +531,75 @@ test.serial('POST returns a 500 response if record creation throws unexpected er
...
+      .where({ name: newRule.name });
+
+    t.false(await ruleModel.exists(newRule.name));
+    t.is(dbRecords.length, 0);
+  } finally {
+    createRuleStub.restore();
+  }
+});
+
+test('POST does not write to DynamoDB if writing to RDS fails', async (t) => {
```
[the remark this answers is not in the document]
botanical: [`29332bd` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/29332bdf9a575edf5381cb1a905ba9c105fb24e4#diff-49139a123843cc0118f2bc2b3275724314181f1de87be662e46c5125aca6c844R539)

packages/api/tests/endpoints/test-rules.js
```diff
@@ -390,16 +531,75 @@ test.serial('POST returns a 500 response if record creation throws unexpected er
...
+    },
+  };
+
+  const response = buildFakeExpressResponse();
+
+  await post(expressRequest, response);
+
+  t.true(response.boom.badImplementation.calledWithMatch('Rule error'));
+
+  const dbRecords = await dbClient.select()
```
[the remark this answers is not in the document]
botanical: [`29332bd` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/29332bdf9a575edf5381cb1a905ba9c105fb24e4#diff-49139a123843cc0118f2bc2b3275724314181f1de87be662e46c5125aca6c844R601)

packages/api/tests/endpoints/test-rules.js
```diff
@@ -56,6 +90,12 @@ test.before(async () => {
   ruleModel = new Rule();
   await ruleModel.createTable();
 
+  collectionsModel = new Collection({ tableName: process.env.CollectionsTable });
+  await collectionsModel.createTable();
+
+  providersModel = new Provider({ tableName: process.env.ProvidersTable });
+  await providersModel.createTable();
```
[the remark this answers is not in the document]
botanical: No, I have removed them [`29332bd` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/29332bdf9a575edf5381cb1a905ba9c105fb24e4#diff-49139a123843cc0118f2bc2b3275724314181f1de87be662e46c5125aca6c844L91), thanks!

packages/api/tests/endpoints/test-rules.js
```diff
@@ -414,6 +614,13 @@ test('PUT replaces a rule', async (t) => {
     .set('Authorization', `Bearer ${jwtAuthToken}`)
     .expect(200);
 
+  const dbRecords = await dbClient.first()
+    .from(tableNames.rules)
+    .where({ name: expectedRule.name });
+
+  t.is(dbRecords.name, expectedRule.name);
+  t.is(dbRecords.workflow, expectedRule.workflow);
+  t.is(dbRecords.enabled, expectedRule.state === 'ENABLED');
```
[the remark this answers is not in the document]
botanical: https://github.com/nasa/cumulus/pull/2012/commits/29332bdf9a575edf5381cb1a905ba9c105fb24e4#diff-49139a123843cc0118f2bc2b3275724314181f1de87be662e46c5125aca6c844R633

packages/api/endpoints/rules.js
botanical: [`ea9c778` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/ea9c7784b4d0e9841c8ebee053f05bbdcce89ce7)

packages/api/endpoints/rules.js
botanical: [`ea9c778` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/ea9c7784b4d0e9841c8ebee053f05bbdcce89ce7)

packages/db/src/rules.ts
botanical: [`ea9c778` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/ea9c7784b4d0e9841c8ebee053f05bbdcce89ce7)

CHANGELOG.md
```diff
@@ -54,6 +54,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     - `@cumulus/message/workflows/getMessageWorkflowStartTime`
     - `@cumulus/message/workflows/getMessageWorkflowStopTime`
     - `@cumulus/message/workflows/getMessageWorkflowName`
+  - **CUMULUS-2199**
+    - Updated Rules endpoint to write to or delete from RDS if a write to or delete from DynamoDB is successful.
```
markdboyd, line 58: I would put this first note under the "Changed" header in the CHANGELOG

packages/api/endpoints/rules.js
```diff
@@ -60,26 +61,28 @@ async function get(req, res) {
...
-          await addToLocalES(record, indexRule);
-        }
-        return res.send({ message: 'Record saved', record });
-      }
-      throw error;
-    }
+    return await dbClient.transaction(async (trx) => {
+      const record = await model.create(rule);
+      const postgresRecord = await translateApiRuleToPostgresRule(record, dbClient);
+      await trx(tableNames.rules).insert(postgresRecord, 'cumulus_id');
```
markdboyd, line 80: I think we need to reverse the order of operations here: do the Postgres insert first and then the Dynamo insert. The reason why is that right now, the Dynamo write could succeed and the Postgres write would fail, meaning the databases would get out of sync. But if you do the Postgres write first, then the Dynamo write is only reached if the Postgres write doesn't error, **and** if the Dynamo write errors, then the transactional Postgres write is rolled back.

packages/api/endpoints/rules.js
```diff
@@ -60,26 +61,28 @@ async function get(req, res) {
...
+
+      if (inTestMode()) await addToLocalES(record, indexRule);
+
+      return res.send({ message: 'Record saved', record });
```
markdboyd, line 84: I don't think we want either of these things to be a part of the transaction, so they should come after. They are not related to the logic of writing to our databases

packages/api/endpoints/rules.js
```diff
@@ -60,26 +61,28 @@ async function get(req, res) {
...
-        const record = await model.create(data);
-
-        if (inTestMode()) {
-          await addToLocalES(record, indexRule);
-        }
-        return res.send({ message: 'Record saved', record });
-      }
-      throw error;
-    }
+    return await dbClient.transaction(async (trx) => {
```
markdboyd, line 77: Nice job with `return await`

packages/api/endpoints/rules.js
```diff
@@ -120,11 +124,18 @@ async function put({ params: { name }, body }, res) {
...
+      const newRule = await model.update(oldRule, body, fieldsToDelete);
+      const newPostgresRecord = await translateApiRuleToPostgresRule(newRule, dbClient);
+
+      if (inTestMode()) await addToLocalES(newRule, indexRule);
+      await trx(tableNames.rules)
+        .insert(newPostgresRecord)
+        .onConflict('name')
+        .merge();
+      return res.send(newRule);
```
markdboyd, line 137: Again, we need to do the Postgres write before the Dynamo write. And let's move the ES write and `return` outside of the transaction

packages/api/endpoints/rules.js
```diff
@@ -154,7 +166,12 @@ async function del(req, res) {
     }
     throw error;
   }
-  await model.delete(record);
+
+  await dbClient.transaction(async (trx) => {
+    await model.delete(record);
+    await trx(tableNames.rules).where({ name }).del();
```
markdboyd, line 172: Flip the order of these two statements

packages/api/tests/endpoints/test-rules.js
```diff
@@ -257,9 +300,59 @@ test('403 error when calling the API endpoint to delete an existing rule without
...
+    name: fakeProvider.id,
+    cm_key_id: fakeProvider.cmKeyId,
+    certificate_uri: fakeProvider.certificateUri,
+    private_key: fakeProvider.privateKey,
+    host: fakeProvider.host,
+    port: fakeProvider.port,
+  };
+
+  const [collectionCumulusId = {}] = await dbClient(tableNames.collections).insert(collectionRecord).returning('cumulus_id');
+  const [providerCumulusId = {}] = await dbClient(tableNames.providers).insert(providerRecord).returning('cumulus_id');
```
markdboyd, line 354: I don't think the `= {}` default assignment is necessary or correct, because the result from a query with `.returning('cumulus_id')` should be an array with the cumulus IDs, not objects

packages/api/tests/endpoints/test-rules.js
```diff
@@ -268,16 +361,54 @@ test('POST creates a rule', async (t) => {
...
 
   newRule.createdAt = record.createdAt;
   newRule.updatedAt = record.updatedAt;
 
-  t.deepEqual(record, newRule);
+  t.is(fetchedPostgresRecord.collection_cumulus_id, collectionCumulusId);
+  t.is(fetchedPostgresRecord.provider_cumulus_id, providerCumulusId);
+
+  t.deepEqual(
+    omit(fetchedPostgresRecord, ['cumulus_id', 'collection_cumulus_id', 'provider_cumulus_id']),
```
markdboyd, line 383: Nit: I would just add the `collection_cumulus_id` and `provider_cumulus_id` to the deepEqual comparison

packages/api/tests/endpoints/test-rules.js
```diff
@@ -287,11 +418,20 @@ test('POST creates a rule that is enabled by default', async (t) => {
     .send(newRule)
     .expect(200);
 
+  const fetchedPostgresRecord = await t.context.dbClient.queryBuilder()
+    .select()
+    .table(tableNames.rules)
+    .where({ name: newRule.name })
+    .first();
+
+  t.is(fetchedPostgresRecord.enabled, true);
```
markdboyd, line 427:
```suggestion
  t.true(fetchedPostgresRecord.enabled);
```

Nit

packages/api/tests/endpoints/test-rules.js
```diff
@@ -203,6 +234,8 @@ test('GET gets a rule', async (t) => {
 
 test('When calling the API endpoint to delete an existing rule it does not return the deleted rule', async (t) => {
   const newRule = fakeRuleFactoryV2();
+  newRule.provider = undefined;
+  newRule.collection = undefined;
```
markdboyd, line 238: Nit: I'd somewhat prefer to do `delete newRule.provider` and `delete newRule.collection`

packages/api/tests/endpoints/test-rules.js
```diff
@@ -390,16 +536,82 @@ test.serial('POST returns a 500 response if record creation throws unexpected er
   }
 });
 
+test.serial('POST does not write to RDS or Dynamo if writing to DynamoDB fails', async (t) => {
```
markdboyd, line 539:
```suggestion
test.serial('POST does not write to RDS or Dynamo if writing to RDS fails', async (t) => {
```

packages/db/src/rules.ts
```diff
@@ -0,0 +1,45 @@
...
+      tableNames.providers,
+      dbClient
+    ) : undefined,
+    collection_cumulus_id: record.collection ? await getRecordCumulusId<PostgresCollectionRecord>(
+      { name: record.collection.name, version: record.collection.version },
+      tableNames.collections,
+      dbClient
+    ) : undefined,
+    meta: record.meta,
+    payload: record.payload as any,
```
markdboyd, line 31: Is `as any` necessary?

packages/api/tests/endpoints/test-rules.js
```diff
@@ -390,16 +536,82 @@ test.serial('POST returns a 500 response if record creation throws unexpected er
...
+  };
+
+  const response = buildFakeExpressResponse();
+
+  await post(expressRequest, response);
+  const dbRecords = await dbClient.select()
+    .from(tableNames.rules)
+    .where({ name: newRule.name });
+
+  t.false(await ruleModel.exists(newRule.name));
```
markdboyd, line 565: This is interesting. I wouldn't expect this to pass based on my understanding of how the transactions work. Let me take a closer look

packages/api/tests/endpoints/test-rules.js
```diff
@@ -414,6 +626,11 @@ test('PUT replaces a rule', async (t) => {
     .set('Authorization', `Bearer ${jwtAuthToken}`)
     .expect(200);
 
+  const dbRecords = await dbClient.first()
+    .from(tableNames.rules)
+    .where({ name: expectedRule.name });
+
+  t.not(dbRecords.queue_url, undefined);
```
markdboyd, line 633: I admit that I have a really hard time following this test's logic, but I **think** we should be testing the opposite and verifying that `queue_url` **is undefined**. 

The test creates the `expectedRule` using the PUT endpoint, and that rule has the `queueUrl` property removed from it. Thus, the `deepEqual` below proves that the `actualRule` returned after the `PUT` no longer has a `queueUrl` property in Dynamo, and nor should it have a `queue_url` property in Postgres. Unless I'm misunderstanding, which is quite possible

packages/api/tests/endpoints/test-rules.js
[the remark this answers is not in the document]
markdboyd: It's fine to add that test, but all we're testing then is that the original rule passed in is returned by the endpoint. It's a fine test to write, but we can really simplify the test logic in that case. And update the test name.

```suggestion
test('PUT returns original rule for rerun', async (t) => {
```

packages/api/tests/endpoints/test-rules.js
```diff
@@ -449,22 +666,96 @@ test('PUT returns 400 for name mismatch between params and payload',
...
+    .send(expectedRule)
+    .expect(200);
+
+  const { body: actualRule } = await request(app)
+    .get(`/rules/${testRule.name}`)
+    .set('Accept', 'application/json')
+    .set('Authorization', `Bearer ${jwtAuthToken}`)
+    .expect(200);
+
+  t.truthy(actualRule);
```
markdboyd, line 689:
```suggestion
  const expectedRule = {
    ...omit(testRule, ['provider', 'collection']),
    action: 'rerun',
  };

  const { body: actualRule } = await request(app)
    .put(`/rules/${expectedRule.name}`)
    .set('Accept', 'application/json')
    .set('Authorization', `Bearer ${jwtAuthToken}`)
    .send(expectedRule)
    .expect(200);

  t.deepEqual(actualRule, testRule);
```

All we need to prove is that what comes back from the PUT request itself is the original rule with no alterations

packages/api/tests/endpoints/test-rules.js
```diff
@@ -449,22 +666,96 @@ test('PUT returns 400 for name mismatch between params and payload',
...
+});
+
+test.serial('PUT does not insert into RDS if replacing onetime rule for rerun fails', async (t) => {
```
markdboyd, line 692: A PUT request with an action of `rerun` doesn't do any writing to Dynamo or Postgres that I can see, so what is this testing?

packages/db/tests/test-rules.js
markdboyd: This one didn't seem to be addressed

packages/api/tests/endpoints/test-rules.js
markdboyd: Ah, so this test is passing because your faked DB client is missing a `transaction` function, which is now required by the endpoint code. I'd run this test through the debugger and inspect the thrown error to see what I mean

packages/db/src/rules.ts
botanical: Yes I believe so because the type is `unknown` which isn't compatible with `object` 
https://github.com/nasa/cumulus/blob/master/packages/types/api/rules.d.ts#L26

packages/api/tests/endpoints/test-rules.js
botanical: True. Let me update to use the correct destructuring.

packages/api/tests/endpoints/test-rules.js
botanical: Yeah, actually your interpretation is one hundred percent correct. It _should_ be checking that `queue_url` is not defined on the postgres record. I think what I meant to check was that it initially had a value for `queue_url` prior to the update call, but that doesn't make sense because I don't insert `testRule` into the pg table. Let me fix this.

packages/api/tests/endpoints/test-rules.js
botanical: Hm I can see how that is confusing because my test statement and setup doesn't make sense for what I intended. What I intended to do was write a test to ensure that no record to written to postgres when a replacing rule's action is 'rerun`.

botanical: I guess I can just combine tests

botanical: I realize now that this test case is actually covered. I'll remove it

botanical: [`717d0e5` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/717d0e50dc816dd503b437d92a583a563a6c0766)

packages/api/tests/endpoints/test-rules.js
botanical: [`717d0e5` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/717d0e50dc816dd503b437d92a583a563a6c0766)

packages/api/tests/endpoints/test-rules.js
botanical: [`717d0e5` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/717d0e50dc816dd503b437d92a583a563a6c0766)

packages/api/tests/endpoints/test-rules.js
botanical: [`717d0e5` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/717d0e50dc816dd503b437d92a583a563a6c0766)

packages/db/src/rules.ts
botanical: [`717d0e5` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/717d0e50dc816dd503b437d92a583a563a6c0766)

packages/api/tests/endpoints/test-rules.js
botanical: [`717d0e5` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/717d0e50dc816dd503b437d92a583a563a6c0766)

packages/api/tests/endpoints/test-rules.js
botanical: [`717d0e5` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/717d0e50dc816dd503b437d92a583a563a6c0766)

packages/api/tests/endpoints/test-rules.js
```diff
@@ -287,13 +406,22 @@ test('POST creates a rule that is enabled by default', async (t) => {
...
 });
 
 test('POST returns a 409 response if record already exists', async (t) => {
-  const newRule = fakeRuleFactoryV2();
+  const { newRule } = t.context;
+  newRule.createdAt = Date.now();
+  newRule.updatedAt = Date.now();
 
-  await ruleModel.create(newRule);
+  await ruleModel.create(newRule, newRule.createdAt, newRule.updatedAt);
```
markdboyd, line 424: Is setting the `createdAt` and `updatedAt` properties important for this test?

packages/api/tests/endpoints/test-rules.js
markdboyd: What else is covering it?

packages/api/tests/endpoints/test-rules.js
```diff
@@ -414,6 +599,11 @@ test('PUT replaces a rule', async (t) => {
     .set('Authorization', `Bearer ${jwtAuthToken}`)
     .expect(200);
 
+  const dbRecords = await dbClient.select()
+    .from(tableNames.rules)
+    .where({ name: expectedRule.name });
+
+  t.is(dbRecords.queue_url, undefined);
```
markdboyd, line 606: I think this is passing for the wrong reason. You need to add a `first()` after the `.where` on the query above in order to get a DB record. Right now `dbRecords` is an array of DB results, on which `queue_url` definitely will not be defined

packages/api/endpoints/rules.js
```diff
@@ -111,17 +119,32 @@ async function put({ params: { name }, body }, res) {
...
     // since body is expected to be a replacement rule, not a partial rule
     const fieldsToDelete = Object.keys(oldRule).filter((key) => !(key in body));
-    const newRule = await model.update(oldRule, body, fieldsToDelete);
-
+    updatedRuleItem = omit(updatedRuleItem, fieldsToDelete);
+    merge(updatedRuleItem, body);
+
+    const stateChanged = body.state && body.state !== oldRule.state;
+    const valueUpdated = body.rule && body.rule.value !== oldRule.rule.value;
+    const record = await model.updateRuleTrigger(updatedRuleItem, stateChanged, valueUpdated);
```
markdboyd, line 138: Can you explain why you moved this logic into the endpoint from the model? Since it's all very specific to Dynamo, it feels like it is better suited for the model code. Also `model.update` had some helpful defensiveness around ensuring that the record was valid before doing some of this record preparation. Plus it seems like we have duplication between the endpoint and model code now

packages/api/endpoints/rules.js
```diff
@@ -60,26 +64,29 @@ async function get(req, res) {
...
-
-        if (inTestMode()) {
-          await addToLocalES(record, indexRule);
-        }
-        return res.send({ message: 'Record saved', record });
-      }
-      throw error;
-    }
+    await dbClient.transaction(async (trx) => {
+      const postgresRecord = await translateApiRuleToPostgresRule(rule, dbClient);
```
markdboyd, line 84: It's subtle: but as this function isn't directly related to writing records, I would move it out above the transaction. The transaction should still fail if this function returns a `postgresRecord` that isn't compatible with Postgres anyway. 

I would make this change throughout the endpoint code

packages/api/tests/endpoints/test-rules.js
```diff
@@ -390,16 +514,77 @@ test.serial('POST returns a 500 response if record creation throws unexpected er
...
 });
 
+test.serial('POST does not write to RDS or DynamoDB if writing to RDS fails', async (t) => {
+  const { newRule, dbClient } = t.context;
+
+  const failingDbClient = {
+    transaction: () => {
+      throw new Error('Insert Rule Error');
+    },
+  };
```
markdboyd, line 524: This mock actually ensures that **neither the Dynamo or RDS write can happen**, since they both happen within a `knex.transaction`. I think you need a mock that looks more like this:

https://github.com/nasa/cumulus/blob/feature-rds/packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-write-execution.js#L530-L539

packages/api/tests/endpoints/test-rules.js
botanical: Ah, nice catch. Another signifier is that the property should be coming back `null` instead of `undefined`. Thanks for pointing this out.

packages/api/tests/endpoints/test-rules.js
botanical: Negatory, removing

packages/api/endpoints/rules.js
```diff
@@ -60,26 +64,29 @@ async function get(req, res) {
...
+  const {
+    model = new models.Rule(),
+    dbClient = await getKnexClient(),
+  } = req.testContext || {};
+
+  let record;
+  const rule = req.body || {};
+  const name = rule.name;
+  rule.createdAt = Date.now();
+  rule.updatedAt = Date.now();
```
markdboyd, line 76: It seems kind of unnecessary to set these properties directly on the rule if they're going to be reset in the `model.create` call

https://github.com/nasa/cumulus/pull/2012/files#diff-2a37aa7b44507f6480bc0edc75d3aa7e70031468d98c108f8aef958a2098ef91R245

So I would just pass in the right values in the call to model.create rather than setting on the incoming rule.object

packages/api/models/rules.js
```diff
@@ -242,8 +242,8 @@ class Rule extends Manager {
       newRuleItem.state = 'ENABLED';
     }
 
-    newRuleItem.createdAt = Date.now();
-    newRuleItem.updatedAt = Date.now();
+    newRuleItem.createdAt = createdAt || Date.now();
+    newRuleItem.updatedAt = updatedAt || Date.now();
```
markdboyd, line 246: These changes should be unit tested in the rules model tests

packages/api/endpoints/rules.js
botanical: I moved this logic into the endpoint from the model because I wanted to ensure that, if a rule type is anything other than a `onetime` rule, that the rule triggers are updated too, but I see now how that is duplicate and that will be taken care of if it reaches `model.update`. I've since removed it.

packages/api/endpoints/rules.js
botanical: Oh that makes sense to me. I will remove it outside of the transaction.

packages/api/models/rules.js
botanical: [`8aedf3f` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/8aedf3f318e5c41e867a5bf29ac0642f85c06c0f)

packages/api/endpoints/rules.js
botanical: [`9e37ed6` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/9e37ed6f9ae36d213d879fe88fcaab071d440c51)

packages/api/tests/endpoints/test-rules.js
botanical: [`9e37ed6` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/9e37ed6f9ae36d213d879fe88fcaab071d440c51)

packages/api/endpoints/rules.js
botanical: [`3a15639` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/3a1563919a247377bb1f57c4fa1ff3599e49762d)

packages/api/endpoints/rules.js
```diff
@@ -60,26 +64,28 @@ async function get(req, res) {
...
+
+  let record;
+  const rule = req.body || {};
+  const name = rule.name;
+
+  if (await model.exists(name)) {
+    return res.boom.conflict(`A record already exists for ${name}`);
+  }
+
+  const postgresRecord = await translateApiRuleToPostgresRule(rule, dbClient);
```
markdboyd, line 80: Sorry to be annoying - but let's move this inside the try but above the transaction. That way if this function did throw an error, we'd get a deterministic endpoint response based on the `catch` logic

packages/api/endpoints/rules.js
```diff
@@ -60,26 +64,28 @@ async function get(req, res) {
...
-        if (inTestMode()) {
-          await addToLocalES(record, indexRule);
-        }
-        return res.send({ message: 'Record saved', record });
-      }
-      throw error;
-    }
+    await dbClient.transaction(async (trx) => {
+      await trx(tableNames.rules).insert(postgresRecord, 'cumulus_id');
+      record = await model.create(rule, rule.createdAt, rule.updatedAt);
```
markdboyd, line 85: `rule.createdAt` is no longer set, so I think this value is undefined and thus the rule is not being created with the same values. I see now why you were setting `rule.createdAt` before `translateApiRuleToPostgresRule`, so we can restore that if you want so that the records are in sync. Sorry for misunderstanding before

packages/api/endpoints/rules.js
botanical: Not annoying at all! that's a good point, will move it.

packages/api/tests/models/rules/test-rules-model.js
```diff
@@ -196,6 +196,20 @@ test.serial('Creating an invalid rule does not create workflow triggers', async
   }
 });
 
+test('Creating a rule without a createdAt or updatedAt date assigns a value to createdAt and updatedAt', async (t) =>{
```
markdboyd, line 199: I think the new functionality that we want to test is that if we **do provide a createdAt/updatedAt** to `rule.create`, then it is used. This test seems confirming the existing behavior before your changes - that createdAt/updatedAt always get populated, but it doesn't test what values are used

packages/api/endpoints/rules.js
botanical: Yeah, that is correct. Sorry I didn't do a great job of explaining my reasoning before. I've since updated it.

packages/api/endpoints/rules.js
```diff
@@ -111,17 +120,28 @@ async function put({ params: { name }, body }, res) {
...
     }
 
+    let updatedRuleItem = cloneDeep(oldRule);
     // Remove all fields from the existing rule that are not supplied in body
     // since body is expected to be a replacement rule, not a partial rule
     const fieldsToDelete = Object.keys(oldRule).filter((key) => !(key in body));
-    const newRule = await model.update(oldRule, body, fieldsToDelete);
-
+    updatedRuleItem = omit(updatedRuleItem, fieldsToDelete);
+    const record = merge(updatedRuleItem, body);
```
markdboyd, line 135: I'm curious whether this is necessary for Postgres. If you have an existing rule record in Postgres, then you submit an update with fields missing, then I think the upsert query would set those missing fields to null. 

This comes down to the difference between Dynamo and Postgres. Dynamo is a schema-less database where we just store objects, so if we want a property removed, we have to do so manually as part of saving. However, in Postgres, all properties have columns that are either null or have a value. So submitting an update with a field missing to `translateApiRuleToPostgresRule` should mean that field just gets set to null, which is all we need.

Thus, I think we just need

```
 const newPostgresRecord = await translateApiRuleToPostgresRule(body, dbClient);
```

I'm curious if the test for removing a field still passes with just this logic

packages/api/tests/models/rules/test-rules-model.js
botanical: [`15f822f` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/15f822f315e854df2abcce13ecf29bb81f5cfd43)

I am wondering if my checks are overkill for `updatedAt`. I want to ensure that the value passed in for `updatedAt` in fact gets updated to another, more current date

botanical: [`77f5f85` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/77f5f859682d26face2bd271929a656456819d8f)

Actually check out that diff instead

packages/api/endpoints/rules.js
botanical: Hmm good point. Looks like you are correct [`c4ffbec` (#2012)](https://github.com/nasa/cumulus/pull/2012/commits/c4ffbec1dd0be748fc8cf8163bf7e91d8a2bd357)

packages/api/tests/models/rules/test-rules-model.js
```diff
@@ -196,6 +196,19 @@ test.serial('Creating an invalid rule does not create workflow triggers', async
   }
 });
 
+test('Creating a rule with a createdAt and updatedAt date assigns the correct values to createdAt and provides a new updatedAt', async (t) => {
+  const newRule = fakeRuleFactoryV2();
+  const currentDate = new Date('1/2/2030').getTime();
```
markdboyd, line 201: No real objection, but why use this future date?

packages/api/tests/models/rules/test-rules-model.js
```diff
@@ -196,6 +196,19 @@ test.serial('Creating an invalid rule does not create workflow triggers', async
...
+  newRule.workflow = workflow;
+
+  delete newRule.collection;
+  delete newRule.provider;
+
+  const rule = await rulesModel.create(newRule, currentDate, currentDate);
+  t.is(rule.createdAt, currentDate);
+  t.not(rule.updatedAt, currentDate);
```
markdboyd, line 209: Shouldn't the `updatedAt` value also be set to the `currentDate`? Unless I'm misunderstanding

packages/api/tests/models/rules/test-rules-model.js
```diff
@@ -196,6 +196,19 @@ test.serial('Creating an invalid rule does not create workflow triggers', async
   }
 });
 
+test('Creating a rule with a createdAt and updatedAt date assigns the correct values to createdAt and provides a new updatedAt', async (t) => {
+  const newRule = fakeRuleFactoryV2();
+  const currentDate = new Date('1/2/2030').getTime();
+  newRule.workflow = workflow;
+
+  delete newRule.collection;
+  delete newRule.provider;
```
markdboyd, line 205: I don't know if this is necessary in a test for just the Dynamo `create`?

packages/api/tests/models/rules/test-rules-model.js
botanical: No, it gets updated to a more recent date later down the code chain: https://github.com/nasa/cumulus/blob/feature-rds/packages/api/models/base.js#L298

packages/api/tests/models/rules/test-rules-model.js
botanical: Sorry, do you mind elaborating? What do you think is unnecessary? Deleting collection and provider?

packages/api/tests/models/rules/test-rules-model.js
botanical: For fun 😮 If that's not a sufficient reason, I can change it do a date you think is more reasonable.

packages/api/tests/models/rules/test-rules-model.js
markdboyd: Ah right. Though I wonder in that case if there's a point in doing tihs?

https://github.com/nasa/cumulus/pull/2012/files#diff-2a37aa7b44507f6480bc0edc75d3aa7e70031468d98c108f8aef958a2098ef91R246

packages/api/tests/models/rules/test-rules-model.js
markdboyd: Deleting the collection/provider is something you're doing in the tests that write to RDS because in RDS a collection/provider is required but you don't want to always pre-create those records in RDS. In Dynamo, collection/provider aren't required per se, so we don't need to delete those attributes of the rule before trying to save. Unless I'm missing something

packages/api/tests/models/rules/test-rules-model.js
botanical: Hm that is a good point. I suppose it doesn't appear to be a point then, because we can just pass `Date.now()` and then it will get updated later anyway.

packages/api/tests/models/rules/test-rules-model.js
botanical: Oh I see. Yes you are totally right, it's not required in the dynamo schema.

CHANGELOG.md
```diff
@@ -54,10 +54,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
     - `@cumulus/message/workflows/getMessageWorkflowStopTime`
     - `@cumulus/message/workflows/getMessageWorkflowName`
+  - **CUMULUS-2199**
+    - Added `translateApiRuleToPostgresRule` to `@cumulus/db` to translate API Rule to conform to Postgres Rule definition.
 
 ### Changed
 
 - **CUMULUS-2185** - RDS Migration Epic
+  - **CUMULUS-2199**
+    - Updated Rules endpoint to write to and delete from RDS if a write to and delete from DynamoDB is successful.
```
markdboyd, line 64: The logic is actually the opposite: the Dynamo write/delete is only attempted if the RDS write/delete is sucessful.
However I don't think the CHANGELOG needs to spell out the exact implementation logic, so I think we can simplify to

```suggestion
    - Updated Rules endpoint to write rules to core PostgreSQL database in addition to DynamoDB and to delete rules from the PostgreSQL database in addition to DynamoDB
```

markdboyd (approved):

PR #2019, opened 2020-12-12
Add retry to ensure localstack is up and running.

flamingbear:
when running the @cumulus/api locally, wrap checkOrCreateTables() in a pRetry so we can start the api at the same time as the Localstack.  This code is only used for our local testing and this change allows us to start both the localstack and api containers at the same time, rather than ensuring the localstack is up and running before starting the api.

Addresses [CUMULUS-2280: migrate to bamboo](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2280)

## Changes

* add retry to ensure localstack is running

## PR Checklist

- [ x ] Update CHANGELOG
- [na ] Unit tests
- [ x ] Adhoc testing
- [ na ] Integration tests


nemreid (approved):

nemreid:
LGTM 👍

PR #2022, opened 2020-12-16
Update async-operation to use 4GB memory.

flamingbear:
This is the change to the Dockerfile, including the parameter being passed to the
node command to increase the heap size.  The terraform change reflects a docker
image already pushed to dockerhub.

**Summary:** Increase memory in node command for async-operation

Addresses [CUMULUS-1: bugfix](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1)

## Changes

* Update Dockerfile to use extra memory
* Update Terraform to use new Dockerhub image (already built and deployed)
* Log size of report before writing.

## PR Checklist

- [x] Update CHANGELOG
- [ NA ] Unit tests
- [ X ] Adhoc testing
- [ NA ] Integration tests


tf-modules/archive/async_operation.tf
```diff
@@ -19,7 +19,7 @@ resource "aws_ecs_task_definition" "async_operation" {
         "value": "${data.aws_region.current.name}"
       }
     ],
-    "image": "cumuluss/async-operation:27",
+    "image": "cumuluss/async-operation:33",
     "memoryReservation": 700,
```
nemreid, line 23: I'm having trouble figuring out if we allow our containers to hit 4GB memory, as this `memoryReservation` entry is the closest thing I can find to a memory limit (although this one is a soft limit). Any idea?

nemreid (approved):
Tentatively approved, but I do have a question..

tf-modules/archive/async_operation.tf
flamingbear: Yup, that's what I found. I assume the memory limit is determined by the EC2 cluster instances that we run the tasks on. That's t3.medium for testing, which has 4GB memory.  So there is a possibility of running out of memory, but I think the update to the image just allows node to expand to 4, not just use 4.  We are going to fix the problem at the root for the reconciliation reports.  So this is kind of a bandaid. Mark was not super happy with this, and maybe we should touch base in an arch meeting to make sure we have the tickets we need and if the Dockerfile change should be merged to master or not.

flamingbear:
This seems like a work around to bad code.  Rather than merge it in, we will close this PR and fix it properly with CUMULUS-2321

PR #2029, opened 2020-12-18
Feature/cumulus 2188 executions migration

npauzenga:
**Summary:** 

Addresses [CUMULUS-2188: Write Dynamo -> RDS migration for Executions](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2188)

## Changes

This creates the `data-migration2` Lambda and implements Execution migrations.

The Lambda function itself is very similar to `data-migration1`. `migrateExecutions` and `migrateExecutionRecord` follow the pattern in `data-migration1`. There is added logic for migrating parent executions in the Lambda and things get a bit different in `db/src/executions.ts`. There are a few additional lookups there for foreign keys.

The other main difference is the infrastructure. This Lambda is intended to be run in ECS and the terraform templates here should deploy it as such.

This is ready for review. I am looking at moving two variables in the terraform templates that we can probably get from the cumulus deployment. Inline notes below.

TODO:

1. Update tfvars to pull from cumulus deploy
2. Update branch and use new PG models where appropriate

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


example/data-migration-ecs/main.tf
```diff
@@ -0,0 +1,30 @@
...
+}
+
+module "data_migration_ecs_service" {
+  source = "https://github.com/nasa/cumulus/releases/download/v4.0.0/terraform-aws-cumulus-ecs-service.zip"
+
+  prefix = var.prefix
+  name   = "ExecutionMigrationService"
+
+  log2elasticsearch_lambda_function_arn = var.log2elasticsearch_lambda_function_arn
+  cluster_arn                           = var.ecs_cluster_arn
```
npauzenga, line 21: These can probably be taken from the cumulus remote state.

example/data-migration-ecs/terraform.tfvars.example
```diff
@@ -0,0 +1,8 @@
+permissions_boundary_arn = "arn:aws:iam::12345:policy/NGAPShRoleBoundary"
+prefix = "PREFIX"
+subnet_ids = ["subnet-12345"]
+vpc_id = "vpc-12345"
+
+data_migration2_function_arn = "arn:aws:lambda:us-east-1:<12345>:function:PREFIX-data-migration2"
+log2elasticsearch_lambda_function_arn = "arn:aws:lambda:us-east-1:<12345>:function:PREFIX-log2elasticsearch"
+ecs_cluster_arn = "arn:aws:ecs:us-east-1:<12345>:cluster/PREFIX-tf-CumulusECSCluster"
```
npauzenga, line 8: Will probably remove. See above.

lambdas/data-migration2/src/executions.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  const updatedRecord = await translateApiExecutionToPostgresExecution(
+    <ExecutionRecord>dynamoRecord, <Knex>knex
+  );
+
+  if (updatedRecord.parent_cumulus_id !== undefined) {
+    // Get parent record
+    // Migrate parent record
+  }
```
npauzenga, line 50: I originally had this logic written in db/src/executions.ts but I don't think that's right. Workin on moving it here.

lambdas/data-migration2/src/executions.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  await knex('executions').insert(updatedRecord);
+};
+
+export const migrateExecutions = async (
```
npauzenga, line 55: This is untested. Next on the list.

packages/db/src/executions.ts
```diff
@@ -0,0 +1,57 @@
...
+  };
+
+  // If we have a parentArn, try a lookup in Postgres. If there's a match, set the parent_cumulus_id
+  if (dynamoRecord.parentArn !== undefined) {
+    const parentId = await getParentExecutionCumulusId(dynamoRecord.parentArn, knex);
+
+    if (parentId !== undefined) {
+      translatedRecord.parent_cumulus_id = parentId;
+    }
+  }
```
npauzenga, line 48: Here we get the parent execution's cumulus_id but not the execution itself. That'll happen in the Lambda.

lambdas/data-migration2/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "@cumulus/errors": "3.0.0",
+    "@cumulus/logger": "3.0.0",
+    "@cumulus/types": "3.0.0",
+    "knex": "0.21.13",
+    "lodash": "^4.17.20",
+    "pg": "^8.3.0"
+  },
+  "devDependencies": {
+    "@cumulus/test-data": "3.0.0",
+    "@cumulus/data-migration1": "3.0.0"
```
npauzenga, line 41: I'm using `data-migration1` in tests. I need to be able to migrate Collections and Async Operations to test fk relationships with executions. Happy to move these somewhere more common if there's a better place. The requires are in test-executions.js:

```
const { migrateAsyncOperationRecord } = require('@cumulus/data-migration1/dist/lambda/async-operations');
const { migrateCollectionRecord } = require('@cumulus/data-migration1/dist/lambda/collections');
```

packages/db/src/executions.ts
```diff
@@ -0,0 +1,71 @@
+import Knex from 'knex';
+
+import { ExecutionRecord } from '@cumulus/types/api/executions';
+import { PostgresExecution } from './types';
+// TODO move to common location
```
npauzenga, line 5: I'm thinking these helpers should be on the new models. Does it seem reasonable to pull in the Postgres `AsyncOperation`, `Collection`, and `Execution` models here? Alternatively some common utils or shared space would do.

packages/db/src/executions.ts
```diff
@@ -0,0 +1,71 @@
...
+    final_payload: JSON.stringify(dynamoRecord.finalPayload),
+    workflow_name: dynamoRecord.name,
+  };
+
+  if (dynamoRecord.timestamp) {
+    translatedRecord.timestamp = new Date(Number(dynamoRecord.timestamp));
+  }
+
+  if (dynamoRecord.collectionId) {
+    // TODO is there a helper for this?
```
npauzenga, line 47: I didn't see anything that translates from the `name___version` format of a collection ID to either an array of those values or a cumulus_id directly. Again, this might be a good small thing to add to the new Collection model unless it exists somewhere already and I'm not seeing it?

packages/db/src/types.ts
```diff
@@ -41,12 +42,27 @@ export interface PostgresCollectionRecord extends PostgresCollection {
...
   arn: string,
   async_operation_cumulus_id?: number,
   collection_cumulus_id?: number,
   parent_cumulus_id?: number,
-  cumulus_version: string,
+  cumulus_version?: string,
+  url?: string,
+  status?: ExecutionRecordStatus,
+  tasks?: Object, // TODO need specific type?
+  error?: Object, // TODO need specific type?
```
npauzenga, line 54: It looks like these are just JSON objects. Should I have different types for them though? Are there keys that these always have?

packages/db/tests/test-executions.js
```diff
@@ -0,0 +1,53 @@
...
+    originalPayload: { testInput: 'originalPayloadValue' },
+    finalPayload: { testOutput: 'finalPayloadValue' },
+    duration: 2,
+  };
+
+  // Note that we are not testing the foreign keys here. Properties like parent_cumulus_id,
+  // collection_cumulus_id, and async_operation_cumulus_id are set in
+  // translateApiExecutionToPostgresExecution using helpers that are tested elsewhere.
+  // The complete execution, including those foreign keys is
+  // tested in the data-migration2 integration tests.
```
npauzenga, line 30: 👋. Open to testing the foreign keys here. I didn't because it requires bringing in and setting up Knex to actually test if the cumulus_ids are picked up correctly. It also requires things like the DynamoDB models for Collections and Async Operations to be imported here. We're doing this in the Lambda's `test-executions.js`.

If it bothers anyone that there are properties not tested in this file, I can definitely look at how to mock some of the cumulus_id stuff.

packages/db/src/executions.ts
```diff
@@ -0,0 +1,71 @@
...
+    ),
+    status: dynamoRecord.status,
+    tasks: JSON.stringify(dynamoRecord.tasks),
+    error: JSON.stringify(dynamoRecord.error),
+    arn: dynamoRecord.arn,
+    duration: dynamoRecord.duration,
+    original_payload: JSON.stringify(dynamoRecord.originalPayload),
+    final_payload: JSON.stringify(dynamoRecord.finalPayload),
+    workflow_name: dynamoRecord.name,
+  };
```
npauzenga, line 40: I don't have a migration for `cumulus_version` here. It looks like we don't have that field in Dynamo and, if I remember correctly, this field will just be `null` if the record was created outside of PG (and migrated using this logic). Is that correct?

packages/db/src/translate/executions.ts
```diff
@@ -0,0 +1,71 @@
+import Knex from 'knex';
+
+import { ExecutionRecord } from '@cumulus/types/api/executions';
+import { PostgresExecution } from '../types';
+// TODO move to common location
+const {
+  getParentExecutionCumulusId,
+  getAsyncOperationCumulusId,
+  getCollectionCumulusId,
+} = require('../../api/lambdas/sf-event-sqs-to-db-records/utils');
```
npauzenga, line 10: My hope is that these can come from the new PG models for Executions, Async Operations, and Collections. Looking at that now as part of merging those changes.

packages/db/src/translate/executions.ts
```diff
@@ -0,0 +1,71 @@
...
+    workflow_name: dynamoRecord.name,
+  };
+
+  if (dynamoRecord.timestamp) {
+    translatedRecord.timestamp = new Date(Number(dynamoRecord.timestamp));
+  }
+
+  if (dynamoRecord.collectionId) {
+    // TODO is there a helper for this?
+    const collectionNameVersionArray = dynamoRecord.collectionId.split('___');
```
npauzenga, line 48: I didn't see one that goes from the `collection___version` string, back to an array or back to a cumulus_id. Again, this will likely end up on the new PG model for Collections I'd think. Looking at it now as part of the merge of the updated rds branch.

packages/db/src/types/execution.ts
```diff
@@ -1,10 +1,20 @@
...
   arn: string,
-  status: string,
   async_operation_cumulus_id?: number,
   collection_cumulus_id?: number,
   parent_cumulus_id?: number,
   cumulus_version?: string,
+  url?: string,
+  status?: ExecutionRecordStatus,
+  tasks?: Object, // TODO need specific type?
+  error?: Object, // TODO need specific type?
```
npauzenga, line 12: Do/should we have specific types for these? Are there keys we know will always be there?

example/data-migration-ecs/terraform.tf.example
```diff
@@ -0,0 +1,8 @@
+terraform {
+  backend "s3" {
+    region         = "us-east-1"
+    bucket         = "PREFIX-state"
+    key            = "PREFIX/data-migration-ecs/terraform.tfstate"
+    dynamodb_table = "PREFIX-tf-locks"
```
botanical, line 6: This is a nit but should this be `PREFIX-locks`?

packages/db/src/types/execution.ts
```diff
@@ -13,4 +23,4 @@ export interface PostgresExecutionRecord extends PostgresExecution {
   cumulus_id: number,
   created_at: Date,
   updated_at: Date,
-}
+}
```
botanical, line 26: Newline

packages/db/src/types/execution.ts
botanical: I could be wrong, but taking a look at the [model](https://github.com/nasa/cumulus/blob/master/packages/api/models/executions.js#L52-L53), there doesn't seem to be specific types for these.

packages/db/src/translate/executions.ts
```diff
@@ -0,0 +1,71 @@
...
+      ) : undefined
+    ),
+    status: dynamoRecord.status,
+    tasks: JSON.stringify(dynamoRecord.tasks),
+    error: JSON.stringify(dynamoRecord.error),
+    arn: dynamoRecord.arn,
+    duration: dynamoRecord.duration,
+    original_payload: JSON.stringify(dynamoRecord.originalPayload),
+    final_payload: JSON.stringify(dynamoRecord.finalPayload),
+    workflow_name: dynamoRecord.name,
```
botanical, line 39: I could be wrong but I think `url` is missing
```suggestion
    workflow_name: dynamoRecord.name,
    url: dynamoRecord.execution,
```

example/data-migration-ecs/variables.tf
```diff
@@ -0,0 +1,38 @@
...
+
+variable "log2elasticsearch_lambda_function_arn" {
+  type    = string
+  default = null
+}
+
+variable "data_migration2_function_arn" {
+  type    = string
+  default = null
+}
```
botanical, line 38: newline

example/data-migration2-tf/main.tf
```diff
@@ -0,0 +1,42 @@
...
+  vpc_id            = var.vpc_id
+  lambda_subnet_ids = var.subnet_ids
+
+  dynamo_tables = data.terraform_remote_state.data_persistence.outputs.dynamo_tables
+
+  rds_security_group_id = data.terraform_remote_state.data_persistence.outputs.rds_security_group
+  rds_user_access_secret_arn = data.terraform_remote_state.data_persistence.outputs.database_credentials_secret_arn
+
+  tags = merge(var.tags, { Deployment = var.prefix })
+}
```
botanical, line 42: newline

example/data-migration2-tf/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "data_migration2_function_arn" {
+  value = module.data_migration2.data_migration2_function_arn
+}
```
botanical, line 3: newline

packages/db/src/types/execution.ts
npauzenga: Ok, yeah that makes sense. It didn't look like we had types for these. Just wasn't sure if I missed 'em. Thanks for the sanity check 😄

example/data-migration-ecs/terraform.tf.example
npauzenga: Yeah, totally agreed that it would probably be clearer as `PREFIX-locks`. I left it as `PREFIX-tf-locks` because it looks like that's the format we're using in our other deploys. `cumulus-tf/terraform.tf.example` for example:

`dynamodb_table = "PREFIX-tf-locks"`

I think if you were to make your prefix `jt-tf` you'd end up with `jt-tf-tf-locks`. I wouldn't think that would matter. It would just look a little wonky.

packages/db/src/types/execution.ts
npauzenga: Ah yup! So that reminds me of a workflow question I had: how/when do you lint? Or actually, what is your pre-commit routine like? Do you run `npm run eslint`, address errors/warnings, and then run `npm test` to run the full test suite? Or do you just use VSCode's linting as you go? 

I missed this (and another lint error that I just fixed/pushed) because I forgot to lint the full project before pushing. I'm wondering if it would be valuable to have the lint step as part of `npm run test` or if I just need to adjust my routine 😄

example/data-migration-ecs/variables.tf
npauzenga: Yeah, it looks like my linter isn't evaluating .tf files. Maybe we should fix that so this throws an error/warning?

example/data-migration2-tf/main.tf
npauzenga: Same as above. Would be cool to have a warning in the console. Maybe I just have a config issue for .tf files.

example/data-migration2-tf/outputs.tf
npauzenga: Yup, same as above.

packages/db/src/translate/executions.ts
npauzenga: Oh! Awesome catch! 

So I missed this because I didn't realize the old Dynamo `execution` field mapped to `url` in PG. I thought `url` was new to the new model (an assumption I didn't realize I had until right now).

Is there a better place to see this kind of thing? What would be nice is sample data or something close to it to verify all of my translations here. What I have been using to verify what the fields are:

- The types/api/executions.d.ts Dynamo model
- The pre-existing execution test data in `fakeExecutionFactoryV2` 
- The schema.svg for the new PG schema

(updating this mapping to your suggested change and updating tests)

packages/db/src/executions.ts
npauzenga: Duplicate comment

packages/db/src/executions.ts
npauzenga: duplicate comment

packages/db/src/executions.ts
npauzenga: I think this actually is on Dynamo as `cumulusVersion: { type: 'string' },`. Found it in schemas.js. Previously I was looking in these two places for the dynamo fields:

- The types/api/executions.d.ts Dynamo model
- The pre-existing execution test data in fakeExecutionFactoryV2

The schemas.js has this additional field that I will add to the translation here. I'll also double-check the others as I wasn't using schema.js.

packages/db/src/translate/executions.ts
botanical: Yeah, I think that is pretty easy to miss! I know I've done that before. 😅 

What I have done when working on similar tickets is having the schema definition for the DynamoDB table and PG table open side by side and creating a mental mapping of them. I also go the extra step to write them down in my notebook to make sure that no field goes unaccounted for (but that's kind of extra and not necessary).

Then I go look for records in DynamoDB to check my understanding and see what actual test records look like. 

For the case of reviewing your PR, I didn't write anything down but just compared the schema.

Not sure if that's helpful but hopefully gives you insight onto how I go about it!

packages/db/src/executions.ts
npauzenga: Corrected in the latest commit. I also double-checked the other fields against shema.js and updated `workflow_name` to set its value from `type` rather than `name`.  The schema for `type` is:
```
type: {
      title: 'The workflow name, e.g. IngestGranule',
      type: 'string',
    },
```

packages/db/src/models/async_operation.ts
```diff
@@ -0,0 +1,15 @@
...
+// eslint-disable-next-line max-len
+class AsyncOperationPgModel extends BasePgModel<PostgresAsyncOperation, PostgresAsyncOperationRecord> {
+  constructor() {
+    super({
+      tableName: tableNames.asyncOperations,
+    });
+  }
+}
+
+export { AsyncOperationPgModel };
```
npauzenga, line 15: This model is actually added in another PR as well:

https://github.com/nasa/cumulus/pull/2033/files#diff-9ea14a7792723723e953031b1d1b7bbc4341c2c8ef86d45ccd7af2e25dba62e2

Adding it here so I can use the base model's `getRecordCumulusId` for async operations.

packages/db/src/translate/executions.ts
npauzenga: On second thought, we shouldn't need this after the migration. I'm just going to leave it as-is rather than put something on the PG model that's Dynamo-specific or further extend the deprecated Dynamo model.

botanical: I just found this `deconstructCollectionId` which is what you may have been looking for:

https://github.com/nasa/cumulus/blob/master/packages/api/lib/utils.js#L34-L40

```
function deconstructCollectionId(collectionId) {
  const [name, version] = collectionId.split('___');
  return {
    name,
    version,
  };
}
```

npauzenga: Yeah, this is exactly what I was looking for. Thanks for finding it, sorry to miss it!

I guess I'm a little hesitant to use it though. Looking at the @cumulus/db package where my code in question lives, we're not currently importing @cumulus/api (where this util lives). I'm a little reluctant to pull in the whole api package  for just this string parsing so I think that might mean moving the `deconstructCollectionId` util to @cumulus/common or something like that. At that point is it worth it for one line that is only used for decoding the deprecated Dynamo field? I guess I'm leaning towards leaving the one line as-is even though it's not very DRY. I also don't feel super strongly so feel free to disagree 😄

lambdas/data-migration2/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@cumulus/data-migration2",
+  "version": "3.0.0",
+  "description": "A Lambda function used for doing data migrations of executions",
```
markdboyd, line 4: Nit: ultimately it won't just be for migrating executions, but also granules/files.

lambdas/data-migration2/src/errors.ts
```diff
@@ -0,0 +1,3 @@
+import { createErrorType } from '@cumulus/errors';
+
+export const RecordAlreadyMigrated = createErrorType('RecordAlreadyMigrated');
```
markdboyd, line 3: I should have just done this to begin with, but since we're using this same error type in data-migration1 and data-migration2, let's move this to being exported from the `errors` package and just use it in both lambdas

lambdas/data-migration2/src/executions.ts
```diff
@@ -0,0 +1,108 @@
...
+      arn: dynamoRecord.arn,
+    })
+    .first();
+  // Throw error if it was already migrated.
+  if (existingRecord) {
+    throw new RecordAlreadyMigrated(`Execution arn ${dynamoRecord.arn} was already migrated, skipping`);
+  }
+
+  const updatedRecord = await translateApiExecutionToPostgresExecution(
+    <ExecutionRecord>dynamoRecord, <Knex>knex
```
markdboyd, line 45: I'm interested why the `<Knex>knex` is necessary? If the type of the variable is already set to Knex in the arguments on line 29?

Similarly, can we set the type of `dynamoRecord` to `ExecutionRecord` in the function signature or does that not work?

lambdas/data-migration2/src/executions.ts
```diff
@@ -0,0 +1,108 @@
...
+    const executionModel = new Execution();
+    const parentExecution = await executionModel.get({ arn: dynamoRecord.parentArn });
+
+    // Migrate parent dynamo record to Postgres and assign parent's cumulus_id to child
+    updatedRecord.parent_cumulus_id = await migrateExecutionRecord(parentExecution, knex);
+  }
+
+  const cumulusId = await knex('executions').insert(updatedRecord).returning('cumulus_id');
+
+  return cumulusId[0];
```
markdboyd, line 61:
```suggestion
  const [cumulusId] = await knex('executions').insert(updatedRecord).returning('cumulus_id');
  return cumulusId;
```

lambdas/data-migration2/tests/test-executions.js
```diff
@@ -0,0 +1,456 @@
...
+
+  t.context.knex = await getKnexClient({
+    env: {
+      ...localStackConnectionEnv,
+      PG_DATABASE: testDbName,
+      migrationDir,
+    },
+  });
+
+  await t.context.knex.migrate.latest();
```
markdboyd, line 113: I might consider using this helper which would let you avoid a lot of this boilerplate:

https://github.com/nasa/cumulus/blob/feature-rds/packages/db/src/test-utils.ts#L21

lambdas/data-migration2/tests/test-executions.js
```diff
@@ -0,0 +1,456 @@
...
+});
+
+test.afterEach.always(async (t) => {
+  await t.context.knex('executions').del();
+  await t.context.knex('collections').del();
+  await t.context.knex('async_operations').del();
```
markdboyd, line 119: Same here, consider using this helper: https://github.com/nasa/cumulus/blob/feature-rds/packages/db/src/test-utils.ts#L38

lambdas/data-migration2/tests/test-executions.js
```diff
@@ -0,0 +1,456 @@
...
+
+test.after.always(async (t) => {
+  await executionsModel.deleteTable();
+  await asyncOperationsModel.deleteTable();
+  await collectionsModel.deleteTable();
+  await rulesModel.deleteTable();
+  await recursivelyDeleteS3Bucket(process.env.system_bucket);
+  await t.context.knex.destroy();
+  await t.context.knexAdmin.raw(`drop database if exists "${testDbName}"`);
+  await t.context.knexAdmin.destroy();
```
markdboyd, line 130: Same here, consider using this helper: https://github.com/nasa/cumulus/blob/feature-rds/packages/db/src/test-utils.ts#L38

lambdas/data-migration2/tests/test-executions.js
markdboyd, line 119: What re-generates these tables so that they exist per test?

lambdas/data-migration2/tests/test-executions.js
```diff
@@ -0,0 +1,456 @@
...
+
+  t.teardown(() => Promise.all([
+    executionsModel.delete({ arn: fakeExecution.arn }),
+    collectionsModel.delete(fakeCollection),
+    asyncOperationsModel.delete({ id: fakeAsyncOperation.id }),
+  ]));
+
+  // migrate existing async operation and collection
+  await migrateAsyncOperationRecord(existingAsyncOperation, t.context.knex);
+  await migrateCollectionRecord(existingCollection, t.context.knex);
```
markdboyd, line 160: I could be wrong, but I'm not sure that we need to create the async operation or collection in Dynamo for the purposes of testing this Lambda. We only need them to exist in Postgres so they can be referenced by the execution via foreign keys, correct?

Separately, I would like to avoid using `migrateAsyncOperationRecord` and `migrateCollectionRecord` in the tests here. The reason why is that now the tests for data-migration2 are dependent on the code for data-migration1, so bugs in data-migration1 code could cause both tests to fail. Instead, I'd probably opt to just create any objects that you need to reference via FKs directly via the models and using these helpers:

https://github.com/nasa/cumulus/blob/feature-rds/packages/db/src/test-utils.ts#L50

Thoughts?

lambdas/data-migration2/tests/test-executions.js
```diff
@@ -0,0 +1,456 @@
...
+  // // remove nullable fields
+  delete newExecution.asyncOperationId;
+  delete newExecution.collectionId;
+  delete newExecution.tasks;
+  delete newExecution.error;
+  delete newExecution.duration;
+  delete newExecution.originalPayload;
+  delete newExecution.finalPayload;
+  delete newExecution.timestamp;
+  delete newExecution.parentArn;
```
markdboyd, line 215: I _think_ we need to add `type` and `cumulusVersion` to these options

lambdas/data-migration2/tests/test-executions.js
```diff
@@ -0,0 +1,456 @@
...
+  const newExecution = fakeExecutionFactoryV2({ parentArn: undefined });
+
+  await migrateExecutionRecord(newExecution, t.context.knex);
+  await t.throwsAsync(
+    migrateExecutionRecord(newExecution, t.context.knex),
+    { instanceOf: RecordAlreadyMigrated }
+  );
+});
+
+test.serial('migrateExecutionRecord skips already migrated record', async (t) => {
```
markdboyd, line 245:
```suggestion
test.serial('migrateExecutions skips already migrated record', async (t) => {
```

packages/db/src/translate/executions.ts
markdboyd: I wouldn't want to move it to `common` because we're trying to keep that package really slim, since it used to be a dumping ground for everything. But we could move it somewhere like this:

https://github.com/nasa/cumulus/blob/feature-rds/packages/message/src/Collections.ts

You'd still be bringing in that whole package just to use that helper though, which is unfortunate. So I could go either way

packages/db/src/translate/executions.ts
```diff
@@ -0,0 +1,88 @@
...
+ *   Knex client
+ * @returns {PostgresExecutionRecord} - converted Execution
+ */
+export const translateApiExecutionToPostgresExecution = async (
+  dynamoRecord: ExecutionRecord,
+  knex: Knex
+): Promise<PostgresExecution> => {
+  const execution = new ExecutionPgModel();
+  const collection = new CollectionPgModel();
+  const asyncOperation = new AsyncOperationPgModel();
```
markdboyd, line 26: Nit: I'd prefer these be named with a suffix of `PgModel` for clarity

packages/db/src/translate/executions.ts
```diff
@@ -0,0 +1,88 @@
...
+    arn: dynamoRecord.arn,
+    duration: dynamoRecord.duration,
+    original_payload: JSON.stringify(dynamoRecord.originalPayload),
+    final_payload: JSON.stringify(dynamoRecord.finalPayload),
+    workflow_name: dynamoRecord.type,
+    url: dynamoRecord.execution,
+    cumulus_version: dynamoRecord.cumulusVersion,
+  };
+
+  if (dynamoRecord.timestamp) {
```
markdboyd, line 49: Nit: It occurs to me that you could handle all of these date fields with a ternary rather than `if` blocks

```
timestamp: dynamoRecord.timestamp ? new Date(Number(dynamoRecord.timestamp)) : undefined,
```

Also I'm curious about the use of the `Number` constructor here? Does this fail without it?

packages/db/src/types/execution.ts
```diff
@@ -1,10 +1,20 @@
...
+
 export interface PostgresExecution {
   arn: string,
-  status: string,
   async_operation_cumulus_id?: number,
   collection_cumulus_id?: number,
   parent_cumulus_id?: number,
   cumulus_version?: string,
+  url?: string,
+  status?: ExecutionRecordStatus,
```
markdboyd, line 10: `status` is not nullable in the PG schema, so it should have to be set even before the record was saved to the database. So it should be required in the interface, I believe

packages/db/tests/test-executions.js
```diff
@@ -0,0 +1,56 @@
+const test = require('ava');
```
markdboyd, line 1: Nit: can we move this test file to the `tests/translate` folder?

packages/db/src/translate/executions.ts
```diff
@@ -0,0 +1,88 @@
...
+  // Map old record to new schema.
+  const translatedRecord: PostgresExecution = {
+    async_operation_cumulus_id: (
+      dynamoRecord.asyncOperationId ? await asyncOperation.getRecordCumulusId(
+        knex,
+        { id: dynamoRecord.asyncOperationId }
+      ) : undefined
+    ),
+    status: dynamoRecord.status,
+    tasks: JSON.stringify(dynamoRecord.tasks),
```
markdboyd, line 38: I'm fairly sure that the `JSON.stringify` is unnecessary when saving objects to the PG database. For some reason it is necessary if you're ever saving an array of values, but that isn't the case for any of these fields

packages/db/tests/test-executions.js
markdboyd: FWIW, you could do a minor refactor of translateApiExecutionToPostgresExecution so that the PG models are optional arguments to the function. Then you could pass "fake" PG models to the tests that return cumulus IDs for the records, so you can test that this function properly assigns cumulus ID values to the record when they're received from the database. The actual logic fetching cumulus IDs from the database is already tested elsewhere, so we don't need to test it again here

For example

https://github.com/nasa/cumulus/blob/feature-rds/packages/db/src/translate/rules.ts#L20
https://github.com/nasa/cumulus/blob/feature-rds/packages/db/tests/translate/test-rules.js#L4

lambdas/data-migration2/src/executions.ts
npauzenga: Yup, that's neater. Or how about:

```
const cumulusId = await knex('executions').insert(updatedRecord).returning('cumulus_id').first();

return cumulusId;
```

?

markdboyd: I don't think `.first()` does anything on an `insert` query. And you still have to destructure the column value out of the query result.

lambdas/data-migration2/tests/test-executions.js
npauzenga: Yeah, the `table.del()` should just clear the table of all rows and retain the table itself. The intention here is to start with an empty table on each test. the `test.after.always` will actually destroy the tables:

`await executionsModel.deleteTable();`

lambdas/data-migration2/tests/test-executions.js
npauzenga: Awesome catch. I just added those fields to the translation and forgot to add them here.

packages/db/src/translate/executions.ts
npauzenga: Yup, absolutely. Hopefully in the main app logic we can avoid naming things like this and just have stuff like `ExecutionModel`. Here though, the whole point is to handle both types so yeah, totally good with anything that clears confusion 😄

lambdas/data-migration2/src/executions.ts
npauzenga: Nope. Edit: This was a lie. `first()` doesn't work on inserts.

packages/db/src/translate/executions.ts
npauzenga: Yup! Definitely. How about if I put all the date-related fields in the main `PostgresExecution` definition but leave the two foreign key fields in `if` blocks? Or I could move those to separate functions to be used in similar ternaries if ya like. My latest push shows it like I'm suggesting.

npauzenga: Oh and you're totally right. the `Number` casting can be removed. I had it there because the test execution I was using before discovering the `fakeExecutionFactoryV2` had that field as a string.

packages/db/src/translate/executions.ts
botanical: This is kind of thinking ahead... but as I'm implementing my work for the granules and files migration, I will need to deconstruct the `collectionId` of a granule to refer to a PG collection record and retrieve the `collection_cumulus_id`. So, we will at least know `deconstructCollectionId` will be used in one other place if you do decide to bring it in 🧐

packages/db/tests/test-executions.js
npauzenga: Oh right! You showed this pattern in a group review a while back. This is exactly the use case for passing mock models around.

I've updated this to:
```
export const translateApiExecutionToPostgresExecution = async (
  dynamoRecord: ExecutionRecord,
  knex: Knex,
  asyncOperationPgModel = new AsyncOperationPgModel(),
  executionPgModel = new ExecutionPgModel()
)
```
All of the Postgres models are instantiated in the function arguments, rather than the function body. I'm still creating an instance of `Logger` in the body because it's not really something we want to mock in the same way. The lesson I'm adopting here is something like "Always accept data models as arguments and set defaults to a new instance of that model for easier testing."

lambdas/data-migration2/src/errors.ts
npauzenga: Yup, makes sense! I moved the `RecordAlreadyMigrated` to the Errors package, removed errors.ts from both data-migration1 and data-migration2 Lambdas and updated the tests.

lambdas/data-migration2/src/executions.ts
npauzenga: Yeah, so the `<Knex>knex` isn't really necessary. I had it that way just because the other argument, `<ExecutionRecord>dynamoRecord` has that type casting and thought it looked more consistent with both arguments set up that way.

The `dynamoRecord` argument _does_ need to be cast as `ExecutionRecord` but we can definitely do it earlier than this call. My latest commit does what you're suggesting in updating the function signature to:

```
export const migrateExecutionRecord = async (
  dynamoRecord: ExecutionRecord,
  knex: Knex
): Promise<number> => { ... }
```
And removing the types from the call to the translate function:

```
  const updatedRecord = await translateApiExecutionToPostgresExecution(
    dynamoRecord, knex
  );
```

In order to make that work we _do_ need to cast `dynamoRecord` as an `ExecutionRecord` because it's being created as an AttributeMap in `migrateExecutions` in this line:

`let record = await searchQueue.peek();`

I think it makes sense to just move the cast to ExecutionRecord down to just after that happens on line 87:

```
try {
      await migrateExecutionRecord(<ExecutionRecord>record, knex);
      migrationSummary.success += 1;
    } catch (error) { ... }
```

tl;dr: We do need to cast the `dynamoRecord` as an `ExecutionRecord` but we can do it earlier and make the changes you're suggesting.

lambdas/data-migration2/tests/test-executions.js
npauzenga: Sold!

lambdas/data-migration2/tests/test-executions.js
npauzenga: Nice. Done.

packages/db/tests/test-executions.js
markdboyd: Yeah, I don't know if I would say **always**. I'm sure there are some people who would question this approach, but it seems like a sane way to allow for easy mocking to me

example/data-migration-ecs/main.tf
Jkovarik: 👍🏻  for /example use lookups where possible.

example/data-migration2-tf/main.tf
```diff
@@ -0,0 +1,42 @@
+terraform {
+  required_providers {
+    aws  = ">= 3.5.0"
+  }
+}
+
+provider "aws" {
+  region  = var.region
```
Jkovarik, line 8: What happens if this is set to `us-west-2`?

lambdas/data-migration2/tests/test-executions.js
npauzenga: I think that's totally right. I'm working on this refactor now + setting up a mock factory for async operations. Originally I was thinking we would need the dynamo records for Collections and Async Operations because we need to set a couple fk fields on the _dynamo_ execution:

```
  // Create new Dynamo execution to be migrated to postgres
  const newExecution = fakeExecutionFactoryV2({
    parentArn: existingExecution.arn,
    collectionId: `${fakeCollection.name}___${fakeCollection.version}`,
    asyncOperationId: fakeAsyncOperation.id,
  });
```
But all of those fields from collections and async operations are also present in the PG table so there's no reason to jump through these hoops and create/migrate Dynamo records.

Thanks for pointing this out. It'll be much cleaner.

npauzenga: Updated the PR. We're no longer creating collections/async operations in dynamo in this test. There's also a new `fakeAsyncOperationRecordFactory` to create the PG record for async operations.

packages/db/src/types/execution.ts
npauzenga: Oh yeah, I missed that. Updated to a required param.

packages/db/tests/test-executions.js
npauzenga: Ah yeah, absolutely.

packages/db/src/translate/executions.ts
npauzenga: Yeah, I believe that's totally right. You should be able to store a JS object in PG. I think it might convert the JS object to a JSON object down the line but either way.

The trick here is our types. I have `JSON.stringify` because some of these fields like `dynamoRecord.tasks` are of type `unknown` on the Dynamo model. I can't assign a field of type `unknown` to a field of type `Object | undefined` as is the case with `tasks` on the PG model, for example. I'm stringifying here as a way of setting the type.

I could maybe convert `dynamoRecord.tasks` (and the rest) to JS objects instead but I think that might add a step and yield the same result as JSON.stringify?

I could also think about changing the type assignment on the Dynamo model. 

Thoughts?

lambdas/data-migration2/package.json
npauzenga: Lies! This dependency will be removed after a conversation below about not creating these dynamo records in test.

Jkovarik (commented):
Realized I had a couple of comments in a draft state, posting a review to expose them.

packages/db/src/translate/executions.ts
markdboyd: So JSON.stringify on `unknown` makes TS recognize it as an `Object`? That is interesting. Regardless, this is a nit, so if it's working, then I just wouldn't bother changing it

example/data-migration-ecs/main.tf
```diff
@@ -0,0 +1,30 @@
+terraform {
```
Jkovarik, line 1: I may just not be up-to-date on the migration setup for our end users, but aren't we going to need to package these as release artifact for the release series where we're supporting dual writes?  

If so, shouldn't be in `tf-modules` and be a configurable module?

example/data-migration-ecs/terraform.tfvars.example
Jkovarik: I'm not sure you actually *do* want to remove this - see my comment re: structure.   

I'm guessing we want a configurable module that then is called by an example project that does the remote state lookups.

CHANGELOG.md
```diff
@@ -38,6 +38,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
   - Add documentation for Getting Started section including FAQs
 
 - **CUMULUS-2185** - RDS Migration Epic
+  - **CUMULUS-2188**
+    - Added `data-migration2` Lambda to be run after `data-migration1`
```
Jkovarik, line 42: This isn't specifically a changelog comment, but I think this needs some sort of README in `lambdas/data-migration2` and/or `example/data-migration/*` for devs prior to the user docs.

example/data-migration2-tf/terraform.tfvars.example
```diff
@@ -0,0 +1,11 @@
+permissions_boundary_arn = "arn:aws:iam::12345:policy/NGAPShRoleBoundary"
```
Jkovarik, line 1: See comments above re: deduplication

example/data-migration2-tf/main.tf
```diff
@@ -0,0 +1,42 @@
...
+}
+
+data "terraform_remote_state" "data_persistence" {
+  backend   = "s3"
+  config    = var.data_persistence_remote_state_config
+  workspace = terraform.workspace
+}
+
+module "data_migration2" {
+  source = "../../lambdas/data-migration2"
```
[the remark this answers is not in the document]
Jkovarik: (and I realize this is following the pattern we have for the data-migration1 deployment)

lambdas/data-migration1/src/async-operations.ts
```diff
@@ -9,8 +9,8 @@ import {
 } from '@cumulus/db';
 import { ApiAsyncOperation } from '@cumulus/types/api/async_operations';
 import Logger from '@cumulus/logger';
+import { RecordAlreadyMigrated } from '@cumulus/errors';
```
Jkovarik, line 12: 👍🏻

lambdas/data-migration2/main.tf
```diff
@@ -0,0 +1,106 @@
...
+
+resource "aws_lambda_function" "data_migration2" {
+  function_name    = "${var.prefix}-data-migration2"
+  filename         = local.lambda_path
+  source_code_hash = filebase64sha256(local.lambda_path)
+  handler          = "index.handler"
+  role             = aws_iam_role.data_migration2.arn
+  runtime          = "nodejs12.x"
+  timeout          = 300
+  memory_size      = 512
```
Jkovarik, line 84: We should consider doing an assessment on mem/performance costs vs disk reads.    If it's linear and the DB can handle it we should consider upping this or possibly making it tuneable.

lambdas/data-migration2/src/executions.ts
```diff
@@ -0,0 +1,108 @@
...
+  });
+
+  const migrationSummary = {
+    dynamoRecords: 0,
+    success: 0,
+    failed: 0,
+    skipped: 0,
+  };
+
+  let record = await searchQueue.peek();
```
Jkovarik, line 81: It may be apparent from later in the PR, but if the runtime env (ecs instance?) hangs here, we lose relevant information on how many records were moved/etc.     Does that matter to users?  Should we log it, or somehow otherwise note progress?

lambdas/data-migration2/src/index.ts
```diff
@@ -0,0 +1,28 @@
...
+export const handler = async (event: HandlerEvent): Promise<string> => {
+  const env = event.env ?? process.env;
+
+  const knex = await getKnexClient({ env });
+
+  try {
+    const executionsMigrationSummary = await migrateExecutions(env, knex);
+
+    return `
+      Migration summary:
```
Jkovarik, line 18: As above, on error we lose this summary - should we consider defensive behavior here and  make sure of logging/dumping/? on an uncaught error?

packages/db/src/types/execution.ts
```diff
@@ -1,10 +1,20 @@
+import { ExecutionRecordStatus } from '@cumulus/types/api/executions';
+
 export interface PostgresExecution {
   arn: string,
```
Jkovarik, line 4: Nit: consider alpha sorting these

Jkovarik (changes requested):
This all looks really good, nice work! 

Adding a review burst as there are some comments to talk about/clarify.   

Remaining: I still want to look closely at the units and deploy/test on a real db prior to approving.

lambdas/data-migration2/main.tf
```diff
@@ -0,0 +1,106 @@
...
+resource "aws_iam_role_policy" "data_migration2" {
+  name   = "${var.prefix}_data_migration2"
+  role   = aws_iam_role.data_migration2.id
+  policy = data.aws_iam_policy_document.data_migration2.json
+}
+
+resource "aws_security_group" "data_migration2" {
+  count = length(var.lambda_subnet_ids) == 0 ? 0 : 1
+
+  name   = "${var.prefix}-data-migration"
```
Jkovarik, line 63:
```suggestion
  name   = "${var.prefix}-data-migration2"
```

This causes a resource name collision if data-migration1 is deployed.

lambdas/data-migration2/main.tf
```diff
@@ -0,0 +1,106 @@
+locals {
+  lambda_path = "${path.module}/dist/webpack/lambda.zip"
+}
+
+data "aws_iam_policy_document" "lambda_assume_role_policy" {
```
Jkovarik, line 5: This concern sits somewhat squarely in between 'ease of use/what does this look as a user package' and 'nit': 

These roles duplicate a lot of the resources required for the migration, which only gets us a seperation-of-access benefit from having mig1 and mig2 have access to different tables.  

I'm wondering if that actually buys us anything over something like using the lambda processing mega-role that's being exported from Core, or at least making these two lambdas a combined tf-module.

CHANGELOG.md
Jkovarik: (and yes I realize it might be just "run the lambda")

lambdas/data-migration2/main.tf
```diff
@@ -0,0 +1,106 @@
...
+      "logs:CreateLogStream",
+      "logs:DescribeLogStreams",
+      "logs:PutLogEvents"
+    ]
+    resources = ["*"]
+  }
+
+  statement {
+    actions = [
+      "dynamodb:Scan",
```
Jkovarik, line 39: I needed to make the following update for the script to run as a lambda in the clear (e.g. not on ECS): 

```suggestion
      "dynamodb:Scan",
      "dynamodb:GetItem",
```

lambdas/data-migration2/main.tf
Jkovarik: Also if we don't want this lambda to run independently, it might be worth explicitly disabling this or lowering the timeout.

packages/db/src/types/execution.ts
npauzenga: 👍

packages/db/src/translate/executions.ts
npauzenga: So circling back to this 😄 

I've updated all of the Execution's usage of `Object` to `object` because `object` will actually give us the validation we want. `Object` will match any primitive object type in addition to object literals. 

The only other place in our PG types where this was an issue was on async operations. That was an easier case because, in the Dynamo to PG translation we weren't trying to go from `unknown` to `object`. In that case it was `string` to `object` so additional updates weren't required. 

Here's the commit if you wouldn't mind a sanity check:

https://github.com/nasa/cumulus/pull/2029/commits/bb393677d17f9a59cb263ede7041b5ed999b5c41

markdboyd: Your changes look good to me, though I question why we're using `uknown` instead of `object` on the Dynamo execution record type for those properties

I just tested this and our Dynamo schema validation prevents saving these fields with those properties using anything other than objects:

```
  const foo = {
    arn: 'fake-arn',
    name: 'name',
    error: 1234,
    tasks: 'not-an-object',
    originalPayload: 'not-an-object',
    finalPayload: 'not-an-object',
  };
  await executionModel.create(foo);
```

So I'd be inclined to update the Dynamo record type to replace `unknown -> object` and simplify this even further

lambdas/data-migration2/main.tf
markdboyd: I think the only reason for separating them was that migration1 can be run easily as just a Lambda given the expected data volume/Lambda runtime, whereas migration2 will have to be an ECS task given the expected time necessary to migrate all granules/files. But both of those migrations could be deployed from a general `data-migration` parent module, I think. 

I'd prefer that and having them share a Lambda role that is defined once than use a "mega role" from core. I think it's just better to only allocate exactly the permissions you need for what you're doing

lambdas/data-migration2/main.tf
markdboyd: The one for data-migration1 should have been named appropriately. Oversight on my part. Though if we're combining deployment of these somehow, this could be shared I suppose

CHANGELOG.md
markdboyd: There is a separate ticket, 2293, for writing the docs on how to run the data migrations

lambdas/data-migration2/main.tf
Jkovarik: Generically, I tend to agree, and think at some point we need to consider a more granular permission structure for the proc lambdas, despite the problems that presents.    As for if that drives user value....

Using a single module certainly reduces "AWS resource litter".

CHANGELOG.md
Jkovarik: I don't believe that addresses the concern - I'm not talking about a comprehensive user doc, more of a "this is how we're expecting to run it" 5-6 line cheat sheet.

packages/db/src/translate/executions.ts
npauzenga: Yup! We had a separate chat about this. To summarize:

I was thinking we shouldn't change the types in `types/api/executions` because those are in prod and presumably used for validation all over the place. Mark pointed out that the Dynamo schema (/api/models/schemas) is actually more strict with an "object" type already. It should be totally fine to make these match and change the Typescript type for the few "unknown" fields on executions.

lambdas/data-migration2/src/executions.ts
```diff
@@ -0,0 +1,108 @@
...
+  Manager.recordIsValid(dynamoRecord, schemas.execution);
+
+  const existingRecord = await knex<PostgresExecutionRecord>('executions')
+    .where({
+      arn: dynamoRecord.arn,
+    })
+    .first();
+  // Throw error if it was already migrated.
+  if (existingRecord) {
+    throw new RecordAlreadyMigrated(`Execution arn ${dynamoRecord.arn} was already migrated, skipping`);
```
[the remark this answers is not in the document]
npauzenga: Yeah, I think it's safe to assume that anything in-flight at the time this migration is run will be written to both databases and therefor either skipped here if the in-flight write comes in before we get there in the migration, or just written normally if not. Does that make sense? Am I thinking about this correctly?

This does touch on two things we'd discussed previously though. Let me make sure I'm on the right page:

1. Upserts. We'd talked about potentially doing upserts here. An upsert in this particular case sounds safer at the expense of time/cost. I don't have a great use case though. Something along the lines of... we have a record already written to dynamo and postgres but an updated version of that same record comes in this queue somehow?  I'm not sure if that's possible with dual writes.
2. Deep compare. Very similar discussion. We'd talked about comparing all properties of a record here as opposed to `WHERE arn = dynamoRecord.arn` as it is now. I think that would just lead to an upsert though.

packages/db/src/translate/executions.ts
```diff
@@ -0,0 +1,87 @@
...
+        { arn: dynamoRecord.parentArn }
+      );
+
+      if (parentId !== undefined) {
+        translatedRecord.parent_cumulus_id = parentId;
+      }
+    } catch (error) {
+      if (error instanceof RecordDoesNotExist) {
+        logger.info(error);
+      }
```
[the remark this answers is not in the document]
npauzenga: Ah yup, good catch. I was thinking `getRecordCumulusId`  would only throw `RecordDoesNotExist` based on its implementation but it could absolutely throw something else if something goes wrong with the PG query itself.

Updated this to throw any other error types but still log and swallow the `RecordDoesNotExist` because we need to keep going in this particular case if we get that error.

example/data-migration-ecs/main.tf
```diff
@@ -0,0 +1,30 @@
...
+  cluster_arn                           = var.ecs_cluster_arn
+  desired_count                         = 1
+  image                                 = "cumuluss/cumulus-ecs-task:1.7.0"
+
+  command = [
+    "cumulus-ecs-task",
+    "--lambdaArn",
+    var.data_migration2_function_arn
+  ]
+}
```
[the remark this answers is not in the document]
npauzenga: Ugh. Sorry. A bunch of these popped up in Jennifer's review too. I'm not set up to correctly lint *.tf or *.example files as it turns out.

example/data-migration2-tf/main.tf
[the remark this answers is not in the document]
npauzenga: Oh good points!

So re: separate deployments for the data-migration lambdas: it seems like it would simplify things for users if we didn't do that and just bundled 'em with example/cumulus-tf. I don't know if we'd want to spin up a container with every cumulus example deploy though so they would still need to deploy `data-migration-ecs` separately I think. But that's an easier lift.

I'm looking for downsides here. I guess these Lambdas are ephemeral and for current users only. If you're deploying cumulus for the first time post-RDS parallel writes, you won't use these data-migration Lambdas. But so long as we don't spin up a container, it shouldn't cost anything so 🤷.

I'm game to bundle these with example/cumulus-tf unless there's something I'm not seeing.

CHANGELOG.md
npauzenga: That totally makes sense. I added 2 READMEs, one for each of the data-migration Lambdas. They're pretty light, just answering the question "What the heck is this?" and giving some details on how it can be run.

lambdas/data-migration2/src/index.ts
npauzenga: TODO: use Logger to log migration summary before return.

lambdas/data-migration2/main.tf
npauzenga: To be addressed in another ticket. Jira # forthcoming.

lambdas/data-migration2/src/executions.ts
npauzenga: To be addressed in another ticket. Jira # forthcoming.

lambdas/data-migration2/src/index.ts
npauzenga: Hmm so actually we do log the migration summary on completion and any time there's a migration error. I'm wondering if that's sufficient knowing that there is work planned to develop additional reporting on the status here (or at least discussing that possibility)?

npauzenga: The relevant logging is here:

https://github.com/nasa/cumulus/pull/2029/files/63551946807ddb72565bcce1c46cbcfe7e7bff10#diff-0c998a701074d6b429e241827a83826f1590334993cafb70f1d57cf79b5b119dR86

The idea here is if there's a problem migrating a record we catch and log the error outside of the Lambda handler in `migrateExecutions`.

example/data-migration2-tf/main.tf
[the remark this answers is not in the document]
npauzenga: Resolved. Thanks for pairing on this!

lambdas/data-migration2/src/executions.ts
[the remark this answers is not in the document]
Jkovarik: "Yeah, I think it's safe to assume that anything in-flight at the time this migration is run will be written to both databases and therefor either skipped here if the in-flight write comes in before we get there in the migration, or just written normally if not. Does that make sense? Am I thinking about this correctly?"

It's not entirely safe -- long running executions (hour(s)) are possible.

packages/db/src/translate/executions.ts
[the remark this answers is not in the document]
Jkovarik: This looks to have been updated in  ab5855f232fc525cfd35ec6fddca4f353b8ad574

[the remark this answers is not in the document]
Jkovarik: 👍🏻

example/data-migration-ecs/main.tf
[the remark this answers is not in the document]
Jkovarik: No worries : 3984de33123e80bf010eb3bdc19f13e10e2e85c

example/data-migration-ecs/terraform.tfvars.example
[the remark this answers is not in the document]
Jkovarik: 3984de33123e80bf010eb3bdc19f13e10e2e85c

CHANGELOG.md
Jkovarik: :100:  Thanks  : 893d5be47f939d027d1567bda00dcde8bc08ff5e

lambdas/data-migration2/src/executions.ts
Jkovarik: Resolving as I know we talked about it in the arch meeting today and a ticket was generated.

Jkovarik (approved):
Thanks!  My concerns have been addressed, and we paired on the updates in: 

7a3b9f3
60da5f9
bb72623

Nice work!

PR #2033, opened 2020-12-22
Refactor sfEventSqsToDbRecords Lambda to use Postgres models

markdboyd:
#2026 should be reviewed first


Jkovarik (approved):
:100: this looks good, nice work!

PR #2034, opened 2020-12-23
CUMULUS-2186 - Write Dynamo -> RDS migration for PDRs

botanical:

Addresses [CUMULUS-2186: Write Dynamo -> RDS migration for PDRs](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2186)

This is a branch off of [this branch](https://github.com/nasa/cumulus/tree/feature/CUMULUS-2188-executions-migration). Depending on the ordering of reviewing and merging, I am open to merging these changes in `feature-rds` if the `feature/CUMULUS-2188-executions-migration` branch is merged first. 

## Changes

* Migrate PDR Records from DynamoDB to RDS logic

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


lambdas/data-migration2/src/pdrs.ts
```diff
@@ -0,0 +1,136 @@
...
+ * @throws {RecordAlreadyMigrated} if record was already migrated
+ */
+export const migratePdrRecord = async (
+  dynamoRecord: AWS.DynamoDB.DocumentClient.AttributeMap,
+  knex: Knex
+): Promise<void> => {
+  // Validate record before processing using API model schema
+  Manager.recordIsValid(dynamoRecord, schemas.pdr);
+
+  const existingRecord = await knex<PostgresPdrRecord>(tableNames.pdrs)
```
markdboyd, line 39: I'd like to replace this with the use of `pdrPgModel.get()` from the db package

lambdas/data-migration2/src/pdrs.ts
```diff
@@ -0,0 +1,136 @@
...
+    throw new RecordAlreadyMigrated(`PDR name ${dynamoRecord.pdrName} was already migrated, skipping.`);
+  }
+
+  const collectionCumulusId = await getRecordCumulusId<PostgresCollectionRecord>(
+    { name: dynamoRecord.collectionId },
+    tableNames.collections,
+    knex
+  );
+
+  const providerCumulusId = await getRecordCumulusId<PostgresProviderRecord>(
```
markdboyd, line 56: Likewise, I'd like to replace these with `collectionPgModel.getRecordCumulusId` and `providerPgModel.getRecordCumulusId` respectively

lambdas/data-migration2/src/pdrs.ts
```diff
@@ -0,0 +1,136 @@
...
+    { name: dynamoRecord.provider },
+    tableNames.providers,
+    knex
+  );
+
+  const executionCumulusId = dynamoRecord.execution
+    ? await getRecordCumulusId<PostgresExecutionRecord>(
```
markdboyd, line 63: Use `executionPgModel.getRecordCumulusId` instead

lambdas/data-migration2/src/pdrs.ts
```diff
@@ -0,0 +1,136 @@
...
+      knex
+    )
+    : undefined;
+
+  // Map old record to new schema.
+  const updatedRecord: PostgresPdr = {
+    name: dynamoRecord.pdrName,
+    provider_cumulus_id: providerCumulusId,
+    collection_cumulus_id: collectionCumulusId,
+    execution_cumulus_id: (executionCumulusId === undefined) ? undefined : executionCumulusId,
```
markdboyd, line 75: I think this conditional is unnecessary? If `executionCumulusId` is `undefined`, then `execution_cumulus_id === undefined` already

```suggestion
    execution_cumulus_id: executionCumulusId,
```

lambdas/data-migration2/src/pdrs.ts
```diff
@@ -0,0 +1,136 @@
...
+    stats: dynamoRecord.stats,
+    address: dynamoRecord.address,
+    original_url: dynamoRecord.originalUrl,
+    timestamp: dynamoRecord.timestamp ? new Date(dynamoRecord.timestamp) : undefined,
+    duration: dynamoRecord.duration,
+    created_at: new Date(dynamoRecord.createdAt),
+    updated_at: dynamoRecord.updated ? new Date(dynamoRecord.updatedAt) : undefined,
+  };
+
+  await knex(tableNames.pdrs).insert(updatedRecord);
```
markdboyd, line 89: We should be using the method for creating records on the PDR PG model. Better yet, let's use the `upsert` method, since it already exists and we decided that our data migrations should be doing upserts

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,336 @@
...
+
+  t.context.knex = await getKnexClient({
+    env: {
+      ...localStackConnectionEnv,
+      PG_DATABASE: testDbName,
+      migrationDir,
+    },
+  });
+
+  await t.context.knex.migrate.latest();
```
markdboyd, line 102: All of this boilerplate can be removed by using `generateLocalTestDb`. See https://github.com/nasa/cumulus/blob/feature-rds/packages/db/tests/models/test-granule-model.js#L20

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,336 @@
...
+test.after.always(async (t) => {
+  await recursivelyDeleteS3Bucket(process.env.system_bucket);
+
+  await providersModel.deleteTable();
+  await collectionsModel.deleteTable();
+  await pdrsModel.deleteTable();
+
+  await t.context.knex.destroy();
+  await t.context.knexAdmin.raw(`drop database if exists "${testDbName}"`);
+  await t.context.knexAdmin.destroy();
```
markdboyd, line 136: Use the `destroyLocalTestDb` helper here. See https://github.com/nasa/cumulus/blob/feature-rds/packages/db/tests/models/test-granule-model.js#L46

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,336 @@
...
+const { getKnexClient, localStackConnectionEnv } = require('@cumulus/db');
+const { fakeCollectionFactory, fakeProviderFactory } = require('@cumulus/api/lib/testUtils');
+
+// eslint-disable-next-line node/no-unpublished-require
+const { migrationDir } = require('../../db-migration');
+const { migratePdrRecord, migratePdrs } = require('../dist/lambda/pdrs');
+const { RecordAlreadyMigrated } = require('../dist/lambda/errors');
+
+const migrateFakeCollectionRecord = async (record, knex) => {
+  const updatedRecord = translateApiCollectionToPostgresCollection(record);
```
markdboyd, line 28: Rather than using a Dynamo fake collection and translating it to Postgres, we can just create a fake Collection PG record directly using `fakeCollectionRecordFactory` from the `db` package. It reduces the boilerplate needed for these tests and makes them a little better isolated

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,336 @@
...
+  const [id] = await knex(tableNames.collections).insert(updatedRecord).returning('cumulus_id');
+  return id;
+};
+
+const fakeEncryptFunction = async () => 'fakeEncryptedString';
+
+const migrateFakeProviderRecord = async (record, knex) => {
+  const updatedRecord = await translateApiProviderToPostgresProvider(record, fakeEncryptFunction);
```
markdboyd, line 36: Likewise, let's use the `fakeProviderRecordFactory` helper here

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,336 @@
...
+  const { knex, testCollection, testProvider, collectionCumulusId, providerCumulusId } = t.context;
+
+  const testPdr = generateTestPdr(testCollection.name, testProvider.id);
+  await migratePdrRecord(testPdr, knex);
+
+  const record = await t.context.knex.queryBuilder()
+    .select()
+    .table(tableNames.pdrs)
+    .where({ name: testPdr.pdrName })
+    .first();
```
markdboyd, line 149: Let's use `pdrPgModel.get()` here

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,336 @@
...
+
+  t.like(
+    omit(record, ['cumulus_id']),
```
markdboyd, line 152: If you use `t.like` for comparison in tests, then you don't need to do an `omit` of properties that exist on the first argument but not the second. That's the whole beauty of `t.like`: it ignores properties that are "extra" and only compares the properties that exist from the second argument on the first:
```suggestion
    record, 
```

https://github.com/avajs/ava/blob/master/docs/03-assertions.md#likevalue-selector-message

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,336 @@
...
+      pan_message: testPdr.PANmessage,
+      stats: testPdr.stats,
+      address: testPdr.address,
+      original_url: testPdr.originalUrl,
+      timestamp: new Date(testPdr.timestamp),
+      duration: testPdr.duration,
+      created_at: new Date(testPdr.createdAt),
+      updated_at: new Date(testPdr.updatedAt),
+    },
+    ['updated_at'])
```
markdboyd, line 170: Do we need to omit `updated_at`? From the code in the test it seems like we know the expected value?

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,336 @@
...
+    .where({ name: testPdr.pdrName })
+    .first();
+
+  t.like(
+    omit(record, ['cumulus_id']),
+    omit({
+      name: testPdr.pdrName,
+      provider_cumulus_id: providerCumulusId,
+      collection_cumulus_id: collectionCumulusId,
+      execution_cumulus_id: testPdr.executionCumulusId ? testPdr.executionCumulusId : null,
```
markdboyd, line 157: It seems like we know that `testPdr.executionCumulusId` is undefined always for this test, so we can always expect this to be `null`? I don't think we need the ternary expression

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,336 @@
...
+  delete testPdr.progress;
+  delete testPdr.stats;
+  delete testPdr.address;
+  delete testPdr.originalUrl;
+  delete testPdr.timestamp;
+  delete testPdr.duration;
+  delete testPdr.updatedAt;
+
+  await migratePdrRecord(testPdr, knex);
+  const record = await t.context.knex.queryBuilder()
```
markdboyd, line 204: Let's use `pdrPgModel.get` throughout the test code

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,336 @@
...
+    .select()
+    .table(tableNames.pdrs)
+    .where({ name: testPdr.pdrName })
+    .first();
+
+  t.like(
+    omit(record, ['cumulus_id']),
```
markdboyd, line 211: We shouldn't need the `omit` here since we're using `t.like`

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,336 @@
...
+      pan_sent: null,
+      pan_message: null,
+      stats: null,
+      address: null,
+      original_url: null,
+      timestamp: null,
+      duration: null,
+      created_at: new Date(testPdr.createdAt),
+      updated_at: new Date(testPdr.updatedAt),
+    }, ['createdAt', 'updatedAt', 'pdrName', 'collection', 'provider', 'collectionId', 'updated_at'])
```
markdboyd, line 229: I guess it's debatable, but the use of `omit` here seems more complicated than just removing the `...testPdr` on line 213 and instead constructing the exact expected object with only the expected properties

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,336 @@
...
+    .first();
+
+  t.like(
+    omit(record, ['cumulus_id']),
+    omit({
+      ...testPdr,
+      name: testPdr.pdrName,
+      provider_cumulus_id: providerCumulusId,
+      collection_cumulus_id: collectionCumulusId,
+      execution_cumulus_id: testPdr.executionCumulusId ? testPdr.executionCumulusId : null,
```
markdboyd, line 217: It seems like this should always be null based on the test setup?

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,336 @@
...
+    privateKey: 'key',
+    cmKeyId: 'key-id',
+    certificateUri: 'uri',
+  });
+
+  t.context.testCollection = testCollection;
+  t.context.testProvider = testProvider;
+
+  t.context.collectionCumulusId = await migrateFakeCollectionRecord(testCollection, t.context.knex);
+  t.context.providerCumulusId = await migrateFakeProviderRecord(testProvider, t.context.knex);
```
markdboyd, line 118: Let's use the PG model create methods here instead of relying on migration logic from other data types

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,336 @@
...
+    encrypted: true,
+    privateKey: 'key',
+    cmKeyId: 'key-id',
+    certificateUri: 'uri',
+  });
+
+  const testPdr = generateTestPdr(testCollection.name, testProvider.id);
+  const anotherPdr = generateTestPdr(anotherCollection.name, anotherProvider.id);
+  await migrateFakeCollectionRecord(anotherCollection, t.context.knex);
+  await migrateFakeProviderRecord(anotherProvider, t.context.knex);
```
markdboyd, line 277: I'd prefer to not be implicitly testing the collection/provider migration code as part of the PDR migration code tests.

Instead, let's use `fakeCollectionRecordFactory` to set `t.context.testCollection` and rely on `test.beforeEach` to manually create those objects using the model methods. And do the same thing for provider

markdboyd (changes requested):
Looks good! A few comments and requested changes

lambdas/data-migration2/src/pdrs.ts
botanical: Yeah, it is unnecessary. Thanks for catching that!

lambdas/data-migration2/src/pdrs.ts
botanical: Definitely agree. Will update

lambdas/data-migration2/tests/test-pdrs.js
botanical: OH, that makes sense a lot of sense

lambdas/data-migration2/tests/test-pdrs.js
botanical: Ah, I need to omit it because the time is slightly off, ie:

```
  -   updated_at: Date 2021-01-22 04:09:04 957ms UTC {},
  +   updated_at: Date 2021-01-22 04:09:04 907ms UTC {},
```

but I will remove `updated_at` from the comparison anyway while leaving the omit.

lambdas/data-migration2/tests/test-pdrs.js
botanical: Yeah, at that point in my code executionCumulusId was always undefined but I've seen added in at least one example where it's not undefined. Thanks for bringing this to my attention! It reminded me that I wanted to add at least one example where it was defined.

lambdas/data-migration2/tests/test-pdrs.js
botanical: https://github.com/nasa/cumulus/pull/2034#discussion_r562376483

lambdas/data-migration2/tests/test-pdrs.js
botanical: It actually wasn't that complicated to remove `...testPdr` and construct the exact expected object. Only thing was `updatedAt` is a nullable field, but I think, upon insertion, `updated_at` will be set. So I need to remove it from the comparison.

lambdas/data-migration2/src/pdrs.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  const existingRecord = await pdrPgModel.get(knex, { name: dynamoRecord.pdrName });
+
+  // Throw error if it was already migrated.
+  if (existingRecord) {
+    throw new RecordAlreadyMigrated(`PDR name ${dynamoRecord.pdrName} was already migrated, skipping.`);
+  }
+  const collectionCumulusId = await collectionPgModel.getRecordCumulusId(
+    knex,
+    { name: dynamoRecord.collectionId }
```
markdboyd, line 51: I don't think this will ever return anything? The `collectionId` from dynamo is a concatenation of a collection name and version something like `name___version`. So you need to deconstruct that into a name and version, then use both of those keys to look up a collection.

lambdas/data-migration2/tests/test-pdrs.js
markdboyd: Hmm, really? I would expect this adding this to the expected object should work:

`updated_at: new Date(testPdr.updatedAt)`

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,344 @@
...
+  await migratePdrRecord(testPdr, knex);
+
+  const record = await pdrPgModel.get(knex, { name: testPdr.pdrName });
+
+  t.like(
+    record,
+    omit({
+      name: testPdr.pdrName,
+      provider_cumulus_id: providerCumulusId,
+      collection_cumulus_id: collectionCumulusId,
```
markdboyd, line 154: I'm curious/surprised this is working, if it is, since the logic for looking up the PDR collection in the actual migration code seems a bit off

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,344 @@
...
+    testProvider,
+  } = t.context;
+
+  const fakeCollection = fakeCollectionRecordFactory();
+  const fakeProvider = fakeProviderRecordFactory();
+  await collectionPgModel.create(knex, fakeCollection);
+  await providerPgModel.create(knex, fakeProvider);
+
+  const testPdr = generateTestPdr(testCollection.name, testProvider.name);
+  const anotherPdr = generateTestPdr(fakeCollection.name, fakeProvider.name);
```
markdboyd, line 283: I could be wrong, but do you really need to create a separate collection and provider for this second PDR? I think you could just re-use the `testCollection` and `testProvider` to reduce the test setup here

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,344 @@
...
```
markdboyd, line 319: Same comment: I think you can just re-use the `testCollection` and `testProvider` here

lambdas/data-migration2/tests/test-pdrs.js
markdboyd: I think it is working because you're passing in a collection name to set the `collectionId` value for the Dynamo record. So then that the value is used to look up the collection by `name` in PG, which works. But in reality, `collectionId` is a concatenation of name/version, like `name___version`, so the tests need to be updated to reflect that

lambdas/data-migration2/src/pdrs.ts
botanical: Ah, that's true.

lambdas/data-migration2/tests/test-pdrs.js
botanical: Yeah, I was curious about that too after seeing your other comment and I think your conclusion is correct. I've since updated the tests

lambdas/data-migration2/tests/test-pdrs.js
botanical: Oh, for some reason I thought that the collection and provider that a PDR references had to be unique between PDRs but that doesn't seem to be the case.

lambdas/data-migration2/tests/test-pdrs.js
botanical: Yeah, I added that and got :
```
  -   updated_at: Date 2021-01-22 20:50:42 618ms UTC {},
  +   updated_at: Date 2021-01-22 20:50:42 585ms UTC {},
```

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -0,0 +1,328 @@
+const test = require('ava');
+const omit = require('lodash/omit');
```
markdboyd, line 2: I think you're going to get a lint error for importing this and not using it, but we'll see

markdboyd (approved):

lambdas/data-migration2/tests/test-pdrs.js
botanical: You are probably right 😬

PR #2037, opened 2021-01-08
[CUMULUS-2167] File PG Schema Change

botanical:

Addresses [CUMULUS-2167: granules/files migration ](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2167)

## Changes

- Change Postgres Schema definition for Files to remove `filename` and only support `file_name`.
- Change Postgres Schema definition for Files to remove `size` to only support `file_size`.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/lambdas/db-migration/src/migrations/20201014105102_create_files_table.ts
+++ b/lambdas/db-migration/src/migrations/20201014105102_create_files_table.ts
@@ -32,8 +32,6 @@ export const up = async (knex: Knex): Promise<void> =>
       .text('key')
       .comment('AWS S3 key file is archived at');
     table
-      .text('name');
-    table
       .text('path')
       .comment('Source file path');
     table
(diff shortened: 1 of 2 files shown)


lambdas/db-migration/src/migrations/20201014105102_create_files_table.ts
```diff
@@ -28,8 +25,6 @@ export const up = async (knex: Knex): Promise<void> =>
     table
       .text('checksum_value')
       .comment('File checksum');
-    table
```
markdboyd, line 31: Based on FileUtils.js, it looks like `name` might be another duplicate of `filename`/`file_name`. Do you mind taking a looking at that?

botanical: Oh, that's true. Good catch! Yes, `name` seems to also be a duplicate, so I've since removed it [`93f0b01` (#2037)](https://github.com/nasa/cumulus/pull/2037/commits/93f0b01ce29fd078dbb1c7a05dac7dcd02e8da18)

markdboyd (approved):

PR #2039, opened 2021-01-11
CUMULUS-2167: Update PostgresFile type and tests

botanical:

Addresses [CUMULUS-2167: Dynamo-> RDS granules/files](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2167)

## Changes

* Update PG schema to archive `data_type` and `version` on a granule
* Update `PostgresGranule` types to remove `filename` and `size` and add `file_size`, and update `PostgresFile`add `date_type` and `version`
* Update `translateApiFiletoPostgresFile` and test

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/lambdas/db-migration/src/migrations/20201014105058_create_granules_table.ts
+++ b/lambdas/db-migration/src/migrations/20201014105058_create_granules_table.ts
@@ -74,12 +74,6 @@ export const up = async (knex: Knex): Promise<void> =>
     table
       .timestamp('timestamp');
     table
-      .text('data_type')
-      .comment('Refers to a collection name');
-    table
-      .text('version')
-      .comment('Refers to a collection version');
-    table
       .unique(['granule_id', 'collection_cumulus_id']);
   });
 
--- a/lambdas/db-migration/src/migrations/20201014105058_create_granules_table.ts
+++ b/lambdas/db-migration/src/migrations/20201014105058_create_granules_table.ts
@@ -74,6 +74,12 @@ export const up = async (knex: Knex): Promise<void> =>
     table
       .timestamp('timestamp');
     table
+      .text('data_type')
+      .comment('Refers to a collection name');
+    table
+      .text('version')
+      .comment('Refers to a collection version');
+    table
       .unique(['granule_id', 'collection_cumulus_id']);
   });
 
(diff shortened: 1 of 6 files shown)


packages/db/src/types/granule.ts
```diff
@@ -23,6 +23,8 @@ export interface PostgresGranule {
   // Processing info from execution
   processing_start_date_time?: Date,
   processing_end_date_time?: Date,
+  data_type?: string,
+  version?: string,
```
botanical, line 27: Is it worthy to mention these changes and the ones above in the changelog? I noticed we didn't do that for https://github.com/nasa/cumulus/pull/2026/files.

lambdas/db-migration/src/migrations/20201014105058_create_granules_table.ts
```diff
@@ -73,6 +73,12 @@ export const up = async (knex: Knex): Promise<void> =>
       .comment('Timestamp for granule production date/time');
     table
       .timestamp('timestamp');
+    table
+      .text('data_type')
+      .comment('Refers to a collection name');
+    table
+      .text('version')
+      .comment('Refers to a collection version');
```
markdboyd, line 81: After looking closer at the code which writes records to Dynamo, it's not setting `data_type` and `version`:

https://github.com/nasa/cumulus/blob/feature-rds/packages/api/models/granules.js#L381

So these values are never making it into the Dynamo table, as far as I can tell. Thus, I'm not sure if we need to save them for Postgres

packages/db/src/translate/file.ts
```diff
@@ -8,12 +8,9 @@ export const translateApiFiletoPostgresFile = (
   bucket: file.bucket,
   checksum_type: file.checksumType,
   checksum_value: file.checksum,
-  // TODO: do we really need both of these properties?
-  filename: file.fileName,
```
markdboyd, line 12: 👍

lambdas/db-migration/src/migrations/20201014105058_create_granules_table.ts
botanical: Thanks for noticing that. I don't think we need to save them now as well. Will update accordingly

markdboyd (approved):

PR #2040, opened 2021-01-12
CUMULUS-2328: Separate s3-credentials deployment to separate module

markdboyd:

Addresses [CUMULUS-2328: Separate s3-credentials deployment to separate module](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2328)

## Changes

* Move definition of resources for /s3credentials endpoint from `distribution` to standalone `s3-credentials` module
* Add deployment of `s3-credentials` module to example deployment
* Remove unused variables

## PR Checklist

- [x] Update CHANGELOG
- [x] Adhoc testing - I tested the /s3credentials endpoint manually after deploying these changes and confirmed that it still works


bamboo/create-release.sh
```diff
@@ -17,7 +17,7 @@ export VERSION=$(jq --raw-output .version lerna.json)
 
 ## Build TF modules that require source building
 
-(cd tf-modules/distribution && ./bin/build-tf-module.sh && cp ./dist/terraform-aws-cumulus-distribution.zip ../../terraform-aws-cumulus-distribution.zip)
+(cd tf-modules/s3-credentials && ./bin/build-tf-module.sh && cp ./dist/terraform-aws-cumulus-s3-credentials.zip ../../terraform-aws-cumulus-s3-credentials.zip)
```
markdboyd, line 20: I tested this script by running it to build this zipfile and it seemed to work

tf-modules/archive/api.tf
```diff
@@ -20,7 +20,6 @@ locals {
       cmr_username                     = var.cmr_username
       CollectionsTable                 = var.dynamo_tables.collections.name
       DISTRIBUTION_ENDPOINT            = var.distribution_url
-      distributionApiId                = var.distribution_api_id
```
markdboyd, line 23: I couldn't find this being used anywhere, and it was necessitating passing another TEA output into the cumulus module, so I just removed it

tf-modules/s3-credentials/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "s3_credentials_redirect_uri" {
+  value = "${var.tea_external_api_endpoint}redirect"
```
markdboyd, line 2: This **was** using the TEA internal endpoint in the original output definition (see `tf-modules/distribution/outputs.tf`), but that doesn't seem like it makes any sense. If the user is proxying their API via cloudfront or anything, then we want to use that to build the URL for the redirect, which is reflected in the TEA external endpoint output.

tf-modules/s3-credentials/variables.tf
```diff
@@ -0,0 +1,79 @@
...
+variable "prefix" {
+  type        = string
+  description = "Resource prefix unique to this deployment"
+}
+
+variable "sts_credentials_lambda_function_arn" {
+  type    = string
+}
+
+variable "tea_api_gateway_stage" {
```
markdboyd, line 12: I was somewhat inclined to rename these `tea_*` variables to remove the `tea_` prefix, since in theory you could pass in values corresponding to **any** API gateway to attach the /s3credentials endpoint. And we just happen to be attaching this endpoint to TEA

markdboyd:
The first time I tried to deploy this, I did get some errors because there are race conditions between when TF destroys the old s3credentials resources and creates the new ones, so you might get errors about "resource already exists". However, they went away if I just re-ran `apply`. 

I could fix the problem entirely by writing up a set of `terraform state mv` commands, but that felt like overkill when there is no downside of having TF destroy/recreate these resources. However, I'm happy to do it if people think it is worthwhile

tf-modules/archive/api.tf
botanical: I also couldn't find this being used anywhere. Nice catch!

example/cumulus-tf/variables.tf
```diff
@@ -350,3 +350,8 @@ variable "ems_deploy" {
   type        = bool
   default     = true
 }
+
+variable "urs_url" {
+  type = string
+  default = "https://uat.urs.earthdata.nasa.gov"
+}
```
botanical, line 357: Nice

tf-modules/s3-credentials/variables.tf
nemreid: I think it's fine either way.

tf-modules/s3-credentials/outputs.tf
nemreid: I _think_ you're right that the external one makes sense.

nemreid (approved):
Confirmed that the build works. No issues with this PR. Nice.

PR #2041, opened 2021-01-12
Release 5.0.0

markdboyd:


npauzenga:
Just did a search for "4.0.0". Do we need to update `tf-modules/ingest/cumulus_version.json`?

`{"cumulus_version":"4.0.0"}`

markdboyd:
@npauzenga That file should not exist for the `master` branch, since it is something for the RDS deployment and only exists for now on the feature branch

botanical:
Checked all the files but was unsure about `packages/message/README.md`. Isn't `packages/message/src/worfklows.ts` in the `feature-rds` branch?

markdboyd:
@botanical great catch! That happened because I switched from the feature branch to do the release but hadn't re-compiled code before generating the docs. I'll fix that

Jkovarik (dismissed):
The doc updates are missing, please fix.

Jkovarik (approved):
LGTM 👍🏻

PR #2045, opened 2021-01-13
[CUMULUS-2344] Remove some restrictions from ES endpoints

bt6453:

Addresses [CUMULUS-2344: Allow reindexing to an existing index](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2344)

## Changes

* The reindex endpoint will allow you to reindex to an endpoint that already exists
* The change-index endpoint will allow you to change to index that does not
  exist. If it doesn't exist it will be created for you.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [ ] Integration tests


laurenfrederick (changes requested):
I have an issue doing the `change-index`, then the `reindex`. On line 82 the `reindex` operation checks that the `sourceIndex` is aliased under the alias. After you do the change-index it's not under the alias anymore. I don't _think_ we need this check there, but haven't had time to completely think through it.

bt6453:
Yeah I'm not sure I understand completely what the alias checks are for. I think users should be able to reindex to any index they want just like using the ES API directly BUT I understand there might be need for some safety checks.

bt6453:
Added a reindexing guide and removed the restrictions around changing index and reindexing.

bt6453:
API docs updates PR: https://github.com/nasa/cumulus-api/pull/296

docs/troubleshooting/reindex-elasticsearch.md
```diff
@@ -0,0 +1,29 @@
...
+id: reindex-elasticsearch
+title: Reindexing Elasticsearch Guide
+hide_title: false
+---
+
+You may find yourself in a situation where you need to reindex your Elasticsearch index. If you have issues with your
+current index, or the mappings for an index have been updated (they do not update automatically), there are two ways
+to get the data synced correctly.
+
+## Change Index and Reindex
```
laurenfrederick, line 11:
```suggestion
## Switch to a new index and reindex
```

docs/troubleshooting/reindex-elasticsearch.md
```diff
@@ -0,0 +1,29 @@
...
+
+The first way involves reindexing from the existing Elasticsearch index to a brand new index. This is done in two steps.
+
+### Change Index
+
+It is recommended to change Cumulus to point to an empty index first. This will allow new data to be passed to the index without
```
laurenfrederick, line 17: Mention that this step will create a new index for you

docs/troubleshooting/reindex-elasticsearch.md
```diff
@@ -0,0 +1,29 @@
...
+hide_title: false
+---
+
+You may find yourself in a situation where you need to reindex your Elasticsearch index. If you have issues with your
+current index, or the mappings for an index have been updated (they do not update automatically), there are two ways
+to get the data synced correctly.
+
+## Change Index and Reindex
+
+The first way involves reindexing from the existing Elasticsearch index to a brand new index. This is done in two steps.
```
laurenfrederick, line 13: We need to mention that for a period of time, while the reindex takes place, you will not see all of your data on the dashboard as it being brought over from the previous index. We should mention both options here. 1. you use the new index and the pros are you don't lose data, but the cons are it messes with your dashboard. 2. you reindex, then switch and the pros are your dashboard looks good, cons are you should stop ingest.

docs/troubleshooting/reindex-elasticsearch.md
```diff
@@ -0,0 +1,29 @@
+---
+id: reindex-elasticsearch
+title: Reindexing Elasticsearch Guide
+hide_title: false
+---
+
+You may find yourself in a situation where you need to reindex your Elasticsearch index. If you have issues with your
+current index, or the mappings for an index have been updated (they do not update automatically), there are two ways
```
laurenfrederick, line 8: Mention that a reindex due to mappings changing will be called out in the Migration Steps of the changelog.

laurenfrederick (changes requested):

bt6453:
Updated with your recommendations.

laurenfrederick (approved):

PR #2051, opened 2021-01-19
CUMULUS-2167: Dynamo -> RDS Granules/Files

botanical:

Addresses [CUMULUS-2167: Dynamo->RDS Granules/files](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2167)

## Changes

* Script to migrate Granules (and files) records from DynamoDB to RDS 
 
## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


botanical:
Questions I have:

+ Should I not write any file records for a granule record that fails? (That's how I've implemented it so far)
+ Should I handle potential duplicate file records? If so, how can I differentiate unique null values when selecting on the Files table using `bucket` and `key` where those unique constraints might be null?

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,269 @@
...
+        fileMigrationSummary.failed += 1;
+        logger.error(
+          `Could not create file record in RDS for file ${file}`,
+          error
+        );
+      }
+    }));
+  });
+
+  return fileMigrationSummary;
```
botanical, line 211: I don't particularly like that this function returns `fileMigrationSummary` but couldn't think of a better way to propagate this information for `migrateGranulesAndFiles`

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,269 @@
...
+
+  /* eslint-disable no-await-in-loop */
+  while (record) {
+    granuleMigrationSummary.dynamoRecords += 1;
+    try {
+      const granuleFileMigrationSummary = await migrateGranuleAndFilesViaTransaction(record, knex);
+      granuleMigrationSummary.success += 1;
+      fileMigrationSummary.dynamoRecords += granuleFileMigrationSummary.dynamoRecords;
+      fileMigrationSummary.success += granuleFileMigrationSummary.success;
+      fileMigrationSummary.failed += granuleFileMigrationSummary.failed;
```
botanical, line 248: Mentioned in my comment https://github.com/nasa/cumulus/pull/2051/files#r561389823 that I don't particularly like this, and I feel like calculating `fileMigrationSummary` here is messy but I am not sure of a cleaner alternative. Open to ideas

botanical:
Marking this as WIP until I fix the merged changes

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,533 @@
...
+  } = t.context;
+
+  await migrateGranuleRecord(testGranule, knex);
+
+  const record = await granulePgModel.get(knex, {
+    granule_id: testGranule.granuleId,
+    collection_cumulus_id: collectionCumulusId,
+  });
+
+  t.like(
```
npauzenga, line 179: Are we using `like` here because of the extra camelCase dynamo fields? In other tests we use an explicit "omit list" which specifies those fields. That might be preferable just so it's super explicit and we don't get unexpected fields? Thoughts? 

See `collectionOmitList` in test-collections.js as an example.

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,533 @@
...
+      success: 2,
+    },
+  });
+  const records = await knex(tableNames.granules);
+  const fileRecords = await knex(tableNames.files);
+  t.is(records.length, 2);
+  t.is(fileRecords.length, 2);
+});
+
+test.serial('migrateGranulesAndFiles processes all non-failing granule records and does not process files of failling granule records', async (t) => {
```
npauzenga, line 471: I've seen the word "failing" on my tests enough to know you have an extra "l" here 😄

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,533 @@
...
+    },
+  });
+  const records = await knex(tableNames.granules);
+  const fileRecords = await knex(tableNames.files);
+  t.is(records.length, 1);
+  t.is(fileRecords.length, 1);
+});
+
+test.serial.skip('migrateFileRecord throws RecordAlreadyMigrated error for already migrated record', async () => {
+});
```
npauzenga, line 533: Should we remove this one?

npauzenga (changes requested):
Awesome work @botanical! I left a few comments but it's mostly clarification and preference-related stuff. Let me know if anything doesn't make sense and feel free to disagree with any of it 😄

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,269 @@
...
+  const executionCumulusId = dynamoRecord.execution
+    ? await getRecordCumulusId<PostgresExecutionRecord>(
+      { arn: dynamoRecord.execution },
+      tableNames.executions,
+      knex
+    )
+    : undefined;
+
+  const existingRecord = await knex<PostgresGranuleRecord>('granules')
+    .where({ granule_id: dynamoRecord.granuleId, collection_cumulus_id: collectionCumulusId })
```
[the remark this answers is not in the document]
botanical: Not a selfish question at all! A `granuleId` isn't unique across files because many files can be associated to a single `granuleId`, so you are correct in that assumption. Though, I am not sure how I am not meeting your suggestion, does `where` not work like I thought? I am looking up a granule by `granuleId` and `collectionCumuludId`. Am I misinterpreting your suggestion? @npauzenga

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,269 @@
...
+    processing_end_date_time: dynamoRecord.processingEndDateTime
+      ? new Date(dynamoRecord.processingEndDateTime) : undefined,
+    processing_start_date_time: dynamoRecord.processingStartDateTime
+      ? new Date(dynamoRecord.processingStartDateTime) : undefined,
+    production_date_time: dynamoRecord.productionDateTime
+      ? new Date(dynamoRecord.productionDateTime) : undefined,
+    timestamp: dynamoRecord.timestamp
+      ? new Date(dynamoRecord.timestamp) : undefined,
+    created_at: new Date(dynamoRecord.createdAt),
+    updated_at: new Date(dynamoRecord.updatedAt),
```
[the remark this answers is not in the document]
botanical: I definitely am not opposed to doing that! In fact, I am in favor of doing it 😃

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,269 @@
...
+  };
+
+  // Validate record before processing using API model schema
+  Manager.recordIsValid(dynamoRecord, schemas.granule);
+  const files = dynamoRecord.files;
+  const granuleId = dynamoRecord.granuleId;
+  const collectionId = dynamoRecord.collectionId;
+
+  await knex.transaction(async () => {
+    await migrateGranuleRecord(dynamoRecord, knex);
```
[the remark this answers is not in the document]
botanical: Yes, if this fails then it doesn't migrate file records. From our last architecture meeting, that was what I believe we concluded 🙂 https://docs.google.com/document/d/12Ss749W8a6wIuIB79Tt3irNoglYxE61JcdW7rXMBS1g/edit

lambdas/data-migration2/src/index.ts
```diff
@@ -21,6 +23,15 @@ export const handler = async (event: HandlerEvent): Promise<string> => {
...
             ${executionsMigrationSummary.failed} records failed
+        Granules:
+          Out of ${granulesAndFilesMigrationSummary.granulesSummary.dynamoRecords} DynamoDB records:
+            ${granulesAndFilesMigrationSummary.granulesSummary.success} records migrated
+            ${granulesAndFilesMigrationSummary.granulesSummary.skipped} records skipped
+            ${granulesAndFilesMigrationSummary.granulesSummary.failed} records failed
+        Files:
+          Out of ${granulesAndFilesMigrationSummary.granulesSummary.dynamoRecords} DynamoDB records:
+            ${granulesAndFilesMigrationSummary.filesSummary.success} records migrated
+            ${granulesAndFilesMigrationSummary.filesSummary.failed} records failed
```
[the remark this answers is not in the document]
botanical: Great question! That, I am unsure of. 

We skip a record if it has already been migrated and I wouldn't want to migrate the same record file multiple times, but I am hung up on differentiating file records that have a null bucket and key.

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,533 @@
...
+// eslint-disable-next-line node/no-unpublished-require
+const { migrationDir } = require('../../db-migration');
+const { migrateGranuleRecord, migrateFileRecord, migrateGranulesAndFiles } = require('../dist/lambda/granulesAndFiles');
+
+const buildCollectionId = (name, version) => `${name}___${version}`;
+
+const dateString = new Date().toString();
+const bucket = cryptoRandomString({ length: 10 });
+
+const generateTestGranule = (collection, executionId, alternateBucket, pdrName, provider) => ({
```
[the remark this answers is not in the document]
botanical: I intentionally didn't use `fakeGranuleFactoryV2` because I want to associate the granule record to a collection, whereas `fakeGranuleFactoryV2` calls `fakeGranuleFactory` and sets the collection id to :
```'fakeCollection___v1'```.

Does that make sense or am I missing something in my reasoning? 🤔

File: lambdas/data-migration2/tests/test-granulesAndFiles.js [hunk 11]
```diff
@@ -0,0 +1,533 @@
...
+  granuleId: cryptoRandomString({ length: 5 }),
+  collectionId: buildCollectionId(collection.name, collection.version),
+  pdrName: pdrName,
+  provider: provider,
+  status: 'running',
+  execution: executionId,
+  cmrLink: cryptoRandomString({ length: 10 }),
+  published: false,
+  duration: 10,
+  files: [
```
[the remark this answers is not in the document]
botanical: Ah, yes, that is a good point.

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: Ah thanks for catching that!

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: Hm yes, I opted to use `like` but perhaps it is better to be explicit and have an `omit list`.

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: I am unsure. https://github.com/nasa/cumulus/pull/2051#discussion_r564089098

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,269 @@
...
+ * @returns {Promise<void>}
+ * @throws {RecordAlreadyMigrated} if record was already migrated
+ */
+export const migrateGranuleRecord = async (
+  dynamoRecord: AWS.DynamoDB.DocumentClient.AttributeMap,
+  knex: Knex
+): Promise<void> => {
+  // Validate record before processing using API model schema
+  Manager.recordIsValid(dynamoRecord, schemas.granule);
+  const [name, version] = dynamoRecord.collectionId.split('___');
```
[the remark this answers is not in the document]
botanical: Yeah, just as you said, I think we can rely on `collectionId` being present since it's required.

botanical:
As I mentioned this morning, I am still going through the variations of logging the file migration summary but I would like the rest of this work to be reviewed (including the current logic I have for logging the file migration summary), so I've marked this as needs review.

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: @npauzenga This is how I ended up refactoring it based on your suggestions [`5a543a5` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/5a543a53520b1e96e50a477902a4ffb14fc71446). In some respects, it is cleaner but I also think it makes the transaction obsolete because the `try/catch` now handles skipping writing file records if a granule record fails. In this commit I've linked I haven't removed the knex transaction but, if we go this route, I will definitely remove it since I don't think it's necessary. Let me know what you think!

lambdas/data-migration1/src/rules.ts
```diff
@@ -114,7 +114,7 @@ export const migrateRules = async (
       } else {
         migrationSummary.failed += 1;
         logger.error(
-          `Could not create rule record in RDS for Dynamo Rule name: ${record.name}, version: ${record.version}}`,
+          `Could not create rule record in RDS for Dynamo Rule name: ${record.name}, version: ${record.version}`,
```
markdboyd, line 117: This looks like a bad copy/paste from the collections migration. Dynamo rules don't have a `version`

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,208 @@
...
+  Manager.recordIsValid(record, schemas.granule);
+  const { name, version } = deconstructCollectionId(record.collectionId);
+  const collectionPgModel = new CollectionPgModel();
+
+  const collectionCumulusId = await collectionPgModel.getRecordCumulusId(
+    knex,
+    { name, version }
+  );
+
+  const existingRecord = await knex<PostgresGranuleRecord>('granules')
```
markdboyd, line 53: Let's use the granule PG model `.get` method here

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,208 @@
...
+    .where({ granule_id: record.granuleId, collection_cumulus_id: collectionCumulusId })
+    .first();
+
+  // Throw error if it was already migrated.
+  if (existingRecord) {
+    throw new RecordAlreadyMigrated(`Granule ${record.granuleId} was already migrated, skipping`);
+  }
+
+  const granule = await translateApiGranuleToPostgresGranule(record, knex, collectionCumulusId);
+  await knex(tableNames.granules).insert(granule);
```
markdboyd, line 63: Should also be using a granule model method, possibly for `upsert`

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,208 @@
...
+    key,
+    granule_cumulus_id: granuleCumulusId,
+    file_size: file.size,
+    checksum_value: file.checksum,
+    checksum_type: file.checksumType,
+    file_name: file.fileName,
+    source: file.source,
+    path: file.path,
+  };
+  await knex(tableNames.files).insert(updatedRecord);
```
markdboyd, line 111: Should be using the files PG model `.upsert` method

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,208 @@
...
+  const [name, version] = collectionId.split('___');
+  const collectionPgModel = new CollectionPgModel();
+  const granulePgModel = new GranulePgModel();
+
+  const collectionCumulusId = await collectionPgModel.getRecordCumulusId(
+    knex,
+    { name, version }
+  );
+
+  const granuleCumulusId = await granulePgModel.getRecordCumulusId(
```
markdboyd, line 91: This lookup is a bit superfluous. We could return the granule's cumulus ID from inside of `migrateGranuleRecord` and pass it into here rather than having to look it up from the dynamo `granuleId`

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,208 @@
...
+  const granuleId = dynamoRecord.granuleId;
+  const collectionId = dynamoRecord.collectionId;
+  const { granulesSummary, filesSummary } = granuleAndFileMigrationSummary;
+
+  try {
+    granulesSummary.dynamoRecords += 1;
+    await migrateGranuleRecord(dynamoRecord, knex);
+    granulesSummary.success += 1;
+    await Promise.all(files.map(async (file : ApiFile) => {
+      filesSummary.dynamoRecords += 1;
```
markdboyd, line 136: So we really need to use a `knex.transaction` here. A try/catch doesn't quite achieve the same thing. With the try/catch as written, if `migrateGranuleRecord` succeeds but then `migrateFileRecord` errors, the granule record that was written by `migrateGranuleRecord` still exists. Thus we have "partial data" where the granule record was written but the files were not, which we previously decided that we do not want.

When you use `knex.transaction` to surround these writes, then if `migrateGranuleRecord` succeeds but `migrateFileRecord` (for any file) fails, then the write for the granule is **not committed**, thus both the granule and files are not written.

Does that make sense?

markdboyd: Also, if we use a transaction, then the success/failure of writes is "all or nothing". Either the granule and all of the file records are written or none of them are. Thus, I don't think you would need to do the `+= 1` operation to count the number of records that succeeded or failed. If the transaction didn't thrown an error, then the granule was a success and all of its files were, otherwise if an error was thrown than they were all failures. It should simplify the logic here a bit

packages/db/src/translate/granules.ts
```diff
@@ -0,0 +1,76 @@
...
+    collection_cumulus_id: collectionCumulusId,
+    published: dynamoRecord.published,
+    duration: dynamoRecord.duration,
+    time_to_archive: dynamoRecord.timeToArchive,
+    time_to_process: dynamoRecord.timeToPreprocess,
+    product_volume: dynamoRecord.productVolume,
+    error: dynamoRecord.error,
+    cmr_link: dynamoRecord.cmrLink,
+    execution_cumulus_id: dynamoRecord.execution
+      ? await getRecordCumulusId<PostgresExecutionRecord>(
```
markdboyd, line 40: Should be using the execution PG model `getRecordCumulusId` method?

packages/db/tests/translate/test-granules.js
```diff
@@ -0,0 +1,97 @@
...
+const { removeNilProperties } = require('@cumulus/common/util');
+const getRecordCumulusId = require('../../dist/database');
+const { translateApiGranuleToPostgresGranule } = require('../../dist/translate/granules');
+
+test('translateApiGranuleToPostgresGranule converts API execution to Postgres', async (t) => {
+  const collectionCumulusId = 1;
+  const providerCumulusId = 2;
+  const executionCumulusId = 3;
+  const pdrCumulusId = 4;
+  const getRecordCumulusIdStub = sinon.stub(getRecordCumulusId, 'getRecordCumulusId').returns(executionCumulusId);
```
markdboyd, line 14: If you make the change in this function to use execution PG model `getRecordCumulusId` in the function, then you don't need this stub and can pass in a fake model like you are for PDRs and providers

lambdas/data-migration2/src/granulesAndFiles.ts
markdboyd: It would also remove the need for the collection lookup in this function

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,208 @@
...
+  const existingRecord = await knex<PostgresGranuleRecord>('granules')
+    .where({ granule_id: record.granuleId, collection_cumulus_id: collectionCumulusId })
+    .first();
+
+  // Throw error if it was already migrated.
+  if (existingRecord) {
+    throw new RecordAlreadyMigrated(`Granule ${record.granuleId} was already migrated, skipping`);
+  }
+
+  const granule = await translateApiGranuleToPostgresGranule(record, knex, collectionCumulusId);
```
markdboyd, line 62: It seems kind of inconsistent to be doing the collection lookup in this function rather than in `translateApiGranuleToPostgresGranule`. If the point of `translateApiGranuleToPostgresGranule` is that it can convert a Dynamo granule record to a PG record, then it needs to do the collection lookup itself

lambdas/data-migration2/tests/test-granulesAndFiles.js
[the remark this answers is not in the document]
markdboyd: It does do that, but `fakeGranuleFactoryV2` also allows you to specify an `options` object that lets you override any of the defaults generated by `fakeGranuleFactory`, so it would be pretty easy to override just the `collectionId` to whatever value that you want

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,538 @@
...
+  await executionsModel.createTable();
+
+  granulesModel = new Granule();
+  await granulesModel.createTable();
+
+  t.context.granulePgModel = new GranulePgModel();
+  t.context.filePgModel = new FilePgModel();
+
+  // Store the CMR password
+  process.env.cmr_password_secret_name = cryptoRandomString({ length: 5 });
```
markdboyd, line 102: I'm definitely missing something, but what makes use of this CMR secret?

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,538 @@
...
+
+  await migrateGranuleRecord(testGranule, knex);
+
+  const record = await granulePgModel.get(knex, {
+    granule_id: testGranule.granuleId,
+    collection_cumulus_id: collectionCumulusId,
+  });
+
+  t.deepEqual(
+    omit(record, ['updated_at', 'cumulus_id']),
```
markdboyd, line 184: The value of `updated_at` should be predictable, so I don't know why we need to omit it. And if you use `t.like` instead of `t.deepEqual`, you don't need to bother with the `omit` because it only compares the properties that exist in the second object

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,538 @@
...
+  delete testGranule.timeToPreprocess;
+  delete testGranule.beginningDateTime;
+  delete testGranule.endingDateTime;
+  delete testGranule.processingStartDateTime;
+  delete testGranule.processingEndDateTime;
+  delete testGranule.lastUpdateDateTime;
+  delete testGranule.timeToArchive;
+  delete testGranule.productionDateTime;
+  delete testGranule.timestamp;
+  delete testGranule.provider;
```
markdboyd, line 282: I think you need to add `version`, `queryFields`, `execution`, and `queryFields` to this list

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,538 @@
...
+
+  await migrateGranuleRecord(testGranule, knex);
+
+  const record = await granulePgModel.get(knex, {
+    granule_id: testGranule.granuleId,
+    collection_cumulus_id: collectionCumulusId,
+  });
+
+  t.deepEqual(
+    omit(record, ['updated_at', 'cumulus_id']),
```
markdboyd, line 292: Same comment as above: `updated_at` should be predictable and you could use `t.like` for comparison

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: Kind of like this? https://github.com/nasa/cumulus/blob/b6542f89bd4d2a5b8be9805930468952b2659c6c/lambdas/data-migration2/src/granulesAndFiles.ts#L137-L154 @markdboyd

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,538 @@
...
+    knex,
+    testGranule,
+  } = t.context;
+
+  const testFile = testGranule.files[0];
+  await migrateGranuleRecord(testGranule, knex);
+  await migrateFileRecord(testFile, testGranule.granuleId, testGranule.collectionId, knex);
+
+  // I am not sure how I can select a file where bucket and key are null
+  const record = await filePgModel.get(knex, {});
```
markdboyd, line 224: Why not just generate a test file where bucket/key will not be null for the purpose of this test? The issue with possible null values for bucket/key is a database level concern that is separate from this code, so I don't know if we need to explicitly test for it. But I could be wrong

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,538 @@
...
+
+test.serial('migrateFileRecord correctly migrates file record', async (t) => {
+  const {
+    filePgModel,
+    knex,
+    testGranule,
+  } = t.context;
+
+  const testFile = testGranule.files[0];
+  await migrateGranuleRecord(testGranule, knex);
```
markdboyd, line 220: I'm somewhat inclined to create the granule manually using the PG model `create` and just pass that into `migrateFileRecord` if you the refactor to it that I proposed above

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,538 @@
...
+      checksum_value: testFile.checksum,
+      checksum_type: testFile.checksumType,
+      key: testFile.key ? testFile.key : null,
+      path: null,
+      file_size: testFile.size.toString(),
+      file_name: testFile.fileName,
+      source: testFile.source,
+      type: testFile.type,
+    },
+    ['type'])
```
markdboyd, line 239: Seems like we can just not add `type` to the object and then remove the `omit`?

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,538 @@
...
+  delete testFile.bucket;
+  delete testFile.checksum;
+  delete testFile.checksumType;
+  delete testFile.fileName;
+  delete testFile.key;
+  delete testFile.path;
+  delete testFile.size;
+  delete testFile.source;
+
+  await migrateGranuleRecord(testGranule, knex);
```
markdboyd, line 352: Again, somewhat inclined to just create the granule with the PG model and pass in that ID to `migrateFileRecord`

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,538 @@
...
+    testExecution2
+  );
+
+  const testCollection2 = fakeCollectionRecordFactory();
+  await collectionPgModel.create(
+    t.context.knex,
+    testCollection2
+  );
+
+  const testGranule2 = generateTestGranule(testCollection2, testExecution2.arn, alternateBucket);
```
markdboyd, line 499: I don't think you really need a new collection, execution, or bucket for the conditions of this test to hold? You just need one granule to fail and the other to succeed

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,538 @@
...
+  );
+
+  const testCollection2 = fakeCollectionRecordFactory();
+  await collectionPgModel.create(
+    t.context.knex,
+    testCollection2
+  );
+
+  const testGranule1 = testGranule;
+  const testGranule2 = generateTestGranule(testCollection2, testExecution2.arn, alternateBucket);
```
markdboyd, line 435: Is using a new collection, execution, and bucket necessary for this test?

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: By `collection lookup` are you referring to retrieving the `collectionCumulusId`? Because I still need that to check for an existing record. I thought I could just pass it into the `translate` function instead of having to run the same lookup but I can see how that isn't generalizable if someone else uses the `translate` function. 

I'd still need to look up `collectionCumulusId` though. @markdboyd

lambdas/data-migration2/src/granulesAndFiles.ts
markdboyd: In this case it was easier to just create a PR to show what I meant:

https://github.com/nasa/cumulus/pull/2056

This PR doesn't take into consideration addressing other feedback that I left but just shows how you could use the transaction

packages/db/src/translate/granules.ts
botanical: [`c58cf1b` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/c58cf1bbdd841e01b1621ee0b002c6a4c0b8d210)

packages/db/tests/translate/test-granules.js
botanical: [`c58cf1b` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/c58cf1bbdd841e01b1621ee0b002c6a4c0b8d210)

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: [`c58cf1b` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/c58cf1bbdd841e01b1621ee0b002c6a4c0b8d210)

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: You're right about `version` and `queryFields`. Taking a closer look at the [granules schema ](https://github.com/nasa/cumulus/blob/master/packages/api/models/schemas.js#L362), it looks like execution is required. Is there a reason why it is not required for our PG schema?

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: Yeah, that is true. I am opting to use `deepEqual` though for this case because I want to ensure I am accounting for all fields in the object. I had initially used `t.like` but changed my mind here https://github.com/nasa/cumulus/pull/2051#discussion_r563918827 

For clarity, you are correct that I don't need to omit `updated_at`

botanical: [`6ad9950` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/6ad99508c7ee725ec8bd8b4c40ca3092192275f4)

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: [`6ad9950` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/6ad99508c7ee725ec8bd8b4c40ca3092192275f4)

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: [`6ad9950` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/6ad99508c7ee725ec8bd8b4c40ca3092192275f4)

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: Yeah 🤦 [`cc2dabc` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/cc2dabc6a12ab1c87781492c881a5b21a5b51ebb)

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: 👀 Ah there I go again creating unnecessary things for my test [`5374077` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/53740775e8529f9099375c223a3b91bb42e27b0e)

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: No. No it isn't 😞 [`5374077` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/53740775e8529f9099375c223a3b91bb42e27b0e)

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: [`c8ed8f2` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/c8ed8f2940a7d3697b4eb7f15c857f5892656ae9)

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: Why possibly? It seems like we should use upsert [`c8ed8f2` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/c8ed8f2940a7d3697b4eb7f15c857f5892656ae9)

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: [`2e2d579` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/2e2d579eabdb74b06e869b01d8e8993acdd98c9e)

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: [`2e2d579` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/2e2d579eabdb74b06e869b01d8e8993acdd98c9e)

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: It turns out that I don't need to use it and I can't quite recall why I thought I needed it in the first place... 🤔 [`f7a166c` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/f7a166c722301e7320ebd417a028d957e385d484)

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: It was mostly because I was running into a PG error but I realized I was trying to find a bucket and key using `undefined` instead of `null` [`0be0714` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/0be07147dfb477fc6c28ccec11645ec6f29b7d33)

lambdas/data-migration2/src/granulesAndFiles.ts
markdboyd: It occurs to me that as long as we have logic to skip records if they already exist in the granules/files tables, then there's no point in using `upsert`, because it'll never try to update an existing record. But that's a question outside the scope of this work and PR

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,199 @@
...
+  granulesSummary: MigrationSummary,
+  filesSummary: MigrationSummary,
+}
+
+/**
+ * Migrate granules record from Dynamo to RDS.
+ *
+ * @param {AWS.DynamoDB.DocumentClient.AttributeMap} record
+ *   Record from DynamoDB
+ * @param {Knex} knex - Knex client for writing to RDS database
```
markdboyd, line 35: In reality this parameter is now a Knex transaction, so we should update the JSdoc

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,199 @@
...
+  const [cumulusId] = await granulePgModel.upsert(knex, granule);
+  return cumulusId;
+};
+
+/**
+ * Migrate File record from a Granules record from DynamoDB  to RDS.
+ *
+ * @param {ApiFile} file - Granule file
+ * @param {number} granuleCumulusId - ID of granule
+ * @param {Knex} knex - Knex client for writing to RDS database
```
markdboyd, line 73: In reality this parameter is now a Knex transaction, so we should update the JSdoc

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,534 @@
...
+
+  process.env.CollectionsTable = cryptoRandomString({ length: 10 });
+  process.env.ExecutionsTable = cryptoRandomString({ length: 10 });
+  process.env.GranulesTable = cryptoRandomString({ length: 10 });
+
+  collectionsModel = new Collection();
+  await collectionsModel.createTable();
+
+  executionsModel = new Execution();
+  await executionsModel.createTable();
```
markdboyd, line 93: Do we need to create the collection and executions Dynamo tables?

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,534 @@
...
+    knex,
+    testGranule,
+  } = t.context;
+
+  await migrateGranuleRecord(testGranule, knex);
+
+  const record = await granulePgModel.get(knex, {
+    granule_id: testGranule.granuleId,
+    collection_cumulus_id: collectionCumulusId,
+  });
```
markdboyd, line 167:
```suggestion
  const [granuleCumulusId] = await migrateGranuleRecord(testGranule, knex);

  const record = await granulePgModel.get(knex, {
    cumulus_id: granuleCumulusId
  });
```

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,534 @@
...
+  const testFile = testGranule.files[0];
+  const granuleCumulusId = await migrateGranuleRecord(testGranule, knex);
+  await migrateFileRecord(testFile, granuleCumulusId, knex);
+
+  const record = await filePgModel.get(knex, { bucket: testFile.bucket, key: testFile.key });
+
+  t.deepEqual(
+    omit(record, fileOmitList),
+    {
+      bucket: testFile.bucket ? testFile.bucket : null,
```
markdboyd, line 215: Do we not have confidence that bucket/key are set from the test setup? Do we need these ternary expressions?

lambdas/data-migration2/tests/test-granulesAndFiles.js
markdboyd: Good point about `execution`. There's no reason that I can think of not to make it required for PG. Maybe worth asking the team though

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,534 @@
...
+  delete testFile.key;
+  delete testFile.path;
+  delete testFile.size;
+  delete testFile.source;
+
+  const granuleCumulusId = await migrateGranuleRecord(testGranule, knex);
+  await migrateFileRecord(testFile, granuleCumulusId, knex);
+
+  // Also unsure of condition for null bucket and key
+  const record = await filePgModel.get(knex, {});
```
markdboyd, line 376: I'm intrigued that this works and why. If you weren't cleaning up the tables between tests, I think it would return any rows where bucket/key are null? Does looking up by `{ bucket: null, key: null }` do the same thing?

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,534 @@
...
+test.serial('migrateGranulesAndFiles processes multiple granules and files', async (t) => {
+  const {
+    knex,
+    testCollection,
+    testExecution,
+    testGranule,
+  } = t.context;
+
+  const alternateBucket = cryptoRandomString({ length: 10 });
+  await s3Utils.createBucket(alternateBucket);
```
markdboyd, line 440: Is this test setup necessary?

packages/db/tests/models/test-granule-model.js
```diff
@@ -219,7 +219,7 @@ test('GranulePgModel.upsert() will not allow a running status to replace a compl
   t.is(record.status, 'completed');
 });
 
-test('GranulePgModel.upsert() will allow a running status to replace a completed status for different execution', async (t) => {
+test.only('GranulePgModel.upsert() will allow a running status to replace a completed status for different execution', async (t) => {
```
markdboyd, line 222: Remove `.only`

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,199 @@
...
+  const existingRecord = await knex<PostgresGranuleRecord>('granules')
+    .where({ granule_id: record.granuleId, collection_cumulus_id: collectionCumulusId })
+    .first();
+
+  // Throw error if it was already migrated.
+  if (existingRecord) {
+    throw new RecordAlreadyMigrated(`Granule ${record.granuleId} was already migrated, skipping`);
+  }
+
+  const granule = await translateApiGranuleToPostgresGranule(record, knex, collectionPgModel);
```
markdboyd, line 63: I suppose that we can do it, but what is the reason for passing `collectionPgModel` into `translateApiGranuleToPostgresGranule`?

lambdas/data-migration2/src/granulesAndFiles.ts
markdboyd: Still needs to be addressed

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: I did it because I'd already instantiated it though I guess I don't need to...

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: No, but I wanted to test a granule with a file that referenced a different bucket. I suppose that isn't necessary.

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: So we can look up a granule by either `granuleId` and `collectionCumulusId` OR `granuleCumulusId`?

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: Turns out we don't

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: Negatory

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: Yeah, it does the same thing! I left it in because I thought it was odd why it works too. Is it better to be explicit and specify null bucket and key? 

Using `bucket: null, key:null` gives the same, and I did that in another test:

https://github.com/nasa/cumulus/blob/jt/CUMULUS-2167/granules-files/lambdas/data-migration2/tests/test-granulesAndFiles.js#L231

I think it's good to look at what's going on here too:
https://github.com/knex/knex/blob/b1117711178e39f81b7a1d3de470a784cc797265/lib/query/querybuilder.js#L394-L401

botanical: I am also wondering if our `get` should not return the first record and should instead return all records that satisfy a given condition. And if we should have another function like `getFirst` to get the first record. What do you think, @markdboyd?

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: [`bc6cc2b` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/bc6cc2bd0d21127123f6a3bbe6cc5477bd3579fc) Don't I need to update the type too?

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: [`393105e` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/393105ed535e7d30036ccdb188bee42b233dd4f5)

Shouldn't I update the type too?

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: [`92f18be` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/92f18be28ea3d207ee6021b5bbf883da51b5cbba)

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: [`bc6cc2b` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/bc6cc2bd0d21127123f6a3bbe6cc5477bd3579fc#diff-91b3c5306441fb2380946f4cabdfa533f0143379d13bebd235dc5cd9af8d8586R53-R56)

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: [`972c441` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/972c441b1b60661185d3edfb5b73855e77d411e4)

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: [`299d0fb` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/299d0fbddc73e17cd00e360974bd4425347f10f3)

lambdas/data-migration2/tests/test-granulesAndFiles.js
markdboyd: Yeah, you can look up any PG record either using its unique column(s) (the combo of `collectionCumulusId` and `granuleId` have a unique constraint for granules) or its primary key, which in this case is the `cumulus_id` column (and is for all of our tables)

lambdas/data-migration2/src/granulesAndFiles.ts
markdboyd: Yup

lambdas/data-migration2/tests/test-granulesAndFiles.js
markdboyd: Great explanation and dive into the Knex code!

Personally, I'd rather leave `get` as returning the first record, since that also matches the behavior from the previous Dynamo  models. And then have a method like `query` for getting all records that match the given query. Though I don't know if we have a use case for such a method yet.

lambdas/data-migration2/src/granulesAndFiles.ts
markdboyd: Yeah I guess there's nothing wrong with it but it seems like in general `translateApiGranuleToPostgresGranule` should be able to be called without an expectation of providing the instantiated model (unless you're doing a fake model for testing)

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,199 @@
...
+      await Promise.all(files.map(async (file : ApiFile) => {
+        try {
+          await migrateFileRecord(file, granuleCumulusId, trx);
+        } catch (error) {
+          logger.error(
+            `Could not create file record in RDS for file ${file}`,
+            error
+          );
+          // Have to re-throw for transaction to fail
+          throw error;
```
markdboyd, line 132: I'm realizing that the try/catch here with the error logging is probably unnecessary, because if an error is thrown within the entire transaction it will be handled and logged on line 145?

markdboyd: And I realize that this code all came from my suggestion 😄

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,510 @@
...
+
+const testDbName = `data_migration_2_${cryptoRandomString({ length: 10 })}`;
+
+test.before(async (t) => {
+  await s3Utils.createBucket(bucket);
+  process.env.stackName = cryptoRandomString({ length: 10 });
+  process.env.system_bucket = bucket;
+
+  process.env.CollectionsTable = cryptoRandomString({ length: 10 });
+  process.env.ExecutionsTable = cryptoRandomString({ length: 10 });
```
markdboyd, line 81: I think these are unnecessary now

lambdas/data-migration2/tests/test-granulesAndFiles.js
markdboyd: This change isn't required, but it doesn't look like it was addressed here

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,510 @@
...
+
+  await migrateGranuleRecord(testGranule, knex);
+
+  const record = await granulePgModel.get(knex, {
+    granule_id: testGranule.granuleId,
+    collection_cumulus_id: collectionCumulusId,
+  });
+
+  t.deepEqual(
+    omit(record, ['cumulus_id']),
```
markdboyd, line 157: Since we could get the cumulus ID from `migrateGranuleRecord` above, we don't have to omit it and could do a direct comparison. But it's not required

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,510 @@
...
+  );
+});
+
+test.serial('migrateGranulesAndFiles skips already migrated granule record', async (t) => {
+  const {
+    knex,
+    testGranule,
+  } = t.context;
+
+  await Promise.all(testGranule.files.map((file) => s3Utils.s3PutObject({
```
markdboyd, line 385: Do we have to stage files in S3 for the migration to work?

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -0,0 +1,510 @@
...
+test.serial('migrateFileRecord correctly migrates file record', async (t) => {
+  const {
+    filePgModel,
+    knex,
+    testGranule,
+  } = t.context;
+
+  const testFile = testGranule.files[0];
+  const granuleCumulusId = await migrateGranuleRecord(testGranule, knex);
+  await migrateFileRecord(testFile, granuleCumulusId, knex);
```
markdboyd, line 195: I'm not going to hold up the review on this, but I think it would be ideal if we weren't testing both `migrateGranuleRecord` and `migrateFileRecord` in this test. We could just create the granule directly, though I suppose we'd have to translate from the Dynamo/API format to the PG format first. So I guess it's not really worth doing

packages/db/tests/translate/test-granules.js
```diff
@@ -0,0 +1,99 @@
...
+    last_update_date_time: new Date(apiGranule.lastUpdateDateTime),
+    processing_end_date_time: new Date(apiGranule.processingEndDateTime),
+    processing_start_date_time: new Date(apiGranule.processingStartDateTime),
+    production_date_time: new Date(apiGranule.productionDateTime),
+    timestamp: new Date(apiGranule.timestamp),
+    created_at: new Date(apiGranule.createdAt),
+    updated_at: new Date(apiGranule.updatedAt),
+  };
+
+  const result = removeNilProperties(
```
markdboyd, line 84: Just curious: what nil properties are we removing? Would there be value in include the null values directly in the comparison?

botanical: Yeah, I realized this was a mindless paste I made. I updated the test [`c69fbd5` (#2051)](https://github.com/nasa/cumulus/pull/2051/commits/c69fbd53ebcfda5050765044a8b6f5c6403526ff)

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: Hmm yes, it can be called without providing the instantiated model. Is it bad to provide it anyway? Or is it purely for testing purposes? Sorry to belabor this but I am not sure what is technically clean code.

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: For some reason that escapes me now I that we needed to do that but it doesn't seem to be the case anymore... 🤔

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: Yeah, I agree... I'm going to remove it

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: I will definitely remove that try/catch

lambdas/data-migration2/tests/test-granulesAndFiles.js
botanical: I'm adding that change

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  granulesSummary.dynamoRecords += 1;
+  filesSummary.dynamoRecords += files.length;
+
+  try {
+    await knex.transaction(async (trx) => {
+      const granuleCumulusId = await migrateGranuleRecord(dynamoRecord, trx);
+      await Promise.all(files.map(async (file : ApiFile) => {
+        await migrateFileRecord(file, granuleCumulusId, trx);
+      }));
```
markdboyd, line 125:
```suggestion
      return Promise.all(files.map(async (file : ApiFile) => migrateFileRecord(file, granuleCumulusId, trx)));
```

markdboyd (approved):

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: I typically am in favor of condensing code to a single line but the line exceeds 100 characters and it doesn't seem feasible to rename anything in the proposed line to condense the character count

markdboyd:
```suggestion
      return Promise.all(files.map(
        async (file : ApiFile) => migrateFileRecord(file, granuleCumulusId, trx)
      ));
```

PR #2052, opened 2021-01-20
CUMULUS-2310: Use valid filename for reconciliation report

jennyhliu:

Addresses [CUMULUS-2310: certain inventory filenames make it impossible to retrieve from the api](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2310)

## Changes

* Use valid filename for reconciliation report

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 
 ### Added
+
 - **CUMULUS-2291**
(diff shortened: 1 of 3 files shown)


flamingbear (approved):
This should work.  Verified that posting to the reconciliation report endpoint with a payload `{reportType: "Inventory", reportName: "01/22/2021 Second Test"}` generated an s3 object with name `01_22_2021 Second Test` and it was retrievable from the dashboard.

PR #2055, opened 2021-01-25
Separate out GNF from Inventory reports

flamingbear:
**Summary:**

Only add a link between granules and files when you are running a Granule Not Found report.

Addresses [CUMULUS-2351: API support to omit okCountByGranules when creating reconciliation reports](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2351)

## Changes

* Only include okCountByGranules (the link between files and granules) when the report type is Granule Not Found.

(there is no doc change as the documentation already describes this behavior https://nasa.github.io/cumulus/docs/next/features/reports#creating-a-report-via-api)

## PR Checklist

- [X] Update CHANGELOG
- [X] Unit tests
- [X] Adhoc testing
- [x] Integration tests


packages/api/lambdas/create-reconciliation-report.js
```diff
@@ -639,13 +651,17 @@ async function createReconciliationReport(recReportParams) {
...
+          const bucketGranuleCount = bucketReport.okCountByGranule[granuleId];
 
-        report.filesInCumulus.okCountByGranule[granuleId] = (currentGranuleCount || 0)
-          + bucketGranuleCount;
-      });
+          report.filesInCumulus.okCountByGranule[granuleId] = (currentGranuleCount || 0)
+            + bucketGranuleCount;
+        });
+      } else {
+        delete report.filesInCumulus.okCountByGranule;
```
kkelly51, line 663: was "okCountByGranule" ever even added if the report type wasn't GNF? Why would we have to delete it?

flamingbear: Yes, it's initialized as an empty set.  We could leave it out here, but then it would exist in the final report as {}

flamingbear: https://github.com/nasa/cumulus/pull/2055/files/fabbcb0e63d8b109425e246b7ff8648a454f429a#diff-61b4ede5087a45c0d1191b106bc92b405c40e436ec41bf30bd9b86ef5bb2d634R156

flamingbear: https://github.com/nasa/cumulus/blob/mhs/CUMULUS-2351/api-options-to-limit-reconciliation-reports/packages/api/lambdas/create-reconciliation-report.js#L156

kkelly51 (approved):

PR #2064, opened 2021-01-29
Demo bug where file verification is not enforced on re-run where duplicateHandling = skip

markdboyd:
- if your file.size and actual s3 file size don’t match AND
- duplicateHandling = skip

THEN sync-granule will error on the first run, but succeed on a successive run

Cause:

- sync-granule code does the file size verification after it has already synced the file: https://github.com/nasa/cumulus/blob/master/tasks/sync-granule/GranuleFetcher.js#L380-L383, thus even if it throws an error because the file size doesn’t match what is on file.size, then the file was still synced to the destination bucket on s3
- thus, if sync-granule runs again, it will defer handling to handleDuplicateFile :https://github.com/nasa/cumulus/blob/master/tasks/sync-granule/GranuleFetcher.js#L371
- and if duplicateHandling = skip, then that function will just return: https://github.com/nasa/cumulus/blob/master/packages/ingest/src/granule.ts#L252

### DIFF
--- a/tasks/sync-granule/tests/test-GranuleFetcher.js
+++ b/tasks/sync-granule/tests/test-GranuleFetcher.js
@@ -365,6 +365,46 @@ test('ingestFile throws an error when no checksum is provided and the size is no
   );
 });
 
+test('ingestFile throws an error on re-run when size is not as expected and duplicateHandling is "skip"', async (t) => {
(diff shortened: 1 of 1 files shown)


markdboyd:
PR was just for demonstration purposes. Closing

PR #2067, opened 2021-02-01
CUMULUS-2328: Separate s3-credentials deployment to separate module

markdboyd:
PR to re-merge changes from #2040 

#2040 was merged, but then had to be reverted because it was causing conflicts with other changes in `master` that would have created deployment headaches for users on release. So this PR should be merged **after the next major release of the current Cumulus version (5.0.1)**.


tf-modules/s3-credentials/main.tf
```diff
@@ -0,0 +1,227 @@
...
+  http_method             = aws_api_gateway_method.s3_credentials_redirect.http_method
+  integration_http_method = "POST"
+  type                    = "AWS_PROXY"
+  uri                     = aws_lambda_function.s3_credentials.invoke_arn
+}
+
+# GET /s3credentialsREADME
+resource "aws_api_gateway_resource" "s3_credentials_readme" {
+  rest_api_id = var.tea_rest_api_id
+  parent_id   = var.tea_rest_api_root_resource_id
```
nemreid, line 187: I can't deploy this because it seems you missed a number of places when you were updating the variables to get rid of the `tea_` prefix.

nemreid (changes requested):
Variables need to be fixed.

botanical (commented):
Do you need to update these docs as well? https://github.com/nasa/cumulus/blob/master/docs/deployment/components.md

tf-modules/s3-credentials/main.tf
```diff
@@ -0,0 +1,227 @@
...
+  integration_http_method = "POST"
+  type                    = "AWS_PROXY"
+  uri                     = aws_lambda_function.s3_credentials.invoke_arn
+}
+
+# GET /redirect
+resource "aws_api_gateway_resource" "s3_credentials_redirect" {
+  rest_api_id = var.rest_api_id
+  parent_id = var.rest_api_root_resource_id
+  path_part   = "redirect"
```
nemreid, line 165: indentation is not very clean here

tf-modules/s3-credentials/README.md
```diff
@@ -0,0 +1,41 @@
+# Cumulus - S3 credentials endpoint module
+
+The Cumulus S3 credentials endpoint Terraform module deploys the S3
+credentials endpoint with a configuration targeted at Cumulus and NGAP.
```
nemreid, line 4: I think it would be nice to have a little bit more detail here. As a stand-alone module, it is less obvious what purpose this serves, so calling out the nature of the credential-dispensing and the additions of `/s3credentials` to the root of the provided API Gateway would be a value-add, IMO.

nemreid (approved):
I have a few 🐜 🐜 🐜 , but overall I'm good with where this is at.

markdboyd:
@botanical 

>Do you need to update these docs as well? https://github.com/nasa/cumulus/blob/master/docs/deployment/components.md

Definitely. VERY good catch

tf-modules/s3-credentials/README.md
markdboyd: Totally. I'll add that

markdboyd: 1f608a2ac8788a4747cad792c7e62261e93bd56c

tf-modules/s3-credentials/main.tf
markdboyd: 1f608a2ac8788a4747cad792c7e62261e93bd56c

botanical (changes requested):
Apologies for not catching this sooner but I think you need to update `example/data-persistence-tf/.terraform/modules/data_persistence/packages/s3-credentials-endpoint/dist/instructions/index.md` as well

botanical (approved):
We talked offline. Docs look good to me

PR #2071, opened 2021-02-02
[CUMULUS-2362] Update logs endpoints to only work with Metrics

bt6453:

Addresses [CUMULUS-2362: Throw error on logs endpoints if Metrics isn't set](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2362)

## Changes

* Logs endpoints throws an error if Metrics is not set and no longer searches logs in Cumulus's ES instance
* Removed tests for testing the endpoint unless Metrics is set

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


dopeters (approved):
This looks good to me. Might be worth throwing a note up in the API docs that the endpoint only works with Metrics

PR #2073, opened 2021-02-02
CUMULUS-2305: Parallel PDR delete

nemreid:

Addresses [CUMULUS-2305: Parallel PDR delete](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2305)

## Changes

* Add PgModel deletion to PDRs endpoint
* Unit tests for:
  * Record in both data stores
  * Record in DDB
  * Record in RDS/Postgres

## PR Checklist

- [x] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/packages/api/endpoints/pdrs.js
+++ b/packages/api/endpoints/pdrs.js
@@ -4,6 +4,10 @@ const router = require('express-promise-router')();
 const {
   deleteS3Object,
 } = require('@cumulus/aws-client/S3');
+const {
+  getKnexClient,
+  PdrPgModel,
+} = require('@cumulus/db');
 const { inTestMode } = require('@cumulus/common/test-utils');
 const { RecordDoesNotExist } = require('@cumulus/errors');
 const Search = require('../es/search').Search;
@@ -66,9 +70,12 @@ async function del(req, res) {
   await deleteS3Object(process.env.system_bucket, pdrS3Key);
 
   const pdrModel = new models.Pdr();
+  const pdrPgModel = new PdrPgModel();
+  const knex = await getKnexClient();
 
   try {
     await pdrModel.delete({ pdrName });
+    await pdrPgModel.delete(knex, { name: pdrName });
 
     if (inTestMode()) {
       const esClient = await Search.es(process.env.ES_HOST);
--- a/packages/api/endpoints/pdrs.js
+++ b/packages/api/endpoints/pdrs.js
@@ -74,8 +74,8 @@ async function del(req, res) {
   try {
     await knex.transaction(async (trx) => {
       await pdrPgModel.delete(trx, { name: pdrName });
-      await pdrModel.delete({ pdrName });
       await deleteS3Object(process.env.system_bucket, pdrS3Key);
+      await pdrModel.delete({ pdrName });
     });
 
     if (inTestMode()) {
--- a/packages/api/endpoints/pdrs.js
+++ b/packages/api/endpoints/pdrs.js
@@ -53,6 +53,8 @@ async function get(req, res) {
   }
 }
 
+const isRecordDoesNotExistError = (e) => e.message.includes('RecordDoesNotExist');
+
 /**
  * delete a given PDR
  *
@@ -69,22 +71,23 @@ async function del(req, res) {
   const pdrPgModel = new PdrPgModel();
   const knex = await getKnexClient();
 
-  await knex.transaction(async (trx) => {
-    await pdrPgModel.delete(trx, { name: pdrName });
-    await pdrModel.delete({ pdrName });
-    await deleteS3Object(process.env.system_bucket, pdrS3Key);
-  }).catch((error) => {
-    res.boom.badRequest(`Error deleting PDR ${pdrName}`, error);
-    throw error;
-  });
+  try {
+    await knex.transaction(async (trx) => {
+      await pdrPgModel.delete(trx, { name: pdrName });
+      await pdrModel.delete({ pdrName });
+      await deleteS3Object(process.env.system_bucket, pdrS3Key);
+    });
 
-  if (inTestMode()) {
-    const esClient = await Search.es(process.env.ES_HOST);
-    await esClient.delete({
-      id: pdrName,
-      index: process.env.ES_INDEX,
-      type: 'pdr',
-    }, { ignore: [404] });
+    if (inTestMode()) {
+      const esClient = await Search.es(process.env.ES_HOST);
+      await esClient.delete({
+        id: pdrName,
+        index: process.env.ES_INDEX,
+        type: 'pdr',
+      }, { ignore: [404] });
+    }
+  } catch (error) {
+    if (!isRecordDoesNotExistError(error)) throw error;
   }
   return res.send({ detail: 'Record deleted' });
 }
--- a/packages/api/endpoints/pdrs.js
+++ b/packages/api/endpoints/pdrs.js
@@ -53,8 +53,6 @@ async function get(req, res) {
   }
 }
 
-const isRecordDoesNotExistError = (e) => e.message.includes('RecordDoesNotExist');
-
 /**
  * delete a given PDR
  *
@@ -67,28 +65,27 @@ async function del(req, res) {
 
   const pdrS3Key = `${process.env.stackName}/pdrs/${pdrName}`;
 
-  await deleteS3Object(process.env.system_bucket, pdrS3Key);
-
   const pdrModel = new models.Pdr();
   const pdrPgModel = new PdrPgModel();
   const knex = await getKnexClient();
 
-  try {
+  await knex.transaction(async (trx) => {
+    await pdrPgModel.delete(trx, { name: pdrName });
     await pdrModel.delete({ pdrName });
-    await pdrPgModel.delete(knex, { name: pdrName });
+    await deleteS3Object(process.env.system_bucket, pdrS3Key);
+  }).catch((error) => {
+    res.boom.badRequest(`Error deleting PDR ${pdrName}`, error);
+    throw error;
+  });
 
-    if (inTestMode()) {
-      const esClient = await Search.es(process.env.ES_HOST);
-      await esClient.delete({
-        id: pdrName,
-        index: process.env.ES_INDEX,
-        type: 'pdr',
-      }, { ignore: [404] });
-    }
-  } catch (error) {
-    if (!isRecordDoesNotExistError(error)) throw error;
+  if (inTestMode()) {
+    const esClient = await Search.es(process.env.ES_HOST);
+    await esClient.delete({
+      id: pdrName,
+      index: process.env.ES_INDEX,
+      type: 'pdr',
+    }, { ignore: [404] });
   }
-
   return res.send({ detail: 'Record deleted' });
 }
 
--- a/packages/api/tests/endpoints/test-pdrs.js
+++ b/packages/api/tests/endpoints/test-pdrs.js
@@ -319,10 +319,7 @@ test('DELETE removes a PDR from RDS and DynamoDB', async (t) => {
     { instanceOf: RecordDoesNotExist }
   );
 
-  t.is(
-    await pdrPgModel.exists(t.context.knex, { name: pdrName }),
-    false
-  );
+  t.false(await pdrPgModel.exists(t.context.knex, { name: pdrName }));
 });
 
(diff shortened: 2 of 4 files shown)


packages/api/endpoints/pdrs.js
```diff
@@ -66,9 +70,12 @@ async function del(req, res) {
   await deleteS3Object(process.env.system_bucket, pdrS3Key);
 
   const pdrModel = new models.Pdr();
+  const pdrPgModel = new PdrPgModel();
+  const knex = await getKnexClient();
 
   try {
     await pdrModel.delete({ pdrName });
+    await pdrPgModel.delete(knex, { name: pdrName });
```
npauzenga, line 78: Is there any value to having these in a transaction where, if the Dynamo delete fails, the PG delete is rolled back?

packages/api/tests/endpoints/test-pdrs.js
```diff
@@ -241,3 +289,93 @@ test('DELETE handles the case where the PDR exists in DynamoDb but not in S3', a
...
+  // Check Dynamo and RDS. The PDR should have been removed from both.
+  await t.throwsAsync(
+    pdrModel.get({ pdrName }),
+    { instanceOf: RecordDoesNotExist }
+  );
+
+  t.is(
+    await pdrPgModel.exists(t.context.knex, { name: pdrName }),
+    false
+  );
```
npauzenga, line 325: Unbearably small nit: I think you could do

```
t.false(await pdrPgModel.exists(t.context.knex, { name: pdrName }))
```
if you were so inclined.

packages/api/tests/endpoints/test-pdrs.js
```diff
@@ -241,3 +289,93 @@ test('DELETE handles the case where the PDR exists in DynamoDb but not in S3', a
...
+});
+
+test('DELETE removes a PDR from RDS only if no DynamoDB match exists', async (t) => {
```
npauzenga, line 328: Ah, this might squash my earlier comment about making it a transaction. So the deletes need to be totally independent? The logic here as I understand it is: 

"Delete the PDR from Dynamo ***and/or*** RDS."

I think that's a little different from what I'm doing with Granules. In that case it's complicated because of files but the way that's written now if the Dynamo record fails to delete, the the PG delete will be rolled back**.

**I'm realizing I don't have a test for this exact case so I could be wrong in what the Granule logic is doing. This is just what was intended.

npauzenga (commented):
Looks great @nemreid! I picked up the review on this because its very similar to the Granule deletes that I'm working on. Just left a couple clarifying questions/comments that could impact either this or my PR.

packages/api/endpoints/pdrs.js
nemreid: I was thinking about that, but that would mean the API has no functionality to delete a record if it doesn't exist in both places. That doesn't seem preferable? We could discuss it with the team as I am a relative newbie to the PG side of things.

packages/api/tests/endpoints/test-pdrs.js
nemreid: Yeah, granules seem more complex because there are dependent records. I'm still not sure there isn't one direction or another where it should be a transaction. Let's discuss in the architecture meeting.

packages/api/endpoints/pdrs.js
botanical: fwiw, I think we need a transaction as Nate described. Though, a DynamoDB delete can't fail if the record isn't found.
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_DeleteItem.html
> Unless you specify conditions, the DeleteItem is an idempotent operation; running it multiple times on the same item or attribute does not result in an error response.

So, I would have a transaction where we try to delete the PDR from PG first, and then attempt to delete from DynamoDB.

botanical: Another thing to keep in mind is that a "failing" PDR delete in the PG table won't return an error, it seems to return 0 if not successful and 1 otherwise.

packages/api/tests/endpoints/test-pdrs.js
botanical: Even more ✨ unbearable ✨ , you could also avoid writing `t.context.knex` twice in the test by doing
```
const { knex } = t.context;
```

packages/api/tests/endpoints/test-pdrs.js
botanical: I commented above but wanted to point out here that I don't think a DynamoDB record could _fail_ to delete since it's idempotent. And, similarly, a PG delete that is unsuccessful returns 0 when it fails, it doesn't error. We could update delete in our base model to throw an error though.

packages/api/endpoints/pdrs.js
```diff
@@ -63,25 +65,27 @@ async function del(req, res) {
...
 
-  try {
+  await knex.transaction(async (trx) => {
+    await pdrPgModel.delete(trx, { name: pdrName });
     await pdrModel.delete({ pdrName });
+    await deleteS3Object(process.env.system_bucket, pdrS3Key);
+  }).catch((error) => {
+    res.boom.badRequest(`Error deleting PDR ${pdrName}`, error);
+    throw error;
+  });
```
markdboyd, line 79: I'm nitpicking, but I'd prefer to use try/catch for consistency with other code. Also, I'm not sure what the effect of doing this is

```
res.boom.badRequest(`Error deleting PDR ${pdrName}`, error);
throw error
```

You need a `return` in front of `res.boom.badRequest` to avoid getting to `throw error`, I believe. But we're not trapping any specific error before we `throw`, so I'm wondering what the intent of this code is

packages/api/endpoints/pdrs.js
```diff
@@ -63,25 +65,27 @@ async function del(req, res) {
...
-    if (inTestMode()) {
-      const esClient = await Search.es(process.env.ES_HOST);
-      await esClient.delete({
-        id: pdrName,
-        index: process.env.ES_INDEX,
-        type: 'pdr',
-      }, { ignore: [404] });
-    }
-  } catch (error) {
-    if (!isRecordDoesNotExistError(error)) throw error;
```
markdboyd, line 82: So this error trap was ensuring that if the error thrown was "RecordDoesNotExist", then the endpoint would return 200. Basically DELETE returns 200 if the record is already deleted. Is that still true?

packages/api/tests/endpoints/test-pdrs.js
```diff
@@ -241,3 +289,84 @@ test('DELETE handles the case where the PDR exists in DynamoDb but not in S3', a
...
+  await pdrPgModel.create(t.context.knex, newPGPdr);
+
+  const response = await request(app)
+    .delete(`/pdrs/${pdrName}`)
+    .set('Accept', 'application/json')
+    .set('Authorization', `Bearer ${jwtAuthToken}`);
+  t.is(response.status, 200);
+
+  // Check Dynamo and RDS. The PDR should have been removed from both.
+  await t.throwsAsync(
```
markdboyd, line 317: Nit: I believe that the Dynamo models also have an `exists` method. But it's not necessary

packages/api/tests/endpoints/test-pdrs.js
markdboyd: @botanical I think we wouldn't want to have our delete logic throw errors for delete requests where the record no longer exists, precisely because then a delete would no longer be idempotent. 

Practically, that means a delete query to PG means the record either gets deleted or already was. Either way, the behavior of delete has been satisfied. 

You could still get an error thrown from the PG query due to deadlocking, database unavailability, or any a number of other reasons.

markdboyd: @nemreid I think this test and the next one are superfluous? They are proving basically that a DELETE request still succeeds (returns 200) and that the records in both systems definitely don't exist even if one or the other record (Dynamo or PG) already didn't exist.

However, the verification that the delete method succeeds whether the record exists or not seems like it should be at the unit test level for the model `delete` methods if it's not already. And then all we need to verify here is that a DELETE endpoint request deletes records from both systems, which the test above verifies

markdboyd: And I don't even think we need unit tests of the  model delete methods to prove that they succeed whether the record exists or not, because that is behavior of the database engines themselves, not our code

packages/api/endpoints/pdrs.js
nemreid: Rolled back that part of the change.

packages/api/endpoints/pdrs.js
nemreid: Undid problematic parts of this change.

packages/api/endpoints/pdrs.js
```diff
@@ -63,12 +67,16 @@ async function del(req, res) {
...
   const pdrModel = new models.Pdr();
+  const pdrPgModel = new PdrPgModel();
+  const knex = await getKnexClient();
 
   try {
-    await pdrModel.delete({ pdrName });
+    await knex.transaction(async (trx) => {
+      await pdrPgModel.delete(trx, { name: pdrName });
+      await pdrModel.delete({ pdrName });
+      await deleteS3Object(process.env.system_bucket, pdrS3Key);
```
markdboyd, line 78: So previously the S3 file was deleted before the dynamo object and now that order is reversed, which means you could have a record deleted from Dynamo but still have an S3 object if the S3 delete failed. That's fine with me as long as it is intentional

nemreid: I wasn't sure about the ordering on that. I do think now it would be preferable to reverse it. I'll push a change here in a minute.

packages/api/endpoints/pdrs.js
npauzenga: I think there is some ambiguity in what we're guarding against with our transactions here. I ran into these same questions on Granule deletes. 

The `Knex.del()` method will returned the number of affected rows. So if no records were deleted, you're totally right. We'll just return 0. Otherwise we'll return the number of deleted records:

>Resolves the promise / fulfills the callback with the number of affected rows for the query.

But I think thats ok and not really what we're trying to guard against here? If a transaction didn't delete anything because there wasn't anything to delete, that transactions should be considered successful. I think where we want to fail the transaction is in cases where there's an actual database error. Like if the PG database is unavailable when we tried to reach it.

So in my mind having a transaction is right and the order that @botanical suggests seems spot-on. We try a PG delete and, if that's successful, we try a Dynamo delete. Dynamo deletes cannot be rolled back so doing that second makes sense.

packages/api/tests/endpoints/test-pdrs.js
```diff
@@ -75,13 +104,32 @@ test.before(async () => {
         .then((record) => indexer.indexPdr(esClient, record, esAlias))
     )
   );
+
+  // Create a PG Collection
+  t.context.collectionCumulusId = 1234;
+  t.context.testPgCollection = fakeCollectionRecordFactory({
+    cumulus_id: t.context.collectionCumulusId,
+  });
```
npauzenga, line 112: I think this is what I had in my test-granules.js. Mark had a good point that this is kind of an anti-pattern. If we get into this habit of setting `cumulus_id` manually we could theoretically run into conflicts if we create a test later that also happens to generate the same cumulus_id. This might be cleaner:

```
  // Create a PG Collection
  t.context.testPgCollection = fakeCollectionRecordFactory();
  const collectionPgModel = new CollectionPgModel();
  [t.context.collectionCumulusId] = await collectionPgModel.create(
    t.context.knex,
    t.context.testPgCollection
  );
```

packages/db/src/models/base.ts
```diff
@@ -53,6 +53,15 @@ class BasePgModel<ItemType, RecordType extends { cumulus_id: number }> {
...
   }
+
+  delete(
+    knexOrTransaction: Knex | Knex.Transaction,
+    params: Partial<RecordType>
+  ) {
+    return knexOrTransaction(this.tableName)
+      .where(params)
+      .del();
+  }
```
npauzenga, line 64: I don't think you need to change this here but my PR will update this to be an async method so we can take advantage of promise rejections.

npauzenga (approved):
Looks great @nemreid! I added one comment about how we're setting up the test PG Collection but its not a deal-breaker. Just passing on a comment Mark made on my PR 😄

botanical (approved):
Is updating the changelog not necessary? Otherwise, ✅

PR #2074, opened 2021-02-03
CUMULUS-2321: reconciliationReport API returns pre-signed URL

jennyhliu:

Addresses [CUMULUS-2321: reconciliationReport endpoint cannot send an "overlarge" json object](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2321)

## Changes

* Updated API endpoint GET `/reconciliationReports/{name}` to return the pre-signed s3 URL in addition to report data
* If the payload size exceeds 6MB, API will return pre-signed s3 URL and an error message

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


flamingbear (approved):
I'm asking for changes, but they're all in the README, so I'll approve up front.  Nice work as always.

PR #2075, opened 2021-02-04
Fix data migrations deploy

markdboyd:
We need to have `data-migration1` be a separately deployable unit since it needs to be deployed and run **before** users deploy the version of `cumulus` which is updated to do parallel writes. Otherwise, the parallel writes will all fail since they can't find dependent objects (e.g. collections) to reference in Postgres.

Moving the deployment of data-migration2 related stuff (Lambda and ECS service) into the `cumulus` module also seemed like a benefit since it means users don't have to manually specify things like the ECS cluster ARN or log2elasticsearch Lambda ARN, since those are known values within the `cumulus` module

I also added some missing Dynamo IAM permissions for the migrations, which surely means that we didn't do a very good job of testing these migrations manually when they were initially reviewed


Jkovarik (approved):
This refactor is 💯 , nice work.   I reviewed, and tested a deployment/test invocation of migration 1 and 2.

PR #2079, opened 2021-02-08
CUMULUS-2376: CMR Revision ID Header

botanical:

Addresses [CUMULUS-2376: As a Cumulus integrator, I need post-to-cmr to pass along an optional cmr-revision-id header so that multiple workflows can update UMM-G.](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2376)

## Changes

* Added cmrRevisionId to PostToCmr as input
- Updated `publishUMMGJSON2CMR` to take in an optional `revisionId` parameter.
- Updated `publishUMMGJSON2CMR` to throw an error if optional `revisionId` does not match resulting revision ID.
- Updated `publishECHO10XML2CMR` to take in an optional `revisionId` parameter.
- Updated `publishECHO10XML2CMR` to throw an error if optional `revisionId` does not match resulting revision ID.
- Updated `publish2CMR` to take in optional `cmrRevisionId`.
- Updated `getWriteHeaders` to take in an optional CMR Revision ID.
- Updated `ingestGranule` to take in an optional CMR Revision ID to pass to `getWriteHeaders`.
- Updated `ingestUMMGranule` to take in an optional CMR Revision ID to pass to `getWriteHeaders`.

For testing, I added unit tests to test changes to `packages/cmrjs/cmr-utils.js` and `tasks/post-to-cmr/index.js` and ran the integration tests.
## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


botanical:
If anyone has any pointers on how I could do adhoc testing for this work, that would be much appreciated.

packages/cmrjs/src/cmr-utils.js
```diff
@@ -140,22 +146,32 @@ async function publishECHO10XML2CMR(cmrFile, cmrClient) {
...
   const res = await cmrClient.ingestUMMGranule(cmrFile.metadataObject);
+
   const conceptId = res['concept-id'];
+  const resultingRevisionId = res['revision-id'];
+
+  const filename = getS3UrlOfFile(cmrFile);
+  const metadataFormat = ummVersionToMetadataFormat(ummVersion(cmrFile.metadataObject));
+  const link = constructCmrConceptLink(conceptId, 'umm_json');
+
+  if (revisionId && (resultingRevisionId !== revisionId)) {
```
markdboyd, line 163: I'd have to ask @laurenfrederick how this actually works, but my understanding (that may be wrong) from the ticket was that CMR itself throws an error if you try to post to CMR using an outdated revision ID

tasks/post-to-cmr/index.js
```diff
@@ -101,10 +101,12 @@ function checkForMetadata(granules, cmrFiles) {
  * @param {boolean} event.config.skipMetaCheck - option to skip Meta file check
  * @param {Object} event.input.granules - Object of all granules where granuleID
  *    is the key
+ * @param {string} event.input.cmrRevisionId - CMR Revision ID
  * @returns {Promise<Object>} the promise of an updated event object
  */
 async function postToCMR(event) {
   // get cmr files and metadata
+  const revisionId = event.input.cmrRevisionId;
```
markdboyd, line 109: I think we need to update the `post-to-cmr` input task schema to note that this an optional input field now?

packages/api/endpoints/granules.js
```diff
@@ -44,6 +44,7 @@ async function list(req, res) {
  * @returns {Promise<Object>} the promise of express response object
  */
 async function put(req, res) {
+  const cmrRevisionId = req.headers['cmr-revision-id'];
```
markdboyd, line 47: I understand the changes to `post-to-cmr` below in the context of the ticket, but I'm not sure I understand why we're updating the granules endpoint to take a CMR revision ID, or the changes to rules. We would need these changes If PODAAC wanted to specify a CMR revision ID as input to a workflow for a granule, but my understanding of the ticket was that the tasks **within** their workflow will pull the CMR revision ID for a granule and send it to post-to-cmr.

markdboyd (commented):
I have a couple fundamental questions about approach before reviewing the code in detail

packages/cmrjs/src/cmr-utils.js
laurenfrederick: I do think that is true. We could verify in the #cmr channel.

packages/api/endpoints/granules.js
botanical: Hm I am sure I am misunderstanding how to achieve this. I was under the assumption that it would work in the following way:
```
PUT granule -> applyWorkflow -> Rule.buildPayload (takes in cmrRevisionId) -> PostToCmr
```

Is what I am missing is the `post-to-cmr` will be invoked in another way?

laurenfrederick: The CMR is usually generated by the end users within their processing tasks. The processing tasks are completely defined by them. In our fake workflow we generate our own. 

I think the use case here is that as they mess with their own CMR files they will tracking the revision-id within their workflow tasks and thus just needs to be an input to the post to cmr task.

botanical:
@markdboyd I think I addressed your concerns. Do you mind reviewing again?

packages/cmrjs/src/cmr-utils.js
markdboyd: @botanical Can you investigate this by asking in the #cmr channel? I don't think we need to throw the error from our code - I think CMR will throw an error if you send a request using an outdated revision ID

packages/cmrjs/tests/cmr-utils/test-reconcilecmrmetadata.js
```diff
@@ -365,3 +365,30 @@ test('publishUMMGJSON2CMR calls ingestUMMGranule with ummgMetadata via valid CMR
   restoreCMR();
   sinon.restore();
 });
+
+test('publishUMMGJSON2CMR calls ingestUMMGranule and throws an error due to a revision ID version conflict', async (t) => {
```
markdboyd, line 369: If I'm correct that our code doesn't need to check for revision ID conflicts, then I think this test is unnecessary

tasks/post-to-cmr/tests/cmr-test.js
```diff
@@ -95,35 +96,66 @@ test.serial('postToCMR throws error if CMR correctly identifies the xml as inval
...
+      {
+        name: 'L0A_RAD_RAW_product_0017-of-0020.cmr.json',
+        bucket: `${t.context.bucket}`,
+        filename: `s3://${t.context.bucket}/L0A_RAD_RAW_product_0017-of-0020.cmr.json`,
+        type: 'metadata',
+        fileStagingDir: 'file-staging/subdir',
+        etag: '"13f2bb38e22496fe9d42e761c42a0e67"',
+      },
+    ],
+  };
```
markdboyd, line 116: I'm trying to understand why we needed to add a second UMM-G granule to the input for this test? Were you trying to show that the revision ID gets used when posting either type of CMR metadata? 

In any case, I'm trying to understand how this test is proving the use of the revision ID for either data type. We've stubbed `ingestUMMGranule` and `ingest` to return the same revision ID as the test setup, so the fact that errors aren't thrown is what proves that it's working? 

Ultimately I think the error throwing is handled on the CMR side, so I think we need to prove is that the revision ID passed as input to this task is received by `ingest` and `ingestUMMGranule` as an argument. And then the unit tests for those functions prove whether it is used correctly.

I would remove the revision ID behavior from this test since it's supposed to be a verification that the basic case of publishing to CMR works. Revision ID is an optional and additional behavior, so I would just add two more small unit tests proving the argument is passed on to `ingest` and `ingestGranule`

packages/cmrjs/src/cmr-utils.js
```diff
@@ -173,17 +189,18 @@ async function publishUMMGJSON2CMR(cmrFile, cmrClient) {
  * @param {string} creds.username - the CMR username, not used if creds.token is provided
  * @param {string} creds.password - the CMR password, not used if creds.token is provided
  * @param {string} creds.token - the CMR or Launchpad token,
+ * @param {string} cmrRevisionId - Optional CMR Revision ID
  * if not provided, CMR username and password are used to get a cmr token
  */
-async function publish2CMR(cmrPublishObject, creds) {
+async function publish2CMR(cmrPublishObject, creds, cmrRevisionId = undefined) {
```
markdboyd, line 195: We need unit tests for `publish2CMR` to prove that the `cmrRevisionId` is passed on to `publishECHO10XML2CMR` and `publishUMMGJSON2CMR`

packages/cmrjs/src/cmr-utils.js
```diff
@@ -140,22 +146,32 @@ async function publishECHO10XML2CMR(cmrFile, cmrClient) {
  * @param {Object} cmrFile.metadataObject - the UMMG JSON cmr metadata
  * @param {Object} cmrFile.granuleId - the metadata's granuleId
  * @param {Object} cmrClient - a CMR instance
+ * @param {string} revisionId - Optional CMR Revision ID
  * @returns {Object} CMR's success response which includes the concept-id
  */
-async function publishUMMGJSON2CMR(cmrFile, cmrClient) {
+async function publishUMMGJSON2CMR(cmrFile, cmrClient, revisionId = undefined) {
```
markdboyd, line 152: We need a unit test here to prove that the cmrRevisionId is correctly passed on to `ingestUMMGranule`

packages/cmrjs/src/cmr-utils.js
```diff
@@ -113,13 +113,19 @@ function granulesToCmrFileObjects(granules) {
  * @param {string} cmrFile.filename - the s3 uri to the cmr xml file
  * @param {string} cmrFile.metadata - granule xml document
  * @param {Object} cmrClient - a CMR instance
+ * @param {string} revisionId - Optional CMR Revision ID
  * @returns {Object} CMR's success response which includes the concept-id
  */
-async function publishECHO10XML2CMR(cmrFile, cmrClient) {
+async function publishECHO10XML2CMR(cmrFile, cmrClient, revisionId = undefined) {
```
markdboyd, line 119: We need a unit test here to prove that the cmrRevisionId is correctly passed on to `ingestGranule`

packages/cmrjs/tests/cmr-utils/test-reconcilecmrmetadata.js
botanical: Ah, I think you are correct. Reading the CMR docs, I believe CMR will throw the error.

tasks/post-to-cmr/tests/cmr-test.js
botanical: > I'm trying to understand why we needed to add a second UMM-G granule to the input for this test? Were you trying to show that the revision ID gets used when posting either type of CMR metadata?

Yes, exactly

markdboyd: OK. Ultimately based on the understanding that CMR is what will throw the error, I think all we need to prove is that we've passed along the revision ID correctly to the code that publishes to CMR. 

So I would revert the changes to this test, which is just supposed to be a basic demonstration that the Lambda handler succeeds, and add one small unit test proving the argument is passed on to publish2Cmr.

packages/cmrjs/src/cmr-utils.js
botanical: I confirmed in the CMR channel

tasks/post-to-cmr/tests/cmr-test.js
botanical: Hm, why is it wrong to update the test to account for ingesting a UMM-G granule?

packages/cmrjs/src/cmr-utils.js
botanical: [`6638647` (#2079)](https://github.com/nasa/cumulus/pull/2079/commits/663864799422184f64130dadd8e00e8930a2606e)

packages/cmrjs/src/cmr-utils.js
botanical: [`4e5ded5` (#2079)](https://github.com/nasa/cumulus/pull/2079/commits/4e5ded590cd0d0e529bd197ab669e5cfe719ec61)

packages/cmrjs/src/cmr-utils.js
botanical: [`4e5ded5` (#2079)](https://github.com/nasa/cumulus/pull/2079/commits/4e5ded590cd0d0e529bd197ab669e5cfe719ec61)

packages/cmrjs/src/cmr-utils.js
```diff
@@ -113,15 +113,23 @@ function granulesToCmrFileObjects(granules) {
...
+  const cmrRevisionId = res.result['revision-id'];
+  const resultLog = `Published ${cmrFile.granuleId} to the CMR. conceptId: ${conceptId}`;
+  const revisionIdLog = `, revisionId: ${cmrRevisionId}`;
 
-  log.info(`Published ${cmrFile.granuleId} to the CMR. conceptId: ${conceptId}`);
+  if (revisionId) {
+    log.info(resultLog + revisionIdLog);
+  } else {
+    log.info(resultLog);
+  }
```
markdboyd, line 132: Code style nit, but I think this is a bit more clear

```
let resultLog = `Published ${cmrFile.granuleId} to the CMR. conceptId: ${conceptId}`;
if (revisionId) resultLog += `, revisionId: ${cmrRevisionId}`;

log.info(resultLog);
```

packages/cmrjs/src/cmr-utils.js
```diff
@@ -140,22 +148,35 @@ async function publishECHO10XML2CMR(cmrFile, cmrClient) {
...
 
-  const res = await cmrClient.ingestUMMGranule(cmrFile.metadataObject);
   const conceptId = res['concept-id'];
+  const cmrRevisionId = res['revision-id'];
+
+  const filename = getS3UrlOfFile(cmrFile);
+  const metadataFormat = ummVersionToMetadataFormat(ummVersion(cmrFile.metadataObject));
+  const link = constructCmrConceptLink(conceptId, 'umm_json');
+  const resultLog = `Published UMMG ${granuleId} to the CMR. conceptId: ${conceptId}`;
+  const revisionIdLog = `, revisionId: ${cmrRevisionId}`;
```
markdboyd, line 165: Same comment as above re: code style

packages/cmrjs/tests/cmr-utils/test-cmr-utils.js
```diff
@@ -729,3 +733,87 @@ test(
...
+    Bucket: bucket,
+    Key: xmlKey,
+    Body: echoMetadataString,
+  });
+
+  const conceptId = 'id204842';
+  const metadataObject = await metadataObjectFromCMRFile(buildS3Uri(bucket, xmlKey));
+  const updatedXmlFile = { ...xmlFile, granuleId: xmlGranuleId, metadataObject };
+  const logInfoSpy = sinon.spy(log, 'info');
+  const ingestGranuleSpy = sinon.stub(CMR.prototype, 'ingestGranule').returns({ result: { 'concept-id': conceptId, 'revision-id': cmrRevisionId } });
```
markdboyd, line 761: So I think that this test (and really unit tests) in general should only spy on their direct dependencies. So this test should only verify that it passes the revision ID into `publishECHO10XML2CMR`, not `ingestGranule` which is used by `publishECHO10XML2CMR`. Then it is the job of the unit test for `publishECHO10XML2CMR` to verify that it passes the revision ID into `ingestGranule`. 

I'm sure we're breaking this rule all over the place, but this should be the goal. Otherwise you're unit testing the dependencies of your code, not the unit of code itself.

packages/cmrjs/tests/cmr-utils/test-cmr-utils.js
```diff
@@ -729,3 +733,87 @@ test(
...
+
+  t.teardown(() => {
+    ingestGranuleSpy.restore();
+    logInfoSpy.restore();
+  });
+  await cmrUtil.publish2CMR(updatedXmlFile, credentials, cmrRevisionId);
+  t.true(logInfoSpy.calledOnceWith(`Published ${updatedXmlFile.granuleId} to the CMR. conceptId: ${conceptId}, revisionId: ${cmrRevisionId}`));
```
markdboyd, line 768: I don't think it's very important to confirm the output of log messages. It also makes tests somewhat brittle because log messages are subject to change. Personally I'd just remove this element from these tests

packages/cmrjs/tests/cmr-utils/test-cmr-utils.js
```diff
@@ -729,3 +733,87 @@ test(
...
+    Bucket: bucket,
+    Key: jsonKey,
+    Body: ummgMetadataString,
+  });
+
+  const conceptId = 'id204842';
+  const metadataObject = await metadataObjectFromCMRFile(buildS3Uri(bucket, jsonKey));
+  const updatedJsonFile = { ...jsonFile, granuleId: jsonGranuleId, metadataObject };
+  const logInfoSpy = sinon.spy(log, 'info');
+  const ingestUMMGranuleSpy = sinon.stub(CMR.prototype, 'ingestUMMGranule').returns({ 'concept-id': conceptId, 'revision-id': cmrRevisionId });
```
markdboyd, line 809: Same feedback on this test: we only need to verify that the revision ID is passed into `publishUMMGJSON2CMR`

tasks/post-to-cmr/tests/cmr-test.js
```diff
@@ -94,36 +95,67 @@ test.serial('postToCMR throws error if CMR correctly identifies the xml as inval
   }
 });
 
-test.serial('postToCMR succeeds with correct payload', async (t) => {
-  const newPayload = t.context.payload;
+test.serial.only('postToCMR succeeds with correct payload', async (t) => {
```
markdboyd, line 98: Remove `.only`

tasks/post-to-cmr/tests/cmr-test.js
markdboyd: I guess it's somewhat philosophical. I think unit tests in general should only be testing one code path through your code, otherwise it's actually a unit test of multiple things, not just one. And the logic for updating UMM-G granules vs ECHO-10 granules are separate code paths, so if we're going to test both here, I'd say they should be separate unit tests.

But more fundamentally, `post-to-cmr` just calls `publish2Cmr`, which is where the logic for ECHO10 vs UMM-G is handled. So I'm not sure `post-to-cmr` itself needs to test the code path for both types of granules in the name of good test isolation. That seems like a job for tests of `publish2Cmr`.

So I'd be in favor of removing the UMM-G testing logic from this test and adding a separate test just to prove that the revision ID from the input to post-to-cmr is passed into publish2Cmr

packages/cmrjs/src/cmr-utils.js
markdboyd: Separately, I think `cmrRevisionId = undefined` is unnecessary. It will be undefined if you don't specify a value at call time

packages/cmrjs/src/cmr-utils.js
```diff
@@ -113,15 +113,23 @@ function granulesToCmrFileObjects(granules) {
  * @param {string} cmrFile.filename - the s3 uri to the cmr xml file
  * @param {string} cmrFile.metadata - granule xml document
  * @param {Object} cmrClient - a CMR instance
+ * @param {string} revisionId - Optional CMR Revision ID
  * @returns {Object} CMR's success response which includes the concept-id
  */
-async function publishECHO10XML2CMR(cmrFile, cmrClient) {
+async function publishECHO10XML2CMR(cmrFile, cmrClient, revisionId = undefined) {
```
markdboyd, line 119: `revisionId = undefined` is unnecessary

packages/cmrjs/src/cmr-utils.js
```diff
@@ -140,22 +148,35 @@ async function publishECHO10XML2CMR(cmrFile, cmrClient) {
  * @param {Object} cmrFile.metadataObject - the UMMG JSON cmr metadata
  * @param {Object} cmrFile.granuleId - the metadata's granuleId
  * @param {Object} cmrClient - a CMR instance
+ * @param {string} revisionId - Optional CMR Revision ID
  * @returns {Object} CMR's success response which includes the concept-id
  */
-async function publishUMMGJSON2CMR(cmrFile, cmrClient) {
+async function publishUMMGJSON2CMR(cmrFile, cmrClient, revisionId = undefined) {
```
markdboyd, line 154: `revisionId = undefined` is unnecessary

markdboyd (changes requested):

packages/cmrjs/src/cmr-utils.js
botanical: For some reason I thought it was necessary to signify that the parameter was optional 🤔

packages/cmrjs/tests/cmr-utils/test-cmr-utils.js
botanical: That makes sense to me. Updated

tasks/post-to-cmr/tests/cmr-test.js
botanical: Ah, that is true. That makes sense to me. I've updated the unit tests so that `publish2CMR` checks for both `ECHO10` and `UMM-G`.

CHANGELOG.md
```diff
@@ -44,6 +44,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 - **CUMULUS-2255**
   - Updated Terraform deployment code syntax for compatibility with version 0.13.6
+- **CUMULUS-2376**
```
markdboyd, line 47: Let's also add a changelog entry under "Added" explaining that `post-to-cmr` now accepts this optional revision ID as part of its task input and will use it when publishing metadata to CMR

packages/cmrjs/src/cmr-utils.js
```diff
@@ -113,15 +113,19 @@ function granulesToCmrFileObjects(granules) {
...
  * @returns {Object} CMR's success response which includes the concept-id
  */
-async function publishECHO10XML2CMR(cmrFile, cmrClient) {
+async function publishECHO10XML2CMR(cmrFile, cmrClient, revisionId) {
   const builder = new xml2js.Builder();
   const xml = builder.buildObject(cmrFile.metadataObject);
-  const res = await cmrClient.ingestGranule(xml);
+  const res = await cmrClient.ingestGranule(xml, revisionId);
   const conceptId = res.result['concept-id'];
+  const cmrRevisionId = res.result['revision-id'];
```
markdboyd, line 124: Does a `revision-id` come back in the result from CMR? If so, what is the reason for using that one over the one provided as an argument? 

If the specified revisionId was wrong, then CMR would error on line 122 and not make it to the log, as I understand it. So this seems slightly unnecessary unless I'm missing something.

packages/cmrjs/tests/cmr-utils/test-cmr-utils.js
```diff
@@ -729,3 +732,159 @@ test(
...
+
+  const metadataObject = await cmrUtil.metadataObjectFromCMRFile(buildS3Uri(bucket, xmlKey));
+  const updatedXmlFile = { ...xmlFile, granuleId: xmlGranuleId, metadataObject };
+  const publishECHO10XML2CMRSpy = sinon.spy(() => Promise.resolve());
+  const revertPublishECHO10XML2CMRSpy = cmrUtil.__set__('publishECHO10XML2CMR', publishECHO10XML2CMRSpy);
+
+  t.teardown(() => {
+    revertPublishECHO10XML2CMRSpy();
+  });
+  await cmrUtil.publish2CMR(updatedXmlFile, credentials, cmrRevisionId);
```
markdboyd, line 764: Since we're mocking `publishECHO10XML2CMR` and spying on it, do we need to go through all of the above setup for fixtures for this test? We could pass in an object only containing an echo 10 `filename` property for the first argument and literally anything for `credentials` and I think this test would still pass. The reason I ask is that if the test conditions can be simplified and yet still provide the same guarantees, it's preferable because it can make what the test is doing more obvious.

The same goes for the test of `publishUMMGJSON2CMR`.

packages/cmrjs/tests/cmr-utils/test-cmr-utils.js
```diff
@@ -729,3 +732,159 @@ test(
...
+  // const cmrFileObject = { filename: 'test.cmr.xml', granuleId: 'testGranuleId' };
+  const bucket = randomString();
+  const payloadPath = path.join(__dirname, 'data', 'payload.json');
+  const rawPayload = fs.readFileSync(payloadPath, 'utf8');
+  const payload = JSON.parse(rawPayload);
+  const xmlFile = payload.input.granules[0].files[3];
+  const xmlGranuleId = payload.input.granules[0].granuleId;
+  const xmlKey = `${xmlGranuleId}.cmr.xml`;
+  const echoMetadataString = fs.readFileSync(path.join(__dirname, 'data/meta.xml'));
+  const cmrRevisionId = 12;
```
markdboyd, line 746: Let's randomize the cmrRevisionId in all these tests using `Math.random()`

packages/cmrjs/tests/cmr-utils/test-reconcilecmrmetadata.js
```diff
@@ -350,7 +350,7 @@ test('publishUMMGJSON2CMR calls ingestUMMGranule with ummgMetadata via valid CMR
     granuleId: 'fakeGranuleID',
   };
   const publishUMMGJSON2CMR = cmrUtils.__get__('publishUMMGJSON2CMR');
-  const ingestFake = sinon.fake.resolves({ result: { 'concept-id': 'fakeID' } });
+  const ingestFake = sinon.fake.resolves({ 'concept-id': 'fakeID', 'revision-id': 123 });
```
markdboyd, line 353: Is it necessary to alter this mock to return a `revision-id`?

packages/cmrjs/tests/cmr-utils/test-reconcilecmrmetadata.js
```diff
@@ -359,7 +359,7 @@ test('publishUMMGJSON2CMR calls ingestUMMGranule with ummgMetadata via valid CMR
   await publishUMMGJSON2CMR(cmrPublishObject, new CmrFake());
 
   // Assert
-  t.true(ingestFake.calledOnceWithExactly(cmrPublishObject.metadataObject));
+  t.true(ingestFake.calledOnceWithExactly(cmrPublishObject.metadataObject, undefined));
```
markdboyd, line 362: We have to assert `undefined` since no revisionId was passed to `publishUMMGJSON2CMR`?

botanical: Yes since the test is checking for `calledOneWithExactly`

packages/cmrjs/tests/cmr-utils/test-reconcilecmrmetadata.js
botanical: It's not and I don't have a good reason for adding it except I thought the next person might find it helpful to know that it could return a revision id.

packages/cmrjs/src/cmr-utils.js
botanical: Oh, that's a good point. I suppose if they are technically the same, we could just use the revision ID the user provides.

packages/cmrjs/tests/cmr-utils/test-cmr-utils.js
botanical: 😱 I didn't realize we could pass anything in for credentials but taking a closer look, that makes sense now since I am mocking the ingest functions.

botanical:
@markdboyd I think I have addressed all of your feedback

packages/cmrjs/tests/cmr-utils/test-cmr-utils.js
```diff
@@ -729,3 +730,67 @@ test(
...
   testMetadataObjectFromCMRFile('s3://bucket/fake.cmr.json')
 );
+
+test.serial('publish2CMR passes cmrRevisionId to publishECHO10XML2CMR', async (t) => {
+  const cmrFileObject = { filename: 'test.cmr.xml', granuleId: 'testGranuleId' };
+  const updatedXmlFile = { ...cmrFileObject, metadataObject: {} };
+
+  const publishECHO10XML2CMRSpy = sinon.spy(() => Promise.resolve());
+  const revertPublishECHO10XML2CMRSpy = cmrUtil.__set__('publishECHO10XML2CMR', publishECHO10XML2CMRSpy);
+  const credentials = await getCmrSettings();
```
markdboyd, line 740: Nit, but avoiding the extra call and implicit test of `getCmrSettings` seems ideal

```suggestion
  const credentials = {};
```

packages/cmrjs/tests/cmr-utils/test-cmr-utils.js
```diff
@@ -729,3 +730,67 @@ test(
...
+  const cmrRevisionId = Math.random;
```
markdboyd, line 741: Sorry, I should have been more clear. `Math.random` isn't a function call, it's just a reference to a function of the `Math` global.

You would need to use: `Math.floor(Math.random() * 100)`

packages/cmrjs/tests/cmr-utils/test-cmr-utils.js
botanical: Ah, nice catch! I thought I got all of them.

packages/cmrjs/tests/cmr-utils/test-cmr-utils.js
botanical: Oh, that makes sense.

packages/cmrjs/tests/cmr-utils/test-cmr-utils.js
```diff
@@ -729,3 +730,67 @@ test(
...
+  t.is(publishUMMGJSON2CMRSpy.getCall(0).args[2], cmrRevisionId);
+});
+
+test.serial('publishECHO10XML2CMR passes cmrRevisionId to ingestGranule', async (t) => {
+  const cmrFileObject = { filename: 'test.cmr.xml', granuleId: 'testGranuleId', metadataObject: {} };
+  const conceptId = randomString();
+  const credentials = {};
+  const cmrRevisionId = Math.floor(Math.random() * 100);
+
+  const ingestGranuleSpy = sinon.stub(CMR.prototype, 'ingestGranule').returns({ result: { 'concept-id': conceptId, 'revision-id': cmrRevisionId } });
```
markdboyd, line 772: Nit: I'm wondering if we should bother including `revision-id` in our mock here. Even though CMR does return that value (is this true?), we're not really relying on that in our code anymore, so I'm hesitant to mock it because it seems like we do need it

tasks/post-to-cmr/tests/cmr-test.js
```diff
@@ -19,7 +19,8 @@ const { isCMRFile } = require('@cumulus/cmrjs');
 const { postToCMR } = require('..');
 
 const result = {
-  'concept-id': 'testingtesting',
+  'concept-id': 'G1222482316-CUMULUS',
+  'revision-id': 123,
```
markdboyd, line 23: Same comment as above: since I don't think we really use revision-id from the return, I'm hesitant to include it in the mock

markdboyd (approved):

PR #2084, opened 2021-02-09
add tags to API gateways and TEA cloudformation stack

markdboyd:
While dealing with cleaning up old stacks, I noticed some resources that didn't have tags, which makes them much harder to find and remove. So I added tags to them.


Jkovarik (approved):
Looks good to me 💯

PR #2090, opened 2021-02-15
Add information on Cumulus sizing configuration to deployment documen…

laurenfrederick:
…tation

**Summary:** Add info on sizing your Cumulus instance to deployment docs

This was an action that came out of the IWG

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


docs/deployment/README.md
```diff
@@ -505,6 +509,37 @@ You should be able to visit the dashboard website at `http://<prefix>-dashboard.
...
+### Elasticsearch
+
+The [mappings file](https://github.com/nasa/cumulus/blob/master/packages/api/models/mappings.json) contains all of the data types that will be indexed into Elasticsearch. Elasticsearch sizing is tied to your archive size, your collections, granules, workflow executions that will be stored.
+
+AWS provides [documentation](https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/sizing-domains.html) on calculating and configuring for sizing.
+
+In addition to size you'll want to consider the [number of nodes](https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es-managedomains-dedicatedmasternodes.html) which determine how the system reacts in the event of a failure.
+
+
+Configuration can be done in the [data persistence module](https://github.com/nasa/cumulus/blob/0e4336d818ad79e259918f1f4b06fa902dd4b69c/tf-modules/data-persistence/variables.tf#L16) and the [cumulus module](https://github.com/nasa/cumulus/blob/0e4336d818ad79e259918f1f4b06fa902dd4b69c/tf-modules/cumulus/variables.tf#L541).
```
markdboyd, line 527: I think I would remove the commit SHA from these links and just point to the versions on the `master` branch? It might be good to reference the variable names since the line numbers could change as we make changes

docs/deployment/README.md
```diff
@@ -249,6 +249,8 @@ In `terraform.tf`, configure the remote state settings by substituting the appro
 
 Fill in the appropriate values in `terraform.tfvars`. See the [data-persistence module variable definitions](https://github.com/nasa/cumulus/blob/master/tf-modules/data-persistence/variables.tf) for more detail on each variable.
 
+Consider [the size of your Elasticsearch cluster](#elasticsearch) when configuring data-persistence.
```
markdboyd, line 252: Is there a reason to not include the sizing information directly in the relevant sections? I worry about it getting buried at the bottom of the document, though it is linked. I don't feel that strongly either way

docs/deployment/README.md
```diff
@@ -505,6 +509,37 @@ You should be able to visit the dashboard website at `http://<prefix>-dashboard.
 
 ---
 
+## Cumulus Instance Sizing
+
+The Cumulus deployment default sizing for Elasticsearch instances, EC2 instances, and Autoscaling Groups are small and design for testing and cost savings. These are likely not suitable for production workloads. Sizing his highly individual and dependent on expected load and archive size.
```
markdboyd, line 514:
```suggestion
The Cumulus deployment default sizing for Elasticsearch instances, EC2 instances, and Autoscaling Groups are small and designed for testing and cost savings. The default settings are likely not suitable for production workloads. Sizing his highly individual and dependent on expected load and archive size.
```

docs/deployment/README.md
```diff
@@ -505,6 +509,37 @@ You should be able to visit the dashboard website at `http://<prefix>-dashboard.
...
+
+> Please be cognizant of costs as any change in size will affect your AWS bill. AWS provides a [pricing calculator](https://calculator.aws/#/) for estimating costs.
+
+### Elasticsearch
+
+The [mappings file](https://github.com/nasa/cumulus/blob/master/packages/api/models/mappings.json) contains all of the data types that will be indexed into Elasticsearch. Elasticsearch sizing is tied to your archive size, your collections, granules, workflow executions that will be stored.
```
markdboyd, line 520:
```suggestion
The [mappings file](https://github.com/nasa/cumulus/blob/master/packages/api/models/mappings.json) contains all of the data types that will be indexed into Elasticsearch. Elasticsearch sizing is tied to your archive size, including your collections, granules, and workflow executions that will be stored.
```

markdboyd (changes requested):
Minor requested changes

docs/deployment/README.md
laurenfrederick: Yes, you're right. Good catch! I don't know how I got the wrong one.

laurenfrederick: Oh ok figured it out. I copied the permalink to that, so that will always point to that even when if something is added to the file and the line number changes. Do you think I should leave it or change it?

docs/deployment/README.md
laurenfrederick: So historically the documentation has gotten out of control and it makes it hard for users who want to just get straight to deployment, especially in sandbox.

docs/deployment/README.md
markdboyd: I guess I worry about using permalinks with SHAs because eventually we're pointing to outdated code which could get a bit confusing. So I think overall I'd rather point to the latest code and call out the variable names

laurenfrederick: Fixed

markdboyd (approved):

PR #2093, opened 2021-02-16
Merge 6.0.0 back to master

markdboyd:


laurenfrederick (approved):

laurenfrederick (changes requested):
So the `release-6.0.x` branch needs `master` merged in to be able to merge.. which will affect the integrity of the 6.0.x branch. What I usually do is create a branch off of 6.0.x and merge that into master so master can be merged into that and then it can be deleted.

PR #2099, opened 2021-02-19
CUMULUS-2355: Add logic to disable s3CredentialsEndpoint

botanical:

Addresses [CUMULUS-2355: Rapidly Disable S3 Credentials](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2355)

## Changes

* Added `DISABLE_S3_CREDENTIALS` environment variable to allow for disabling of S3 credentials endpoint lambda without redeploying.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [ ] Integration tests

## Testing notes

Tested by setting `DISABLE_S3_CREDENTIALS` to `true` on `s3-credentials-endpoint` lambda.
Screen Shot 2021-02-26 at 11 55 59 AM

Tested by setting `DISABLE_S3_CREDENTIALS` to `false` and by removing the environment variable and confirmed that I receive expected output.


CHANGELOG.md
```diff
@@ -56,6 +56,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     custom security group mappings
 - **CUMULUS-2402**
   - Officially expose `sftp()` for use in `@cumulus/sftp-client`
+- **CUMULUS-2355**
+  - Add logic to disable `s3CredentialsEndpoint` based upon value for environment variable `DISABLE_S3_CREDENTIALS`. If set to "true", it will
+    disable the lambda.
```
markdboyd, line 61: This isn't quite true - if the environment variable is set then it will disable the dispensing of S3 credentials from the endpoint, but the Lambda will still run. Also we should give the name of the Lambda for context

```suggestion
  - Added logic to disable `/s3Credentials` endpoint based upon value for environment variable `DISABLE_S3_CREDENTIALS`. If set to "true",  the endpoint will not dispense S3 credentials and instead return a message indicating that the endpoint has been disabled.
```

botanical: Oh, that's true. I will update for clarity

packages/s3-credentials-endpoint/tests/test-index.js
```diff
@@ -342,3 +342,21 @@ test('displayS3Credentials fills template with correct distribution endpoint.',
...
+
+  const response = await request(distributionApp)
+    .get('/s3credentials')
+    .set('Accept', 'application/json')
+    .set('Cookie', [`accessToken=${accessTokenRecord.accessToken}`])
+    .expect(503);
+
+  t.is(response.status, 503);
+  t.teardown(() => {
+    process.env.DISABLE_S3_CREDENTIALS = undefined;
```
markdboyd, line 360:
```suggestion
    delete process.env.DISABLE_S3_CREDENTIALS ;
```

packages/s3-credentials-endpoint/tests/test-index.js
```diff
@@ -342,3 +342,21 @@ test('displayS3Credentials fills template with correct distribution endpoint.',
...
+  const accessTokenRecord = fakeAccessTokenFactory({ username });
+  await accessTokenModel.create(accessTokenRecord);
+
+  const response = await request(distributionApp)
+    .get('/s3credentials')
+    .set('Accept', 'application/json')
+    .set('Cookie', [`accessToken=${accessTokenRecord.accessToken}`])
+    .expect(503);
+
+  t.is(response.status, 503);
```
markdboyd, line 358: Given that the ticket AC mentions that the endpoint when disabled should respond with an informational message, I think we should test that in this unit test. Generally I don't like testing endpoint response messages because its brittle but in this case the ticket seems to call for it.

CHANGELOG.md
markdboyd: Also, it seems like documentation on this is called for somewhere, but we don't seem to have any public docs about managing this endpoint for integrators. Failing that, maybe adding a README to `packages/s3-credentials-endpoint` that calls it out would be good

markdboyd: Actually, since this is an NGAP-only feature, we have a wiki page that would be worth updating instead:

https://wiki.earthdata.nasa.gov/display/CUMULUS/Cumulus+Direct+S3+Access

markdboyd:
💯 great job on the manual testing. Just a few small comments here

packages/s3-credentials-endpoint/tests/test-index.js
botanical: [`e47343b` (#2099)](https://github.com/nasa/cumulus/pull/2099/commits/e47343b5c6d79c1a4e9c81f16100ada84039450d)

packages/s3-credentials-endpoint/tests/test-index.js
botanical: [`e47343b` (#2099)](https://github.com/nasa/cumulus/pull/2099/commits/e47343b5c6d79c1a4e9c81f16100ada84039450d)

CHANGELOG.md
botanical: [`fc51a9a` (#2099)](https://github.com/nasa/cumulus/pull/2099/commits/fc51a9a280c17bb4b8c023ed01d7e027f62336dc)

botanical: Should I update the [wiki page](https://wiki.earthdata.nasa.gov/display/CUMULUS/Cumulus+Direct+S3+Access) after the feature goes out? I don't want to cause confusion by adding it prematurely. @markdboyd

markdboyd: Ah good call. Yeah I guess we'll have to just update it after this change has been released

packages/s3-credentials-endpoint/tests/test-index.js
markdboyd: I **think** it's possible to just check the value of `response.message` and not have to compare against the full JSON response

markdboyd: Sorry, I should have been more explicit. You can check `response.body.message` to get the message value that you want to test directly:

https://github.com/nasa/cumulus/blob/master/packages/api/tests/endpoints/test-granules.js#L341

botanical: [`0ba4cb5` (#2099)](https://github.com/nasa/cumulus/pull/2099/commits/0ba4cb51ffe4fec17799f3419c73f7c327f2824e)

CHANGELOG.md
```diff
@@ -56,6 +56,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     custom security group mappings
 - **CUMULUS-2402**
   - Officially expose `sftp()` for use in `@cumulus/sftp-client`
+- **CUMULUS-2355**
```
markdboyd, line 59: This CHANGELOG note is nested under the 6.0.0 release right now, but it should be moved up to the "Unreleased" section

CHANGELOG.md
```diff
@@ -24,9 +24,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - `<prefix>-lambda-api-gateway` IAM role used by API Gateway Lambda now supports accessing all buckets defined in your `buckets` variable except "internal" buckets
 - **CUMULUS-2362**
   - Logs endpoints only work with Metrics set up
-
 - **CUMULUS-2350**
   - Updates the examples on the `/s3credentialsREADME`, to include Python and JavaScript code demonstrating how to refresh the s3credential for programatic access.
+- **CUMULUS-2355**
```
markdboyd, line 29: This nodte is still nested under the release 7.0.0  instead of "Unreleased"

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
+
+### Changes
```
markdboyd, line 9:
```suggestion
### Changed
```

Nit: for consistency

markdboyd (approved):

PR #2107, opened 2021-02-22
Adds SIT var for FAKE_PROVIDER_CONFIG_BUCKET

flamingbear:
Override SIT variable in script rather than in bamboo branch config.

There's no tests, except the SIT Tests (master) plan in bamboo will pass after this is merged.


Jkovarik (approved):
👍🏻

PR #2108, opened 2021-02-22
Feature/cumulus 2309 granule bulk deletes

npauzenga:

Addresses [CUMULUS-2309: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2309)

## Changes

* Updates the bulk delete logic to handle PG files

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [ ] Integration tests


packages/api/lambdas/bulk-operation.js
```diff
@@ -118,26 +128,67 @@ function applyWorkflowToGranules({
...
+      if (dynamoGranule && dynamoGranule.published && forceRemoveFromCmr) {
+        updateResponse = await unpublishGranule(knex, dynamoGranule);
+      }
+
+      // Delete the Dynamo Granule, the Postgres Granule (if one was found),
+      // and associated files.
+      await deleteGranuleAndFiles({
+        knex,
+        dynamoGranule: updateResponse ? updateResponse.dynamoGranule : dynamoGranule,
+        pgGranule: updateResponse ? updateResponse.pgGranule : pgGranule,
```
npauzenga, line 188: pgGranule could be undefined here and `deleteGranuleAndFiles` should handle that.

packages/api/lib/granule-delete.js
```diff
@@ -64,7 +64,7 @@ const deleteGranuleAndFiles = async ({
         }
       );
 
-      await granulePgModel.delete(trx, { granule_id: pgGranule.granule_id });
+      await granulePgModel.delete(trx, { cumulus_id: pgGranule.cumulus_id });
```
npauzenga, line 67: Already updated in https://github.com/nasa/cumulus/pull/2091

Made the change here to test changes.

packages/api/lib/granule-remove-from-cmr.js
```diff
@@ -60,15 +62,19 @@ const unpublishGranule = async (knexOrTransaction, granule) => {
...
     if (!(error instanceof RecordDoesNotExist)) {
       throw error;
     }
   }
 
-  return granuleModelClient.update({ granuleId: granule.granuleId }, { published: false }, ['cmrLink']);
+  return {
+    dynamoGranule: await granuleModelClient.update({ granuleId: granule.granuleId }, { published: false }, ['cmrLink']),
+    pgGranule: pgGranule,
+  };
```
npauzenga, line 77: pgGranule might be undefined here and that should be expected.

Previously we weren't using the return value so this change is geared specifically for bulk deletes.

packages/api/tests/db-data-helpers/create-test-data.js
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Helper for creating a granule, a parent collection,
+ * and files belonging to that granule (in S3 and PG)
+ * @param {Knex} dbClient - Knex client
+ * @param {number} collectionId - collectionId for the granule's parent collection
+ * @param {number} collectionCumulusId - cumulus_id for the granule's parent collection
+ * @param {boolean} published - if the granule should be marked published to CMR
+ * @returns {Object} fake granule object
+ */
+async function createGranuleAndFiles({
```
npauzenga, line 40: Moving this to a separate file because we're using it in 3 places. This could definitely be made more generic. Right now it does a lot. If this proves valuable it could be split up to handle things a little nicer.

packages/api/tests/lambdas/test-bulk-operation.js
```diff
@@ -157,6 +207,7 @@ test('bulk operation lambda throws error for unknown event type', async (t) => {
   }));
 });
 
+// This test must run for the following tests to pass
```
npauzenga, line 210: Adding this because I was confused why `only` was failing. We need this test to set ENVs.

packages/api/tests/lib/test-granule-remove-from-cmr.js
```diff
@@ -36,44 +31,6 @@ test.before(async (t) => {
   process.env.GranulesTable = randomString();
   await new Granule().createTable();
 
-  testCumulusMessage = {
```
npauzenga, line 39: This is all just cleanup. Not related to the functional changes in this PR.

packages/api/lambdas/bulk-operation.js
```diff
@@ -118,26 +128,67 @@ function applyWorkflowToGranules({
...
+        pgGranule = await granulePgModel.get(knex,
+          {
+            granule_id: granuleId,
+            collection_cumulus_id: collectionCumulusId,
+          });
+      } catch (error) {
+        if (!(error instanceof RecordDoesNotExist)) {
+          throw error;
+        }
+      }
```
npauzenga, line 173: Yeah. I know we don't really want to get these records before deleting. Totally open to refactoring suggestions here. This was a compromise to try and avoid refactoring a ton of stuff from the previous work.

packages/api/tests/endpoints/test-granules.js
```diff
@@ -737,15 +639,13 @@ test('DELETE returns 404 if granule does not exist', async (t) => {
   t.true(response.body.message.includes('No record found'));
 });
 
-test('DELETE deleting an existing granule that is published will fail and not delete records', async (t) => {
-  const { s3Buckets, newGranule } = await createGranuleAndFiles({
+test.serial('DELETE deleting an existing granule that is published will fail and not delete records', async (t) => {
```
npauzenga, line 642: This is hopefully temporary. I'm seeing a failure in CI that I don't see locally in these tests. I'm trying to run these in serial to see if that's the issue. If it is, there might be something in the way I'm creating test granules that needs to change.

packages/api/tests/db-data-helpers/create-test-data.js
```diff
@@ -0,0 +1,178 @@
+'use strict';
```
npauzenga, line 1: Having this helper in /tests/ is, I think, causing problems. The test runner is trying to find tests in here and failing, which I think might be causing a failure on the unit test step in CI. I'm moving this to `/db/src/test-data.ts` but open to other opinions if that doesn't seem right.

packages/api/tests/endpoints/test-granules.js
npauzenga: This does seem to alleviate the test failure in CI (though they are still failing due to the test helper mentioned below). I'll see if I can get these running in parallel.

packages/api/lambdas/bulk-operation.js
Jkovarik: This goes back a bit to the arch meeting, but yes, the only way to do this other than a direct cumulus_id reference would be to do the work to do a delete with the unique constrained keys, which we do have here........

The functional problem I see with this approach is: 

1)  Getting the Dynamo Record
2) Pulling the collection/granule cumulus_id based on that 
3) Optionally - Unpublish (getting the updated dynamo record and/or PG record)
4) Delete Granules and Files (using the value from 3 if present *else* 2

We should reorder this such that it doesn't bother doing the get queries if unpublish is going to happen.

packages/api/lambdas/bulk-operation.js
```diff
@@ -118,26 +128,67 @@ function applyWorkflowToGranules({
...
+
+      // Using the Dynamo record as the primary source, unpublish it from
+      // CMR if it's published and we need to force-remove it.
+      let updateResponse;
+
+      if (dynamoGranule && dynamoGranule.published && forceRemoveFromCmr) {
+        updateResponse = await unpublishGranule(knex, dynamoGranule);
```
Jkovarik, line 180: Consider: 
```suggestion
        { pgGranule, dynamoGranule } = await unpublishGranule(knex, dynamoGranule);
```

with the refactor below to just remove the ternary.

packages/api/lib/create-test-data.js
```diff
@@ -0,0 +1,181 @@
...
+const { s3PutObject } = require('@cumulus/aws-client/S3');
+const {
+  FilePgModel,
+  GranulePgModel,
+  CollectionPgModel,
+} = require('@cumulus/db');
+
+const { constructCollectionId } = require('@cumulus/message/Collections');
+
+// PG mock data factories
```
Jkovarik, line 13: Nit:   Consider renaming the factory method (and following) instead of the comments  if this isn't clear from the code itself.

Jkovarik (commented):
Initial comments, continuing review

packages/api/lambdas/bulk-operation.js
```diff
@@ -118,26 +128,67 @@ function applyWorkflowToGranules({
...
+
+      let dynamoGranule;
+      let pgGranule;
+
+      // Try to get the Dynamo record. If it cannot be found, just log
+      // the error and skip it.
       try {
-        const granule = await granuleModel.getRecord({ granuleId });
-        if (granule.published && forceRemoveFromCmr) {
-          await granuleModel.unpublishAndDeleteGranule(granule);
```
Jkovarik, line 129: This should be removed from the granule model if it's no longer being used.

packages/api/tests/lib/test-granule-remove-from-cmr.js
Jkovarik: What caused the cleanup?

Jkovarik (changes requested):
This work looks really good - just a couple of concerns/questions to address.

packages/api/tests/lib/test-granule-remove-from-cmr.js
npauzenga: Yeah, this was a test I wrote in my previous Granule Delete PR. There were a couple points during that review where I moved tests around between files, split tests up among different files, and generally copy-pasted things around. This stuff was cruft left over, likely because it was copied from somewhere else where it was actually used. Here it's not necessary.

packages/api/lib/granule-remove-from-cmr.js
npauzenga: 👍 sorry, I'm psyched to have docstrings on this project but not used to it so I tend to forget 😄 

https://github.com/nasa/cumulus/pull/2108/commits/4b99df2da3ba283bd0b0d8caabdf69d77eafc902

packages/api/tests/lambdas/test-bulk-operation.js
npauzenga: Yeah, agreed. Leaving this comment open. I'll circle back and address after the feedback that's relevant to my changes.

packages/api/lambdas/bulk-operation.js
```diff
@@ -118,26 +128,67 @@ function applyWorkflowToGranules({
...
   const deletedGranules = [];
   await pMap(
     granuleIds,
     async (granuleId) => {
+      const collectionPgModel = new CollectionPgModel();
+
+      let dynamoGranule;
+      let pgGranule;
+
+      // Try to get the Dynamo record. If it cannot be found, just log
```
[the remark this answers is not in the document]
npauzenga: That's fair. I worry in situations like this where we have a function performing 4+ steps with a bunch of conditionals. I think that's more of a sign that the function needs to be broken up. As part of addressing another suggestion I refactored a bit to hopefully make this function a little clearer:

https://github.com/nasa/cumulus/pull/2108/commits/4b006bd0b60a9c8fc412c8e66795ea5a61184280

packages/api/lambdas/bulk-operation.js
npauzenga: Totally makes sense. See what you think of this:

https://github.com/nasa/cumulus/commit/4b006bd0b60a9c8fc412c8e66795ea5a61184280

The idea here is to get the Dynamo Granule and check if it's published right away. If it is and we're doing a "forceRemoveFromCMR", we'll do the update on both Dynamo and PG records (if we have them). Then we can use the response from the update if it went through. 

This should bail earlier if a granule is published and save a PG fetch request if the granule was updated.

npauzenga: Also note the new private function `_getPgGranuleByCollection`. This lives in the bulk-operation Lambda but I hemmed and hawed a bit over where to put it. The other option is to drop it in /lib somewhere, likely either granules.js or a new file. But then I don't know if I (or anyone else) would know to look for it there if they needed it. I also don't know if this will ever be used again outside of this Lambda.

packages/api/lambdas/bulk-operation.js
```diff
@@ -118,26 +128,67 @@ function applyWorkflowToGranules({
 async function bulkGranuleDelete(payload) {
   const granuleIds = await getGranuleIdsForPayload(payload);
```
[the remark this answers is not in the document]
npauzenga: 👍 https://github.com/nasa/cumulus/pull/2108/commits/803052b8f4938e4f365817b9f13995e43f2a6675

packages/api/lambdas/bulk-operation.js
npauzenga: 👍 https://github.com/nasa/cumulus/pull/2108/commits/08b8bc49c34667d535be1d29018301e37c94e13b

This seems nicer. I ALWAYS forget JS's 'destructuring without declaration' syntax with the parens. We have those dynamoGranule/pgGranule vars in the upper scope so this should be a reassignment. Let me know if you had something else in mind!

packages/api/lib/create-test-data.js
npauzenga: Yeah, I think it makes sense to call the new PG factories stuff like `fakeGranuleRecordFactory` as, _eventually_, they'll be the only factories we care about. It's just in this interim where we have factories for both Dynamo and PG that it's sort of confusing.

packages/api/lib/granule-remove-from-cmr.js
```diff
@@ -41,17 +41,19 @@ const unpublishGranule = async (knexOrTransaction, granule) => {
...
     const collectionCumulusId = await collectionPgModel.getRecordCumulusId(
       knexOrTransaction,
       deconstructCollectionId(granule.collectionId)
     );
 
-    await granulePgModel.update(
+    [pgGranule] = await granulePgModel.update(
       knexOrTransaction,
       {
         granule_id: granule.granuleId,
```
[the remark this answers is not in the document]
npauzenga: That's a really good point. Whatever we do for deletes should probably also be on updates/upserts. 

@markdboyd did you end up settling on an approach for validating keys on delete? I still like the idea of only using `cumulus_id` from the perspective of the `delete` method but there is a definite downside. If our code doesn't pass the cumulus_id everywhere we need it, we'll need to add queries to get the ID every time.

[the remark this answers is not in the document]
markdboyd: I didn't decide on an approach but I was planning a follow up PR, notionally as part of 2367, to address it

packages/api/lib/granules.js
```diff
@@ -3,7 +3,6 @@
 const awsClients = require('@cumulus/aws-client/services');
 const isInteger = require('lodash/isInteger');
 const isNil = require('lodash/isNil');
-
```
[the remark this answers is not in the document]
npauzenga: 👍 https://github.com/nasa/cumulus/pull/2108/commits/e848394138aaa6eca0eae985cc93d67dc84fb04a

packages/api/tests/lambdas/test-bulk-operation.js
```diff
@@ -255,31 +306,56 @@ test.serial('bulk operation BULK_GRANULE applies workflow to granule IDs returne
   });
 });
 
-test.serial('bulk operation BULK_GRANULE_DELETE deletes listed granule IDs', async (t) => {
+test.serial('bulk operation BULK_GRANULE_DELETE deletes listed granule IDs from Dynamo and PG', async (t) => {
```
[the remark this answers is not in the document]
npauzenga: Yup, I enjoy clarity 😄 

https://github.com/nasa/cumulus/pull/2108/commits/64194aa0432717216f609b8aa3c4a9381660dcec

packages/api/tests/lambdas/test-bulk-operation.js
```diff
@@ -374,38 +455,70 @@ test.serial('bulk operation BULK_GRANULE_DELETE deletes granule IDs returned by
...
-    envVars,
-    payload: {
-      ids: [
-        granules[0].granuleId,
-        granules[1].granuleId,
-      ],
-      forceRemoveFromCmr: true,
-    },
-  });
+  sinon.stub(
```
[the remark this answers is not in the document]
npauzenga: Yup, totally agree. In hindsight this is a bit messy. I tossed a question in IX and Mark had a good suggestion. Rather than mocking `unpublishGranule` here, it might be better to test the `bulkGranuleDelete` logic separate from the Lambda handler. Then, if we do that, we can pass a mock of `unpublishGranule` directly to `bulkGranuleDelete`.

That conversation is here:

https://eosdis.slack.com/archives/CLBC362EQ/p1614625915373300

And the change is here:

https://github.com/nasa/cumulus/pull/2108/commits/4ad56143a43c5f26f78f5bd873e547f8bd00ca43

See what ya think!

packages/api/lambdas/bulk-operation.js
npauzenga: Yeah, I have a lot of squeamishness about removing Dynamo methods. My fear is that they're exposed and used somewhere that I'm not seeing. But that shouldn't be possible if I can't find references.

Getting over it 😄 

https://github.com/nasa/cumulus/pull/2108/commits/bb49238ba4da92a741dcc8bb32d3b23872ceaf13

packages/api/lib/granule-remove-from-cmr.js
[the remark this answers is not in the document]
npauzenga: Should this perhaps be a separate ticket so that we can address it on deletes as well? Also happy to do it here but I just want to be sure we have a consistent approach.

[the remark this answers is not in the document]
markdboyd: I had hesitated on going the "another ticket" approach for fear of the work getting lost. I might take a look at seeing if I can at least work something up for "delete" this morning, but if I'm not getting anywhere I can create a ticket

packages/api/lambdas/bulk-operation.js
```diff
@@ -102,6 +107,11 @@ function applyWorkflowToGranules({
   return Promise.all(applyWorkflowRequests);
 }
 
+// FUTURE: the Dynamo Granule is currently the primary record driving the
```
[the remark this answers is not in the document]
npauzenga: 👍 created this ticket in Phase 3 to capture it. I'd imagine we'll have several more like this:

https://bugs.earthdata.nasa.gov/browse/CUMULUS-2426

packages/api/tests/lambdas/test-bulk-operation.js
npauzenga: Here's a quick update that should fix it. The problematic test here is already deleting ENVs before trying to re-add them so I believe we can just add them normally in the `before` and expect the test to remove and add them.

https://github.com/nasa/cumulus/pull/2108/commits/01e8c77c48c0d0b824440494c0c879e1ba0468f9

packages/api/lambdas/bulk-operation.js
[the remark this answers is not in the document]
Jkovarik: 👍🏻  nice

packages/api/lambdas/bulk-operation.js
Jkovarik: You're actually correct this should be *deprecated* using @cumulus/common/util/deprecate as it's publically exported. 

There's a bit of absurdity to that stance given our efforts to remove Dynamo, but it might be worth a brief discussion.  

Regardless of our choice here we should include a CL entry as well.

packages/api/lib/granule-remove-from-cmr.js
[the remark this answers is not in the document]
Jkovarik: I think so long as we've captured the concern it doesn't need to be in this PR, but we absolutely shouldn't consider Phase 1 done until we address. this.

packages/api/tests/endpoints/test-granules.js
Jkovarik: I'm assuming this still needs to be addressed, will comment below

packages/api/tests/lambdas/test-bulk-operation.js
Jkovarik: 👍🏻  that works for now.

packages/api/lambdas/bulk-operation.js
Jkovarik: 👍🏻  no this looks right.   Nit: I don't think you need the parens, although. that looks updated so it's possible that was sorted anyway.

Jkovarik: This note is really weird, it's in a commit ref.

packages/api/lambdas/bulk-operation.js
Jkovarik: Nice work!   I'm ok with YAGNI-ing the concern re: _getPgGranuleByCollection.

packages/api/lib/granule-remove-from-cmr.js
```diff
@@ -63,7 +65,7 @@ const unpublishGranule = async (knexOrTransaction, granule) => {
         published: false,
         cmr_link: undefined,
       },
-      ['cumulus_id']
+      ['cumulus_id', 'published']
```
Jkovarik, line 68: Interesting -- I see why this was. added in deleteGranulesAndFiles... was this fixing an integration test?

Jkovarik (approved):
Thanks, all my changes (aside the new concern about deprecation have been addressed.  Great work!

packages/api/lib/granule-remove-from-cmr.js
npauzenga: This is added here so that we can remove a previous `pgGranule.get()` call. After some refactoring the only thing that `get` was giving us that we didn't already have was `published`. This change just makes the `update` method return that field so we can eliminate the `get`. This was the commit:

4b006bd0b60a9c8fc412c8e66795ea5a61184280

npauzenga: Yeah, it was either an integration test or a unit test that started failing after the most recent refactors. Either way, the update here needs to return `published` because this would be used for `deleteGranuleAndFiles` if we got that far. 

The test I was working on for `bulkGranuleDelete when forceRemove is true` started throwing `You can't delete a published granule', indicating I wasn't passing this properly.

packages/api/lambdas/bulk-operation.js
npauzenga: Yeah, I think I agree with your initial instinct to just remove this one. If deleting it involved changing a dozen references over to this new logic, that might not be something we'd want to tackle without a lot of consideration. In this case though the only reference to `unpublishAndDeleteGranule` is here and we've removed it. I think it's safe to delete it from the model entirely.

Updated the changelog to call it out:
https://github.com/nasa/cumulus/pull/2108/commits/3ad39cc1a62ed481e419e57387fc4f7622b0ac59

packages/api/tests/endpoints/test-granules.js
npauzenga: Yeah, I've been poking at this a bit. I'm thinking these might legitimately need to be run in serial. There are 4 tests here for delete that each create Dynamo, Postgres, and S3 buckets and records and delete all of it. I'm wondering if the problem is that running these in parallel causes issues where one test expects the state of the tables one way but another test is inserting/deleting records. Particularly this test highlighted where delete fails...

I'm tempted to keep these in serial. Conceptually it sorta makes sense to me. What do you think?

CHANGELOG.md
```diff
@@ -130,6 +130,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     - Changed `DELETE /pdrs/{pdrname}` API behavior to also delete record from Postgres database.
   - **CUMULUS-2309**
     - Changed `DELETE /granules/{granuleName}` API behavior to also delete record from Postgres database.
+    - Changed `Bulk operation BULK_GRANULE_DELETE` API behavior to also delete records from Postgres database.
+    - Removed `@cumulus/api/models/granule.unpublishAndDeleteGranule` in favor of `@cumulus/api/lib/granule-remove-from-cmr.unpublishGranule` and `@cumulus/api/lib/granule-delete.deleteGranuleAndFiles`.
```
Jkovarik, line 134: This is *probably* technically a breaking change as api/models/* is a public interface.

packages/api/lambdas/bulk-operation.js
Jkovarik: The issue being we have a precedent of *always* deprecating anything that's a 'public'/exported interface, and api/models/* is.

CHANGELOG.md
npauzenga: 👍 https://github.com/nasa/cumulus/pull/2108/commits/803265db1779356e6df3caf97943af834939085e

PR #2114, opened 2021-02-24
CUMULUS-2424: Update queue-pdrs to have ChildWorkflowMeta

nemreid:

## Changes

* ...

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


example/spec/parallel/ingest/ingestFromPdrWithChildWorkflowMetaSpec.js
```diff
@@ -0,0 +1,181 @@
...
+        uploadTestDataToBucket(
+          config.bucket,
+          unmodifiedS3Data,
+          testDataFolder
+        ),
+        addCollections(config.stackName, config.bucket, collectionsDir, testSuffix, testId),
+        addProviders(config.stackName, config.bucket, providersDir, config.bucket, testSuffix),
+      ]);
+
+      addedCollection = populatePromises[2][0];
```
npauzenga, line 112: Super small nit but maybe we could destructure the return from `Promise.all` and switch up the order so that we don't have to index into it here. Something like:

```
const [addedCollection, ...rest] = await Promise.all([
        addCollections(config.stackName, config.bucket, collectionsDir, testSuffix, testId),
        updateAndUploadTestDataToBucket(
          config.bucket,
          s3data,
          testDataFolder,
          [
            { old: 'cumulus-test-data/pdrs', new: testDataFolder },
            { old: 'DATA_TYPE = MOD09GQ;', new: `DATA_TYPE = MOD09GQ${testSuffix};` },
          ]
        ),
        uploadTestDataToBucket(
          config.bucket,
          unmodifiedS3Data,
          testDataFolder
        ),
        addProviders(config.stackName, config.bucket, providersDir, config.bucket, testSuffix),
      ]);
```

example/spec/parallel/ingest/ingestFromPdrWithChildWorkflowMetaSpec.js
```diff
@@ -0,0 +1,181 @@
...
+  let beforeAllFailed;
+  let config;
+  let pdrFilename;
+  let provider;
+  let testDataFolder;
+  let testSuffix;
+  let workflowExecution;
+  let addedCollection;
+  let executionNamePrefix;
+  let queuePdrsOutput;
```
npauzenga, line 76: I'm just noticing this about Jasmine (still a newbie). Do we not have the concept of a `context`? Any time we want to refer to and share variables outside of the scope of the `it` block we need to declare it in a higher scope?

example/spec/parallel/ingest/ingestFromPdrWithChildWorkflowMetaSpec.js
```diff
@@ -0,0 +1,181 @@
...
+
+  it('results in an IngestGranule workflow execution', async () => {
+    if (beforeAllFailed) fail('beforeAll() failed');
+    else {
+      const executionArn = queuePdrsOutput.payload.running[0];
+      await expectAsync(waitForStartedExecution(executionArn)).toBeResolved();
+    }
+  });
+
+  it('passes through childWorkflowMeta to the IngestGranule execution', async () => {
```
npauzenga, line 176: Do we need the `beforeAll` check in all tests in the `describe` block?

`if (beforeAllFailed) fail('beforeAll() failed');`

I'm thinking it's on the previous two tests because these tests are run in parallel and any of them could be the first executed after `beforeAll`?

npauzenga (changes requested):
This looks greatn @nemreid! I just had a couple clarifying questions and a super small nit.

example/spec/parallel/ingest/ingestFromPdrWithChildWorkflowMetaSpec.js
nemreid: That is correct.

example/spec/parallel/ingest/ingestFromPdrWithChildWorkflowMetaSpec.js
nemreid: Yes, the purpose is to *stop* the test from running if the beforeAll fails, so we need that clause in every test. Jasmine does *not* stop testing because test setup failed.

example/spec/parallel/ingest/ingestFromPdrWithChildWorkflowMetaSpec.js
nemreid: https://github.com/nasa/cumulus/pull/2114/commits/b2951e1aeffe4f886a6a361104dbc8c01edc988f

example/spec/parallel/ingest/ingestFromPdrWithChildWorkflowMetaSpec.js
npauzenga: Gotchya. So then we would need it in this test too?

npauzenga (approved):
Looks great @nemreid! I just left one comment above that might need a tweak (unless I'm confused).

example/spec/parallel/ingest/ingestFromPdrWithChildWorkflowMetaSpec.js
nemreid: Fair point. Good catch. https://github.com/nasa/cumulus/pull/2114/commits/cd6bd67522c5a94620f8fee05ab67f2d4d8dca5e

PR #2115, opened 2021-02-24
[CUMULUS-2383] Handle CMR 5XX Errors in PostToCmr

dopeters:
**Summary:** Handle CMR internal errors in PostToCMR

Addresses [CUMULUS-2383: As a Cumulus integrator, I want to configure my workflows to take a certain action when CMR is down](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2383)

## Changes

* The base CMR calls in PostToCMR happen in `validate` and `validateUMMG`. In those functions, when the statusCode is in the 500s, we are throwing a specific CMRInternalError.
* We'll be able to use this error type in the catch field of the workflow to take whatever action we'd like (see: https://docs.aws.amazon.com/step-functions/latest/dg/tutorial-handling-error-conditions.html)
* Added some tests to ensure that the error is being passed up to `PostToCmr` correctly.
* Also, removed the `ValidationError` in `cmr-client` so that we're using the one from the errors package instead

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [ ] Integration tests


flamingbear (approved):
Nice job. Looks good to me. 👍 💯

PR #2117, opened 2021-02-25
Fix issue with syncing 0 byte files in sync-granule

markdboyd:
We were running into the issue on MAAP of being unable to sync 0 byte files to S3 for `sync-granule`. This PR fixes that issue.

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Adhoc testing
- [x] Integration tests


example/spec/parallel/SyncGranuleTask/SyncZeroByteFileSpec.js
```diff
@@ -0,0 +1,162 @@
...
+
+const { deleteCollection } = require('@cumulus/api-client/collections');
+const { deleteProvider } = require('@cumulus/api-client/providers');
+const { lambda } = require('@cumulus/aws-client/services');
+const S3 = require('@cumulus/aws-client/S3');
+
+const { loadConfig } = require('../../helpers/testUtils');
+const { fetchFakeS3ProviderBuckets } = require('../../helpers/Providers');
+
+describe('The SyncGranule task with a 0 byte file to be synced', () => {
```
markdboyd, line 20: An integration test is necessary because unit tests with Localstack actually can't reproduce the issue, most likely because Localstack doesn't actually replicate S3's behavior for handling multipart uploads

packages/ingest/test/test-S3ProviderClient.js
```diff
@@ -110,6 +108,26 @@ test.serial('S3ProviderClient.sync syncs a file to a target S3 location', async
   );
 });
 
+test.serial('S3ProviderClient.sync syncs a 0 byte file', async (t) => {
```
markdboyd, line 111: This test doesn't really prove anything since it passed against Localstack even before the fixes were made. However, in case Localstack does ever update their multipart upload handling to match real S3 behavior, this will be a good test to have.

example/spec/parallel/SyncGranuleTask/SyncZeroByteFileSpec.js
```diff
@@ -0,0 +1,162 @@
...
+        (functionConfig.Timeout + 10) * 1000
+      );
+    } catch (error) {
+      beforeAllFailed = true;
+      throw error;
+    }
+  });
+
+  it('succeeds', async () => {
+    if (beforeAllFailed) fail('beforeAll() failed');
```
npauzenga, line 126: Just curious. I see this pattern in all of our integration tests. Is the idea here that we want to get to an actual test failure because if a `before` block fails it'll stop the test runner? Or not report the failure correctly?

example/spec/parallel/SyncGranuleTask/SyncZeroByteFileSpec.js
```diff
@@ -0,0 +1,162 @@
...
+        await S3.s3ObjectExists({
+          Bucket: parsedFileUrl.Bucket,
+          Key: parsedFileUrl.Key,
+        })
+      ).toBeTrue();
+    }
+  });
+
+  afterAll(async () => {
+    await pAll(
```
npauzenga, line 149: Also curious: I'm not familiar with `p-all`. It looks like the benefit here is that we can try to resolve all of the promises with `{ stopOnError: false}` rather than stopping at the first rejection. This way we can get an aggregate of all the results and have better error logging. Is that about right? Is there another reason it's used here in place of Promise.all?

packages/ingest/test/test-S3ProviderClient.js
npauzenga: Would it make sense to have this test check that `s3CopyObject ` was used in place of `multipartCopyObject` when a 0 byte files was uploaded? Tests really shouldn't know about the function's implementation like that but maybe it would be worthwhile to ensure that we're using the new correct code path if this is always going to work locally?

Thinking out loud: If Localstack ever changes to adopt the correct S3 behavior, I guess we'd still always use your new logic here so this test might not actually help us in that case. Does that make sense? I could be missing the point.

npauzenga (changes requested):
Looks great @markdboyd!

example/spec/parallel/SyncGranuleTask/SyncZeroByteFileSpec.js
markdboyd: Yeah, we want to run as much of the cleanup as possible, and `p-all` with `{ stopOnError: false }` will run all of the promises even if any one of them fails, whereas `Promise.all` would stop as soon as any one of the promises failed.

example/spec/parallel/SyncGranuleTask/SyncZeroByteFileSpec.js
markdboyd: Yeah, Jasmine will still try to run the tests even if there is an error in `beforeAll`:

https://github.com/jasmine/jasmine/issues/1533

So the test would fail because the necessary setup in beforeAll did not complete, but it would be completely non-obvious that is what happened. To make it more obvious, we're trapping the error in beforeAll and manually failing the tests to say that is what happened.

The better solution is upgrading our integration tests to use a new test runner, since there seems to be no momentum towards fixing this in Jasmine. But that's obviously a big effort.

npauzenga: Good to know! Thanks for the explanation. Also glad there's a workaround 😄

packages/ingest/test/test-S3ProviderClient.js
```diff
@@ -110,6 +108,26 @@ test.serial('S3ProviderClient.sync syncs a file to a target S3 location', async
...
+    Body: '',
+  });
+
+  const { s3uri, etag } = await s3ProviderClient.sync({
+    destinationBucket: t.context.targetBucket,
+    destinationKey: targetKey,
+    fileRemotePath: t.context.sourceKey,
+  });
+  t.truthy(s3uri, 'Missing s3uri');
+  t.truthy(etag, 'Missing etag');
```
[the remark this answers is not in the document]
markdboyd: That message will be printed if the assertion fails to give you more context about why the test failed. It's not required by any means but seems helpful enough

packages/ingest/test/test-S3ProviderClient.js
markdboyd: Hmm, interesting thought. I'm inclined to say, like you did, that we don't want the unit test for S3ProviderClient.sync to know which code path is hit necessarily. In some cases we do assert something about the code path for a unit test because creating the system state necessary for a unit test is impractical, so mocking and verifying that the mock was used is preferable. But in this case we're actually staging a fixture for a 0 byte file and proving that `sync` handles it properly. 

I don't know if that answers the question

npauzenga: Yeah, I think what you have here is right and the alternative is ...undesirable. It just feels a little wonky that we could revert your code changes and this test would still pass because Localstack's behavior doesn't match the real deal.

This test 100% does what it says it's going to though: `S3ProviderClient.sync syncs a 0 byte file` and there's an integration test to double-check in a deployed environment.

I think it's good. I can't think of a non-wonky way to validate that your workaround specifically works and I'm not sure we need to validate that here.

packages/ingest/src/S3ProviderClient.ts
```diff
@@ -79,17 +79,44 @@ class S3ProviderClient implements ProviderClient {
   ): Promise<{s3uri: string, etag: string}> {
     const { fileRemotePath, destinationBucket, destinationKey, bucket } = params;
     const sourceBucket = bucket || this.bucket;
+    const sourceKey = fileRemotePath;
+
     try {
+      const sourceObject = await S3.headObject(sourceBucket, sourceKey);
```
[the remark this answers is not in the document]
markdboyd: So originally, `multipartCopyObject` was calling `headObject` that could throw the `NoSuchKey` or `NotFound` errors. However, I'm not sure if other S3 functions called for multipart uploads inside `multipartCopyObject` may also throw those errors and thus whether we want that call in the try/catch. I can't find any AWS docs that make it super clear. 

As a result I'm a little hesitant to make the change, but am I misunderstanding?

packages/aws-client/src/S3.ts
```diff
@@ -908,6 +908,7 @@ export const multipartCopyObject = async (
     sourceKey: string,
     destinationBucket: string,
     destinationKey: string,
+    sourceObject?: AWS.S3.HeadObjectOutput,
```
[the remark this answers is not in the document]
markdboyd: 071a93a21627a4c3b48d0bc0ba268fd0d142f164

packages/ingest/src/S3ProviderClient.ts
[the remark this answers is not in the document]
npauzenga: Ok, that's totally fine with me then. This comment was based on a suspicion that the `catch` here was  really only meant for `headObject` (and maybe `s3CopyObject`) and that it was the only one that would possibly throw `NoSuchKey` or `NotFound`.

If `multipartCopyObject` could potentially throw those errors and we want to guard against that (as it looks like we would) that's fine.

npauzenga (approved):

npauzenga:
Looks good to me @markdboyd there's just one comment from Jonathan remaining:

https://github.com/nasa/cumulus/pull/2117/files#r583816983

example/spec/parallel/SyncGranuleTask/SyncZeroByteFileSpec.js
```diff
@@ -0,0 +1,162 @@
...
+  let prefix;
+  let provider;
+  let syncGranuleOutput;
+  let parsedFileUrl;
+
+  const filename = '0byte.dat';
+
+  beforeAll(async () => {
+    try {
+      const config = await loadConfig();
```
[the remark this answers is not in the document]
markdboyd: Yeah, I want to hold off on this for now, but generally speaking there's a lot of cleanup in the integration tests to do

Jkovarik (approved):
@npauzenga that comment was a take/leave, there's. no reason not to merge and sort that in another PR.

PR #2118, opened 2021-02-26
replaces a sequence of calls to cmr with a single compound call.

flamingbear:
Addresses [CUMULUS-2412: Collections endpoint times out after Elasticsearch changes](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2412)

## Changes

* Main change is that the /collections endpoint has been modified.  
  ** Before it made a cmr.searchCollections() call for every collection to append MMTLinks. 
  ** Now it is updated to make a single compound call to CMR before appending the MMTLink information

## PR Checklist

- [X] Update CHANGELOG
- [X] Unit tests
- [X] Adhoc testing
- [n/a] Integration tests (no additional ones needed)


packages/api/tests/lib/test-mmt.js
```diff
@@ -3,33 +3,89 @@
...
 
+/**
+ *  Fakes a CMR return feed where each item in the list contains an
+ *  collection_id labeled "id" that is made of the short_name and version
+ * @param {Array<Object>} list - parsed list of collection results
+ * @returns {Object} Fake CMR responses
+ */
+const cmrReturnsWithIds = async (list) => {
+  const entry = list.map((l) => ({
+    id: `${l.short_name}-${l.version}`,
```
jennyhliu, line 25: The cmr collection id is something like `C987654321-CUMULUS` and doesn't contain shortname and versionid.  How would a cmr collection item not have 'collection' id?

packages/cmrjs/src/cmr-utils.js
```diff
@@ -881,6 +882,55 @@ async function reconcileCMRMetadata({
...
+ * This query is a compound conditional using JSONQueryLanguage supported by CMR.
+ * This returns every collection that matches any of the short_name version pairs provided.
+ * the final query should be like
+ *  {"condition":
+ *   { "or": [{ "and": [{"short_name": "sn1"}, {"version": "001"}] },
+ *            { "and": [{"short_name": "sn2"}, {"version": "006"}] },
+ *            { "and": [{"short_name": "sn3"}, {"version": "001"}] },
+ *            .... ] } }
+ *
+ * @param {Array<Object>} results - objects with keys "short_name" and "version"
```
jennyhliu, line 896: 👍

packages/api/tests/lib/test-mmt.js
flamingbear: This is just a fake function. It is used to test that whatever is in the ID is used in the MMTLink that is created from it.  I'm using the shortName and version just for convenience in the test you can see here: https://github.com/nasa/cumulus/blob/mhs/CUMULUS-2412/update-cmr-client-mmt-information/packages/api/tests/lib/test-mmt.js#L105-L115


I actually don't know if CMR can return a collection without a collection id. These are just updated tests that I had written before.  Oh, but now that I look at them, what I would expect is a return that is empty from CMR.  I will make this function return empty objects as if the collection wasn't found in CMR.  https://github.com/nasa/cumulus/blob/mhs/CUMULUS-2412/update-cmr-client-mmt-information/packages/api/tests/lib/test-mmt.js#L38-L41. Would that be better?

flamingbear: changed in 8f87a62

jennyhliu (approved):
LGTM

PR #2125, opened 2021-03-02
Changes waitForAsyncOperationStatus to fail if status not reached.

flamingbear:
**Summary:** Increases test times for AsyncOperation and ReconciliationReports

Addresses [CUMULUS-2420 : Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2420)

## Changes

* waitForAsyncOperationStatus now throws an Error when the desired status is not reached during the specified wait period.
   - Before this change, this routine just returned the state of operation after the retries were expired. This led to lots of errors that were `Error: Expected 'RUNNING' to be 'SUCCESS'.`
* Changed the default values for waitForAsyncOperationStatus retries.
   - Originally this was set as 10 retries sampling every 2 seconds (20 sec total). Inspection shows that this isn't enough time for the AsyncOperation tests to complete every time. The defaults have been changed to use an exponential backoff and 15 retries.  This leads to 15 attempts over 60 Seconds.
   - This increase should help the tests pass. If they fail by not reaching the correct status, we get a better error message and we can tune the defaults.
* Changed the retry values in CreateReconciliationReportSpec and IngestGranuleSuccessSpec from retries: 100 (100 retries in 200 seconds [3.3 min] with no backoff) to use an exponential backoff and 70 retries which results in 70 attempts in ~6.3 min. This duration increase should fix the types of recReport errors that were occurring with errors like `Expected 'RUNNING' to be 'SUCCESS'.`
* Wraps the `waitForAsyncOperationStatus` in try/catch to fail with an explicit and more descriptive error.

## PR Checklist

- [X] Update CHANGELOG
- [n/a] Unit tests
- [X] Adhoc testing
- [X] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - `<prefix>-lambda-api-gateway` IAM role used by API Gateway Lambda now supports accessing all buckets defined in your `buckets` variable except "internal" buckets
 - **CUMULUS-2355**
   - Added logic to disable `/s3Credentials` endpoint based upon value for environment variable `DISABLE_S3_CREDENTIALS`. If set to "true",  the endpoint will not dispense S3 credentials and instead return a message indicating that the endpoint has been disabled.
+- **CUMULUS-2420**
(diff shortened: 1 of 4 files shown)


jennyhliu (approved):
Looks good👍

PR #2127, opened 2021-03-02
CUMULUS-2346: Create an ORCA specific API endpoint for recovery status

jennyhliu:

Addresses [CUMULUS-2346: Create an ORCA specific API endpoint for recovery status](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2346)

## Changes

* Added orca API endpoint to `@cumulus/api` to get recovery status
* Remove orca step from IngestAndPublishGranule workkflow, and create a separate ingest workflow with orca step for testing. Remove the `excludeFileTypes` from the test collections which are not used for orca testing
* Update the default orca `database_name` and `database_app_user`, they are not actually configurable.

Note:
There is a orca ticket: ORCA-145 - Fix issue with ORCA deployments that fail to update database passwords when changed for deployment.  So if you changed the database password, and your orca integration test fails with authentication error, you need to update the password manually.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Adhoc testing
- [x] Integration tests


packages/api-client/src/orca.ts
```diff
@@ -0,0 +1,31 @@
...
+}): Promise<ApiGatewayLambdaHttpProxyResponse> => {
+  const { prefix, query, callback = invokeApi } = params;
+
+  return callback({
+    prefix: prefix,
+    payload: {
+      httpMethod: 'GET',
+      resource: '/{proxy+}',
+      path: '/orca/recovery',
+      queryStringParameters: query,
```
kkelly51, line 28: This query will work with both options, correct? A full job so just asyncOperationId, then a granuleId + asyncOperationId

packages/api/endpoints/orca.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * List recovery request status
+ *
+ * @param {Object} req - express request object
+ * @param {Object} res - express response object
+ * @returns {Promise<Object>} the promise of express response object
+ */
+async function listRequests(req, res) {
+  const query = req.query || {};
```
kkelly51, line 20: And if no query is provided, what do we expect to get back? Did we talk about this option with them? I think we said it would always at least include a jobId

packages/api-client/src/orca.ts
jennyhliu: Yes.  But currently orca ignores asyncOperationId.

packages/api/endpoints/orca.js
jennyhliu: no query parameter will return all the requests I think.  It depends how the request_status lambda is implemented.  Will can we get a jobId?

example/data/collections/L2_HR_PIXC-000/L2_HR_PIXC-000.json
```diff
@@ -8,9 +8,6 @@
   "reportToEms": false,
   "url_path": "{cmrMetadata.Granule.Collection.ShortName}___{cmrMetadata.Granule.Collection.VersionId}/{substring(file.name, 0, 3)}",
   "sampleFileName": "L2_HR_PIXC_product_0001-of-4154.h5",
-  "meta": {
-    "excludeFileTypes": [".cmr.json", ".cmr.xml", ".h5", ".hdf", ".jpg", ".md5", ".met", ".nc", ".txt"]
```
kkelly51, line 12: Why did you get rid of this meta stuff in all these json files?

kkelly51 (approved):

example/data/collections/L2_HR_PIXC-000/L2_HR_PIXC-000.json
jennyhliu: Before this PR, IngestAndPublishGranule workflow has copy_to_glacier step, and excludeFileTypes was added to prevent granule files get copied to glacier.  Now only IngestAndPublishGranuleWithOrca has orca step, and this workflow is invoked in the OrcaBackupAndRecoverySpec.

PR #2129, opened 2021-03-04
Feature/cumulus 2385 add Postgres File constraints

npauzenga:
**Summary:** 

Addresses [CUMULUS-2385: Resolve PG files constraints](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2385)

The idea here is to make these fields required after proving that a huge dataset like PODAAC's does not have any files that are missing them.

We're still using the `getBucket` and `getKey` helpers in our data migration Lambdas so it will still look to `file.filename` if it can't find a bucket or key on the Dynamo record. I don't see a problem with continuing to support that even if we're fairly confident everything will have actual `key` and `bucket` properties.

## Changes

* Updates Postgres File schema to require (make notNullable) the `bucket` and `key` fields

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


CHANGELOG.md
```diff
@@ -118,14 +118,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
+    - Changed Postgres Schema definition for Files to remove `filename` and `name` and only support `file_name`.
+    - Changed Postgres Schema definition for Files to remove `size` to only support `file_size`.
+    - Changed `PostgresFile` to remove duplicate fields `filename` and `name` and rename `size` to `file_size`.
   - **CUMULUS-2305**
     - Changed `DELETE /pdrs/{pdrname}` API behavior to also delete record from Postgres database.
   - **CUMULUS-2309**
     - Changed `DELETE /granules/{granuleName}` API behavior to also delete record from Postgres database.
     - Changed `Bulk operation BULK_GRANULE_DELETE` API behavior to also delete records from Postgres database.
+  - **CUMULUS-2385**
+    - Changed Postgres `File` schema to require `bucket` and `key` fields.
```
npauzenga, line 130: I don't think this is technically a breaking change because RDS hasn't gone out, right?

botanical: Yeah that sounds correct to me

botanical (changes requested):
Looks good! I think we need to update `bucket` and `key` on `PostgresFile` as well in `packages/db/src/types/file.ts`

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -228,41 +228,6 @@ test.serial('migrateFileRecord correctly migrates file record', async (t) => {
...
-      checksum_value: null,
-      checksum_type: null,
-      key: null,
-      path: null,
-      file_size: null,
-      file_name: testFile.fileName,
-      source: null,
-    }
-  );
-});
```
botanical, line 264: Yay for removing code 😁

npauzenga:
> Looks good! I think we need to update `bucket` and `key` on `PostgresFile` as well in `packages/db/src/types/file.ts`

Oh right! The TypeScript type definitions. Whoops, good catch!

https://github.com/nasa/cumulus/pull/2129/commits/30bc79652791ef586b8b050866e2465f158c4061

botanical (approved):
Looks like the change I mentioned was addressed. Nice work and testing! 😀 😎

packages/db/src/translate/file.ts
```diff
@@ -5,12 +5,12 @@ import { PostgresFile } from '../types/file';
 export const translateApiFiletoPostgresFile = (
   file: ApiFile
 ): Omit<PostgresFile, 'granule_cumulus_id'> => ({
-  bucket: file.bucket,
+  bucket: file.bucket || '',
```
markdboyd, line 8: What is the cause for this? I guess a `bucket` and `key` that are empty strings are technically valid, but it's going to make looking up records uniquely quite difficult.

I guess more precisely: why are we allowing empty strings based on what we found from the investigation of the PODAAC data?

npauzenga: Yup! Apologies, I wanted to check that this actually did what I thought it would before commenting. Wasn't trying to slip this through 😄 

The intention in this PR is to make file.bucket and file.key required fields in the Postgres schema because the PODAAC data indicates these files will likely always be present.  We also need to update the `PostgresFile` TypeScript type definition to make these fields required.

The trick here is that the Dynamo `ApiFile` TypeScript type definition does NOT have bucket/key as required fields. I think that's correct; those fields are not required by Dynamo so making them required on the `ApiFile` doesn't feel right. It seems like the TypeScript definition should match the table schema. What that means here is that we're trying to assign the Dynamo `string or undefined` type to the Postgres `string` type. Which isn't possible. This change defaults to an empty string if bucket or file is undefined to make the TypeScript compiler happy.

We could talk about making the bucket/key properties required on `ApiFile`. That's what I initially did that caused the integration test failures. In retrospect I don't think that's what we want regardless of test failures but I could be wrong.

This could be worth a team discussion.

npauzenga: Put more succinctly, the question here is:

What should we do if we find a file without a bucket or key during Dynamo to Postgres migration?

Based on the PODAAC data, this may never be the case (it isn't in PODAAC's).

npauzenga: Discussed offline: we will throw an error here if `bucket` or `key` is undefined. Will reply with commit # once all integration tests are verified.

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
```diff
@@ -147,7 +147,26 @@ const generateFileRecord = ({ file, granuleCumulusId }) => ({
...
+  const errors = [];
+
+  const generatedFiles = files.map((file) => {
+    let generatedFile;
+
+    try {
+      generatedFile = generateFileRecord({ file, granuleCumulusId });
+    } catch (error) {
+      errors.push(error);
+    }
```
npauzenga, line 160: Collecting errors here so that we don't stop on the first one and continue to generate any other valid files.

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
```diff
@@ -245,10 +266,30 @@ const writeGranuleAndFilesViaTransaction = async ({
...
+      });
+    } catch (error_) {
+      await granulePgModel.upsert(
+        trx,
+        {
+          ...granuleRecord,
+          status: 'failed',
+          error: {
+            Error: 'Failed writing files to Postgres.',
+            Cause: error_.message,
```
npauzenga, line 288: We could also include ALL of the errors from `filesAndErrors.errors` here. This would just be the first one that triggered the `catch`.

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-write-granules.js
```diff
@@ -574,3 +577,75 @@ test('writeGranules() does not persist records to Dynamo or Postgres if Postgres
     await t.context.granulePgModel.exists(knex, { granule_id: granuleId })
   );
 });
+
+test.serial('writeGranules() writes a granule and marks as failed if any file writes fail', async (t) => {
```
npauzenga, line 581: These have to be run in serial because reasons. I'm guessing because a few of these tests directly update the `cumulusMessage`. I'll see if I can clean this up before merge.

packages/db/src/translate/file.ts
npauzenga: Well, it's several commits 😄. `translateApiFiletoPostgresFile` now throws an error and is handled in write-granules.js.

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
markdboyd: I'd recommend rewriting this entire function using `p-map` like so:

```
return pMap(
   files,
   (file) => generateFileRecord({ file, granuleCumulusId }),
   {
     stopOnError: false,
   }
 )
```

It's a little cleaner and `stopOnError: false` automatically handles the behavior of waiting for all promises to settle then rejecting with an aggregate error for any errors

https://github.com/sindresorhus/p-map#stoponerror

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
markdboyd: Is catching the error from the record **generation** the right approach? I would think that we want to catch errors from writing any of the files to the DB, which would fail on schema validation for missing bucket/key. Does `generateFileRecords` actually throw errors if it can't generate a bucket/key?

You would also need to refactor `writeFilesViaTransaction` so that it doesn't fail on the first failing file write. Right now it's using Promise.all so one failed file write fails them all

packages/db/src/translate/file.ts
```diff
@@ -1,16 +1,21 @@
...
-  checksum_type: file.checksumType,
-  checksum_value: file.checksum,
-  file_name: file.fileName,
-  file_size: file.size,
-  key: file.key,
-  path: file.path,
-  source: file.source,
-});
+): Omit<PostgresFile, 'granule_cumulus_id'> => {
+  if (!file.bucket || !file.key) throw new ValidationError(`bucket and key properties are required: ${JSON.stringify(file)}`);
```
markdboyd, line 9: I don't think we need or want this. This is essentially replicating the enforcement of the schema from the database in code. I think it is preferable to let the database enforce that on write and not have to duplicate the logic in code

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
markdboyd: I understand now why this is happening based on the logic in `translateApiFiletoPostgresFile`. See my comments below though about how I think we should change this.

markdboyd (changes requested):

packages/db/src/translate/file.ts
npauzenga: Right, we talked about this recently. Taking a look at removing it and leaning on the Postgres schema.

packages/db/src/translate/file.ts
npauzenga: Going to resolve this discussion as it's ~4 refactors out of date and likely to cause confusion.

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
npauzenga: This was refactored here: https://github.com/nasa/cumulus/pull/2129/commits/bf8eff5425d57c2c3b944ae7fce6b078ef29e050

We're now using `pMap` for the upserts, not the file generation. Also removed the async/await on the file generation here:

https://github.com/nasa/cumulus/pull/2129/commits/d32bb5a6b24e871241c8f69a20094250453057b1

Shouldn't need it because the translate function isn't async.

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
```diff
@@ -144,38 +145,28 @@ const generateFileRecord = ({ file, granuleCumulusId }) => ({
...
-  trx,
+  knex,
   filePgModel = new FilePgModel(),
-}) =>
-  Promise.all(fileRecords.map(
-    (fileRecord) => filePgModel.upsert(trx, fileRecord)
-  ));
+}) => pMap(
+  fileRecords,
+  async (fileRecord) => filePgModel.upsert(knex, fileRecord)
```
npauzenga, line 168: Note we're using `knex` instead of `trx` here. This is based on our finding that a Knex transaction always fails whenever any part of it fails, even if we catch the error.

I updated the tests to try a granule with 10 files (and adhoc tested with 100). It performs fine but it will get slower once you hit 100+ files. Want to flag this in case it seems like something worth addressing now.

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
npauzenga: Updated this whole thing here: https://github.com/nasa/cumulus/pull/2129/commits/bf8eff5425d57c2c3b944ae7fce6b078ef29e050

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-write-granules.js
npauzenga: The `writeGranules` tests really do appear to need to be run in serial. We're writing, updating, and throwing errors based on database writes. I'm thinking running these in parallel causes collisions or unexpected database state and that's why I'm seeing intermittent failures when not run in `serial`.

packages/db/src/translate/file.ts
npauzenga: Updated here: https://github.com/nasa/cumulus/pull/2129/commits/bf8eff5425d57c2c3b944ae7fce6b078ef29e050

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -230,41 +230,6 @@ test.serial('migrateFileRecord correctly migrates file record', async (t) => {
   );
 });
 
-test.serial('migrateFileRecord correctly migrates file record with null bucket and key', async (t) => {
```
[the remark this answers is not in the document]
npauzenga: Yup, that's a good point. Added a very similar test to look for `filename` in place of `bucket` and `key`:
https://github.com/nasa/cumulus/pull/2129/commits/74a32af2f097b80c25f1203a1ea1277acbbfe6dc

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
```diff
@@ -236,24 +260,86 @@ const writeGranuleAndFilesViaTransaction = async ({
...
+          Error: 'Failed writing files to Postgres.',
+          Cause: error_.message,
+        },
+      }
+    );
+  }
+};
+
+/**
+ * Create Postgres File objects from a granule's files
```
markdboyd, line 322: The returned files here aren't Postgres file objects. They're just File API objects that have been fleshed out with some extra properties that they may have been missing

```suggestion
 * Transform granule files to latest file API structure
```

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
```diff
@@ -236,24 +260,86 @@ const writeGranuleAndFilesViaTransaction = async ({
...
+ *
+ * @param {Object} params
+ * @param {Object} params.granule - An API granule object
+ * @param {Object} params.provider - An API provider object
+ *
+* @returns {Promise<Array>} - A list of file objects once resolved
+ */
+const _generateFiles = async ({
```
markdboyd, line 330: Also can we rename this function? I'm getting confused between `generateFileRecords` and `generateFiles`. Maybe this one can be `transformApiFiles` or something. That's not very descriptive, but it gets the basic idea. Maybe we should also rename `generateFileRecords` to `generateFilePgRecords` for additional clarity

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
```diff
@@ -236,24 +260,86 @@ const writeGranuleAndFilesViaTransaction = async ({
...
-  return writeFilesViaTransaction({
-    fileRecords,
-    trx,
+  try {
+    await _writeFiles({
+      fileRecords,
+      knex,
+    });
+  } catch (error_) {
+    const granule = await granulePgModel.get(knex, { cumulus_id: granuleCumulusId });
```
markdboyd, line 305: I think we should log the error here just for additional visibility

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
```diff
@@ -236,24 +260,86 @@ const writeGranuleAndFilesViaTransaction = async ({
...
 
-  return writeFilesViaTransaction({
-    fileRecords,
-    trx,
+  try {
+    await _writeFiles({
+      fileRecords,
+      knex,
+    });
+  } catch (error_) {
```
markdboyd, line 304: Nit: There's no `error` variable in the outer scope here, so I would rename this to just `error`

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
markdboyd: Hmm, yeah. So the problem is that if we used `trx` here, then as soon as one of the file writes failed, the others would fail. 

Is that a problem though? It's okay if only some of the file writes succeed. I think the main thing is that we want the granule to update to `failed` even if some/all of the file writes fail. Meaning that the granule write has to use `knex` and not `trx`. But maybe we can use `trx` for the file writes.

I'm mentioning it because I think using `knex` vs `trx` can result in a proliferation of database connections, which if you're running thousands of workflows to ingest data (very realistic), we could eat up the number of allowed connections

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
```diff
@@ -236,24 +260,86 @@ const writeGranuleAndFilesViaTransaction = async ({
...
+    const granule = await granulePgModel.get(knex, { cumulus_id: granuleCumulusId });
+
+    granulePgModel.upsert(
```
markdboyd, line 307: We either need to `await` or `return` here. Otherwise we're not guaranteeing that this promise resolves before the function returns

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-write-granules.js
markdboyd: Interesting. All the tests were passing as non-serial before though, so the reason for the intermittent failures must be caused by this new test? I don't care a lot whether we run them in serial or not, but I do want to understand what test interdependencies we do/don't have

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-write-granules.js
```diff
@@ -119,6 +120,12 @@ test.beforeEach(async (t) => {
   );
 });
 
+test.afterEach.always(async (t) => {
+  await t.context.knex(tableNames.files).del();
```
markdboyd, line 124: Generally speaking I prefer that tests clean up exactly what they created rather than wiping tables between tests, but it's not a hill that I will die on

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-write-granules.js
```diff
@@ -574,3 +584,85 @@ test('writeGranules() does not persist records to Dynamo or Postgres if Postgres
...
+
+  t.true(await granuleModel.exists({ granuleId }));
+
+  const pgGranule = await t.context.granulePgModel.get(knex, { granule_id: granuleId });
+
+  t.is(pgGranule.status, 'failed');
+  t.deepEqual(pgGranule.error.Error, 'Failed writing files to Postgres.');
+});
+
+test.serial('writeGranules() writes all valid files if any non-valid file fails', async (t) => {
```
markdboyd, line 621: I think that this would be better as a unit test of `_writeFiles` directly. Since that is where the relevant logic under test here lives

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
npauzenga: Right. The question here is: would we be ok failing ALL file writes for a granule if ANY file write fails? 

The logic we'd talked about (which I think makes a lot of sense) is basically storing any files that we can, even if some others fail. There doesn't seem to be a way to do that without making each file it's own independent transaction and I totally agree that that seems scary. I've verified that it _works_ as-is for 100+ files but I think, at the very least, once you get into the thousands you'd see a performance hit compared to transactions (unverified but seems really probable).

So from the implementation side it does seem like a better idea to use a transaction but that's different business logic.

Thoughts? Also looping in @Jkovarik because this is sort of a compromise if we change it.

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
npauzenga: Yup, here's the latest call stack:

1. writeGranules
    2. _writeGranule
        3. _generateFilesFromGranule
        4. _writeGranuleViaTransaction
        5. _writeGranuleFiles
            7. _generateFilePgRecords
                8. generateFilePgRecord 
            9. _writeFiles

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
```diff
@@ -165,7 +165,8 @@ const _writeFiles = async ({
   filePgModel = new FilePgModel(),
 }) => pMap(
   fileRecords,
-  async (fileRecord) => filePgModel.upsert(knex, fileRecord)
+  async (fileRecord) => filePgModel.upsert(knex, fileRecord),
+  { stopOnError: false }
```
npauzenga, line 169: This is interesting. It was actually doing what we wanted without this line (the default is `true`). Making this flag `false` gets us a more useful `AggregateError` though. 

Either way it does all of the upserts even if one fails in the middle.

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
npauzenga: I think we've landed on an agreement here. We will keep this as-is where each file write uses the Knex client directly, rather than a transaction. In testing it appears to not cause a massive influx of database connections and it gets the file integrity we're after.

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
npauzenga: 👍  https://github.com/nasa/cumulus/pull/2129/commits/da067e78a61e25edbaae29605b7b0717608059f1

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
npauzenga: Yup, refactored here:

https://github.com/nasa/cumulus/pull/2129/commits/6d7f94af14d8b5c9b1d7baedf131e409a71a24b9

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-write-granules.js
npauzenga: I'm happy to fight about it 😄 . What I like about the `afterEach nuke the entire table` approach is that you're guaranteed a fresh slate. I do hear ya though. If you decide to feel strongly about it, I'm happy to relent.

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-write-granules.js
npauzenga: Yeah so I'm going back-and-forth on how to structure these tests and which functions should be private v. public. What I like about testing the logic from `writeGranules` is that we're also getting the `generateFilePgRecords` logic under test here as well. That would previously throw errors for non-valid files and has been updated.

The tests for this file are currently set up to only test the exported functions, which I think is cool. I could make this change and export `_writeFiles` (minus the underscore) but I also like limiting our exports here.

Thoughts? I don't feel super strongly. Just strongly enough to want to discuss first 😄

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
npauzenga: Updated!

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-write-granules.js
markdboyd: The thing i don't like about that is that then you don't know exactly what the state is created by the code under test or the test setup itself. So your tests can be passing even the logic under test is leaving around some state that you don't know about/expect

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-write-granules.js
markdboyd: I don't think I would worry about whether we have to export an extra function to test it or not. 

I see what you mean `writeGranules` testing the logic for generating and writing the records, so that's fine. Though it could equally be achieved by unit tests for both of those functions. But that gets into unit testing philosophy and it's not that important here

markdboyd (approved):

PR #2131, opened 2021-03-08
Add documentation on validating and resuming an Elasticsearch reindex

laurenfrederick:
This is what Nga and I did to fix her reindex issues, so I wanted to add it to the docs.

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


docs/troubleshooting/reindex-elasticsearch.md
```diff
@@ -24,8 +23,138 @@ This will point Cumulus to the index in Elasticsearch that will be used when ret
 
 The reindex operation will take the data from one index and copy it into another index. The reindex operation can be found [here](https://nasa.github.io/cumulus-api/#reindex)
 
+### Reindex status
```
npauzenga, line 26: Super small nit/thought: I'm wondering if this could be an addition to the `Reindex from the old index to the new index` sub-heading above rather than a separate section? I like that the main section, `Switch to a new index and Reindex` has two sub-sections that are explicitly called out in the first sentence. That section would read:

**Reindex from the old index to the new index**
The reindex operation will take the data from one index and copy it into another index. The reindex operation can be found here. Reindexing is a long-running operation. The reindex-status endpoint can be used to monitor the progress of the operation.

docs/troubleshooting/reindex-elasticsearch.md
```diff
@@ -24,8 +23,138 @@ This will point Cumulus to the index in Elasticsearch that will be used when ret
...
+      "docs": {
+        "count": 715949,
+        "deleted": 140191
+      }
+    }
+  }
+}
+```
+
+To further drill down into what is missing, login to the Kibana instance (found in the Elasticsearch section of the AWS console) and run the following command replacing `<index>` with your index name.
```
npauzenga, line 73: Super small nit: "log in" instead of "login".

npauzenga (approved):
This looks great to me @laurenfrederick! I had two super small things but consider them optional.

docs/troubleshooting/reindex-elasticsearch.md
laurenfrederick: Ah yep that makes sense. I made that change.

docs/troubleshooting/reindex-elasticsearch.md
laurenfrederick: 👍

PR #2132, opened 2021-03-09
[CUMULUS-2371] Archive Incoming SQS Messages

botanical:

Addresses [CUMULUS-2371: Archive incoming SQS Messages](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2371)

## Changes

* Added `archiveMessage` function to `sqs-message-consumer` to archive source SQS messages. Uses `MessageId` on message as key.
* Add deleteArchivedMessage function to `sqs-message-remover`
* Note that `sqs-message-consumer` reads messages from Queues defined in Rules where the state is set to `ENABLED`. 
 
## Testing
* Added unit tests
* Added integration tests to check that unprocessed messages are archived to S3.
* Did adhoc testing to ensure that message files exist in the path: `${stackName}/archived-incoming-messages/${message.MessageId}` and that the message ID corresponds to messages on the queue's DLQ. 

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [x] Integration tests


lambdas/sqs-message-remover/tests/test-index.js
```diff
@@ -179,17 +185,35 @@ test('sqsMessageRemover lambda does nothing for a workflow message when eventSou
...
+  }).promise();
+
+  await s3PutObject({
+    Bucket: process.env.system_bucket,
+    Key: sqsMessage.MessageId,
+    Body: JSON.stringify(sqsMessage.Body),
+  });
+
+  // Check that item exists in S3
+  const item = await s3().getObject({
```
botanical, line 205: I elected to call `getObject` directly instead of calling `getJsonS3Object` from `@cumulus/aws-client/S3` because `getS3Object` is deprecated.

lambdas/sqs-message-remover/tests/test-index.js
```diff
@@ -201,7 +225,16 @@ test('sqsMessageRemover lambda removes message from queue when workflow succeede
   t.is(numberOfMessages.numberOfMessagesAvailable, 0);
   t.is(numberOfMessages.numberOfMessagesNotVisible, 0);
 
-  await awsServices.sqs().deleteQueue({ QueueUrl: sqsQueues.queueUrl }).promise();
+  // Check that item does not exist in S3 and therefore throws an error
+  await t.throwsAsync(s3().getObject({
```
botanical, line 229: Weirdly, this doesn't throw... If I call to delete the message directly, it throws. The logs demonstrate that the path is `updateSqsQueue -> deleteArchivedMessages` without erroring, but the object in the bucket actually isn't deleted.

botanical: After looking into this, the reason why the object can still be retrieved is that deletion is eventually consistent so there isn't a guarantee that, immediately after deletion, the object is not retrievable.

packages/api/tests/lambdas/test-sqs-message-consumer.js
```diff
@@ -362,3 +363,63 @@ test.serial('SQS message consumer queues correct number of workflows for rules m
...
+    Bucket: process.env.system_bucket,
+    Key: keyTwo,
+  }).promise(), { code: 'NoSuchKey' });
+
+  t.teardown(async () => {
+    await cleanupRulesAndQueues(rules, queues);
+  });
+});
+
+test.serial('archiveMessage archives all SQS messages', async (t) => {
```
markdboyd, line 404: What is this test specifically verifying vs the previous test? It says "archive all messages", perhaps implying that it should be testing that multiple messages get archived, but it's not setting up multiple messages? If it were, that might be a valuable test

packages/ingest/test/test-consumer.js
```diff
@@ -10,11 +10,13 @@ const { Consumer } = require('../consumer');
...
 const timeToReceiveMessages = 200; // ms
 const timeLimitModifier = 50;
+const sqsMessage = { Body: 'message', MessageId: 'id' };
 let testConsumer;
 
 async function stubReceiveSQSMessages(_url, { numOfMessages }) {
   await delay(timeToReceiveMessages);
-  return Array.apply(undefined, { length: numOfMessages }).map(() => 'i am a message'); // eslint-disable-line prefer-spread
+  // eslint-disable-next-line prefer-spread
+  return Array.apply(undefined, { length: numOfMessages }).map(() => sqsMessage);
```
markdboyd, line 19: No objection here, but why the change?

packages/api/lambdas/sqs-message-consumer.js
```diff
@@ -61,23 +87,26 @@ async function processQueues(event, dispatchFn) {
...
  * @param {string} queueUrl - Queue URL for incoming message
  * @param {Object} message - incoming queue message
  * @returns {Promise} - promise resolved when the message is dispatched
  */
-function dispatch(queueUrl, message) {
+async function dispatch(queueUrl, message) {
   const messageReceiveCount = Number.parseInt(message.Attributes.ApproximateReceiveCount, 10);
   const rulesForQueue = this.rulesForQueue;
+  log.info(`Archiving messages from queue ${queueUrl}`);
+  await archiveMessage(message);
```
markdboyd, line 109: Can you refresh my memory: did we decide that we only want to implement this message archiving for SQS rules?

packages/api/lambdas/sqs-message-consumer.js
```diff
@@ -61,23 +87,26 @@ async function processQueues(event, dispatchFn) {
       visibilityTimeout,
       deleteProcessedMessage: false,
     });
-    log.info(`processing queue ${queueUrl}`);
 
-    const messageConsumerFn = dispatchFn.bind({ rulesForQueue });
+    log.info(`Processing queue ${queueUrl}`);
+    const messageConsumerFn = dispatchFn.bind({ queueUrl, rulesForQueue });
```
markdboyd, line 92: I don't think we need to add `queueUrl` to this `bind` expression. `queueUrl` is already passed to the message consumer function:

https://github.com/nasa/cumulus/blob/master/packages/ingest/src/consumer.ts#L44

CHANGELOG.md
```diff
@@ -257,6 +257,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
   - **CUMULUS-2468** - Added `@cumulus/aws-client/DynamoDb.parallelScan` helper to perform [parallel scanning on DynamoDb tables](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Scan.html#Scan.ParallelScan)
   - **CUMULUS-2507**
     - Updated granule record write logic to set granule status to `failed` in both Postgres and DynamoDB if any/all of its files fail to write to the database.
+- **CUMULUS-2371**
```
markdboyd, line 260: These changelog notes need to be moved under `Unreleased`. Currently this is under release 9.0.0

lambdas/sqs-message-remover/src/index.js
```diff
@@ -4,17 +4,40 @@ const get = require('lodash/get');
...
+ * Deletes archived SQS Message from S3
+ *
+ * @param {Object} messageId - SQS message ID
+ * @returns {void}
+ */
+async function deleteArchivedMessage(messageId) {
+  const bucket = envUtils.getRequiredEnvVar('system_bucket', process.env);
+  const stackName = envUtils.getRequiredEnvVar('stackName', process.env);
+  const key = `${stackName}/archived-incoming-messages/${messageId}`;
+  if (bucket && key) {
```
markdboyd, line 30: I think the `if` statement here is unnecessary if we're already doing a try/catch (which is good). The try/catch will handle logging the error if either bucket/key is missing. And actually, with the `if` statement, it's perhaps less clear what happened because no error gets logged

lambdas/sqs-message-remover/src/index.js
```diff
@@ -4,17 +4,40 @@ const get = require('lodash/get');
...
 
 const logger = new Logger({ sender: '@cumulus/sqs-message-remover' });
 
+/**
+ * Deletes archived SQS Message from S3
+ *
+ * @param {Object} messageId - SQS message ID
+ * @returns {void}
+ */
+async function deleteArchivedMessage(messageId) {
```
markdboyd, line 26: I think it might be good to move this function and `archiveMessage` to the `ingest` package (`ingest/sqs.ts`?) and add unit tests for them explicitly. The other reason I like having those two functions in the same file is that we can centralize this logic for getting the message key into a third function:

`${stackName}/archived-incoming-messages/${messageId}`

Otherwise, it just seems too easy for bugs to happen with a typo in the S3 key in either function.

Would also rename `archiveMessage` to `archiveSqsMessageToS3` and `deleteArchivedMessage` to `deleteArchivedMessageFromS3`

example/spec/parallel/testAPI/sqsRuleSpec.js
```diff
@@ -238,5 +241,14 @@ describe('The SQS rule', () => {
     it('messages are picked up and removed from source queue', async () => {
       await expectAsync(waitForQueueMessageCount(queues.sourceQueueUrl, 0)).toBeResolved();
     });
+
+    it('stores incoming messages on S3', async () => {
```
markdboyd, line 245: So we can only test that the message exists on S3 in the case of an unprocessable message, otherwise it would have been deleted from S3 possibly before this test could check for it?

markdboyd: Should we possible clean up this message manually then in the `afterAll` for this block?

markdboyd (changes requested):
This looks great! A few requested changes but really nice work!

example/spec/parallel/testAPI/sqsRuleSpec.js
botanical: Oh, cleaning up the message afterwards is a good point. Yes, we can only check that it exists if it is unprocessable.

packages/ingest/test/test-consumer.js
botanical: It made more sense to me because it specifies disabling that rule for the next line `eslint-disable-next-line`, but I am not sure it's a good enough reason to keep the change

packages/api/lambdas/sqs-message-consumer.js
botanical: We want to archive all messages but this ticket targets SQS messages. There is two other tickets for handling SNS and Kinesis ones. 
https://bugs.earthdata.nasa.gov/browse/CUMULUS-2370
https://bugs.earthdata.nasa.gov/browse/CUMULUS-2372

example/spec/parallel/testAPI/sqsRuleSpec.js
botanical: [`5223038` (#2132)](https://github.com/nasa/cumulus/pull/2132/commits/522303829adf7b5c63fec589dced18ebcf084cf4)

lambdas/sqs-message-remover/src/index.js
botanical: [`5223038` (#2132)](https://github.com/nasa/cumulus/pull/2132/commits/522303829adf7b5c63fec589dced18ebcf084cf4)

lambdas/sqs-message-remover/src/index.js
botanical: [`5223038` (#2132)](https://github.com/nasa/cumulus/pull/2132/commits/522303829adf7b5c63fec589dced18ebcf084cf4)

CHANGELOG.md
botanical: [`5223038` (#2132)](https://github.com/nasa/cumulus/pull/2132/commits/522303829adf7b5c63fec589dced18ebcf084cf4)

packages/api/lambdas/sqs-message-consumer.js
botanical: [`5223038` (#2132)](https://github.com/nasa/cumulus/pull/2132/commits/522303829adf7b5c63fec589dced18ebcf084cf4)

lambdas/sqs-message-remover/src/index.js
```diff
@@ -97,6 +103,7 @@ function handler(event) {
 }
 
 module.exports = {
+  deleteArchivedMessage,
```
markdboyd, line 106: This export no longer refers to anything

lambdas/sqs-message-remover/tests/test-index.js
```diff
@@ -233,3 +241,38 @@ test('sqsMessageRemover lambda updates message visibilityTimeout when workflow f
...
+  const key = `${process.env.stackName}/archived-incoming-messages/${eventSource.messageId}`;
+
+  await s3PutObject({
+    Bucket: process.env.system_bucket,
+    Key: key,
+    Body: msgBody,
+  });
+
+  await updateSqsQueue(eventMessage);
+  await t.throwsAsync(s3().getObject({
```
markdboyd, line 274: Nit: prefer `getObject` from `aws-client/s3`

packages/ingest/src/sqs.ts
```diff
@@ -0,0 +1,56 @@
+'use strict';
+
+import Logger from '@cumulus/logger';
+import { SQSMessage } from '@cumulus/aws-client/SQS';
+const { envUtils } = require('@cumulus/common');
+const { s3PutObject } = require('@cumulus/aws-client/S3');
+const { deleteS3Object } = require('@cumulus/aws-client/S3');
+const logger = new Logger({ sender: '@cumulus/ingest/sqs' });
```
markdboyd, line 8: Prefer to use `import` style statements in TS code

CHANGELOG.md
```diff
@@ -62,6 +62,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- **CUMULUS-2371**
+  - Added `archiveMessage` to `sqs-message-consumer` which archives all incoming SQS messages to S3.
+  - Added `deleteArchivedMessage` to `sqs-message-remover` which deletes archived SQS message from S3 once it has been processed.
```
markdboyd, line 67: These are no longer quite accurate. We should note the behavior change in each lambda, but these functions were not added to those Lambdas

example/spec/parallel/testAPI/sqsRuleSpec.js
```diff
@@ -238,5 +244,14 @@ describe('The SQS rule', () => {
...
+      const keyOne = `${config.stackName}/archived-incoming-messages/${messageId}`;
```
markdboyd, line 249: We could use the helper to get this key, now that we have it

packages/api/lambdas/sqs-message-consumer.js
```diff
@@ -61,23 +63,26 @@ async function processQueues(event, dispatchFn) {
...
  *
  * @param {string} queueUrl - Queue URL for incoming message
  * @param {Object} message - incoming queue message
  * @returns {Promise} - promise resolved when the message is dispatched
  */
-function dispatch(queueUrl, message) {
+async function dispatch(queueUrl, message) {
   const messageReceiveCount = Number.parseInt(message.Attributes.ApproximateReceiveCount, 10);
   const rulesForQueue = this.rulesForQueue;
+  log.info(`Archiving messages from queue ${queueUrl}`);
```
markdboyd, line 84: I would remove this log from here and put it in `archiveSqsMessageToS3`, if anything. Because the log is specific to the logic in that function itself

packages/ingest/test/test-sqs.js
```diff
@@ -0,0 +1,89 @@
...
+      Bucket: process.env.system_bucket,
+      Key: k,
+    })).promise()));
+
+  const msgBody = (m) => t.truthy(JSON.parse(JSON.parse(m.Body.toString())));
+
+  items.every(msgBody);
+});
+
+test.serial('deleteArchivedMessages deletes archived message in S3', async (t) => {
```
markdboyd, line 59:
```suggestion
test.serial('deleteArchivedMessageFromS3 deletes archived message in S3', async (t) => {
```

packages/ingest/package.json
```diff
@@ -37,6 +37,7 @@
   "author": "Cumulus Authors",
   "license": "Apache-2.0",
   "dependencies": {
+    "@cumulus/api": "8.1.0",
```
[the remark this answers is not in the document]
botanical: It's for 
```
const { createSqsQueues } = require('@cumulus/api/lib/testUtils');
``` 
in the unit test. I could just copy this function. Why do we want to avoid this dependency? Should we just move that function outside of this package?

packages/ingest/src/sqs.ts
```diff
@@ -0,0 +1,56 @@
...
+
+function getS3KeyForArchivedMessages(stackName: string, messageId?: string) {
```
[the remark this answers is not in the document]
botanical: It's optional because the Message interface has `MessageId` as optional. Though, it seems like all messages should be guaranteed a message ID. Will update.

packages/ingest/package.json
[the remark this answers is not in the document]
markdboyd: We want to avoid the dependency because `packages/api` is a huge package which itself has a lot of dependencies, which has a lot of undesirable side effects for bootstrap time, compiled code size, or even security vulnerability. Also `packages/api` depends on `packages/ingest`, so you can't really create this dependency, otherwise you now have a circular dependency.

I'd just some manually replicate `createSqsQueues` is doing for the sake of the tests in this package. Also be sure to check whether you really need everything that function is doing

packages/ingest/test/test-sqs.js
```diff
@@ -0,0 +1,89 @@
...
+  process.env.system_bucket = randomString();
+  process.env.stackName = randomString();
+  await createBucket(process.env.system_bucket);
+});
+
+test.after.always(async () => {
+  await recursivelyDeleteS3Bucket(process.env.system_bucket);
+});
+
+test.serial('archiveSqsMessageToS3 archives all SQS messages', async (t) => {
```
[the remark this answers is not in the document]
botanical: That sounds reasonable to me. For the `test-sqs-message-consumer` test, should I be spying on `archiveSqsMessageToS3` as well as checking on the contents of S3? I wrote a test that sends several SQS messages and checks that all messages are archived in S3 but can definitely add one that spies on `archiveSqsMessageToS3`

packages/api/tests/lambdas/test-sqs-message-consumer.js
botanical: Oh, gotcha

lambdas/sqs-message-remover/tests/test-index.js
botanical: Like this? [`b43c09c` (#2132)](https://github.com/nasa/cumulus/pull/2132/commits/b43c09c2e91fbdf652ce65d88a0bb760a3df8639) I don't know if I prefer it but don't feel too strongly.

CHANGELOG.md
botanical: [`c8bed0a` (#2132)](https://github.com/nasa/cumulus/pull/2132/commits/c8bed0a333136af6c382568ed4831b83d0b662ef)
Apologies, I updated this in two different commits

example/spec/parallel/testAPI/sqsRuleSpec.js
botanical: Does this require publishing the ingest package again? @markdboyd

packages/api/lambdas/sqs-message-consumer.js
botanical: Makes sense to me

lambdas/sqs-message-remover/tests/test-index.js
```diff
@@ -7,6 +7,12 @@ const get = require('lodash/get');
 const awsServices = require('@cumulus/aws-client/services');
 const { receiveSQSMessages } = require('@cumulus/aws-client/SQS');
 const { randomString } = require('@cumulus/common/test-utils');
+const {
+  createBucket,
+  s3PutObject,
+} = require('@cumulus/aws-client/S3');
+const { s3 } = require('@cumulus/aws-client/services');
+const { getObject } = require('@cumulus/aws-client/S3');
```
markdboyd, line 15: We're already requiring `@cumulus/aws-client/S3` above, so we should add `getObject` to what is required from there

packages/ingest/test/test-sqs.js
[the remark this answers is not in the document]
markdboyd: I'm not sure if you should need to use spies at all. It seems like you could mostly repeat the setup of the `test.serial('processQueues archives messages from the ENABLED sqs rule only` test, except modify it to create multiple messages for the enabled SQS rule, then check to ensure that those messages end up on S3

packages/ingest/src/sqs.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  if (!message.MessageId) {
+    const error = new Error(`MessageId on message ${message} required but not found.`);
+    logger.error(error.message);
+    throw error;
+  }
+
+  const key = getS3KeyForArchivedMessage(stackName, message.MessageId);
+  const body = JSON.stringify(message.Body);
+  logger.info(`Archiving messages from queue ${queueUrl}`);
```
markdboyd, line 34:
```suggestion
  logger.info(`Archiving message ${message.MessageId} from queue ${queueUrl}`);
```

lambdas/sqs-message-remover/tests/test-index.js
```diff
@@ -233,3 +242,38 @@ test('sqsMessageRemover lambda updates message visibilityTimeout when workflow f
...
+  const sqsOptions = { numOfMessages: 10, visibilityTimeout: 120, waitTimeSeconds: 20 };
+  const receiveMessageResponse = await receiveSQSMessages(sqsQueues.queueUrl, sqsOptions);
+  const { MessageId: messageId, ReceiptHandle: receiptHandle } = receiveMessageResponse[0];
+
+  const eventSource = createEventSource({ messageId, receiptHandle, queueUrl: sqsQueues.queueUrl });
+  const eventMessage = createCloudwatchEventMessage({
+    status: 'SUCCEEDED',
+    eventSource,
+  });
+  const key = `${process.env.stackName}/archived-incoming-messages/${eventSource.messageId}`;
```
markdboyd, line 266: It should be safe to use the helper to get the key in this unit test

packages/ingest/test/test-sqs.js
```diff
@@ -0,0 +1,107 @@
...
+    Key: key,
+    Body: JSON.stringify(sqsMessage.Body),
+  });
+
+  // Check that item exists in S3
+  const item = await s3().getObject({
+    Bucket: process.env.system_bucket,
+    Key: key,
+  }).promise();
+  t.truthy(item);
```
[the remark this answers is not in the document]
botanical: Oh, didn't know about `s3ObjectExists`, that is useful - thanks!

packages/api/tests/lambdas/test-sqs-message-consumer.js
```diff
@@ -362,3 +364,85 @@ test.serial('SQS message consumer queues correct number of workflows for rules m
...
+  await handler(event);
+
+  const item = await s3().getObject({
+    Bucket: process.env.system_bucket,
+    Key: enabledQueueKey,
+  }).promise();
+
+  t.deepEqual(message, JSON.parse(JSON.parse(item.Body.toString())));
+
+  await t.throwsAsync(s3().getObject({
```
markdboyd, line 396: Use `t.false` with `s3ObjectExists`. The reason I'm asking for this is that it's more clear what the test is checking for and we don't need to compare against a specific error coe

packages/api/tests/lambdas/test-sqs-message-consumer.js
```diff
@@ -362,3 +365,85 @@ test.serial('SQS message consumer queues correct number of workflows for rules m
...
+  );
+  const deriveKey = (m) => getS3KeyForArchivedMessage(stackName, m.MessageId);
+  const keys = messages.map((m) => deriveKey(m));
+
+  // Send message to DISABLED queue
+  const secondMessage = await SQS.sendSQSMessage(
+    queues[1].queueUrl,
+    { testdata: randomString() }
+  );
+  const deadLetterKey = getS3KeyForArchivedMessage(stackName, secondMessage.MessageId);
```
markdboyd, line 427: I don't think this is necessary for this test.  We just want to prove that multiple messages are handled for the enabled rule. The previous test already proves that only messages for an enabled rule are handled.

packages/api/tests/lambdas/test-sqs-message-consumer.js
```diff
@@ -362,3 +365,85 @@ test.serial('SQS message consumer queues correct number of workflows for rules m
...
+
+  await handler(event);
+
+  const item = await s3().getObject({
+    Bucket: process.env.system_bucket,
+    Key: enabledQueueKey,
+  }).promise();
+
+  t.deepEqual(message, JSON.parse(JSON.parse(item.Body.toString())));
```
markdboyd, line 395: Why do we have to do a double `JSON.parse`?

packages/api/tests/lambdas/test-sqs-message-consumer.js
```diff
@@ -362,3 +365,85 @@ test.serial('SQS message consumer queues correct number of workflows for rules m
...
+
+  await handler(event);
+
+  const items = await Promise.all(keys.map(async (k) =>
+    (s3().getObject({
+      Bucket: process.env.system_bucket,
+      Key: k,
+    })).promise()));
+
+  const msgBody = (m) => t.truthy(JSON.parse(JSON.parse(m.Body.toString())));
```
markdboyd, line 437: If we're going to compare against every message, we might as well check that it has the exact expected content. Otherwise just doing a truthy comparison isn't very meaningful

Or we can just do an exists check here since the previous test verifies that the correct message is archived

packages/api/tests/lambdas/test-sqs-message-consumer.js
markdboyd: I guess my question then is: does it have to be that way? Or is it only that way because of the code we've written?

Do we need to do a `JSON.stringify` on `message.Body` here?

https://github.com/nasa/cumulus/pull/2132/files?file-filters%5B%5D=.ts#diff-38eb9e3df7df0195f5bf20728f154f3d8d7ec21729d9884b010fae57d2e6c703R33

It seems like the message body returned from SQS should already be a string? 

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_Message.html

So if it was JSON it would already have been stringified, meaning we don't need to stringify it again?

packages/ingest/test/test-sqs.js
```diff
@@ -0,0 +1,108 @@
...
+  const body = { testdata: randomString() };
+  const message = await SQS.sendSQSMessage(
+    queues.queueUrl,
+    body
+  );
+  const messages = await receiveSQSMessages(
+    queues.queueUrl,
+    { numOfMessages: 1, visibilityTimeout: 5 }
+  );
+  const key = `${process.env.stackName}/archived-incoming-messages/${message.MessageId}`;
```
markdboyd, line 66: Seems like you could use the helper to get these keys in the unit tests?

markdboyd (approved):

PR #2138, opened 2021-03-11
CUMULUS-2428: Enable using custom queue for bulk operations

markdboyd:

Addresses [CUMULUS-2428: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2428)

## Changes

* Update granules bulk operation (`/granules/bulk`) to accept a `queueUrl' property
* Update bulk operations lambda to use queueUrl, if provided, when scheduling bulk granule workflows

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Create corresponding cumulus-api docs upgrade
- [x] Unit tests
- [ ] Adhoc testing
- [x] Integration tests


CHANGELOG.md
```diff
@@ -6,9 +6,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### BREAKING CHANGES
+
+- **CUMULUS-2428**
+  - Changed `/granules/bulk` to use `queueUrl` property instead of a `queueName` property for setting the queue to use for scheduling bulk granule workflows
```
markdboyd, line 12: I hate flagging this as a breaking change since technically the `queueName` property wasn't actually being used at all by the code, but technically this is a breaking change to the payload expected/used by our API

example/spec/parallel/testAPI/bulkOperationsSuccessSpec.js
```diff
@@ -0,0 +1,293 @@
...
+    });
+
+    it('returns a status code of 202', () => {
+      expect(beforeAllSucceeded).toBeTrue();
+      expect(postBulkGranulesResponse.statusCode).toEqual(202);
+    });
+
+    it('returns an Async Operation Id', () => {
+      expect(beforeAllSucceeded).toBeTrue();
+      expect(postBulkOperationsBody.id).toMatch(/[\da-f]{8}(?:-[\da-f]{4}){3}-[\da-f]{12}/);
```
npauzenga, line 209: It looks like we're using this regex in a couple other tests (e.g. `bulkdDeleteFailureSpec`). Could we put it in a test utils somewhere?

example/spec/parallel/testAPI/bulkOperationsSuccessSpec.js
```diff
@@ -0,0 +1,293 @@
...
+
+      expect(getAsyncOperationResponse.statusCode).toEqual(200);
+      expect(getAsyncOperationBody.status).toEqual('SUCCEEDED');
+
+      let output;
+      try {
+        output = JSON.parse(getAsyncOperationBody.output);
+      } catch (error) {
+        throw new SyntaxError(`getAsyncOperationBody.output is not valid JSON: ${getAsyncOperationBody.output}`);
+      }
```
npauzenga, line 265: Is it possible to have an `asyncOperation` with an invalid `output` if the record was written correctly and `SUCCEEDED`? Do/should we check on the write that the `output` is parsable JSON?

packages/api-client/src/granules.ts
```diff
@@ -330,6 +330,38 @@ export const listGranules = async (params: {
...
+  return callback({
+    prefix: prefix,
+    payload: {
+      httpMethod: 'POST',
+      resource: '/{proxy+}',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      path: '/granules/bulk',
+      body: JSON.stringify(body),
```
npauzenga, line 360: Would `object` be a more accurate type for `body`? I guess `stringify` doesn't care and will just make anything a string so it probably doesn't matter.

packages/api/lambdas/bulk-operation.js
```diff
@@ -80,18 +80,17 @@ function applyWorkflowToGranules({
   granuleIds,
   workflowName,
   meta,
-  queueName,
+  queueUrl,
+  granuleModel = new GranuleModel(),
```
npauzenga, line 84: 👍

npauzenga (commented):
Looks great to me! Just had a couple questions.

example/spec/parallel/testAPI/bulkOperationsSuccessSpec.js
markdboyd: I'm not sure if I understand - but by surrounding the JSON.parse in a try/catch and throwing an error if the JSON.parse throws, we are ensuring that the output is parsable JSON

CHANGELOG.md
markdboyd: As I thought about it more, the fact that `queueName` wasn't being used due to a bug is kind of a detail. It's still a breaking change to the API either way, so as a user if you did have code making requests with `queueName`, you would have to change it to `queueUrl`.

npauzenga: Yeah, totally agree. If it was exposed this should be marked as a breaking change.

example/spec/parallel/testAPI/bulkOperationsSuccessSpec.js
npauzenga: Yeah, I don't think it would need to be a separate check.

npauzenga (approved):
Looks great. Thanks @markdboyd!

PR #2156, opened 2021-03-26
CUMULUS-2266: Execution/PDR/Granule writes should not write to DLQ for out of order processing

nemreid:

Addresses [CUMULUS-2266: Execution/PDR/Granule writes should not write to DLQ for out of order processing](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2266)

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


packages/api/models/granules.js
```diff
@@ -650,7 +649,15 @@ class Granule extends Manager {
       updateParams.ConditionExpression += ' and #execution <> :execution';
     }
 
-    return this.dynamodbDocClient.update(updateParams).promise();
+    try {
+      return await this.dynamodbDocClient.update(updateParams).promise();
+    } catch (error) {
+      if (error.name && error.name.includes('ConditionalCheckFailedException')) {
```
markdboyd, line 655: A delayed `running` event isn't the only possible cause of the conditional check exception though? It could also be because of the `createdAt` timestamp condition. I guess that only makes the log message invalid, but not the general logic?

nemreid: True. I can update the log message.

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-index.js
```diff
@@ -411,3 +413,31 @@ test('Lambda sends message to DLQ when writeRecords() throws an error', async (t
...
+    pdrModel,
+    knex,
+    pdrName,
+    granuleId,
+  } = t.context;
+
+  const pdrPgModel = new PdrPgModel();
+  const granulePgModel = new GranulePgModel();
+
+  cumulusMessage.meta.status = 'completed';
```
markdboyd, line 430: So the test setup is "out of order" because the "completed" message is written before the "running" one. And we're just proving that an error is not thrown?

What is guaranteeing that the first message is "older" than the second one? Is that guaranteed? Is it important for this test?

nemreid: I will add a second unit test that tests the createdAt condition.

nemreid: https://github.com/nasa/cumulus/pull/2156/commits/1027bd6a4454d070ed340d950ee02e03372f90d2

packages/api/models/granules.js
nemreid: https://github.com/nasa/cumulus/pull/2156/commits/1027bd6a4454d070ed340d950ee02e03372f90d2

markdboyd (approved):

PR #2169, opened 2021-04-07
[CUMULUS-1808] Expanded error messaging when disabling rule

dopeters:

Addresses [CUMULUS-1808: Permission is not found when attempting to disable an sns rule from the dashboard.](https://bugs.earthdata.nasa.gov/browse/CUMULUS-1808)

## Changes

* Adds additional info to error message

This error appears when the trigger has been manually deleted from the lambda. Deleting the trigger also deletes the subscription. In order to get around this error, the user will probably have to manually delete the rule from Dynamo. I considered catching the error and letting the disable/delete happen anyway so that the user can avoid manual intervention. However, if the error pops up for a different reason, it would hide the error and potentially cause other issues. Since this has only been reported a couple of times, rather than change any base behavior, I just added info to the error that's reported to the user.

To test on the dashboard: create/enable an SNS rule. Find the associated lambda and delete the SNS trigger (if you go to the sns topic, you can find the lambda in the subscription). On dashboard, attempt to disable the rule. You should see the error with additional messaging. 

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [ ] Integration tests


packages/api/lib/errors.js
```diff
@@ -32,10 +35,26 @@ class EarthdataLoginError extends Error {
   }
 }
 
+const resourceNotFoundInfo = 'Check if topic subscription and/or lambda trigger have been manually deleted from AWS. If so, rule may need to be manually disabled/deleted.';
```
flamingbear, line 38: This warning may be too specific/directive.
```suggestion
const resourceNotFoundInfo = 'One solution may be to check if topic subscription and/or lambda trigger have been manually deleted from AWS. If so, rule may need to be manually disabled/deleted.';
```

flamingbear: I'm saying this because I was naively editing a kinesis rule to be SNS, and got your error.  I of course can't reproduce it now.

flamingbear (approved):
This looks good to me, I did have a question about the specificity of the extra information added, so you can take or leave that advice.  I'll approve either way.

packages/api/lib/errors.js
dopeters: Yeah, there probably are more ways for this to crop up, so having it be less specific makes sense

PR #2170, opened 2021-04-07
Migration code changes

markdboyd:


lambdas/postgres-migration-async-operation/package.json
```diff
@@ -28,11 +28,11 @@
...
   "dependencies": {
-    "@cumulus/api": "3.0.0",
+    "@cumulus/api": "7.1.0",
     "@cumulus/async-operations": "3.0.1-alpha.1",
-    "@cumulus/logger": "3.0.0"
+    "@cumulus/logger": "7.1.0"
   },
   "devDependencies": {
-    "@cumulus/test-data": "3.0.0"
+    "@cumulus/test-data": "7.1.0"
```
botanical, line 36: thanks for updating these!

lambdas/data-migration1/package.json
```diff
@@ -37,6 +36,7 @@
     "pg": "^8.3.0"
   },
   "devDependencies": {
+    "@cumulus/api": "7.1.0",
```
botanical, line 39: What is the main distinction between `dependencies` and `devDependencies`? Did it affect why you moved `@cumulus/api`?

botanical (commented):
What about the tests in `lambdas/data-migration1/tests/test-async-operations.js`: `migrateAsyncOperations processes all non-failing records` 
and `lambdas/data-migration1/tests/test-providers.js`: 
`'migrateProviderRecord throws error on invalid source data from Dynamo`
`migrateProviders processes all non-failing records`?

lambdas/data-migration1/package.json
markdboyd: This sums it up better than I would: https://nodejs.dev/learn/npm-dependencies-and-devdependencies

> Development dependencies are intended as development-only packages, that are unneeded in production. For example testing packages, webpack or Babel.

With the changes that I made, `@cumulus/api` is no longer used directly by the lambda code, it's only used by unit test code, so it's a dev dependency for running tests but not for the code in production

markdboyd:
@botanical

Good call on the `migrateProviderRecord throws error on invalid source data from Dynamo` test. The name of that test certainly no longer made sense, though I felt like it was still a valuable test with a minor change. Let me know what you think

4f6f995de6b209b44fcac7376679bded77ac927e

As for the `migrateProviders processes all non-failing records` and `migrateAsyncOperations processes all non-failing records` tests? They are passing and their logic makes sense to me unless I'm missing something

markdboyd:
I also updated the providers and rules migration to use the translate logic from the db package, which greatly simplifies the data migration logic:

5cbf68147fe75b2dfc20969f0e0cf3bcb4f0ac56
78648b201e25a5629f53a695648c3252ba103eca

botanical (approved):
Apologies for not elaborating on my last comment. It was more a comment about if we needed to persist those tests as well because there's no longer a concept of a "failing"/invalid record, if that makes sense. But otherwise, everything checks out and looks good.

Jkovarik (approved):

PR #2174, opened 2021-04-08
Move Dynamo table names for API lambdas into SSM parameter

markdboyd:
I started getting this error when trying to deploy the feature branch:

```
Error: Error modifying Lambda Function (mboyd-test-tf-ApiEndpoints) configuration : InvalidParameterValueException: Lambda was unable to configure your environment variables because the environment variables you have provided exceeded the 4KB limit.
```

The cause of the error is that the stringified version of all of our environment variables for the API lambdas are greater than 4096 bytes, which is not surprising since we've added a few environment variables to those Lambdas in this feature branch

The solution @Jkovarik  and I settled on as a temporary fix is to move some of the variables (right now only Dynamo table names) into a Systems Manager parameter, then read that parameter when starting up the API, and add the table names as environment variables.

### DIFF
--- a/packages/api/app/index.js
+++ b/packages/api/app/index.js
@@ -11,6 +11,9 @@ const morgan = require('morgan');
 const awsServerlessExpress = require('aws-serverless-express');
 const awsServerlessExpressMiddleware = require('aws-serverless-express/middleware');
 
+const { services } = require('@cumulus/aws-client');
+const { MissingRequiredEnvVar } = require('@cumulus/errors');
+
 const router = require('./routes');
 const { jsonBodyParser } = require('./middleware');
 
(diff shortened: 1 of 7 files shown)


packages/api/app/index.js
```diff
@@ -58,14 +61,36 @@ app.use((err, _req, res, _next) => {
...
+
   // workaround to support multiValueQueryStringParameters
   // untill this is fixed: https://github.com/awslabs/aws-serverless-express/issues/214
   const modifiedEvent = {
     ...event,
     queryStringParameters: event.multiValueQueryStringParameters || event.queryStringParameters,
   };
-  awsServerlessExpress.proxy(server, modifiedEvent, context);
+  // see https://github.com/vendia/serverless-express/issues/297
+  return new Promise((resolve, reject) => {
```
markdboyd, line 87: Version 4.0.0+ of the serverless-express package would allow us to use an async handler for this Lambda without a workaround, but would require other updates to our Lambda code:

https://github.com/vendia/serverless-express/blob/mainline/CHANGELOG.md#400-2021-02-01
https://github.com/vendia/serverless-express/blob/mainline/UPGRADE.md

And since this is a fix directed at the environment variable problem, I didn't want to blow up the scope here to do a bunch of API endpoint refactors

tf-modules/archive/api.tf
```diff
@@ -1,13 +1,28 @@
...
+    AccessTokensTable          = var.dynamo_tables.access_tokens.name
+    RulesTable                 = var.dynamo_tables.rules.name
+    AsyncOperationsTable       = var.dynamo_tables.async_operations.name
+    CollectionsTable           = var.dynamo_tables.collections.name
+    ExecutionsTable            = var.dynamo_tables.executions.name
+    GranulesTable              = var.dynamo_tables.granules.name
+    PdrsTable                  = var.dynamo_tables.pdrs.name
+    ProvidersTable             = var.dynamo_tables.providers.name
+    ReconciliationReportsTable = var.dynamo_tables.reconciliation_reports.name
+    RulesTable                 = var.dynamo_tables.rules.name
```
botanical, line 14: I think you can remove this last RulesTable since you have it on line 6

botanical (approved):
I pulled this down and deployed successfully and confirmed that parameter store value was correct.

PR #2175, opened 2021-04-09
CUMULUS-2452: Supporting migrating granules by collection

markdboyd:

Addresses [CUMULUS-2452: Supporting migrating granules by collection](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2452)

## Changes

* Add support for migrating only certain granules by `granuleId` or `collectionId`
* Add support for running migration for only certain data types as part of data-migration2

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Adhoc testing - Have not done adhoc testing and may wait until [CUMULUS-2468]((https://bugs.earthdata.nasa.gov/browse/CUMULUS-2468) is addressed and running migrations against large sets of data may be more efficient


packages/aws-client/src/DynamoDbSearchQueue.ts
```diff
@@ -1,17 +1,20 @@
...
 class DynamoDbSearchQueue {
   private readonly dynamodbDocClient: AWS.DynamoDB.DocumentClient;
-  private readonly searchType: 'scan';
-  private readonly params: AWS.DynamoDB.DocumentClient.ScanInput;
+  private readonly searchType: searchType;
+  private readonly params: searchParams;
   private items: Array<AWS.DynamoDB.DocumentClient.AttributeMap|null>;
 
-  constructor(params: AWS.DynamoDB.DocumentClient.ScanInput, searchType: 'scan' = 'scan') {
+  constructor(params: AWS.DynamoDB.DocumentClient.ScanInput, searchType: searchType = 'scan') {
```
markdboyd, line 17: Probably need more unit tests for the `query` case?

packages/db/src/models/granule.ts
```diff
@@ -34,18 +34,23 @@ export default class GranulePgModel extends BasePgModel<PostgresGranule, Postgre
           updated_at: granule.updated_at,
           created_at: granule.created_at,
         })
-        .where(knexOrTrx.raw(`${this.tableName}.created_at <= to_timestamp(${translateDateToUTC(granule.created_at)})`))
+        .where(knexOrTrx.raw(`${this.tableName}.created_at <= to_timestamp(${translateDateToUTC(granule.created_at)})`));
+
+      if (executionCumulusId) {
```
markdboyd, line 39: The only place where `executionCumulusId` could be undefined is from the data migrations

lambdas/data-migration2/src/index.ts
```diff
@@ -5,43 +5,71 @@ import { migrateExecutions } from './executions';
...
-    const executionsMigrationSummary = await migrateExecutions(env, knex);
-    const granulesAndFilesMigrationSummary = await migrateGranulesAndFiles(env, knex);
-    const pdrsMigrationSummary = await migratePdrs(env, knex);
-
-    const summary = `
+    let summary = `
       Migration summary:
+    `;
+
+    if (migrationsToRun.includes('executions')) {
```
markdboyd, line 32: Adding the ability to run only certain parts of data-migration2 seemed like a necessity for re-run scenarios where users only want to retry migrating certain failed executions or granules. 

Adding this configurability is a bit dangerous, as if users were to configure data-migration2 to only do `granules` on the **first run**, then they would all fail as they could not refer to the executions that weren't migrated. But overall I think the configurability is desirble and we just need to document appropriately

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -179,13 +180,45 @@ export const migrateGranuleAndFilesViaTransaction = async (
...
+        ':granuleId': granuleSearchParams.granuleId,
+      },
+    };
+  } else if (granuleSearchParams.collectionId) {
+    dynamoSearchType = 'query';
+    extraSearchParams = {
+      IndexName: 'collectionId-granuleId-index',
+      KeyConditionExpression: 'collectionId = :collectionId',
+      ExpressionAttributeValues: {
+        ':collectionId': granuleSearchParams.collectionId,
```
markdboyd, line 210: I opted to support `collectionId` (concatenation of collection name/version) as the supported filter value since that is how the value shows up in Dynamo, so they can copy it straight from Dynamo records into the payload for this migration

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -416,6 +427,10 @@ test.serial('migrateGranuleRecord updates an already migrated record if the upda
     cumulus_id: granuleCumulusId,
   });
 
+  t.teardown(async () => {
```
markdboyd, line 430: A lot of these type of changes are just improvements to our test cleanup

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -488,8 +506,40 @@ test.serial('migrateGranulesAndFiles skips already migrated granule record', asy
...
-  const records = await knex(tableNames.granules);
+  const records = await t.context.granulePgModel.search(t.context.knex, {});
+  t.is(records.length, 1);
+
+  t.teardown(async () => {
+    await t.context.granulePgModel.delete(t.context.knex, { cumulus_id: records[0].cumulus_id });
+  });
+});
+
+test.serial('migrateGranulesAndFiles processes granule with no files', async (t) => {
```
markdboyd, line 517: I noticed that this wasn't previously supported and it seems like a possible scenario since `files` aren't a required property of granules in Dynamo

packages/aws-client/src/DynamoDbSearchQueue.ts
markdboyd: Please advise whether 8e6b65ded63953aa3ad426a1e956efce18cdc6e3 is sufficient

lambdas/postgres-migration-async-operation/src/index.ts
```diff
@@ -1,28 +1,31 @@
...
     stackName: process.env.stackName,
     systemBucket: process.env.system_bucket,
     dynamoTableName,
     knexConfig: process.env,
     useLambdaEnvironmentVariables: true,
   }, models.AsyncOperation);
 
-  logger.info('Completed Postgres Migration Async Operation');
-  return asyncOperation;
+  logger.info(`Started async operation ${asyncOperationId} for Data Migration 2`);
```
markdboyd, line 29: The previous log message wasn't really accurate. The async operation has not completed when this line is reached, it has just been started and an ID has been returned. The operation completes asynchronously to the process that starts it

botanical: That's a good point. Thanks for updating this to be more accurate!

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -156,7 +145,10 @@ test.serial('migrateGranuleRecord correctly migrates granule record', async (t)
     testGranule,
   } = t.context;
 
-  const [granuleCumulusId] = await migrateGranuleRecord(testGranule, knex);
+  const granuleCumulusId = await migrateGranuleRecord(testGranule, knex);
+  t.teardown(async () => {
```
Jkovarik, line 149: Mentioning here so I don't forget -- this pattern is interesting, although I'm used to seeing it revert spies/mocks,  but if I'm reading right, the order reverses in ava 4.... should we have `reverseTeardowns` enabled?

https://github.com/avajs/ava/blob/main/docs/02-execution-context.md

lambdas/data-migration2/tests/test-granulesAndFiles.js
Jkovarik: Do you mean "covered" or "supported"?

lambdas/data-migration2/src/index.ts
```diff
@@ -1,52 +1,70 @@
...
+      migrationSummary.executions = {
+        total_dynamo_db_records: executionsMigrationResult.total_dynamo_db_records,
+        migrated: executionsMigrationResult.migrated,
+        skipped: executionsMigrationResult.skipped,
+        failed: executionsMigrationResult.failed,
+      };
+    }
+
+    if (migrationsToRun.includes('granules')) {
+      const { granulesResult, filesResult } = await migrateGranulesAndFiles(
```
Jkovarik, line 37: If we're eventually going to be writing these outputs to S3, should we be recording the filter options in the output somehow?   Else if we run a bunch of these reports it might be hard to tell which collection you just tried to migrate.

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -20,15 +20,15 @@ import {
   PostgresUpdateFailed,
 } from '@cumulus/errors';
 
-import { MigrationSummary } from './types';
+import { GranuleDynamoSearchParams, MigrationResult } from '@cumulus/types/migration';
```
Jkovarik, line 23: 🤔  It's out of scope for this ticket, but I'm wondering if we should consider making the granuleId search param a *list* of granuleIds, as if there are failures to target it would allow users to munge their failure results and batch submit a targeted list of failed granules.

lambdas/data-migration2/tests/test-index.js
```diff
@@ -120,128 +108,300 @@ test.after.always(async (t) => {
...
-    source: 'source',
-  });
-
-  const fakeExecution = {
-    arn: randomId('arn'),
-    duration: 180.5,
-    name: randomId('name'),
-    execution: randomId('execution'),
+  const collectionId = `${testCollection.name}___${testCollection.version}`;
+  const fakeExecution = fakeExecutionFactoryV2({
```
Jkovarik, line 121: Nice refactor

lambdas/data-migration2/tests/test-index.js
```diff
@@ -68,49 +71,34 @@ test.before(async (t) => {
...
-  });
-  await executionPgModel.create(
-    t.context.knex,
-    testExecution
-  );
-  t.context.testExecution = testExecution;
 });
 
 test.after.always(async (t) => {
-  await t.context.granulesModel.deleteTable();
```
Jkovarik, line 108: With the preface that I'm probably  misunderstanding something - I think it's fine to have individual record teardowns in the test, but what's the reason for not also removing the dynamo table entirely to ensure that a teardown/kill -9/ctrl c mash doesn't foul up the dynamo env in localstack?

lambdas/data-migration2/tests/test-index.js
```diff
@@ -120,128 +108,300 @@ test.after.always(async (t) => {
...
-      pdrs: {
-        failed: 0,
-        migrated: 1,
-        skipped: 0,
-        total_dynamo_db_records: 1,
-      },
-    },
-  };
-  t.deepEqual(call, expected);
+test.serial('handler migrates only granules if configured', async (t) => {
```
Jkovarik, line 252: Nice thorough tests!

Jkovarik (changes requested):
This is really excellent!   I have a couple of questions/comments for your review, let me know if you have questions or want to pair to look over it.

lambdas/data-migration2/src/pdrs.ts
```diff
@@ -3,13 +3,11 @@ import Knex from 'knex';
 import DynamoDbSearchQueue from '@cumulus/aws-client/DynamoDbSearchQueue';
 import Logger from '@cumulus/logger';
 import {
-  getRecordCumulusId,
```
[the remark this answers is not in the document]
markdboyd: Yeah I plan to do a cleanup PR at some point (obviously very soon) to remove this sort of stuff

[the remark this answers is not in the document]
Jkovarik: :100: this PR is already pretty large so that's fine 👍🏻

lambdas/data-migration2/tests/test-granulesAndFiles.js
markdboyd: In a lot of cases (almost all?) I only have one `teardown` per test, but generally yes we would probably want to enable `reverseTeardowns`

Also, guess who submitted that feature to Ava:

https://github.com/avajs/ava/pull/2505/files

Our very own @yjpa7145

lambdas/data-migration2/tests/test-index.js
markdboyd: I think I read these wrong as deleting PG tables, not dynamo tables. I can restore them

lambdas/data-migration2/src/index.ts
markdboyd: Not a bad idea, but I assume I would need for #2179 to merge first before I could add in that information?

lambdas/data-migration2/src/granulesAndFiles.ts
markdboyd: That's a good idea. I'll update

packages/aws-client/tests/test-DynamoDbSearchQueue.js
```diff
@@ -99,3 +99,40 @@ test.serial('DynamoDbSearchQueue can handle paging', async (t) => {
...
+      Item: {
+        bucket: { S: bucket },
+        key: { S: randomString() },
+      },
+    }).promise()));
+
+  const queue = new DynamoDbSearchQueue(
+    {
+      TableName: t.context.tableName,
+      Limit: 2,
```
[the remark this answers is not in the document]
markdboyd: The `Limit` parameter to a Dynamo scan controls how many items are returned by an individual scan operation. By default, Dynamo will keep scanning until it has 1 MB of data, then return it, by you can specify `Limit` so that each scan operation only evaluates that many records before returning. 

In essence, doing this forces our test to do "paging" of the Dynamo results, which ensures that it is working for the "query" mode. But there's no reason it shouldn't, since that part of the code isn't dependent on the "mode"

packages/db/src/lib/granule.ts
```diff
@@ -19,7 +19,7 @@ import { GranulesExecutionsPgModel } from '../models/granules-executions';
 export const upsertGranuleWithExecutionJoinRecord = async (
   knexTransaction: Knex.Transaction,
   granule: PostgresGranule,
-  executionCumulusId: number,
+  executionCumulusId?: number,
```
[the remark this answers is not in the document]
markdboyd: I noticed that too, which was odd

lambdas/data-migration2/src/index.ts
Jkovarik: Not necessarily, if we updated the output to have a collection key, or something like that.       If we're planning on including it in the *filename* then I guess that should probably be coordinated with @botanical depending on merge order.      

Given you can't have a *list* of collections it might not make sense to have it in the report object itself, but I wouldn't hate it either given it would allow programmatic consumption of some range of report files.        Embedding it in the S3 key probably also suffices to address the concern though, insofar as it is a concern.

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -55,14 +55,15 @@ export const migrateGranuleRecord = async (
...
-  // Schema validation on Dynamo record will fail if `record.execution`
-  // does not exist
-  const executionCumulusId = await executionPgModel.getRecordCumulusId(
-    knex,
-    {
-      url: record.execution,
-    }
-  );
+  // It's possible that very old records could have this field be undefined
+  const executionCumulusId = record.execution
```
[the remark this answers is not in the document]
markdboyd: I'm not sure that I follow. 

The reason for this change is that previously, we had been doing schema validation on the Dynamo granule record, and since `execution` is required as of the current Dynamo schema, that means the record would either fail the validation or would have an execution.

But since we removed that Dynamo schema validation across the board, I realized we needed to clean up the code here to not assume that an `execution` exists for a granule. Since they could have old granule records from before `execution` was required in the Dynamo schema.

The confusing thing is that `execution` from the Dynamo granule records is/should be a link to a Step Function execution URL, so that's how we have to look for the execution record in PG

lambdas/data-migration2/src/granulesAndFiles.ts
markdboyd: Apparently you can’t do an IN/CONTAINS operator in a query against a Dynamo:

https://stackoverflow.com/a/32751240

So it’s not really easy to allow people to specify a list of granule IDs  for the migration without doing it via a scan, which then eliminates the efficiency benefit. 

It also seems like given the forthcoming speed improvements from CUMULUS-2468, asking people to re-run an entire set of granules via the collection may not be unreasonable. But I'm open to other ideas

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -164,68 +165,100 @@ export const migrateGranuleAndFilesViaTransaction = async (
...
+
+  if (granuleSearchParams.granuleId) {
+    dynamoSearchType = 'query';
+    extraSearchParams = {
+      KeyConditionExpression: 'granuleId = :granuleId',
+      ExpressionAttributeValues: {
+        ':granuleId': granuleSearchParams.granuleId,
+      },
+    };
+  } else if (granuleSearchParams.collectionId) {
```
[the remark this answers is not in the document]
markdboyd: I'm inclined towards documentation, since failing feels a bit unnecessary, but open to pushback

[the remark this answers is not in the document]
markdboyd: I updated the AC for CUMULUS-2293

lambdas/data-migration2/src/index.ts
markdboyd: Now that I look more closely, #2179 is really just producing a report of errors. It's not storing a general migration summary to S3.

So I can definitely add this information to the migration result information, but for now that's just going to show up in the CW log unless we do some further work. Thoughts?

packages/aws-client/src/DynamoDbSearchQueue.ts
markdboyd: The test coverage report for this file specifically was:

` DynamoDbSearchQueue.ts   |      76 |    70.59 |      80 |   71.43 | 32-39`

So the only uncovered lines are 32-39, which are nothing related to what I changed

packages/db/src/models/granule.ts
markdboyd: d26488cb7be5f7a1d13c2910275e3aaef1acf4a7

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -164,68 +165,100 @@ export const migrateGranuleAndFilesViaTransaction = async (
...
+      KeyConditionExpression: 'collectionId = :collectionId',
+      ExpressionAttributeValues: {
+        ':collectionId': granuleSearchParams.collectionId,
+      },
+    };
+  }
+
+  const searchQueue = new DynamoDbSearchQueue(
+    {
+      ...defaultSearchParams,
```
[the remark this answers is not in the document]
markdboyd: fdd33c811926fa49db663c81f5feae1612135fd3

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -164,68 +165,100 @@ export const migrateGranuleAndFilesViaTransaction = async (
...
 
-  return { granulesSummary, filesSummary };
+  return { granulesResult, filesResult };
 };
 
 export const migrateGranulesAndFiles = async (
   env: NodeJS.ProcessEnv,
-  knex: Knex
-): Promise<GranulesAndFilesMigrationSummary> => {
+  knex: Knex,
```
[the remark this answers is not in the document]
markdboyd: 5a403eb9f42971aba1c58e0908b14d55a39ffc58

lambdas/data-migration2/src/index.ts
markdboyd: I made some initial changes: 4be478fc232b9bb9de014b6d5777dbf78ae86de1

But as I said, for now these would only show up in CW

lambdas/data-migration2/tests/test-index.js
markdboyd: 8dabf1d2f23bb1e0f95dbee0a761a2cf1eb979d6

lambdas/data-migration2/src/granulesAndFiles.ts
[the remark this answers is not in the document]
Jkovarik: 👍🏻

lambdas/data-migration2/src/index.ts
Jkovarik: Right, re: 2179, that's just errors.    *frown* 

Given that, I think adding it to the object provides needed transparency on what the report output is referring to and we can punt actual outputs to a file as out-of-scope/YAGNI here.

lambdas/data-migration2/src/granulesAndFiles.ts
Jkovarik: I think it's worth discussing but that clearly exceeds the scope needed for this ticket/change.

lambdas/data-migration2/src/granulesAndFiles.ts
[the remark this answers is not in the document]
Jkovarik: Looks good, thanks!

lambdas/data-migration2/src/granulesAndFiles.ts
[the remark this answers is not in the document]
Jkovarik: Sorry, I was trying to recall/presuming that 

`   // It's possible that very old records could have this field be undefined ` 

was the reason the value also wasn't required in the postgres schema.   Thanks for the clarification on the other points.

lambdas/data-migration2/tests/test-granulesAndFiles.js
Jkovarik: Fair enough - given the tests here, it's fine not to immediately sort it, but we should make sure we do.  I'll see about surfacing that.

packages/aws-client/tests/test-DynamoDbSearchQueue.js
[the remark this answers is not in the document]
Jkovarik: Sorry, I was really asking why it wasn't, say, 1 or 10 for the above purpose.    It's inconsequential in this context, was just curious.

Jkovarik (approved):
Thanks for addressing my concerns!    Looks good to me.

PR #2179, opened 2021-04-13
[CUMULUS-2453] Store errors for executions and granules and files in a JSON file

botanical:

Addresses [CUMULUS-2453: Investigate data-migration2 performance testing execution migration failures](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2453)

## Changes

* Create `storeErrors` function which stores errors in system bucket
* Modify tf files to account for use of stackName and system bucket
* Packaged and deployed changes and this is an example of the output:
`data-migration2-executions-errors.json`
* Updated `granulesAndFiles.ts` and `executions.ts` to call `storeErrors`
```
{ "errors": [
"Could not create execution record in RDS for Dynamo execution arn arn:aws:states:us-east-1:[redacted]:execution:jtran-int-tf-DiscoverAndQueuePdrsExecutionPrefix:40fcf686-5e77-453f-bf7c-a4b0b4becc77:, Cause: RecordDoesNotExist: Record in collections with identifiers {\"name\":\"MOD09GQ_test-jtran-int-tf-IngestFromPdr-1614183240869\",\"version\":\"006\"} does not exist.",
"Could not create execution record in RDS for Dynamo execution arn arn:aws:states:us-east-1:[redacted]:execution:jtran-int-tf-DiscoverGranules:082dede3-5633-4225-b044-188de6314b20:, Cause: RecordDoesNotExist: Record in collections with identifiers {\"name\":\"collection-name-8daff3ef4a\",\"version\":\"collection-version-f074cc489b\"} does not exist."
]}
```
`data-migraiton2-granulesAndFiles-errors.json`
```
{ "errors": [
"Could not create granule record and file records in RDS for DynamoDB Granule granuleId: L2_HR_PIXC_product_0001-of-4154_test-jtran-int-tf-KinesisTestTrigger-1611879735452 with files [{\"bucket\":\"not-exist-bucket\",\"fileName\":\"L2_HR_PIXC_product_0001-of-4154.h5\",\"size\":50,\"checksum\":\"cd043418380acdf69b2fb80649357cda\",\"checksumType\":\"md5\",\"source\":\"s3://cumulus-test-sandbox-internal/somepath/L2_HR_PIXC_product_0001-of-4154.h5\",\"type\":\"data\"}], Cause: RecordDoesNotExist: Record in collections with identifiers {\"name\":\"L2_HR_PIXC_test-jtran-int-tf-KinesisTestTrigger-1611879735452\",\"version\":\"000\"} does not exist.",
"Could not create granule record and file records in RDS for DynamoDB Granule granuleId: L2_HR_PIXC_product_0001-of-4154_test-jtran-int-tf-KinesisTestTrigger-1611628271474 with files [{\"bucket\":\"not-exist-bucket\",\"fileName\":\"L2_HR_PIXC_product_0001-of-4154.h5\",\"size\":50,\"checksum\":\"cd043418380acdf69b2fb80649357cda\",\"checksumType\":\"md5\",\"source\":\"s3://cumulus-test-sandbox-internal/somepath/L2_HR_PIXC_product_0001-of-4154.h5\",\"type\":\"data\"}], Cause: RecordDoesNotExist: Record in collections with identifiers {\"name\":\"L2_HR_PIXC_test-jtran-int-tf-KinesisTestTrigger-1611628271474\",\"version\":\"000\"} does not exist."
]}
```

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [ ] Integration tests


lambdas/data-migration2/src/storeErrors.ts
```diff
@@ -0,0 +1,25 @@
+import { s3PutObject } from '@cumulus/aws-client/S3';
+
+/**
+ * Store migration errors in JSON file in S3
+ * @param {string} bucket
+ * @param {string[]} message
+ * @param {string} recordClassification
```
botanical, line 7: Definitely open to better names because I am not sure this is it

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -139,16 +140,22 @@ export const migrateFileRecord = async (
...
  * @returns {Promise<MigrationSummary>} - Migration summary for granules and files
  */
 export const migrateGranuleAndFilesViaTransaction = async (
   dynamoRecord: AWS.DynamoDB.DocumentClient.AttributeMap,
   granuleAndFileMigrationSummary: GranulesAndFilesMigrationSummary,
   knex: Knex,
-  loggingInterval: number
+  loggingInterval: number,
+  stackName?: string,
+  bucket?: string
```
botanical, line 153: I recognize that this function's parameter list is getting longer and I am wondering if I've now added too many parameters to it. Suggestions or comments welcome

lambdas/data-migration2/src/storeErrors.ts
Jkovarik: I spent longer than I should trying to improve on that, and came up empty.   This seems fine, particularly in this case as it really doesn't present a readability concern.

lambdas/data-migration2/src/storeErrors.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @param {string | undefined} stackName
+ * @returns {void}
+ */
+export const storeErrors = async (
+  bucket: string,
+  message: string[],
+  recordClassification: string,
+  stackName?: string) => {
+  const file = `{
```
Jkovarik, line 17: Consider using `putJsonS3Object` from the S3 client.

lambdas/data-migration2/main.tf
```diff
@@ -89,10 +89,12 @@ resource "aws_lambda_function" "data_migration2" {
...
       databaseCredentialSecretArn = var.rds_user_access_secret_arn
-      ExecutionsTable = var.dynamo_tables.executions.name
-      GranulesTable = var.dynamo_tables.granules.name
-      PdrsTable = var.dynamo_tables.pdrs.name
-      dbHeartBeat = var.rds_connection_heartbeat
+      ExecutionsTable             = var.dynamo_tables.executions.name
+      GranulesTable               = var.dynamo_tables.granules.name
+      PdrsTable                   = var.dynamo_tables.pdrs.name
+      dbHeartBeat                 = var.rds_connection_heartbeat
+      system_bucket               = var.system_bucket
```
Jkovarik, line 96: I'm wondering if we should use a more generic configuration variable name here as the output doesn't *have* to go to the system bucket, so much as *a* bucket.

lambdas/data-migration2/src/executions.ts
```diff
@@ -79,6 +82,8 @@ export const migrateExecutions = async (
     failed: 0,
     skipped: 0,
   };
+  const errorFile = [];
```
Jkovarik, line 85: Nit: errorMessages or errors?    It's not really a 'file' or filestream object, so much as an array of errorMessages.

lambdas/data-migration2/src/granulesAndFiles.ts
Jkovarik: Generally when they get > 3 or so arguments, it's worth considering making it a params object.   In this case, *particularly* as you have stackName/Bucket as optional values.

Jkovarik: So something like 

```
const someFunc = async ( params: { 
  foo: string, 
  bar: string 
}) => { 
  const { foo, bar } = params;
......
```

lambdas/data-migration2/src/executions.ts
```diff
@@ -97,13 +102,15 @@ export const migrateExecutions = async (
         migrationSummary.skipped += 1;
       } else {
         migrationSummary.failed += 1;
-        logger.error(
-          `Could not create execution record in RDS for Dynamo execution arn ${record.arn}:`,
-          error
-        );
+        errorMessage = `Could not create execution record in RDS for Dynamo execution arn ${record.arn}:`;
+        errorFile.push(`Error: ${error} ${errorMessage}`);
```
Jkovarik, line 106: Could this approach exceed the lambda/container's memory capacity for a reasonably large numbers of skips/errors (?) and if so I'm wondering if we should consider using a write stream or another defensive approach?

Jkovarik (changes requested):
This looks really good!      I wanted to deploy/run and ran into some stack issues, so commenting with a few items for you to look at, will pick it back up in the AM re: deployment/testing.

lambdas/data-migration2/src/executions.ts
```diff
@@ -97,13 +102,15 @@ export const migrateExecutions = async (
...
+        logger.error(errorMessage, error);
       }
     }
 
+    if (bucket) {
```
markdboyd, line 111: Is there a reason for this logic to be optional?

lambdas/data-migration2/src/executions.ts
```diff
@@ -67,6 +68,8 @@ export const migrateExecutions = async (
   knex: Knex
 ): Promise<MigrationSummary> => {
   const executionsTable = envUtils.getRequiredEnvVar('ExecutionsTable', env);
+  const bucket = process.env.system_bucket;
+  const stackName = process.env.stackName;
```
markdboyd, line 72: I'm inclined to use `envUtils.getRequiredEnvVar` to fetch both of these values. They should always be set, since we're the ones setting them on the lambda deployment config, so treating them as required should be safe

lambdas/data-migration2/src/executions.ts
```diff
@@ -79,6 +82,8 @@ export const migrateExecutions = async (
     failed: 0,
     skipped: 0,
   };
+  const errorFile = [];
+  let errorMessage;
```
markdboyd, line 86: Why are we initializing the errorMessage variable here? The variable seems to be only used inside the else block below, so I would prefer to use `const` inside that block

lambdas/data-migration2/src/executions.ts
markdboyd: +1. I think using a write stream would be safer.

lambdas/data-migration2/src/storeErrors.ts
markdboyd: I would ask for descriptions of each param in the JSdoc though

lambdas/data-migration2/tests/test-storeErrors.js
```diff
@@ -0,0 +1,38 @@
...
+  const recordClassification = 'classification';
+  await storeErrors(process.env.system_bucket, 'message', recordClassification, process.env.stackName);
+  const filename = `data-migration2-${recordClassification}-errors.json`;
+  const key = `${process.env.stackName}/${filename}`;
+
+  const item = await s3().getObject({
+    Bucket: process.env.system_bucket,
+    Key: key,
+  }).promise();
+  const messageBody = JSON.parse(item.Body);
```
markdboyd, line 35: Use the `getJsonS3Object` helper from the aws-client package

lambdas/data-migration2/src/executions.ts
markdboyd: More fundamentally, because of the value of this error logging for users, I don't think it should be optional

lambdas/data-migration2/tests/test-executions.js
```diff
@@ -514,3 +515,38 @@ test.serial('migrateExecutions logs summary of migration for a specified logging
   await migrateExecutions(process.env, t.context.knex);
   t.true(logSpy.calledWith('Batch of 1 execution records processed, 1 total'));
 });
+
+test.serial('migrateExecutions calls storeError when system bucket is provided', async (t) => {
```
markdboyd, line 519:
```suggestion
test.serial('migrateExecutions writes errors to S3 object', async (t) => {
```

lambdas/data-migration2/tests/test-granulesAndFiles.js
```diff
@@ -590,6 +591,36 @@ test.serial('migrateGranulesAndFiles processes all non-failing granule records a
   t.is(fileRecords.length, 1);
 });
 
+test.serial('migrateGranulesAndFiles calls storeError when system bucket is provided', async (t) => {
```
markdboyd, line 594:
```suggestion
test.serial('migrateGranulesAndFiles writes errors to an S3 object', async (t) => {
```

lambdas/data-migration2/tests/test-executions.js
```diff
@@ -514,3 +515,38 @@ test.serial('migrateExecutions logs summary of migration for a specified logging
...
+  ]));
+
+  await migrateExecutions(process.env, t.context.knex);
+
+  // Check that error file exists in S3
+  const item = await s3().getObject({
+    Bucket: process.env.system_bucket,
+    Key: key,
+  }).promise();
+  const messageBody = JSON.parse(item.Body);
```
markdboyd, line 549: Use `getJsonS3Object` helper

lambdas/data-migration2/src/storeErrors.ts
markdboyd: Also, `migrationName`?

lambdas/data-migration2/tests/test-executions.js
```diff
@@ -514,3 +515,38 @@ test.serial('migrateExecutions logs summary of migration for a specified logging
...
+  const parentExecution = fakeExecutionFactoryV2({
+    parentArn: undefined,
+    asyncOperationId: cryptoRandomString({ length: 10 }),
+  });
+  const childExecution = fakeExecutionFactoryV2({ parentArn: parentExecution.arn });
+
+  await Promise.all([
+    // Have to use Dynamo client directly because creating
+    // via model won't allow creation of an invalid record
+    dynamodbDocClient().put({
```
markdboyd, line 531: I was curious and I took a quick look to find out if using `dynamodbDocClient().put` was really necessary. It is not. You can just use `executionsModel.create`. Furthermore, if the purpose of this test is just to ensure that an error is written for one failure, then we don't need the parent/child setup. Just creating an execution that tries to reference a non-existent async operation would be sufficient

lambdas/data-migration2/tests/test-executions.js
```diff
@@ -514,3 +515,38 @@ test.serial('migrateExecutions logs summary of migration for a specified logging
...
+  await migrateExecutions(process.env, t.context.knex);
+
+  // Check that error file exists in S3
+  const item = await s3().getObject({
+    Bucket: process.env.system_bucket,
+    Key: key,
+  }).promise();
+  const messageBody = JSON.parse(item.Body);
+
+  t.truthy(messageBody.errors[0]);
```
markdboyd, line 551: A truthy evaluation isn't telling us how many errors were written, which seems like a more relevant confirmation that things worked as expected? So maybe we should test the length of the written errors

lambdas/data-migration2/src/storeErrors.ts
```diff
@@ -0,0 +1,27 @@
...
+    "errors": ${JSON.stringify(message)}
+  }`;
+  const filename = `data-migration2-${recordClassification}-errors.json`;
```
markdboyd, line 20: I think that the filename should be timestamped, so that each run of the migration has a separate error log that you can refer to.

lambdas/data-migration2/src/storeErrors.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @param {string[]} message
+ * @param {string} recordClassification
+ * @param {string | undefined} stackName
+ * @returns {void}
+ */
+export const storeErrors = async (
+  bucket: string,
+  message: string[],
+  recordClassification: string,
+  stackName?: string) => {
```
markdboyd, line 16: Why is `stackName` optional?

lambdas/data-migration2/src/storeErrors.ts
```diff
@@ -0,0 +1,27 @@
...
+  const file = `{
+    "errors": ${JSON.stringify(message)}
+  }`;
+  const filename = `data-migration2-${recordClassification}-errors.json`;
+  const key = `${stackName}/${filename}` || filename;
```
markdboyd, line 21: What is the logic here doing? Effectively, even if `stackName` was undefined, you'd be doing:

```
`/${filename}` || filename
```

Whether filename has a value or not, nothing after the `||` is ever reached, so I think it's unnnecessary

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -172,10 +179,13 @@ export const migrateGranuleAndFilesViaTransaction = async (
...
-      logger.error(
-        `Could not create granule record and file records in RDS for DynamoDB Granule granuleId: ${dynamoRecord.granuleId} with files ${dynamoRecord.files}`,
-        error
-      );
+      errorMessage = `Could not create granule record and file records in RDS for DynamoDB Granule granuleId: ${dynamoRecord.granuleId} with files ${JSON.stringify(dynamoRecord.files)}`;
+      errorFile.push(`Error: ${error} ${errorMessage}`);
+      logger.error(errorMessage, error);
+    }
+
+    if (bucket) {
```
markdboyd, line 187: Out of curiosity: is there any way once this file is written that the user will easily be able to find groups of "related" errors? Such as "X number of granules failed bc of this missing collection", etc. 

Is it easy enough to somehow parse that information from the generated report? 

I think it might be possible to generate a report that "classifies" the errors by cause by throwing/catching specific error types, but not sure if that's overkill

lambdas/data-migration2/README.md
```diff
@@ -43,4 +43,6 @@ Migration summary:
         $ 0 records failed
 ```
 
-Records skipped indicates if a record was already migrated (which the lambda will throw with a `RecordAlreadyMigrated` error.)
\ No newline at end of file
+Records skipped indicates if a record was already migrated (which the lambda will throw with a `RecordAlreadyMigrated` error.)
+
+For `executions`, `granules`, and `files`, all data migration error messages will be written to JSON files `data-migration2-execution-errors.json` and `data-migration2-granulesAndFiles-errors.json` and uploaded to S3.
```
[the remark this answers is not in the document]
botanical: Ah, that's a good point! Will update

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -135,20 +136,35 @@ export const migrateFileRecord = async (
...
+    granuleAndFileMigrationSummary,
+    knex,
+    loggingInterval,
+    bucket,
+    stackName,
+  } = params;
   const files = dynamoRecord.files;
   const { granulesSummary, filesSummary } = granuleAndFileMigrationSummary;
+  const errorFile = [];
+  let errorMessage;
```
Jkovarik, line 167: This doesn't appear to be used anywhere but in the catch below, consider moving it to around L182 where it's used rather than defining it here.

lambdas/data-migration2/src/executions.ts
```diff
@@ -67,6 +68,8 @@ export const migrateExecutions = async (
   knex: Knex
 ): Promise<MigrationSummary> => {
   const executionsTable = envUtils.getRequiredEnvVar('ExecutionsTable', env);
+  const bucket = envUtils.getRequiredEnvVar('system_bucket', env);
+  const stackName = envUtils.getRequiredEnvVar('stackName', env);
```
Jkovarik, line 72: Side note/:ant: For the purposes of this PR this is fine given the nature of this code.  

I've been trying to avoid creating situations where we're implicitly relying on env variables deeper in the code for switchable configuration where we don't have to, as those are essentially >globals<.

We might consider putting this block in the main lambda function and then include these as a parameter/part of a params object in `migrateExecutions`, `migrateGranulesAndFiles`, and `migratePdrs` for example.  

Again, not *this* PR, but something to consider for maintenance reasons in the future/a future arch discussion.

Jkovarik (changes requested):
Just a couple of additional comments, nothing significant over the prior review.     Let me know if you have questions!

lambdas/data-migration2/src/executions.ts
botanical: Not any good reason

lambdas/data-migration2/src/executions.ts
botanical: Agreed

botanical:
@Jkovarik , @markdboyd I think I addressed most of the feedback. Doing adhoc testing has revealed that the generated JSON file for the granulesAndFiles errors are missing a `]}` at the end of the file so it cannot be parsed to JSON. I am not sure why it isn't being written because I definitely write it https://github.com/nasa/cumulus/pull/2179/files#diff-91b3c5306441fb2380946f4cabdfa533f0143379d13bebd235dc5cd9af8d8586R304 so another pair of eyes on that would be much appreciated.

lambdas/data-migration2/README.md
```diff
@@ -43,4 +43,6 @@ Migration summary:
         $ 0 records failed
 ```
 
-Records skipped indicates if a record was already migrated (which the lambda will throw with a `RecordAlreadyMigrated` error.)
\ No newline at end of file
+Records skipped indicates if a record was already migrated (which the lambda will throw with a `RecordAlreadyMigrated` error.)
```
markdboyd, line 46: FWIW the above report summary structure is no longer accurate based on your previous refactors. To be honest I'm not sure if it's necessary to include a preview of the lambda output in the README

markdboyd: In fact a lot of the README above is now inaccurate based on how this Lambda gets run. We don't have to clean it up in this PR necessarily, but need to make sure that we don't forget

botanical: That's a good point. As for the lambda output, why do you think it's unnecessary to not include a preview of the output? I don't have strong views either way.

markdboyd: Just because it seems like documentation that gets easily out of date and I'm not sure that it's the job of a README to give a preview of Lambda output. I see the README as describing mostly what something is and how to use it, if that is even applicable

lambdas/data-migration2/src/storeErrors.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @param {string[]} params.filename - Write Stream filename
+ * @param {string} params.recordClassification - Classification of record
+ * @param {string} params.stackName - User stack name/prefix
+ * @param {string | undefined} params.timestamp - Timestamp for unit testing
+ * @returns {void}
+ */
+export const storeErrors = async (params: {
+  bucket: string,
+  filename: string,
+  recordClassification: string,
```
markdboyd, line 19: Nit: I'm gonna recommend `migrationName` as this variable name vs `recordClassification`, since I think that's a clearer explanation of what it is

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -266,14 +284,25 @@ export const migrateGranulesAndFiles = async (
...
     migrationResult.filesResult = result.filesResult;
 
     await searchQueue.shift();
     record = await searchQueue.peek();
+
+    if (record) {
+      errorFileWriteStream.write(',\n');
+    }
   }
+  errorFileWriteStream.write('\n]}');
```
markdboyd, line 304:
```suggestion
  errorFileWriteStream.end('\n]}');
```

Based on https://nodejs.org/api/stream.html#stream_writable_end_chunk_encoding_callback

lambdas/data-migration2/README.md
```diff
@@ -43,4 +43,6 @@ Migration summary:
...
+
+For `executions`, `granules`, and `files`, all data migration error messages will be written to JSON files `data-migration2-execution-errors.json` and `data-migration2-granulesAndFiles-errors.json` and uploaded to S3 to your configured system bucket at `${prefix}/data-migration-${type}-errors.json`.
```
markdboyd, line 48: These filenames would have timestamps in them as well

lambdas/data-migration2/src/storeErrors.ts
```diff
@@ -0,0 +1,34 @@
+import moment from 'moment';
```
markdboyd, line 1: We're using moment but I don't see that we've added it as a dependency to the package.json for the Lambda?

lambdas/data-migration2/package.json
```diff
@@ -11,7 +11,7 @@
     "clean": "rm -rf dist",
     "build-lambda-zip": "cd dist/webpack && node ../../../../bin/zip.js lambda.zip index.js",
     "package": "npm run clean && npm run prepare && npm run webpack && npm run build-lambda-zip",
-    "test": "../../node_modules/.bin/ava",
+    "test": "../../node_modules/.bin/ava --serial",
```
markdboyd, line 14: I would prefer not to set this here in the package config and instead just add `.serial` to every test where we need it

lambdas/data-migration2/src/storeErrors.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Store migration errors JSON file on S3.
+ *
+ * @param {Object} params
+ * @param {string} params.bucket - Name of S3 bucket where file will be uploaded
+ * @param {string[]} params.filepath - Write Stream file path
+ * @param {string} params.migrationName - Name of migration
+ * @param {string} params.stackName - User stack name/prefix
+ * @param {string | undefined} params.timestamp - Timestamp for unit testing
+ * @returns {void}
```
markdboyd, line 15:
```suggestion
 * @returns {Promise<void>}
```

lambdas/data-migration2/src/storeErrors.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const { s3 } = require('@cumulus/aws-client/services');
+const fs = require('fs');
+
+/**
+ * Store migration errors JSON file on S3.
+ *
+ * @param {Object} params
+ * @param {string} params.bucket - Name of S3 bucket where file will be uploaded
+ * @param {string[]} params.filepath - Write Stream file path
```
markdboyd, line 11:
```suggestion
 * @param {string} params.filepath - Write Stream file path
```

lambdas/data-migration2/README.md
markdboyd:
```suggestion
For `executions`, `granules`, and `files`, all data migration error messages will be written to JSON files `data-migration2-execution-errors.json` and `data-migration2-granulesAndFiles-errors.json` and uploaded to S3 to your configured system bucket at `${prefix}/data-migration-${type}-errors-${timestamp}.json`.
```

lambdas/data-migration2/src/storeErrors.ts
```diff
@@ -0,0 +1,34 @@
...
+  bucket: string,
+  filepath: string,
+  migrationName: string,
+  stackName: string,
+  timestamp?: string,
+}) => {
+  const { bucket, filepath, migrationName, stackName, timestamp } = params;
+  const fileKey = `data-migration2-${migrationName}-errors`;
+  const dateString = timestamp || moment.utc().format('YYYY-MM-DD_HH:MM:SS');
+  const key = `${stackName}/${fileKey}_${dateString}.json`;
```
markdboyd, line 27: Nit: let's be consistent in our use of dashes vs underscores

```suggestion
  const key = `${stackName}/${fileKey}-${dateString}.json`;
```

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -230,6 +245,9 @@ export const migrateGranulesAndFiles = async (
     granulesResult: granuleMigrationResult,
     filesResult: fileMigrationResult,
   };
+  const filepath = 'granulesAndFilesMigrationErrorLog.json';
+  const errorFileWriteStream = fs.createWriteStream(filepath);
+  errorFileWriteStream.write('{ "errors": [\n');
```
markdboyd, line 250: I'm on the fence, but I might abstract this logic into a `createErrorFileWriteStream` helper in the `storeErrors.ts` file. Just make it accept `migrationName` as an argument.

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -266,14 +284,25 @@ export const migrateGranulesAndFiles = async (
...
     migrationResult.filesResult = result.filesResult;
 
     await searchQueue.shift();
     record = await searchQueue.peek();
+
+    if (record) {
+      errorFileWriteStream.write(',\n');
+    }
   }
+  errorFileWriteStream.end('\n]}');
```
markdboyd, line 304: Likewise, this could be a `closeErrorFileWriteStream` helper

CHANGELOG.md
```diff
@@ -152,10 +152,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     - Added `postgres-migration-async-operation` lambda to start an ECS task to run a the `data-migration2` lambda.
     - Updated `async_operations` table to include `Data Migration 2` as a new `operation_type`.
     - Updated `cumulus-tf/variables.tf` to include `optional_dynamo_tables` that will be merged with `dynamo_tables`.
+  - **CUMULUS-2453**
+    - Created `storeErrors` function which stores errors in system bucket.
+    - Updated `executions` and `granulesAndFiles` to call `storeErrors` to store migration errors.
```
markdboyd, line 157:
```suggestion
    - Updated `executions` and `granulesAndFiles` data migrations to call `storeErrors` to store migration errors.
```

CHANGELOG.md
```diff
@@ -152,10 +152,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
+    - Added `system_bucket` variable to `data-migration2`.
   - **CUMULUS-2451**
     - Added summary type file `packages/db/src/types/summary.ts` with `MigrationSummary` and `DataMigration1` and `DataMigration2` types.
     - Updated `data-migration1` and `data-migration2` lambdas to return `MigrationSummary` objects.
     - Added logging for every batch of 100 records processed for executions, granules and files, and PDRs.
+    - Removed `RecordAlreadyMigrated` logs in `data-migration1` and `data-migration2`
```
markdboyd, line 163: This line seems to duplicate the one below it?

lambdas/data-migration2/src/executions.ts
```diff
@@ -97,16 +111,21 @@ export const migrateExecutions = async (
         migrationResult.skipped += 1;
       } else {
         migrationResult.failed += 1;
-        logger.error(
-          `Could not create execution record in RDS for Dynamo execution arn ${record.arn}:`,
-          error
-        );
+        const errorMessage = `Could not create execution record in RDS for Dynamo execution arn ${record.arn}:`;
+        errorFileWriteStream.write(JSON.stringify(`Error: ${error} ${errorMessage}`));
```
markdboyd, line 115:
```suggestion
        errorFileWriteStream.write(JSON.stringify(`${errorMessage}, Cause: ${error}`));
```

lambdas/data-migration2/package.json
botanical: Hm, I am willing to remove this but doing so doesn't achieve the same thing.

markdboyd: Fair enough, but if we need it then we have some unexpected interference among our tests and I'd like to understand why

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -230,6 +245,9 @@ export const migrateGranulesAndFiles = async (
     granulesResult: granuleMigrationResult,
     filesResult: fileMigrationResult,
   };
+  const filepath = 'granulesAndFilesMigrationErrorLog.json';
```
markdboyd, line 248: This filename (here and in the executions migration) should probably be timestamped. The race condition you're seeing is probably due to this filename (which is the same throughout all tests) is getting unlinked while it's still being written to by some tests

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: Do you mind expanding on why a helper to end the write stream is useful? It would only be one line, right? I am

CHANGELOG.md
botanical: [`c6c1f32` (#2179)](https://github.com/nasa/cumulus/pull/2179/commits/c6c1f32eff6ceb5882a6dbb716014d89e99b8866#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edL164)

lambdas/data-migration2/README.md
botanical: [`c6c1f32` (#2179)](https://github.com/nasa/cumulus/pull/2179/commits/c6c1f32eff6ceb5882a6dbb716014d89e99b8866#diff-4c92a92285abf37daad11f4a41b7ac4bf43e8649a39a04347600c035825cd2e9R20)

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: I ended up adding a end function to help ensure that the stream closes based on some feedback I got from @Jkovarik [`963d9f7` (#2179)](https://github.com/nasa/cumulus/pull/2179/commits/963d9f7adde0399981f6a5ecdb6b5ee44eca2086)

lambdas/data-migration2/src/storeErrors.ts
botanical: https://github.com/nasa/cumulus/pull/2179/files#diff-56c5b1137921652643573bffebcd7fae07a919d6517811b49ab0f84759ba10beR44

lambdas/data-migration2/src/storeErrors.ts
botanical: https://github.com/nasa/cumulus/pull/2179/files#diff-56c5b1137921652643573bffebcd7fae07a919d6517811b49ab0f84759ba10beR40

lambdas/data-migration2/src/storeErrors.ts
botanical: https://github.com/nasa/cumulus/pull/2179/files#diff-56c5b1137921652643573bffebcd7fae07a919d6517811b49ab0f84759ba10beR56

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: https://github.com/nasa/cumulus/pull/2179/files#diff-56c5b1137921652643573bffebcd7fae07a919d6517811b49ab0f84759ba10beR15

CHANGELOG.md
botanical: https://github.com/nasa/cumulus/pull/2179/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR157

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -175,12 +186,12 @@ export const migrateGranuleAndFilesViaTransaction = async (
...
     } else {
+      const errorMessage = `Could not create granule record and file records in RDS for DynamoDB Granule granuleId: ${dynamoRecord.granuleId} with files ${JSON.stringify(dynamoRecord.files)}`;
       granulesResult.failed += 1;
       filesResult.failed += files.length;
-      logger.error(
-        `Could not create granule record and file records in RDS for DynamoDB Granule granuleId: ${dynamoRecord.granuleId} with files ${dynamoRecord.files}`,
-        error
-      );
+
+      stream.write(JSON.stringify(`Error: ${error} ${errorMessage}`));
```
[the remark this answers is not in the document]
botanical: https://github.com/nasa/cumulus/pull/2179/commits/7ea4144651c1375087efef89938fe560d9b30d29

lambdas/data-migration2/src/executions.ts
botanical: https://github.com/nasa/cumulus/pull/2179/commits/7ea4144651c1375087efef89938fe560d9b30d29

lambdas/data-migration2/src/granulesAndFiles.ts
botanical: The race condition didn't go away with timestamped names but it occurred significantly less. What I ended up implementing was a function to close the write stream and ensure it finishes [`963d9f7` (#2179)](https://github.com/nasa/cumulus/pull/2179/commits/963d9f7adde0399981f6a5ecdb6b5ee44eca2086#diff-56c5b1137921652643573bffebcd7fae07a919d6517811b49ab0f84759ba10beR29)

lambdas/data-migration2/README.md
```diff
@@ -17,30 +17,8 @@ The resources migrated in Data Migration 2 are:
...
-        $ 0 records failed
-```
-
-Records skipped indicates if a record was already migrated (which the lambda will throw with a `RecordAlreadyMigrated` error.)
\ No newline at end of file
+The result will be a migration summary object.
+
+Records skipped indicates if a record was already migrated (which the lambda will throw with a `RecordAlreadyMigrated` error.)
+
+For `executions`, `granules`, and `files`, all data migration error messages will be written to JSON files `data-migration2-execution-errors-${timestamp}.json` and `data-migration2-granulesAndFiles-errors-${timestamp}.json` and uploaded to S3 to your configured system bucket at `${prefix}/data-migration-${type}-errors-${timestamp}.json`.
```
markdboyd, line 24:
```suggestion
For `executions`, `granules`, and `files`, all data migration error messages will be written to JSON files `data-migration2-execution-errors-${timestamp}.json` and `data-migration2-granulesAndFiles-errors-${timestamp}.json` and uploaded to S3 to your configured system bucket at `${prefix}/data-migration2-${type}-errors-${timestamp}.json`.
```

lambdas/data-migration2/src/storeErrors.ts
```diff
@@ -0,0 +1,63 @@
...
+export const storeErrors = async (params: {
+  bucket: string,
+  filepath: string,
+  migrationName: string,
+  stackName: string,
+  timestamp?: string,
+}) => {
+  const { bucket, filepath, migrationName, stackName, timestamp } = params;
+  const fileKey = `data-migration2-${migrationName}-errors`;
+  const dateString = timestamp || moment.utc().format('YYYY-MM-DD_HH:MM:SS');
```
markdboyd, line 55: Can we avoid using moment and introducing the dependency by doing `new Date().toISOString()`?

markdboyd: It also might be nice for this function to log the name of the object it wrote to S3 for traceability

markdboyd (approved):
Had a couple more nits but I'll approve pending their resolution

Jkovarik (changes requested):
This all looks good, but on verification, I'm seeing output like: 

```
 nsidc-jkovarik  :45 ->cat data-migration2-granulesAndFiles-errors-2021-04-27T13\:41\:31.225Z.json  |more
{ "errors": [
,
,
,
,
,
,
"Could not create granule record and file records in RDS for DynamoDB Granule granuleId: MOD09GQ.A0370686.tq5B1P.006.4526292427657 with files 
<snip> 
```

which is resulting in: 

 nsidc-jkovarik  :46 ->cat data-migration2-granulesAndFiles-errors-2021-04-27T13\:41\:31.225Z.json  |jq
parse error: Expected value before ',' at line 2, column 1

I think we need to add a skip for entries without errors/filter the results that are resulting in an empty error value.

botanical:
> This all looks good, but on verification, I'm seeing output like:
> 
> ```
>  nsidc-jkovarik  :45 ->cat data-migration2-granulesAndFiles-errors-2021-04-27T13\:41\:31.225Z.json  |more
> { "errors": [
> ,
> ,
> ,
> ,
> ,
> ,
> "Could not create granule record and file records in RDS for DynamoDB Granule granuleId: MOD09GQ.A0370686.tq5B1P.006.4526292427657 with files 
> <snip> 
> ```
> 
> which is resulting in:
> 
> nsidc-jkovarik :46 ->cat data-migration2-granulesAndFiles-errors-2021-04-27T13:41:31.225Z.json |jq
> parse error: Expected value before ',' at line 2, column 1
> 
> I think we need to add a skip for entries without errors/filter the results that are resulting in an empty error value.

Thanks for the feedback @Jkovarik! That was a necessary callout. I added some tests for skipped records and failed records and changed the implementation to use a package `JSONStream`'s [stringify](https://www.npmjs.com/package/JSONStream#jsonstreamstringifyopen-sep-close ) function which handles the delimiting and stores the errors in an array.

lambdas/data-migration2/src/storeErrors.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @param {string | undefined} timestamp - Timestamp for unit testing
+ * @returns {Object}                     - Object containing error write streams and file path
+ */
+export const createErrorFileWriteStream = (migrationName: string, timestamp?: string) => {
+  const dateString = timestamp || new Date().toISOString();
+  const filepath = `${migrationName}ErrorLog-${dateString}.json`;
+  const errorFileWriteStream = fs.createWriteStream(filepath);
+  const jsonWriteStream = JSONStream.stringify();
+  jsonWriteStream.pipe(errorFileWriteStream);
+  errorFileWriteStream.write('{ "errors": \n');
```
markdboyd, line 24: This is a little confusing to me. We're writing to `errorFileWriteStream` directly, even though it seems like what we want to do is write to the JSONstream, which is piped to the error file already.

I think we could avoid this if we customized the open/close options that we pass when initializing the JSONStream, like so:

```suggestion
  const jsonWriteStream = JSONStream.stringify('{"errors": [\n', '\n,\n', '\n]}\n');
  jsonWriteStream.pipe(errorFileWriteStream);
  errorFileWriteStream.write('{ "errors": \n');
```

If we do this, then `closeErrorFileWriteStream` should also be unnecessary, if I'm understanding the docs for JSONstream correctly, because the whole point of passing a custom "close" string when setting up the JSON stream is letting it handle things for you automatically

botanical: Thanks for the feedback! I am skeptical that I can let `stringify` end the stream, so I am still making an explicit call to end the JSON stream but I removed the closeErrorFileWriteStream function. [`3e09741` (#2179)](https://github.com/nasa/cumulus/pull/2179/commits/3e09741a7febe8ee0442fa9d57d6ff5a9d2a8910)

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -175,12 +188,12 @@ export const migrateGranuleAndFilesViaTransaction = async (
...
     } else {
+      const errorMessage = `Could not create granule record and file records in RDS for DynamoDB Granule granuleId: ${dynamoRecord.granuleId} with files ${JSON.stringify(dynamoRecord.files)}`;
       granulesResult.failed += 1;
       filesResult.failed += files.length;
-      logger.error(
-        `Could not create granule record and file records in RDS for DynamoDB Granule granuleId: ${dynamoRecord.granuleId} with files ${dynamoRecord.files}`,
-        error
-      );
+
+      stream.write(JSON.stringify(`${errorMessage}, Cause: ${error}`));
```
markdboyd, line 195: I think the JSON.stringify here is unnecessary. Part of what the JSONStream is doing for you is this stringification automatically.

With the JSON.stringify in place, when I inspect the local contents of the error file, I see:

```
{
  "errors": [
    "\"Could not create granule record and file records in RDS for DynamoDB Granule granuleId: 9a2c4 with files [{\\\"bucket\\\":\\\"ba67b2bd6a\\\",\\\"fileName\\\":\\\"MOD09GQ.A4369670.7bAGCH.006.0739896140643.hdf\\\",\\\"size\\\":1098034,\\\"checksum\\\":\\\"checkSum01\\\",\\\"checksumType\\\":\\\"md5\\\",\\\"source\\\":\\\"source\\\",\\\"type\\\":\\\"data\\\",\\\"key\\\":\\\"fc75692596\\\"}], Cause: RecordDoesNotExist: Record in collections with identifiers {\\\"name\\\":\\\"6fe57\\\",\\\"version\\\":\\\"0.0.0\\\"} does not exist.\"",
    "\"Could not create granule record and file records in RDS for DynamoDB Granule granuleId: f0095 with files [{\\\"bucket\\\":\\\"ba67b2bd6a\\\",\\\"fileName\\\":\\\"MOD09GQ.A4369670.7bAGCH.006.0739896140643.hdf\\\",\\\"size\\\":1098034,\\\"checksum\\\":\\\"checkSum01\\\",\\\"checksumType\\\":\\\"md5\\\",\\\"source\\\":\\\"source\\\",\\\"type\\\":\\\"data\\\",\\\"key\\\":\\\"851a0de80f\\\"}], Cause: RecordDoesNotExist: Record in collections with identifiers {\\\"name\\\":\\\"dd859\\\",\\\"version\\\":\\\"0.0.0\\\"} does not exist.\""
  ]
}
```

You can see that it is unnecessarily double-escaping the contents of each error message, which is less readable.

With the JSON.stringify removed, I see:

```
{
  "errors": [
    "Could not create granule record and file records in RDS for DynamoDB Granule granuleId: a7c71 with files [{\"bucket\":\"e824412d07\",\"fileName\":\"MOD09GQ.A4369670.7bAGCH.006.0739896140643.hdf\",\"size\":1098034,\"checksum\":\"checkSum01\",\"checksumType\":\"md5\",\"source\":\"source\",\"type\":\"data\",\"key\":\"f7d6ebe20e\"}], Cause: RecordDoesNotExist: Record in collections with identifiers {\"name\":\"a07de\",\"version\":\"0.0.0\"} does not exist.",
    "Could not create granule record and file records in RDS for DynamoDB Granule granuleId: 24935 with files [{\"bucket\":\"e824412d07\",\"fileName\":\"MOD09GQ.A4369670.7bAGCH.006.0739896140643.hdf\",\"size\":1098034,\"checksum\":\"checkSum01\",\"checksumType\":\"md5\",\"source\":\"source\",\"type\":\"data\",\"key\":\"90520d5818\"}], Cause: RecordDoesNotExist: Record in collections with identifiers {\"name\":\"6207d\",\"version\":\"0.0.0\"} does not exist."
  ]
}
```

Which is I think more what we want

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -140,20 +143,30 @@ export const migrateFileRecord = async (
...
-export const migrateGranuleAndFilesViaTransaction = async (
+export const migrateGranuleAndFilesViaTransaction = async (params: {
   dynamoRecord: AWS.DynamoDB.DocumentClient.AttributeMap,
   granuleAndFileMigrationResult: GranulesAndFilesMigrationResult,
   knex: Knex,
-  loggingInterval: number
-): Promise<GranulesAndFilesMigrationResult> => {
-  const files = dynamoRecord.files ?? [];
+  loggingInterval: number,
+  stream: any,
```
markdboyd, line 159: I think we can give this a type of `Writable` from the `stream` package. Also, we should give this variable a more descriptive name, like `errorsLogWriteStream`

lambdas/data-migration2/src/storeErrors.ts
```diff
@@ -0,0 +1,58 @@
...
+ * @returns {Object}                     - Object containing error write streams and file path
+ */
+export const createErrorFileWriteStream = (migrationName: string, timestamp?: string) => {
+  const dateString = timestamp || new Date().toISOString();
+  const filepath = `${migrationName}ErrorLog-${dateString}.json`;
+  const errorFileWriteStream = fs.createWriteStream(filepath);
+  const jsonWriteStream = JSONStream.stringify('{"errors": [\n', '\n,', '\n]}\n');
+  jsonWriteStream.pipe(errorFileWriteStream);
+
+  return { jsonWriteStream, errorFileWriteStream, filepath };
```
markdboyd, line 24: We don't need to include `errorFileWriteStream` in the return here

lambdas/data-migration2/src/storeErrors.ts
```diff
@@ -0,0 +1,58 @@
...
+  const dateString = timestamp || new Date().toISOString();
+  const key = `${stackName}/${fileKey}-${dateString}.json`;
+
+  await s3().putObject({
+    Bucket: bucket,
+    Key: key,
+    Body: fs.createReadStream(filepath),
+  }).promise();
+
+  logger.info(`Stored error log file with key ${key} to bucket ${bucket}.`);
```
markdboyd, line 56:
```suggestion
  logger.info(`Stored error log file on S3 at s3://${bucket}/${key}`);
```

This would make it easier for someone to copy/paste the S3 location into their terminal and use it with a CLI command

lambdas/data-migration2/src/granulesAndFiles.ts
```diff
@@ -266,14 +292,21 @@ export const migrateGranulesAndFiles = async (
...
+      loggingInterval,
+      stream: jsonWriteStream,
+    });
     migrationResult.granulesResult = result.granulesResult;
     migrationResult.filesResult = result.filesResult;
 
     await searchQueue.shift();
     record = await searchQueue.peek();
   }
+  jsonWriteStream.end();
```
markdboyd, line 308: It seems like we might need to bring back the close stream helper. I'm getting intermittent unit test failures locally without it which I think are related. Sorry for asking you to remove that.

We just don't need to pass any arguments to the `stream.end()` call.

lambdas/data-migration2/src/storeErrors.ts
botanical: Ooh that is smart!

lambdas/data-migration2/src/storeErrors.ts
Jkovarik: @markdboyd   Are you suggesting the explicit close/await on the finished stream isn't required?    Apologies if that's obtuse, I'm not sure I'm following though.

Jkovarik (approved):
Thanks for addressing my concerns, this looks good 👍🏻

PR #2188, opened 2021-04-22
CUMULUS-2469: Standup skeleton distribution API

jennyhliu:

Addresses [CUMULUS-2469: Standup skeleton distribution API](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2469)

## Changes

*  Added `tf-modules/cumulus_distribution` module to standup a skeleton distribution api

Testing:
* `example/cumulus-tf` deployment should output distribution_api_redirect_uri and distribution_api_uri which are from the new distribution api
* test the endpoint 'profile' and others in the ticket

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


jennyhliu:
`tf-modules/tea-distribution` are from old `tf-modules/distribution`, nothing really changed there.  `tf-modules/distribution` are the new module.

laurenfrederick (commented):
Can you please change the `outputs.tf` of `cumulus-tf` to point to TEA distribution? Or maybe output both?

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
   - The `@cumulus/api-client.granules.getGranule` function takes a `query` parameter which can be used to
   request additional granule information.
   - Published `@cumulus/api@7.2.1-alpha.0` for dashboard testing
+- **CUMULUS-2469**
+  - Added `tf-modules/cumulus_distribution` module to standup a skeleton distribution api
```
kkelly51, line 17: Yeah I guess this is fine. coffee_distribution would have been better

example/spec/parallel/orca/OrcaBackupAndRecoverySpec.js
```diff
@@ -222,6 +225,7 @@ describe('The S3 Ingest Granules workflow', () => {
         { granuleId },
         'completed'
       );
+      await waitForGranuleRecordsInList(config.stackName, [granuleId]);
```
kkelly51, line 228: Was this part of this ticket or a quick fix from something else?

packages/api/app/distribution-routes.js
```diff
@@ -90,7 +90,38 @@ async function ensureAuthorizedOrRedirect(req, res, next) {
   return next();
 }
 
+// TODO implement each method with CSDAP authentication
```
kkelly51, line 93: I'm a little unsure if this is correct place to put these. This file was originally just figuring out if they're authorized and sending them to EDLogin. Do we want to make a new file for a place for this stuff, or are we going to change the purpose of this file

kkelly51 (changes requested):
Have some questions, Not necessarily changes

example/spec/parallel/orca/OrcaBackupAndRecoverySpec.js
jennyhliu: I had a test failure, and it appears that the record is not in es yet, this is just to make sure the record has been added to es.  It's not part of the ticket, just a fix for race condition.

packages/api/app/distribution-routes.js
jennyhliu: It's currently not used except for local distribution api (not sure who uses it), and we can make local distribution api to work for cognito (since auth part is ignored).

kkelly51 (commented):
Deployed correctly with no errors

packages/api/app/distribution-routes.js
kkelly51: Okay, I guess I'm fine with that if that's going to be the plan. We might decided to reorganize later anyway

kkelly51 (approved):

PR #2190, opened 2021-04-23
Cleanup DB code

markdboyd:


packages/db/src/database.ts
```diff
@@ -6,12 +6,6 @@ import { tableNames } from './tables';
 
 export const isRecordDefined = <T>(record: T) => record !== undefined;
 
-export const doesRecordExist = async<T>(
-  params: Partial<T>,
-  knex: Knex,
-  tableName: string
-): Promise<boolean> => isRecordDefined(await knex<T>(tableName).where(params).first());
-
```
botanical, line 14: Should you remove `getRecordCumulusId` as well?

markdboyd: Yup, good catch!

markdboyd: 46572d3010a007041ed4b38d9098800f68513566

packages/api/src/lib/granule-delete.d.ts
```diff
@@ -0,0 +1,2 @@
+export {};
+//# sourceMappingURL=granule-delete.d.ts.map
```
botanical, line 2: What is this file?

markdboyd: A compiled TS file that I definitely didn't mean to commit. I'll fix

markdboyd: 3136e699e4bd7953ba57aaaab21286b4383ca2f5

packages/db/src/database.ts
```diff
@@ -5,33 +5,3 @@ import { RecordDoesNotExist } from '@cumulus/errors';
 import { tableNames } from './tables';
```
botanical, line 5: I think you can remove these imports now

botanical (approved):
I have one comment about removing the imports in `packages/db/src/database.ts`, but other than that ✅

PR #2209, opened 2021-05-03
[CUMULUS-2472] Implement Cognito Client

npauzenga:
**Summary:** 

Addresses [CUMULUS-2472: Implement Cognito Client]https://bugs.earthdata.nasa.gov/browse/CUMULUS-2472)

## Changes

* Renamed `@cumulus/earthdata-login-client` to `@cumulus/oauth-client`
* Moved existing EDL client to `ouathclient/EarthDataLoginClient`
* Created new `CognitoClient`
* Created parent `OauthClient` class that both EDL and Cognito clients inherit from

NOTE: CI is currently failing. Working on it but this is probably ready for review in the meantime.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [ ] Integration tests


packages/oauth-client/src/CognitoClient.ts
```diff
@@ -0,0 +1,145 @@
...
+
+const httpErrorToCognitoError = (httpError: HTTPError) => {
+  const response = <CognitoErrorResponse>httpError.response;
+
+  if (response.body && response.body.error) {
+    return new CognitoError(
+      response.body.error,
+      response.body.error_description
+    );
+  }
```
npauzenga, line 26: This is a bit different from the `EarthdataLoginError` approach. In that case we're looking for a couple specific `response.body.error` cases and throwing `EarthdataLoginErrors` with our own language.

I think it makes sense here (and really in both cases) to just use the error and error_description that's returned from the auth API rather than conforming. I didn't want to mess with the EDL client error handling though so this would be a difference between the clients.

I'm open to updating the error handling in the EDL client but that may be outside of the scope here. I could also do what we're doing there in the CognitoClient if that seems more appropriate.

packages/oauth-client/src/CognitoClient.ts
```diff
@@ -0,0 +1,145 @@
...
+   * @param {string} authorizationCode - an OAuth2 authorization code
+   * @returns {Promise<Object>} access token information
+   */
+  async getAccessToken(authorizationCode: string): Promise<Object> {
+    try {
+      return await super.getAccessToken(authorizationCode);
+    } catch (error) {
+      throw httpErrorToCognitoError(error);
+    }
+  }
```
npauzenga, line 126: Both clients have a couple functions overriding the parent class. The only reason I'm doing this is to preserve the difference in error handling between the EDL client and the Cognito client. 

Open to discussion here. If we made the EDL client and the Cognito client handle errors in the same way we could throw a generic `OAuthError` or something from the parent class and remove these overrides.

packages/oauth-client/src/EarthdataLoginClient.ts
```diff
@@ -0,0 +1,180 @@
...
+      return new EarthdataLoginError('InvalidToken', 'Invalid token');
+    case 'token_expired':
+      return new EarthdataLoginError('TokenExpired', 'The token has expired');
+    default:
+      return new EarthdataLoginError(
+        'UnexpectedResponse',
+        `Unexpected response: ${httpError.response.body}`
+      );
+  }
+};
```
npauzenga, line 30: This is the existing error handling I mentioned in previous comments. It's moved to a new file but otherwise unchanged.

packages/oauth-client/src/EarthdataLoginClient.ts
```diff
@@ -0,0 +1,180 @@
...
+          'InvalidResponse',
+          'Response from Earthdata Login was not valid JSON'
+        );
+      }
+
+      if (isHttpForbiddenError(error)) {
+        throw httpErrorToEarthdataLoginError(error);
+      }
+
+      throw error;
```
npauzenga, line 135: Again, untouched from the original implementation. I'm tempted to look at this error handling for EDL but I'm a bit concerned about the scope of this work.

packages/oauth-client/src/OAuthClient.ts
```diff
@@ -0,0 +1,228 @@
...
+    const response = await this.requestAccessToken(authorizationCode);
+
+    return {
+      accessToken: response.body.access_token,
+      refreshToken: response.body.refresh_token,
+      username: response.body.endpoint.split('/').pop(),
+      // expires_in value is in seconds
+      expirationTime: Math.floor(Date.now() / 1000) + response.body.expires_in,
+    };
+  }
```
npauzenga, line 130: No error handling on the parent class. These methods should just throw and rely on the inheriting class to handle.

packages/oauth-client/src/OAuthClient.ts
```diff
@@ -0,0 +1,228 @@
...
+    if (!clientPassword) throw new TypeError('clientPassword is required');
+    this.clientPassword = clientPassword;
+
+    if (!loginUrl) throw new TypeError('loginUrl is required');
+    validateUrl(loginUrl);
+    this.loginUrl = loginUrl;
+
+    if (!redirectUri) throw new TypeError('redirectUri is required');
+    validateUrl(redirectUri);
+    this.redirectUri = redirectUri;
```
npauzenga, line 71: We could remove this validation. The child class should validate it's params before calling `super`. Didn't seem to hurt to leave it here but open to removing.

packages/oauth-client/tests/test-EarthdataLoginClient.js
```diff
@@ -1,16 +1,12 @@
-import cryptoRandomString from 'crypto-random-string';
-import nock from 'nock';
-import test from 'ava';
-import { URL, URLSearchParams } from 'url';
+const cryptoRandomString = require('crypto-random-string');
```
npauzenga, line 1: Note that I updated this test from TypeScript to JS. Partially for consistency, partially because it's a pain to have to recompile the file every time.

packages/oauth-client/src/CognitoClient.ts
jennyhliu: Instead of add the override method to the children class, in the super class, we can do:

async getAccessToken(authorizationCode: string): Promise<Object> {
    try {
    const response = await this.requestAccessToken(authorizationCode);
  catch {
    throw httpErrorToAuthError(error);
  }

And add `httpErrorToAuthError` to both super and children classes.

packages/oauth-client/src/EarthdataLoginClient.ts
```diff
@@ -0,0 +1,180 @@
...
+  /**
+   * Given an authorization code, request an access token and associated
+   * information from the EarthdataLogin service. This overrides the
+   * base class for better, EarthdataLogin-specific errors.
+   *
+   * See OAuthClient.getAccessToken(authorizationCode).
+   *
+   * @param {string} authorizationCode - an OAuth2 authorization code
+   * @returns {Promise<Object>} access token information
+   */
```
jennyhliu, line 148: Same comments as fore Cognito client.  This override method can be removed if we override methods to convert the error.

packages/oauth-client/src/EarthdataLoginClient.ts
```diff
@@ -0,0 +1,180 @@
...
+   *
+   * const oAuth2Provider = new EarthdataLoginClient({
+   *   clientId: 'my-client-id',
+   *   clientPassword: 'my-client-password',
+   *   earthdataLoginUrl: 'https://earthdata.login.nasa.gov',
+   *   redirectUri: 'http://my-api.com'
+   * });
+   */
+
+  constructor(
```
jennyhliu, line 63: Same comments as for Cognito client.
1. Use `loginUrl` instead of `earthdataLoginUrl`
2. Use object deconstruction to get the parameters.

packages/oauth-client/src/OAuthClient.ts
jennyhliu: I think we only need to validate in parent class.

packages/oauth-client/src/CognitoClient.ts
jennyhliu: This would do in parent:
  httpErrorToAuthError(error) {
    this.httpErrorToAuthError(error);
  }

jennyhliu: Actually, should be throw **this.**httpErrorToAuthError(error);
`this` will refer to children's method

npauzenga: This is a great idea. Much cleaner.

I moved `httpErrorToAuthError` to a class a method on `OAuthClient`, `CognitoClient`, and `EarthdataLoginClient`. The Cognito and EDL clients no longer have the methods overriding the parent class. 

We could take this further but see what you think of these changes:
https://github.com/nasa/cumulus/pull/2209/commits/95794893c9faf1488569fedd4f3dc702116fd453

packages/oauth-client/README.md
```diff
@@ -0,0 +1,259 @@
...
+```
+## Cognito Usage Example
+
+```js
+const { CognitoClient } = require('@cumulus/oauth-client');
+
+const client = new CognitoClient({
+  clientId: 'my-client-id',
+  clientPassword: 'my-client-password',
+  cognitoLoginUrl: 'https://earthdata.login.nasa.gov',
```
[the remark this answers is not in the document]
npauzenga: Good catch. I'm using the SIT URL here. I'm not sure what the prod URL will end up being but as an example this might be fine?

https://github.com/nasa/cumulus/pull/2209/commits/b9e78a517a7e1328aa133e7cb916023196469f79

packages/oauth-client/src/CognitoClient.ts
```diff
@@ -0,0 +1,145 @@
...
+   *   cognitoLoginUrl: 'https://earthdata.login.nasa.gov',
+   *   redirectUri: 'http://my-api.com'
+   * });
+   */
+
+  constructor(
+    params: {
+      clientId: string,
+      clientPassword: string,
+      cognitoLoginUrl: string,
```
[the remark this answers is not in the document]
npauzenga: Oh sure! I changed `cognitoLoginUrl` to `loginUrl` in a previous commit. Here's the update to call `super(params)`:

https://github.com/nasa/cumulus/pull/2209/commits/588192a5c9bb53f6177dd7c055803b7dfc544cf4

packages/oauth-client/src/CognitoClient.ts
```diff
@@ -0,0 +1,145 @@
...
+      redirectUri: string
+    }
+  ) {
+    if (!params.clientId) throw new TypeError('clientId is required');
```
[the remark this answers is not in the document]
npauzenga: Yup, that's much cleaner. The `EarthdataLoginClient` still includes the original assignments because we have that annoying `earthdataLoginUrl` to preserve the original interface. I did remove the validation though. The CognitoClient is much cleaner:

https://github.com/nasa/cumulus/pull/2209/commits/f23d283114a5f63c54b14a6b644ab76b5889f103

packages/oauth-client/src/EarthdataLoginClient.ts
npauzenga: Yup, good plan. Addressed in previous commits.

packages/oauth-client/src/EarthdataLoginClient.ts
npauzenga: So I was debating this change. It would be cleaner to use `loginUrl` but I was reluctant to change the interface for `EarthdataLoginClient`, which uses `earthdataLoginUrl`. I suppose it isn't that risky if the change is only limited to our code here in the core repo. I can update all instances of the `EarthdataLoginClient`. Is there a possibility that `EarthdataLoginClient` is used outside of our code though? It's in a public package so change from `earthdataLoginUrl` to `loginUrl` would be a breaking change. 

Thoughts? I'm not opposed, just a little concerned about the implications.

packages/oauth-client/package.json
```diff
@@ -16,11 +16,14 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
   "files": [
-    "dist/src/**/*.js",
-    "dist/src/**/*.d.ts"
+    "dist/index.js",
+    "dist/index.d.ts"
```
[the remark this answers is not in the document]
npauzenga: https://github.com/nasa/cumulus/pull/2209/commits/95794893c9faf1488569fedd4f3dc702116fd453

packages/oauth-client/tests/test-CognitoClient.js
```diff
@@ -0,0 +1,373 @@
...
+      pass: cognitoClient.clientPassword,
+    })
+    .reply(responseStatus, responseBody);
+};
+
+test.before(() => {
+  nock.disableNetConnect();
+});
+
+test('The CognitoClient constructor throws a TypeError if clientId is not specified', (t) => {
```
jennyhliu, line 65: test-OAuthClient has these TypeError tests and some other functions , no added value to tests in all three test files

packages/oauth-client/src/CognitoClient.ts
```diff
@@ -0,0 +1,97 @@
...
+type CognitoErrorResponse = Response<{
+  error: string,
+  error_description: string,
+}>;
+
+/**
+ * A client for the Cognito API
+ */
+export class CognitoClient extends OAuthClient {
+  readonly clientId: string;
```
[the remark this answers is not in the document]
npauzenga: Yup! https://github.com/nasa/cumulus/pull/2209/commits/1b28242b3fd655ef11c1bc0eff49e6bf94ec7110

Renaming that one property really cleans things up.

packages/oauth-client/tests/test-CognitoClient.js
npauzenga: Yup, good call. I removed the duplicate constructor tests but left tests for `refreshAccessToken()` and `getAccessToken()` in both the test-CognitoClient and test-EarthdataLoginClient. Those two functions are in the parent class but we handle errors slightly differently and that's what those tests are looking for.

https://github.com/nasa/cumulus/pull/2209/commits/de8c1308412e3003874eef5be6078f16bef6e978

packages/oauth-client/README.md
```diff
@@ -0,0 +1,259 @@
...
+  clientPassword: 'my-client-password',
+  cognitoLoginUrl: 'https://auth.csdap.sit.earthdatacloud.nasa.gov/',
+  redirectUri: 'http://my-api.com'
+});
+```
+
+## API
+
+**Kind**: global class
+
```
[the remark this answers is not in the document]
npauzenga: So I didn't actually realize that these docs were generated by JSdoc 😳. This should be much better:

https://github.com/nasa/cumulus/pull/2209/commits/7e4bf9ad171f4f8d7254c232d2f682a7138d9c9e

packages/oauth-client/README.md
```diff
@@ -0,0 +1,254 @@
...
+The examples below document these common methods as well as methods specific to the child classes, e.g.
+`cognitoClient.getUserInfo(accessToken)`.
+
+## Earthdata Login Usage Example
+
+```js
+const { EarthdataLoginClient } = require('@cumulus/oauth-client');
+
+const client = new EarthdataLogin({
```
jennyhliu, line 37:
```suggestion
const client = new EarthdataLoginClient({
```

packages/oauth-client/README.md
```diff
@@ -0,0 +1,254 @@
...
+
+### oAuthClient.getAccessToken(authorizationCode) ⇒ <code>Promise.&lt;Object&gt;</code>
+Given an authorization code, request an access token and associated
+information from the login service.
+
+Returns an object with the following properties:
+
+- accessToken
+- refreshToken
+- username
```
jennyhliu, line 181: For Cognito, the api doesn't return `username` or `endpoint`(where the username is extracted from)

npauzenga: Huh. Curious that I wasn't seeing failures there. That should blow up if `endpoint` is undefined. I was testing Cognito from the unit tests against the actual API but maybe I had some response handling in there that masked this.

In any case, here's a fix to make `username` undefined if `endpoint` isn't available. I could also remove `username` from the response in this case but, because this is on the parent `OAuthClient` I was thinking it might be best to not do that in case EarthdataLogin ever returns an invalid endpoint. I think it'd be confusing in that case to just not have any hint of `username`. Open to push-back though.

https://github.com/nasa/cumulus/pull/2209/commits/1af43af0ec6b4689d27f1d6bcdc5ab37d4fd83fb

npauzenga:
Closing this PR. It will be re-opened as a new PR to try and fix the Snyk failures.

PR #2210, opened 2021-05-03
execution name fix to avoid dropping message with bad cumulus_meta

nemreid:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests

### DIFF
--- a/packages/api/tests/lambdas/test-write-db-dlq-records-to-s3.js
+++ b/packages/api/tests/lambdas/test-write-db-dlq-records-to-s3.js
@@ -184,6 +184,14 @@ test('unwrapDeadLetterCumulusMessage unwraps an AWS states message within an SQS
   t.deepEqual(unwrapDeadLetterCumulusMessage(testSqsMessage), cumulusMessage);
 });
 
+test('unwrapDeadLetterCumulusMessage returns wrapped message on error', (t) => {
+  const invalidMessage = {
+    eventSource: 'aws:sqs',
+    detail: {},
+  };
+  t.deepEqual(unwrapDeadLetterCumulusMessage(invalidMessage), invalidMessage);
+});
+
 test('unwrapDeadLetterCumulusMessage returns an non-unwrappable message', (t) => {
   const testMessage = {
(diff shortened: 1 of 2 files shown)


botanical (approved):
Looks good to me

nemreid:
@botanical I've updated it based on some findings of actual messages I saw being written to my dead letter archive, to add unwrapping logic. Please re-review when able.

packages/api/tests/lambdas/test-write-db-dlq-records-to-s3.js
```diff
@@ -103,3 +107,89 @@ test.serial('write-db-dlq-records-to-s3 throws error if system bucket is not def
...
+    },
+  };
+  const testSqsMessage = {
+    sourceEvent: 'aws:sqs',
+    body: JSON.stringify(testStatesMessage),
+  };
+  t.deepEqual(unwrapDeadLetterCumulusMessage(testSqsMessage), cumulusMessage);
+});
+
+test('unwrapDeadLetterCumulusMessage returns an non-unwrappable message', (t) => {
```
botanical, line 187: 🐜 
```suggestion
test('unwrapDeadLetterCumulusMessage returns input as-is if it cannot be unwrapped', (t) => {
```
or something like that? 🤔

botanical (commented):
I'm not sure if this case is possible but what if we have:
```
const testStatesMessage = {
    source: 'aws.states',
    detail: {
      output: null,
    },
  };
  ```
  
  Then `unwrappedDeadLetterCumulusMessage` will return:
  `JSON.parse(output || input);
Uncaught SyntaxError: Unexpected token u in JSON at position 0`
🤔 

So it looks like these are the different execution events that can occur:
- started
- succeeded
- failed
- timed out
- aborted

and no input looks like it's just an empty object string `"{}"`

so perhaps the test case above is not possible.

botanical (approved):
Approved but I had one concern in the comment I left

PR #2224, opened 2021-05-12
[CUMULUS-2471] Add /s3CredentialsReadme to distribution API

dopeters:

Addresses [CUMULUS-2471: Add s3CredentialsREADME code to distribution API](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2471)

## Changes

* Moved `displayS3CredentialInstructions` out of `s3-credentials-endpoint` package and into `api` package
* Updated webpack config to use `html-loader` so that we can add the html file to the dependency list
	* This is helpful because then the html file is added to the bundle rather than having to rely on the file existing in the file structure of where the function is being called

Endpoint should work for both tea and new distribution. Updates to the README itself are out of scope of this ticket

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Adhoc testing
- [ ] Integration tests


packages/api/endpoints/s3credentials-readme/index.js
```diff
@@ -0,0 +1,17 @@
+const template = require('lodash/template');
+const { inTestMode } = require('@cumulus/common/test-utils');
+const instructions = inTestMode() ? `<a href="${process.env.DISTRIBUTION_ENDPOINT}s3credentials" target="_blank">${process.env.DISTRIBUTION_ENDPOINT}s3credentials</a>` : require('./instructions/index.html');
```
dopeters, line 3: `ava` was not liking the html import here, so this was my workaround. Tried a lot of other things to get it working. Open to suggestions

kkelly51 (approved):
Good to go 👍

PR #2227, opened 2021-05-14
Implement ESLint require-await

nemreid:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


example/lambdas/s3AccessTest/index.js
```diff
@@ -91,7 +91,7 @@ async function handler(event) {
     'write-object': testWrite,
     'list-objects': testList,
   };
-  return testChoices[testName](s3, params);
+  return await testChoices[testName](s3, params);
```
npauzenga, line 94: Just to make sure I'm understanding, this is redundant, right? We want to always require `await` but in cases like this we don't actually expect it to function differently. Is that right?

nemreid: So long as the thing we are calling returns a Promise, that is correct.

lambdas/postgres-migration-count-tool/test/test-utils.js
```diff
@@ -62,12 +62,12 @@ test('buildCollectionMappings returns the expected mappings', async (t) => {
   ];
 
   const collectionModelMock = {
-    get: async (_knex, collection) => {
+    get: (_knex, collection) => {
       if (collection.name === 'BADCOLLECTION') {
-        throw new Error('Danger Will Robinson');
+        return Promise.reject(new Error('Danger Will Robinson'));
```
botanical, line 67: 😂 had to look up the `Danger Will Robinson` reference

lambdas/postgres-migration-count-tool/test/test-utils.js
```diff
@@ -62,12 +62,12 @@ test('buildCollectionMappings returns the expected mappings', async (t) => {
   ];
 
   const collectionModelMock = {
-    get: async (_knex, collection) => {
+    get: (_knex, collection) => {
```
botanical, line 65: Does this not need to be async if it returns a Promise?

packages/api/lambdas/payload-logger.js
```diff
@@ -28,7 +28,7 @@ async function kinesisEventLogger(event, logger = log) {
  * @returns {void} returns nothing
  */
 async function handler(event) {
-  return kinesisEventLogger(event);
+  return await kinesisEventLogger(event);
```
botanical, line 31: We don't need to `await` this since it's not async, right? And it doesn't return a promise, `kinesisEventLogger` returns an object.

packages/api/tests/lib/test-mmt.js
```diff
@@ -20,30 +20,30 @@ const log = mmt.__get__('log');
  * @param {Array<Object>} list - parsed list of collection results
  * @returns {Object} Fake CMR responses
```
botanical, line 21:
```suggestion
 * @returns {Promise<Object>} Fake CMR responses
```

packages/api/tests/lib/test-mmt.js
```diff
@@ -20,30 +20,30 @@ const log = mmt.__get__('log');
...
     version_id: l.version,
   }));
-  return { feed: { entry } };
+  return Promise.resolve({ feed: { entry } });
 };
 
 /**
  * Fakes a CMR return feed where no items in the list have a collection_id.
  * @param {Array<Object>} list - parsed list of collection results
  * @returns {Object} Fake CMR responses
```
botanical, line 35:
```suggestion
 * @returns {Promise<Object>} Fake CMR responses
```

packages/db/src/models/base.ts
```diff
@@ -125,11 +125,11 @@ class BasePgModel<ItemType, RecordType extends { cumulus_id: number }> {
    * @param {ItemType} item - A record to insert into the DB
    * @returns {Promise<number[]>} List of IDs of the inserted records
```
botanical, line 126: Does this still return a promise if we are awaiting `knexOrTransaction`?

packages/db/src/models/granule.ts
```diff
@@ -26,7 +26,7 @@ export default class GranulePgModel extends BasePgModel<PostgresGranule, Postgre
     knexOrTransaction: Knex | Knex.Transaction,
     params: PostgresGranuleUniqueColumns | { cumulus_id: number }
   ): Promise<number> {
```
botanical, line 28: Also this one, does it still return a promise?

packages/ingest/src/SftpProviderClient.js
```diff
@@ -115,7 +115,7 @@ class SftpProviderClient {
    * @returns {Promise<string>} the path that the file was saved to
```
botanical, line 115: Does this return just a string now?

packages/integration-tests/api/executions.js
```diff
@@ -15,7 +15,7 @@ const executionsApi = require('@cumulus/api-client/executions');
  */
```
botanical, line 15: Should it return an object?

packages/integration-tests/api/granules.js
```diff
@@ -15,12 +15,12 @@ const granulesApi = require('@cumulus/api-client/granules');
  */
```
botanical, line 15: Is this returning a Promise still accurate?

packages/integration-tests/api/rules.js
```diff
@@ -32,7 +32,7 @@ async function postRule(params) {
  */
```
botanical, line 32: And this promise?

tasks/files-to-granules/index.js
```diff
@@ -103,7 +103,7 @@ exports.filesToGranules = filesToGranules;
  *                              See schemas/output.json for detailed output schema
  */
 async function handler(event, context) {
-  return cumulusMessageAdapter.runCumulusTask(
+  return await cumulusMessageAdapter.runCumulusTask(
```
botanical, line 106: the return result will be an object, right?

packages/db/src/models/base.ts
botanical: Nevermind! It does.

packages/api/lambdas/payload-logger.js
nemreid: Yeah.. This one was a bit of a compromise. `async` is not strictly necessary here, but not having it will necessitate changing the [lambda handler from async-style to callback-style](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html) (`function handler(event, context, callback) {`. At least in my testing, I do get the expected return value from `await`ing a non-async function. If you think it should be fully converted to callback style I can do that.

botanical: That compromise seems fine to me. Thanks for explaining!

botanical (approved):
Thanks for addressing my feedback and thanks for your thorough work on this! This looks good to me.

PR #2234, opened 2021-05-18
CUMULUS-2208: Implement ES replication for providers, collections, rules, PDRs

markdboyd:

Addresses [CUMULUS-2208: Implement ES replication](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2208)

## Changes

* Updated collection endpoints to create/update/delete records directly in Elasticsearch
* Updated provider endpoints to create/update/delete records directly in Elasticsearch
* Updated rules endpoints to create/update/delete records directly in Elasticsearch
* Updated PDRs endpoint to delete records directly in Elasticsearch
* Removed trigger for `dbIndexer` Lambda from DynamoDB table streams:
    - providers
    - rules
    - pdrs
    - collections

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


packages/api/endpoints/collections.js
```diff
@@ -216,24 +226,45 @@ async function put(req, res) {
...
           index: process.env.ES_INDEX,
           type: 'collection',
+          refresh: inTestMode(),
         }, { ignore: [404] });
+      });
+    } catch (innerError) {
+      // Delete is idempotent, so there may not be a DynamoDB
+      // record to recreate
+      if (existingCollection) {
+        await collectionsModel.create(existingCollection);
```
npauzenga, line 264: In other PRs we talked about not needing a similar catch on `delete` but I think it's necessary here. We have the `esClient.delete` coming after both model deletes so if _that_ fails, we might have already deleted the Dynamo collection and will need to re-create.

packages/api/endpoints/pdrs.js
```diff
@@ -63,28 +63,46 @@ const isRecordDoesNotExistError = (e) => e.message.includes('RecordDoesNotExist'
...
+        await deleteS3Object(process.env.system_bucket, pdrS3Key);
+        await pdrModel.delete({ pdrName });
+        await esClient.delete({
+          id: pdrName,
+          index: process.env.ES_INDEX,
+          type: 'pdr',
+          refresh: inTestMode(),
+        }, { ignore: [404] });
+      });
+    } catch (innerError) {
```
npauzenga, line 99: Do we need to do anything here with `deleteS3Object`? Or is it understood that if there's an error we will have already deleted the S3 file?

packages/api/endpoints/pdrs.js
```diff
@@ -96,4 +114,7 @@ router.get('/:pdrName', get);
 router.get('/', list);
 router.delete('/:pdrName', del);
 
-module.exports = router;
+module.exports = {
+  del,
+  router,
+};
```
npauzenga, line 120: Not sure why we're exporting `del` separately here. Will likely become clear as I go through the rest 😄

packages/api/app/routes.js
```diff
@@ -52,10 +52,10 @@ router.use('/granules', ensureAuthorized, granules);
 router.use('/granule-csv', ensureAuthorized, granuleCsv);
 
 // provider endpoints
-router.use('/providers', ensureAuthorized, providers);
+router.use('/providers', ensureAuthorized, providers.router);
 
 // pdr endpoints
-router.use('/pdrs', ensureAuthorized, pdrs);
+router.use('/pdrs', ensureAuthorized, pdrs.router);
```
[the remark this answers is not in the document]
npauzenga: Ah I see we've changed the export in this PR 🤔

packages/api/endpoints/providers.js
```diff
@@ -217,4 +255,9 @@ router.delete('/:id', del);
 router.post('/', post);
 router.get('/', list);
 
-module.exports = router;
+module.exports = {
+  del,
+  post,
+  put,
+  router,
+};
```
npauzenga, line 263: Same comment as above. Curious why this is necessary.

packages/api/endpoints/rules.js
```diff
@@ -67,29 +66,36 @@ async function get(req, res) {
...
+
+    try {
+      await knex.transaction(async (trx) => {
+        await rulePgModel.create(trx, postgresRule);
+        record = await ruleModel.create(apiRule);
+        await indexRule(esClient, record, process.env.ES_INDEX);
+      });
+    } catch (innerError) {
+      // Clean up DynamoDB record in case of any failure
+      await ruleModel.delete(apiRule);
```
npauzenga, line 96: Is this right? If `await rulePgModel.create(trx, postgresRule);` fails the transaction will fail and throw. This would cause the Dynamo model to delete even if though we didn't try to create it I think.

npauzenga: This would apply to all of the POST endpoints.

packages/api/endpoints/rules.js
```diff
@@ -123,29 +137,33 @@ async function put({ params: { name }, body }, res) {
...
+
+    try {
+      await knex.transaction(async (trx) => {
+        await rulePgModel.upsert(trx, postgresRule);
+        newRule = await ruleModel.update(oldRule, apiRule, fieldsToDelete);
+        await indexRule(esClient, newRule, process.env.ES_INDEX);
+      });
+    } catch (innerError) {
+      // Revert Dynamo record update if any write fails
+      await ruleModel.create(oldRule);
```
npauzenga, line 163: Same as above. If `await rulePgModel.upsert(trx, postgresRule);` fails I think this would try to recreate an existing rule. This would apply to all `put` endpoints.

packages/api/es/search.js
```diff
@@ -228,6 +228,11 @@ class BaseSearch {
     return resp;
   }
 
+  async exists(id) {
+    const response = await this.get(id);
+    return response.detail !== 'Record not found';
```
npauzenga, line 233: Is there a chance `response` could be undefined and throw an error?

packages/api/lambdas/db-indexer.js
```diff
@@ -125,6 +125,8 @@ function getParentId(type, record) {
  * @returns {Promise<Object>} elasticsearch response
  */
 function performIndex(indexFnName, esClient, data) {
+  // TODO: I don't think `process.env.ES_INDEX` is ever set and can be
+  // removed
```
npauzenga, line 129: Want to do that here or hold off?

packages/api/lib/testUtils.js
```diff
@@ -393,6 +403,121 @@ async function getSqsQueueMessageCounts(queueUrl) {
...
+  await indexPdr(esClient, originalPdr, process.env.ES_INDEX);
+  const originalEsRecord = await esPdrsClient.get(
+    originalPdr.pdrName
+  );
+  return {
+    originalDynamoPdr,
+    originalPgRecord,
+    originalEsRecord,
+  };
+};
```
npauzenga, line 519: Nice. I really like these helpers. Standardizing our test data creation is 👍

packages/message/src/Collections.ts
```diff
@@ -17,6 +17,8 @@ type CollectionInfo = {
   version: string
 };
 
+const getCollectionIdSeparator = () => '___';
```
npauzenga, line 20: Maybe a variable instead of a function?

tf-modules/archive/db_indexer.tf
```diff
@@ -91,28 +79,6 @@ resource "aws_lambda_event_source_mapping" "granules_table_db_indexer" {
   batch_size        = 10
 }
 
-data "aws_dynamodb_table" "pdrs" {
```
npauzenga, line 94: Why were these removed? I'm sure it makes sense, just curious.

npauzenga (changes requested):

packages/api/app/routes.js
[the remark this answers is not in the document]
markdboyd: Yeah, I changed some of the exports to allow testing the endpoint handlers directly, which allows us to inject mocks for testing those endpoint handlers

packages/api/endpoints/collections.js
```diff
@@ -125,27 +125,28 @@ async function post(req, res) {
...
+        dynamoRecord = await collectionsModel.create(
+          omit(collection, 'dataType')
+        );
+        // process.env.ES_INDEX is only used to isolate the index for
+        // each unit test suite
+        await indexCollection(esClient, dynamoRecord, process.env.ES_INDEX);
+      });
+    } catch (innerError) {
+      // Clean up DynamoDB collection record in case of any failure
+      await collectionsModel.delete(collection);
```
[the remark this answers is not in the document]
markdboyd: Yup, exactly correct

packages/api/endpoints/collections.js
markdboyd: Exactly

packages/api/endpoints/pdrs.js
markdboyd: Yeah, I'm not sure what we can do about it. But really this PR is only concerned with the affects of adding the ES write into the transaction and handling any side effects of that. 

But once an object is deleted from s3, you can't recreate it without having the contents of the original file. We could perhaps add something to handle it, but it gets pretty complicated with 4 operations in the transaction.

packages/api/endpoints/pdrs.js
markdboyd: Exporting the endpoint handler directly allows us to do some mock injection in unit testing using `req.testContext` and more efficiently test all the different delete cases

packages/api/endpoints/rules.js
markdboyd: >This would cause the Dynamo model to delete even if though we didn't try to create it I think.

It will certainly try to do a delete even though we didn't succeed in creating it, but since DELETE is idempotent, trying to delete something that doesn't exist shouldn't be an issue. It is perhaps wasteful, but to handle it more elegantly we'd have to inspect the type of error that we catch which adds a lot of complexity.

In general these extra `catch` blocks are kinda unfortunate, but it's a necessary evil with three operations in the transaction

packages/api/endpoints/rules.js
markdboyd: So it's true that if `rulePgModel.upsert` fails, then it is going to run `ruleModel.create(oldRule)` to restore the original Dynamo record. Truthfully, this is unnecessary if it's the PG write that failed, because it never updated the Dynamo record. However, it shouldn't be harmful to issue a request to restore the original record even if it was never updated.

It's important to note that our Dynamo model `create` methods are actually doing a PUT under the hood, so they will either create a record or update an existing one.

Thoughts?

packages/api/lambdas/db-indexer.js
markdboyd: I'd like to hold off for now since it's not required for my work and I want to avoid unintended consequences

tf-modules/archive/db_indexer.tf
markdboyd: Currently, our data makes it from Dynamo to Elasticsearch via these Dynamo table streams. So every time we update a record in Dynamo, it writes that update to the table stream, which triggers the dbIndexer Lambda to write that data to Elasticsearch.

However, now that we are writing to Elasticsearch directly from the endpoint, we don't need to rely on these streams and the dbIndexer Lambda to write our data to Elasticsearch. So I'm removing these mappings for each data type as I update the endpoint logic

packages/api/endpoints/pdrs.js
markdboyd: So I took a quick look at adding something to recreate the S3 file if any of the delete operations fail, but it's the same thing where isolating whether it was the S3 delete that failed or something else would be pretty hard, so we'd have to just do a blanket PUT of the S3 file if any of the delete operations failed. Thoughts?

packages/api/app/routes.js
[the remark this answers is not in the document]
npauzenga: Neat! Makes sense to me.

packages/api/endpoints/pdrs.js
npauzenga: Yeah, hmm recreating the S3 files seems less than desirable. Could `deleteS3Object` be the last operation in the transaction? That way we've definitely removed the PDR from PG, dynamo, and ES before deleting from S3. It's a tradeoff because then, if S3 delete fails we have a file in S3 that doesn't exist anywhere else. I suppose it's a question of which is preferable as a worst-case scenario.

packages/api/endpoints/rules.js
npauzenga: I see what you mean. I think you're right. If line 91 or 92 fails, we'll try to delete a Dynamo Rule that doesn't exist and it'll just return 0 rows. That doesn't seem so bad.

packages/api/endpoints/rules.js
npauzenga: That makes sense. What if `indexRule` fails though? In that case will we throw and overwrite the updated rule with the old one? If yes, would we also want to do the same for PG? Or if `indexRule` fails will it not throw 🤔?

markdboyd: > In that case will we throw and overwrite the updated rule with the old one?

Yes, exactly

> If yes, would we also want to do the same for PG

We don't need to manually overwrite for PG, because the PG write only gets committed to the database if everything in the transaction succeeds. 

> Or if indexRule fails will it not throw

I would expect it to throw if there is any error

There is a test for the `if Elasticsearch write fails` here:

https://github.com/nasa/cumulus/blob/e586d77c0cd38756ccca92c149e03e7228c80d4c/packages/api/tests/endpoints/test-rules.js#L881

And if Dynamo write fails:

https://github.com/nasa/cumulus/blob/e586d77c0cd38756ccca92c149e03e7228c80d4c/packages/api/tests/endpoints/test-rules.js#L753

And if PG write fails:

https://github.com/nasa/cumulus/blob/e586d77c0cd38756ccca92c149e03e7228c80d4c/packages/api/tests/endpoints/test-rules.js#L819

packages/message/src/Collections.ts
markdboyd: 5594949704c311196374f5473a4d50a28363529d

packages/api/endpoints/pdrs.js
markdboyd: How about this?

https://github.com/nasa/cumulus/pull/2234/commits/47973f35030a7f4f3e6c6254c1e6fbbe1a9b6613

packages/api/endpoints/rules.js
npauzenga: Awesome. Thanks for taking the time to step through it. Seems totally sane.

packages/api/endpoints/pdrs.js
npauzenga: Nice. Resolved with : https://github.com/nasa/cumulus/pull/2234#commitcomment-51897225

npauzenga (approved):

PR #2238, opened 2021-05-19
Start elasticsearch's indexFromDatabase asyncOperation with correct parameters.

flamingbear:

Addresses CUMULUS-2520: error passing parameters to startAsyncOperation

## Changes

* passes correct payload and params to startAsyncOperation

## PR Checklist

- [ X ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


packages/api/endpoints/elasticsearch.js
```diff
@@ -236,11 +236,11 @@ async function indexFromDatabase(req, res) {
       },
       esHost: process.env.ES_HOST,
       esRequestConcurrency: process.env.ES_CONCURRENCY,
-      stackName,
-      systemBucket,
-      dynamoTableName: tableName,
-      knexConfig,
     },
+    stackName,
```
markdboyd, line 240: Do we have/can we add some sort of test to make sure that we don't break this again? It's hard because the endpoint just triggers an async operation, which is probably mocked in the endpoint test, so we'd need an integration test to prove that an index from database operation can be started?

flamingbear: Yeah I can do that.

flamingbear: If you've got ideas on how to clean this up after it's started. I'm all ears.  I don't see any stop task or delete index operations I have access too.

example/spec/parallel/testAPI/esIndexFromDatabase.js
```diff
@@ -0,0 +1,36 @@
...
+
+describe('Elasticsearch endpoint', () => {
+  describe('Index From Database ', () => {
+    beforeAll(async () => {
+      testConfig = await loadConfig();
+      prefix = testConfig.stackName;
+      indexName = randomId('estestindex').toLocaleLowerCase();
+    });
+    afterAll(async () => {
+      // delete index? how do we clean up
```
flamingbear, line 19: This is very unsatisfying.

markdboyd (approved):

PR #2242, opened 2021-05-19
Fix npm build by removing build script

botanical:
I noticed an issue with this lambda when running `npm run build` at the root level of Cumulus and am opening this PR to implement the fix. 

## Changes

* Remove `build` script in `postgres-migration-async-operation/package.json`
* Reorder scripts object
* Update packages to current version `9.0.1`

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Adhoc testing
- [ ] Integration tests


lambdas/postgres-migration-async-operation/package.json
```diff
@@ -11,15 +11,14 @@
   "types": "./dist/lambda/index.d.ts",
   "scripts": {
     "clean": "rm -rf dist",
-    "build": "rm -rf dist && mkdir dist && ../../node_modules/.bin/webpack",
```
markdboyd, line 14: So I ran `npm run build` in this directory locally and got

```
RROR in Entry module not found: Error: Can't resolve './dist/lambda/index.js' in '/Users/mboyd/development/cumulus/lambdas/postgres-migration-async-operation'
```

And I think the cause is that the TS code isn't compiled before it starts to do the webpack build. If that is true, then we don't need to remove `build`, we just need to add `npm run tsc` before the webpack command

botanical: Oh, interesting! I didn't realize the error was caused by the TS code not being compiled before the webpack build. Thanks!

markdboyd: I just checked out the code locally and it fixes `npm run build` in this package, but I notice that the build in the API package also fails

```
ERROR in ./lambdas/sqs-message-consumer.js                                                                                              │ERROR in ./lambdas/sqs-message-consumer.js
Module not found: Error: Can't resolve '@cumulus/ingest/sqs' in '/Users/mboyd/development/cumulus/packages/api/lambdas'                 │Module not found: Error: Can't resolve '@cumulus/ingest/sqs' in '/Users/mboyd/development/cumulus/packages/api/lambdas'
 @ ./lambdas/sqs-message-consumer.js 25:4-34                                                                                            │ @ ./lambdas/sqs-message-consumer.js 25:4-34
                                                                                                                                        │npm ERR! code ELIFECYCLE
lerna ERR! npm run build stderr:                                                                                                        │npm ERR! errno 2
npm ERR! code ELIFECYCLE                                                                                                                │npm ERR! @cumulus/api@9.0.1 build: `../../node_modules/.bin/webpack`
npm ERR! errno 2                                                                                                                        │npm ERR! Exit status 2
npm ERR! @cumulus/api@9.0.1 build: `../../node_modules/.bin/webpack`                                                                    │npm ERR!
npm ERR! Exit status 2                                                                                                                  │npm ERR! Failed at the @cumulus/api@9.0.1 build script
```

I think the reason is that there is no `build` script in the `ingest` package, and the API package depends on being able to reference the compiled TS code in the `ingest` package. Should be as simple as adding a `build` script to the `ingest` package to run the TS compilation, but if you would rather not address in this PR I can do it as a follow up

packages/ingest/package.json
```diff
@@ -6,6 +6,7 @@
     "node": ">=12.18.0"
   },
   "scripts": {
+    "build": "rm -rf dist && mkdir dist && npm run prepare && npm run webpack",
```
markdboyd, line 9: There is no `npm run webpack` in this package

example/package.json
```diff
@@ -45,30 +45,30 @@
...
-    "@cumulus/move-granules": "8.1.0",
-    "@cumulus/parse-pdr": "8.1.0",
-    "@cumulus/pdr-status-check": "8.1.0",
-    "@cumulus/post-to-cmr": "8.1.0",
-    "@cumulus/queue-granules": "8.1.0",
-    "@cumulus/queue-pdrs": "8.1.0",
-    "@cumulus/sf-sqs-report": "8.1.0",
-    "@cumulus/sync-granule": "8.1.0",
-    "@cumulus/test-processing": "8.1.0"
+    "@cumulus/async-operations": "9.0.1-alpha.1",
```
markdboyd, line 48: Should just be `9.0.1` since that package version is released now. The alpha package was for pre-release

packages/ingest/package.json
botanical: Hm I wonder how this worked when I tested it. Let me fix this, thanks for pointing this out!

example/package.json
botanical: AH gotcha

packages/ingest/package.json
markdboyd: So I'm not sure how this is working, because there is no `webpack.config.js` in this directory. But you're right that it is working. However, since there is no webpack config for this package, we don't need it. The only compilation needed for this package is the TS compilation. So we should just remove `npm run webpack`

botanical: That is interesting... Okay, I will remove the `npm run webpack` part of it

markdboyd (approved):

PR #2251, opened 2021-05-21
CUMULUS-2208: Write async operations to ES

markdboyd:
BLOCKED on #2233 and #2245 

Addresses [CUMULUS-2208: Implement RDS -> ES replication](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2208)

## Changes

* Update logic creating async operations in `async-operations` package to write record directly to ES
* Update logic in async operations Docker image (packages/api/ecs/async-operation) to update record directly in ES
* Update endpoint DELETE logic to delete record directly from ES
* Update unit tests

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Adhoc testing


packages/api/lib/testUtils.js
```diff
@@ -214,13 +222,15 @@ function fakeExecutionFactory(status = 'completed', type = 'fakeWorkflow') {
 function fakeAsyncOperationFactory(params = {}) {
   const asyncOperation = {
     taskArn: randomId('arn'),
-    id: randomId('id'),
+    id: uuidv4(),
```
botanical, line 225: nice

packages/api/tests/endpoints/test-elasticsearch.js
```diff
@@ -271,6 +271,7 @@ test.serial('Reindex request returns 400 with the expected message when source i
     index: defaultIndexName,
     body: { mappings },
   });
+  t.teardown(() => esClient.indices.delete({ index: defaultIndexName }));
```
botanical, line 274: Doesn't the test essentially do this on line 276?

packages/async-operations/src/async_operations.ts
```diff
@@ -95,6 +104,53 @@ export const startECSTask = async ({
   }).promise();
 };
 
+export const createAsyncOperation = async (
+  params: {
+    createObject: ApiAsyncOperation,
```
botanical, line 109: Curious to hear why you named this `createObject` 🤔

packages/es-client/indexer.js
```diff
@@ -372,6 +411,33 @@ function deletePdr({
...
+ *
+ * @param  {Object} params
+ * @param  {Object} params.esClient - ElasticSearch Connection object
+ * @param  {string} params.id - the async operation ID
+ * @param  {string[]} [params.ignore] - Array of response codes to ignore
+ * @param  {string} params.index - Elasticsearch index alias (default defined in search.js)
+ * @param  {string} params.type - Elasticsearch type (default: asyncOperation)
+ * @returns {Promise} Elasticsearch response
+ */
+function deleteAsyncOperation({
```
botanical, line 425: Is there a reason why we don't need to write a unit test for this function?

botanical (approved):
I pulled this down and tested that an async operation was correctly written to Elasticsearch, DynamoDB, and PostgreSQL and verified that it is deleted in parallel from Elasticsearch, DynamoDB, and PostgreSQL.

packages/api/tests/endpoints/test-elasticsearch.js
markdboyd: Yeah, that's bad duplication. I will remove.

packages/async-operations/src/async_operations.ts
markdboyd: Gives it some delineation between the fact that this the object represents the async operation that we are going to create, but hasn't been created yet until this function returns

packages/es-client/indexer.js
markdboyd: These functions are just shallow calls directly through to `deleteRecord`, but I'll add some unit tests

PR #2262, opened 2021-06-02
add unit test for index from database

markdboyd:


packages/async-operations/src/async_operations.ts
```diff
@@ -130,6 +132,7 @@ export const startAsyncOperation = async (params: {
   stackName: string,
   systemBucket: string,
   useLambdaEnvironmentVariables?: boolean,
+  startEcsTaskFunc?: () => StartEcsTaskReturnType
```
markdboyd, line 135: This is the magic which allows us to now pass in a mock for this function that starts ECS tasks. So for unit tests we can mock just this behavior and test that everything else for starting the async operation is setup properly

packages/api/tests/endpoints/test-elasticsearch.js
```diff
@@ -643,3 +666,21 @@ test.serial('request to /elasticsearch/index-from-database endpoint returns 503
     asyncOperationStartStub.restore();
   }
 });
+
+test.serial('indexFromDatabase request completes successfully', async (t) => {
```
markdboyd, line 670: I confirmed that this test was failing prior to your changes and passing after them

PR #2274, opened 2021-06-09
CUMULUS-2208: Write executions directly to ES

markdboyd:

Addresses [CUMULUS-2208: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2208)

## Changes

* ...

## Todo

- [x] Update execution delete logic to delete from ES once #2283 is merged

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


packages/es-client/tests/test-es-indexer.js
```diff
@@ -353,6 +353,33 @@ test.serial('delete a provider record', async (t) => {
   );
 });
 
+test.serial('delete a execution record', async (t) => {
```
botanical, line 356:
```suggestion
test.serial('delete an execution record', async (t) => {
```
🐜

packages/api/endpoints/executions.js
```diff
@@ -79,10 +81,35 @@ async function del(req, res) {
...
+        index: process.env.ES_INDEX,
+        ignore: [404],
+      });
+    });
+  } catch (error) {
+    // Delete is idempotent, so there may not be a DynamoDB
+    // record to recreate
+    if (apiExecution) {
+      await executionModel.create(apiExecution);
+    }
```
botanical, line 110: Why do we need to recreate the dynamodb record? Wouldn't the execution model throw an error on line 86 if the record doesn't exist?

markdboyd: The need to recreate the DynamoDB record comes from the transactional delete logic, which could happen this way:

- PG delete succeeds
- DynamoDB delete succeeds
- ES delete throws an error

In that case, the PG delete will get rolled back, but you can't roll back a DynamoDB delete. So the only way to "restore" the Dynamo record and thus have all three systems in the original state where the records exist is to manually recreate the DynamoDB record.

botanical (changes requested):
After pulling this down, deploying, and testing, I wasn't able to get any execution records stored in DynamoDB, ES, or PG. 

This message from CW logs might be a good starting point:
```
Memory Size: 256 MB	Max Memory Used: 256 MB
```

CHANGELOG.md
```diff
@@ -39,6 +39,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     - Updated logic for rules API POST/PUT/DELETE to create/update/delete records directly in Elasticsearch in parallel with updates to DynamoDb/PostgreSQL
     - Updated logic for providers API POST/PUT/DELETE to create/update/delete records directly in Elasticsearch in parallel with updates to DynamoDb/PostgreSQL
     - Updated logic for PDRs API DELETE to delete records directly in Elasticsearch in parallel with deletes to DynamoDB/PostgreSQL
+    - `sfEventSqsToDbRecords` Lambda now writes following data directly to Elasticsearch in parallel with writes to DynamoDB/PostgreSQL:
+      - executions
```
botanical, line 43: Should you add a note about the logic update for API DELETE on executions?

packages/es-client/indexer.js
```diff
@@ -372,6 +372,33 @@ function deletePdr({
   });
 }
 
+/**
+ * Deletes the execution in ElasticSearch
```
botanical, line 376:
```suggestion
 * Deletes the execution in Elasticsearch
```
🐜

packages/es-client/indexer.js
```diff
@@ -372,6 +372,33 @@ function deletePdr({
...
+ *
+ * @param  {Object} params
+ * @param  {Object} params.esClient - ElasticSearch Connection object
```
botanical, line 379:
```suggestion
 * @param  {Object} params.esClient - Elasticsearch Connection object
```

tf-modules/archive/sf_event_sqs_to_db_records.tf
```diff
@@ -168,7 +169,7 @@ resource "aws_lambda_function" "sf_event_sqs_to_db_records" {
   handler          = "index.handler"
   runtime          = "nodejs12.x"
   timeout          = local.sf_event_sqs_lambda_timeout
-  memory_size      = 256
+  memory_size      = 512
```
botanical, line 172: 👍

botanical (approved):
After pulling down, deploying, and testing, this all checks out. I had a few 🐜 nit comments but otherwise looks good ✅

PR #2282, opened 2021-06-11
[CUMULUS-2475][CUMULUS-2476] GET and HEAD endpoints

dopeters:
**Summary:** Add GET and HEAD endpoints to distribution API.

Addresses 
- [CUMULUS-2475: Implement GET endpoint](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2475)
- [CUMULUS-2476: Implement HEAD endpoint](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2476)

## Changes

* CUMULUS-2475
Adds GET endpoint
To test: 
go to `<dist-api-url>/bucket/key` in browser
Public files should not require auth
Private files should require auth (will redirect you to Cognito)

* CUMULUS-2476
Adds HEAD Endpoint
To test:
1. Basic Request: 
```sh
curl --location --head '<dist-api-url>/path/to/science-file.hdr'
```
2. With Range Headers
```sh
curl --location --head '<dist-api-url>/path/to/science-file.hdr' \
--header 'Range: bytes=0-1023'
```
3. Basic Request (Authenticated)
```sh
curl --location --head '<dist-api-url>/path/to/science-file.hdr' \
--header 'Cookie: accessToken=<access-token-from-cognito>'
```
4. Range Requests (Authenticated)
```sh
curl --location --head '<dist-api-url>/path/to/science-file.hdr' \
--header 'Range: bytes=0-1023' \
--header 'Cookie: accessToken=<access-token-from-cognito>'
```   

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


tf-modules/cumulus_distribution/iam.tf
```diff
@@ -46,6 +46,17 @@ data "aws_iam_policy_document" "lambda_distribution_api_gateway_policy" {
     resources = [aws_dynamodb_table.access_tokens.arn]
   }
 
+  statement {
+    actions = [
+      "s3:GetObject*",
```
jennyhliu, line 51: It seems we only need `"s3:GetObject*"` action based on my testing.

jennyhliu:
Have you compared the s3signed url generated from our endpoint with the one from tea distribution?  Just to make sure we don't miss anything since Cloud Metrics parses the s3 access log from distribution.

tf-modules/cumulus_distribution/iam.tf
dopeters: Ah yeah, that makes sense. Removed everything else

dopeters:
> Have you compared the s3signed url generated from our endpoint with the one from tea distribution? Just to make sure we don't miss anything since Cloud Metrics parses the s3 access log from distribution.

Yep did some testing and we're passing all the same params. The only major difference I'm seeing is that for an unauthenticated user, we're passing `A-userid: unauthenticated user` whereas TEA is passing `A-userid: None`. Not sure if we should change that to match or if it's ok as is.

kkelly51 (approved):
Tested the s3signedUrl between tea and cumulus dist. Everything looks good

PR #2292, opened 2021-06-16
fix write-granules logging

markdboyd:


Jkovarik (approved):

packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js
```diff
@@ -270,14 +270,14 @@ const _writeGranuleViaTransaction = async ({
   );
   // Ensure that we get a granule ID for the files even if the
   // upsert query returned an empty result
-  const granuleCumulusId = getGranuleCumulusIdFromQueryResultOrLookup({
+  const granuleCumulusId = await getGranuleCumulusIdFromQueryResultOrLookup({
```
markdboyd, line 273: This is particularly important. Otherwise we're not returning an actual value necessarily

PR #2300, opened 2021-06-21
Deploy Cognito instead of TEA

npauzenga:
**Summary:** 

Addresses [CUMULUS-2482: Deployment can deploy either distribution app](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2482)

## Changes

This allows the user to deploy either the new Cumulus Distribution API or TEA. To test:

1. Deploy the branch from `example` as-is. The default distribution is now Cumulus Distribution. It should deploy successfully and you should be able to ingest and publish a granule to CMR to see the correct distribution URL. Contact me for help with this.
2. Switch the variables in `example/cumulus-tf/main.tf` starting on line 167 from Cognito to TEA and deploy again. This should deploy TEA. Run another ingest/publish and note that the distribution URL is different and points to TEA.

Note that there will be additional user-facing changes in `cumulus-template-deploy`. In that repo we will be deploying TEA as the default and Cognito will be optional.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


npauzenga:
This will likely cause integration test failures. I'd like to get a second opinion on the approach here in the meantime but there will likely be additional commits to address failed tests.

example/spec/parallel/distributionAPI/distributionSpec.js
```diff
@@ -24,7 +24,7 @@ const s3Data = [
   '@cumulus/test-data/granules/MOD09GQ.A2016358.h13v04.006.2016360104606.hdf.met',
 ];
 
-describe('Distribution API', () => {
+xdescribe('Distribution API', () => {
```
npauzenga, line 27: We want to skip these tests because they're TEA-specific. Again, we could delete.

tf-modules/cumulus/variables.tf
```diff
@@ -125,11 +125,13 @@ variable "system_bucket" {
 variable "tea_external_api_endpoint" {
   description = "Thin Egress App external endpoint URL"
   type        = string
+  default     = null
```
npauzenga, line 128: We don't use these for Cumulus Distribution. We could comment/delete them instead of making them optional like this.

tf-modules/cumulus/distribution.tf
```diff
@@ -1,31 +1,33 @@
-module "distribution" {
-  source = "../distribution"
+# Uncomment if using TEA for distribution
```
npauzenga, line 1: This file and `tf-modules/cumulus/outputs.tf` might actually need to be uncommented as we're still intending to support TEA as the default distribution in `cumulus-template-deploy`. If this module and it's outputs are removed here I think that would break the template. If we uncomment this and leave it in we'd always be deploying the distribution module even if we don't use it though.

tf-modules/cumulus/outputs.tf
```diff
@@ -1,9 +1,16 @@
 # API outputs
 
-output "distribution_bucket_map" {
-  value = module.distribution.distribution_bucket_map
-}
+# TEA-Specific outputs. Uncomment to use TEA distribution
```
npauzenga, line 3: See above. We probably need these values active.

tf-modules/cumulus/distribution.tf
npauzenga: Ideally I think the `cumulus-template-deploy` would just have this change:

https://github.com/nasa/cumulus-template-deploy/pull/63/files

This would require uncommenting the `distribution` module and the outputs below. We would be deploying both but maybe that's acceptable.

example/cumulus-tf/main.tf
```diff
@@ -163,17 +163,19 @@ module "cumulus" {
...
+  # tea_api_gateway_stage         = local.tea_stage_name
+  # tea_rest_api_id               = module.thin_egress_app.rest_api.id
+  # tea_rest_api_root_resource_id = module.thin_egress_app.rest_api.root_resource_id
+  # tea_internal_api_endpoint     = module.thin_egress_app.internal_api_endpoint
+  # tea_external_api_endpoint     = module.thin_egress_app.api_endpoint
 
   log_destination_arn = var.log_destination_arn
 
+  # Cumulus Distribution settings. Remove/comment if not using Cognito.
+  tea_external_api_endpoint = var.cumulus_distribution_url
```
npauzenga, line 177: Talked with Lauren a bit about the naming here. It would be a breaking change to rename `tea_external_api_endpoint` but that's something we could consider.

tf-modules/cumulus/variables.tf
markdboyd: We can't delete them from the `cumulus` module if we want to support both TEA and cumulus distribution simultaneously, which it seems like we do? Because we're not forcing an upgrade?

example/cumulus-tf/outputs.tf
```diff
@@ -6,17 +6,21 @@ output "archive_api_redirect_uri" {
   value = module.cumulus.archive_api_redirect_uri
 }
 
-output "distribution_url" {
-  value = module.thin_egress_app.api_endpoint
-}
+# TEA-Specific outputs. Uncomment to use TEA distribution
```
markdboyd, line 9: I'm not sure I understand the point of commenting out all of the TEA code for our example deployments versus just deleting it. Whether the code is just commented or deleted, everyone's deployments will get updated to use cumulus distribution by default. And at some point carrying forward a bunch of commented code is just technical debt.

And if someone really needs/wants to use TEA, there is always git history. 

Is there something I'm missing here? Otherwise I think we just should just be deleting all this code vs commenting out

example/cumulus-tf/main.tf
markdboyd: What deploys the cumulus distribution? Where does the value of `cumulus_distribution_url` come from?

I ask because I don't see what in this PR is actually deploying the Cognito auth API?

tf-modules/cumulus/variables.tf
markdboyd: I'm saying that we have to keep these variables and make them optional like this if we want to keep supporting TEA and new Cognito auth

tf-modules/cumulus/distribution.tf
markdboyd: So this file is just including the `distribution` in the `cumulus` module. And the `distribution` module is not deploying TEA, it's just attaching some extra functionality to it, namely the S3 credentials endpoint. So whether the code in here is commented/deleted or not, TEA gets deployed from `example/cumulus/thin_egress_app.tf`.

Still, if people are using TEA, we want to have this functionality enabled.

I would probably go with a different approach here of just making the whole module conditional using a `count`. It looks like `distribution/variables.tf` says that `tea_internal_api_endpoint` is a required variable for that module, so you could make this module conditional on that like so

`count: var.tea_internal_api_endpoint !== null ? 1 : 0`

example/README.md
```diff
@@ -155,6 +155,21 @@ Run `terraform apply`.
 
 This will deploy your data-migration resources.
 
+## Choosing a distribution API
```
markdboyd, line 158: Who are these instructions for specifically? Are these intended to be user facing?

markdboyd: If so, I think we should be writing a general upgrade note document (e.g. https://nasa.github.io/cumulus/docs/upgrade-notes/migrate_tea_standalone) about using new Cumulus distribution and how to remove TEA. I don't think we should generally be putting any user facing document in our `example` folder versus the public docs for discoverability.

markdboyd (changes requested):
I have a few questions about the approach here. Happy to chat offline if that is easier

example/cumulus-tf/main.tf
jennyhliu: I think the value should be the output of the cumulus_distribution module:
`tea_external_api_endpoint = module.cumulus_distribution.api_uri`
cumulus_distribution_url is the input of the example/cumulus-tf/terraform.tfvars, and is optional

tf-modules/cumulus/distribution.tf
jennyhliu: Since we need to support both TEA and cumulus_distribution,  is it possible to deploy both distribution endpoints and point distribution only to cumulus_distribution?  This way, if we need to update `s3credentials` functionalities, we can easily test it out for both distribution endpoints.

example/spec/parallel/distributionAPI/distributionSpec.js
npauzenga: Yeah so this is a gray area because we support both TEA and Cognito but we only deploy Cognito _by default_ with the `example` deployments. So what if a change needs to happen involving the previous distribution API (with TEA)? In that case I would imagine the developer would want to deploy their stack with TEA and run this test suite against it. 

But then, is that likely to happen? Are we likely to have a change to TEA OR is the developer likely to remember this test suite? Maybe not. Maybe we remove these tests because the odds of needing them in the future are low enough. I'm conflicted. Thoughts? Am I thinking about this wrong?

tf-modules/cumulus/variables.tf
npauzenga: Right. I think this is the best way to go. My worse idea above was leaving them as-is (not optional) and adding a note to the Cognito deploy instructions to delete them if you use Cognito. Which seems more confusing.

tf-modules/cumulus/distribution.tf
npauzenga: This is helpful, thanks both for chiming in. 

Mark, to your point it totally makes sense that this doesn't deploy TEA. I also have the same pattern of commenting out everything in `example/cumulus/thin_egress_app.tf` that will also need to change either to a conditional or to always be included.

Jenny, if we left this `distribution` module as it was and always deploy it, would we also need to always deploy `example/cumulus/thin_egress_app.tf`? 

In my mind the options are:

1) Always deploy TEA and Cognito by leaving `tf-modules/cumulus/distribution.tf` and `example/cumulus/thin_egress_app.tf` uncommented. The switch between Cognito and TEA would just be handled in cumulus-tf/main.tf.

2) Conditionally switch both `tf-modules/cumulus/distribution.tf` as Mark was suggesting, and `example/cumulus/thin_egress_app.tf` as well.

example/cumulus-tf/thin_egress_app.tf
```diff
@@ -1,79 +1,81 @@
-locals {
```
jennyhliu, line 1: instead of commenting everything out, can we just rename this file to thin_egress_app.tf.example?

tf-modules/cumulus/distribution.tf
jennyhliu: I like the idea '2. Conditionally switch'

example/cumulus-tf/main.tf
npauzenga: Ah, good catch. Tested this and it gives us the same URL:

https://github.com/nasa/cumulus/pull/2300/commits/2b07efe5b665ff107490cd5a0419e962e15ceb42

example/cumulus-tf/outputs.tf
npauzenga: Yup, totally valid. I had that question in my PR description here because it's a little messy. The only reason I went this way was because I could link to files locally in the README and in the moment you need to switch from one to the other, that's easier. But for all other reasons it might be a worse approach.

That said, this will be pending the outcome of the above discussions about deploying TEA AND Cognito vs. conditionally including one vs. the other. These outputs would likely also need to be conditional if we go that route.

example/README.md
npauzenga: This was only meant for developers deploying from example for CI/testing/etc. There will be user-facing docs that will call out instructions for adapting the template deploy. We'll also note the downtime a switch will cause.

I wanted to get the approach here hammered out first to know what to say there but I'll update that PR as well once we're set.

example/cumulus-tf/thin_egress_app.tf
npauzenga: Oh yeah, that's a great idea! See the thread above re: conditionals vs. deploying both but this seems like a solid third option.

example/cumulus-tf/outputs.tf
npauzenga: Jenny had a good suggestion: https://github.com/nasa/cumulus/pull/2300#discussion_r662480391. The outputs here might still need to be conditional but we could switch the other tea-only files to `xxx.tf.example` instead of commenting out OR deleting.

tf-modules/cumulus/distribution.tf
npauzenga: 3. Per Jenny's suggestion: https://github.com/nasa/cumulus/pull/2300#discussion_r662480391 make these `tf-modules/cumulus/distribution.tf.example` and `example/cumulus/thin_egress_app.tf.example` to not always deploy them but leave them available and obvious.

jennyhliu: We can't do `tf-modules/cumulus/distribution.tf.example` since it's part of the tf-modules, and users won't update it.

example/cumulus-tf/outputs.tf
markdboyd: >e moment you need to switch from one to the other, that's easier.

Do we need to switch from one to the other? Because we're supporting both for some time period? 

If so, then I guess going with the `example` file approach feels better than commented or deleted code. 

But yeah the outputs here would need to be conditional not commented or deleted

example/spec/parallel/distributionAPI/distributionSpec.js
markdboyd: It feels like being in this limbo between Cumulus distribution and TEA is what complicates it. If we really are supporting both, then I guess just disabling this test is a fine approach for now. 

But overall it feels like technical debt we're holding on to "just in case"

tf-modules/cumulus/outputs.tf
markdboyd: Yes they'll need to be conditional

tf-modules/cumulus/distribution.tf
markdboyd: If we are intending to support both TEA and cumulus distribution in the next release, which is what I gather from other comments in this PR, then we can't comment the contents of this file. 

Because deploying the `distribution` module is necessary for Cumulus to work with TEA. It creates the S3 credentials endpoint on TEA and the TEA bucket map cache file.

Thus, I think the only option is to make this module conditional on whether TEA seems to be deployed, which maybe we can infer from the presence of the `var.tea_internal_api_endpoint` variable

The question about what to do with `example/cumulus/thin_egress_app.tf` and relates more to whether we want to deploy both TEA and cumulus distribution for our example deployments

example/spec/parallel/distributionAPI/distributionSpec.js
npauzenga: Yeah, I hear ya. I'd like to ping @laurenfrederick here if she has a second. I know the plan right now is to continue to support (and default to) TEA for the cumulus template deploys. Is there a plan to eventually deprecate TEA and the old distribution API in favor of the new one? If yes then we can add that context as a comment when we disable tests and stuff like this because we'll delete them eventually.

Otherwise, if we're always going to support both it does seem likely that we'll need a way to test TEA. That might mean that we need some better way to deploy and test both in our integration tests. Thoughts?

tf-modules/cumulus/outputs.tf
```diff
@@ -1,9 +1,16 @@
 # API outputs
 
+# TEA-Specific outputs.
+
 output "distribution_bucket_map" {
-  value = module.distribution.distribution_bucket_map
+  value = var.tea_internal_api_endpoint != null ? module.distribution[0].distribution_bucket_map : null
```
npauzenga, line 6: Let me know what you think of this. Outputs on conditionally created modules aren't the most seamless. A couple things to note:

1. These outputs don't disappear if we're not using TEA. They're just `null`.
2. module.distribution is a list now that it has a `count` param
3. I think the only reason this works at all is because the `distribution` module is now a list with either 1 or 0 members. The Terraform compiler would complain if `module.distribution` doesn't exist at all but it's fine if it's an empty list. It's a little weird.

tf-modules/cumulus/distribution.tf
```diff
@@ -1,4 +1,5 @@
 module "distribution" {
+  count = var.tea_internal_api_endpoint != null ? 1 : 0
```
npauzenga, line 2: This makes `distribution` conditional and changes its shape. It's now a list so outputs need to reference `module.distribution[x]`

example/cumulus-tf/outputs.tf
```diff
@@ -6,6 +6,8 @@ output "archive_api_redirect_uri" {
   value = module.cumulus.archive_api_redirect_uri
 }
 
+# TEA-Specific outputs
+
```
npauzenga, line 10: Note that these are not optional. We're always deploying `thin_egress_app` with this configuration. It doesn't get switched off when using cumulus_distribution. We just deploy both.

tf-modules/cumulus/variables.tf
npauzenga: Resolving this. Seems like we're on the same page and there aren't changes since this thread.

tf-modules/cumulus/distribution.tf
npauzenga: The latest changes to this PR (sorry, it's spread over a few commits) make `distribution` conditional and `thin_egress_app` remains unchanged and is always deployed. The switch between the two distribution APIs is controlled by main.tf.

tf-modules/cumulus/outputs.tf
npauzenga: And they are: https://github.com/nasa/cumulus/pull/2300/files#diff-c6fa420df06c6c4749e5fcb023d8b4a4096a0050767a3814d021d039571717ebR3

example/cumulus-tf/outputs.tf
npauzenga: The way this is configured now, the outputs in `example` are left as-is. `thin_egress` app will always deploy with `example` deployments but we can switch the distribution API we're using in main.tf. 

This was sort of the original "can we just deploy both in `example`" thought. I _think_ that seems reasonable versus having `thin_egress_app` conditional along with all of its resources and these outputs returning `null` if it's not deployed. Both are a little messy though so open to pushback.

npauzenga: It's also one less step. I guess that's what pushed me to do it this way. If you want to switch between the two distribution APIs you only need to switch the assignments in main.tf. If we do something like `thin_egress_app.tf.example`, that's another step. Again, though, I'm obviously waffling 😄

tf-modules/cumulus/outputs.tf
markdboyd: 2 and 3 are expected results of making the distribution module conditional. And I think 1 is fine, but need to think about it

example/cumulus-tf/terraform.tfvars.example
```diff
@@ -67,7 +67,7 @@ csdap_client_password = "csdap client password"
 ## url for uat: "https://auth.csdap.uat.earthdatacloud.nasa.gov"
 csdap_host_url = "csdap host url"
 
-# Optional
+# Optional. Uncomment if using Cumulus Distribution.
```
markdboyd, line 70: Where does this variable come from or get used?

example/spec/parallel/distributionAPI/distributionSpec.js
markdboyd: It sounds like we're keeping both for now, so we can keep the tests

CHANGELOG.md
```diff
@@ -79,6 +79,11 @@ the error object:
 `@cumulus/api-client/granules.deleteGranule` to control the retry
 behavior
 
+### Changed
```
markdboyd, line 82: This header is a duplicate. There is already a "Changed" header above

example/cumulus-tf/outputs.tf
markdboyd: Separate comment: can we change these output names? Otherwise it's a bit confusing which output is for TEA vs cumulus distribution

`distribution_api_uri` -> `cumulus_distribution_api_uri`
`distribution_api_redirect_uri` -> `cumulus_distribution_api_redirect_uri`

I might even go all the way and do these all well:

`distribution_url` -> `tea_distribution_url`
`distribution_redirect_uri` -> `tea_distribution_redirect_uri`
`s3_credentials_redirect_uri` -> `tea_s3_credentials_redirect_uri`

Since these outputs are from `example/cumulus-tf`, they're not user facing so it should be safe to change them. It just helps avoid confusion for us

npauzenga: Oh yeah, good point. I guess these really aren't used elsewhere other than the console output on a deploy.

https://github.com/nasa/cumulus/pull/2300/commits/be47b39f61c79ac38d4323f7e778168ece608501

example/cumulus-tf/terraform.tfvars.example
npauzenga: It looks like this is used in `cumulus_distribution` and comes from the output of your deploy. I have it set in my terraform.tfvars like this:

`cumulus_distribution_url = "https://ah5f8qo0v1.execute-api.us-east-1.amazonaws.com:7000/dev/"`

I'm not totally sure why it's set up this way but it does seem weird that the output of the module is used to populate a field on that module...

CHANGELOG.md
npauzenga: 👍 https://github.com/nasa/cumulus/pull/2300/commits/ccf13ed4065724577e5bad76f9f1870f00f69f41

example/cumulus-tf/terraform.tfvars.example
jennyhliu: Similar to distribution_url, cumulus_distribution_url is used to overwrite the original output of the cumulus_distribution module, and it allows you to specify a different url like cloudfront url or a one with port number.
So if you rename these two variables, make sure the 'overwrite' functionalities still work.

npauzenga: Ah, that's helpful. Thanks @jennyhliu! I was unclear why we would take the output of the module and give it back to the input but that makes sense.

@markdboyd is that what you were getting at or did you have other concerns?

example/cumulus-tf/outputs.tf
jennyhliu: I think it's better to change the input variable `distribution_url -> tea_distribution_url` as well if you change the output.
https://github.com/nasa/cumulus/blob/master/example/cumulus-tf/variables.tf#L178
https://github.com/nasa/cumulus/blob/791a955480b12faa1d511c4d7abcdfae0f177124/example/cumulus-tf/thin_egress_app.tf#L41

npauzenga: Yup! Good idea. Here's the update:

https://github.com/nasa/cumulus/pull/2300/commits/10d32874983563d5b405b21586555b8eb30f025f

I also added `tea_distribution_url` to the tfvars.example:

https://github.com/nasa/cumulus/pull/2300/commits/00c539d3e79b7f8c1e0ea123ad137ba77d1afa1e

Do my comments there make sense? Do you always need to uncomment and use one those variables? Is there a situation where you actually don't need to overwrite the URL to add a port?

jennyhliu: Both `tea_distribution_url` and `cumulus_distribution_url` are optional.  You don't have to overwrite in the case the original url is accessible.

example/deployments/cumulus/jl-tf.tfvars
```diff
@@ -41,5 +41,5 @@ key_name      = "jl"
 include_orca = true
 drop_database = "False"
 
-distribution_url = "https://jwhwz6qg4j.execute-api.us-east-1.amazonaws.com:7000/DEV/"
+tea_distribution_url = "https://e5wdgj4q4i.execute-api.us-east-1.amazonaws.com/DEV/"
```
jennyhliu, line 44:
```suggestion
tea_distribution_url = "https://jwhwz6qg4j.execute-api.us-east-1.amazonaws.com:7000/DEV/"
```

npauzenga: Ah yeah, I think I replaced this with mine, sorry about that. I'm curious: why do you have these variables in `deployments/cumulus/jl-tf.tfvars`? Is it so that your CI stack is more easily testable if you want to do something like manually test auth? (assuming this is your CI stack, could be wrong)

jennyhliu: I use the same stack for local build and testing.

npauzenga: Cool, that makes sense. I might update my tfvars so I can do the same in a future PR 😄

docs/deployment/README.md#cumulus-instance-sizing
```diff
@@ -441,8 +441,8 @@ Outputs:
 
 archive_api_redirect_uri = https://abc123.execute-api.us-east-1.amazonaws.com/dev/token
 archive_api_uri = https://abc123.execute-api.us-east-1.amazonaws.com/dev/
-distribution_redirect_uri = https://abc123.execute-api.us-east-1.amazonaws.com/DEV/login
-distribution_url = https://abc123.execute-api.us-east-1.amazonaws.com/DEV/
+tea_distribution_redirect_uri = https://abc123.execute-api.us-east-1.amazonaws.com/DEV/login
+tea_distribution_url = https://abc123.execute-api.us-east-1.amazonaws.com/DEV/
```
markdboyd, line 445: So these output names wouldn't change for end users unless you make a corresponding change in the template-depoy repo, which it doesn't appear that you have:

https://github.com/nasa/cumulus-template-deploy/pull/63/files

Whether we should is another question. But if end users are only deploying TEA or cumulus, it's probably not as essential

example/README.md
```diff
@@ -155,6 +155,11 @@ Run `terraform apply`.
...
+
+The steps below assume you will be using the [Cumulus Distribution API](./cumulus-tf/cumulus_distribution.tf). If you would prefer
+to use the Thin Egress App (TEA), uncomment the TEA-specific resources in the "cumulus" module in [example/cumulus-tf/main.tf](./cumulus-tf/main.tf).
```
markdboyd, line 161: If you prefer to use TEA and not the cumulus distribution API, you also have to disable it, correct?

docs/deployment/README.md#cumulus-instance-sizing
npauzenga: Ah ok that makes sense. Just the outputs from `example` have changed.

example/README.md
npauzenga: You definitely could. My thinking was that, because TEA was the default, if you should choose to deploy TEA instead we'd still deploy the `cumulus_distribution` module but we just wouldn't use it. I guess I was thinking that way because we have integration tests looking for the Cumulus Distribution module and not TEA so your deployment of TEA seemed more likely to be temporary. That only applies to CI deployments from `example` though, you could be deploying using TEA for other testing purposes and then you'd have a useless Cumulus Distribution module taking up space.

🤔 maybe you're right and just a note here to comment it would suffice and cover both bases.

markdboyd: You at least need to change the language to clarify that uncommenting TEA will end up deploying both, not one or the other. And you're not really "choosing a distribution API" at that point but just enabling TEA.

npauzenga: 👍  See what you think of this: https://github.com/nasa/cumulus/pull/2300/commits/dddcc836c7c312f8110060951d759c40b996a92f

markdboyd:
```suggestion
to use the Thin Egress App (TEA), uncomment the TEA-specific variables in the "cumulus" module in [example/cumulus-tf/main.tf](./cumulus-tf/main.tf).
```

markdboyd: Looks good! Just one minor nit

npauzenga: Ah yeah, a `resource` is a thing in Terraform. Good call.

example/README.md
```diff
@@ -155,6 +155,17 @@ Run `terraform apply`.
 
 This will deploy your data-migration resources.
 
+## Enable a distribution API
+
+The steps below assume you will be using the [Cumulus Distribution API](./cumulus-tf/cumulus_distribution.tf) and no additional changes are required unless you want to use the Thin Egress App instead. If you would prefer
+to use the Thin Egress App (TEA), uncomment the TEA-specific variables in the "cumulus" module in [example/cumulus-tf/main.tf](./cumulus-tf/main.tf).
+
+**Note:** If you choose to deploy TEA by making the above change in [example/cumulus-tf/main.tf](./cumulus-tf/main.tf), both TEA and the Cumulus Distribution API
+will be deployed but only TEA will be enabled for use. If you wish to use TEA and don't want to deploy both, comment or delete the following:
```
markdboyd, line 164:
```suggestion
**Note:** Both TEA and the Cumulus Distribution API are deployed by default but if you make the above change in [example/cumulus-tf/main.tf](./cumulus-tf/main.tf), only TEA will be enabled for use. If you wish to use TEA and don't want to deploy both, comment or delete the following:
```

Nitpicking, but this clarified it for me. The key is that you're not deploying TEA when making the above change because it is still deployed by default, you're just making it functional

markdboyd (approved):
I tested by pulling down the code and deploying with the default Cumulus distribution and even tested a deployment where TEA was not deployed (by moving thin_egress_app.tf to thin_egress_app.tf.example) and it all worked. Great job!

PR #2314, opened 2021-06-25
CUMULUS-2327: API Provisioned Concurrency

nemreid:

Addresses [CUMULUS-2327: As a cumulus integrator, I'd like to see reserved concurrency or some other mechanism in place to protect critical control of cumulus](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2327)

## Changes

* Set API provisioned concurrency to 5. Reserved concurrency was not used as that setting caps the maximum concurrency, whereas provisioned concurrency does not.

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/tf-modules/archive/api.tf
+++ b/tf-modules/archive/api.tf
@@ -156,6 +156,7 @@ resource "aws_lambda_function" "api" {
   role             = aws_iam_role.lambda_api_gateway.arn
   runtime          = "nodejs12.x"
   timeout          = 100
+  publish          = true
   environment {
     variables = merge(local.api_env_variables, {"auth_mode"="public"})
   }
@@ -169,6 +170,14 @@ resource "aws_lambda_function" "api" {
       security_group_ids = concat(local.lambda_security_group_ids, [var.rds_security_group])
     }
   }
+}
+
+resource "aws_lambda_provisioned_concurrency_config" "api_provisioned_concurrency" {
+  function_name = aws_lambda_function.api.function_name
+  provisioned_concurrent_executions = 5
+  qualifier = aws_lambda_function.api.version
+
+  depends_on = [aws_lambda_function.api]
 }
 
 data "aws_iam_policy_document" "private_api_policy_document" {
--- a/tf-modules/archive/api.tf
+++ b/tf-modules/archive/api.tf
@@ -172,12 +172,21 @@ resource "aws_lambda_function" "api" {
   }
 }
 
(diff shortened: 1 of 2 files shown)


nemreid:
New change made includes updating TF config to point the API gateway to a lambda alias invoke ARN. Need to do ad-hoc testing to ensure API lambda remains reachable and invoke via alias works.

tf-modules/archive/api.tf
```diff
@@ -171,6 +172,23 @@ resource "aws_lambda_function" "api" {
...
+  description      = "Cumulus API with provisioned concurrency"
+  function_name    = aws_lambda_function.api.function_name
+  function_version = aws_lambda_function.api.version
+
+  depends_on = [aws_lambda_function.api]
+}
+
+resource "aws_lambda_provisioned_concurrency_config" "api_provisioned_concurrency" {
+  function_name = aws_lambda_alias.provisioned_api.function_name
+  provisioned_concurrent_executions = 5
```
npauzenga, line 186: Is there any reason we might want this as a configurable `var`?

npauzenga (commented):
Nice @nemreid! It does indeed look like this adds provisioned concurrency. I'm curious how it solves the problem of concurrent timeouts when the concurrency limit is reached though. I'm sure it does, I just haven't used it this way. I've only gone down this road if I'm having a problem with cold starts and need to ensure that some seldom-used Lambda stays warm to deliver a quick response.

If we have a provisioned concurrency of x, does that mean that, even if you hit your concurrency limit, those x invocations will always be available? I was under the impression that your concurrency is all in one bucket. If you set provisioned concurrency, it'll keep that number of invocations from taking time to spin up but won't impact your concurrency limit. If you reach or exceed that limit you'll need reserved concurrency for that Lambda to basically steal from the larger, account-level pool of concurrency.

That's my understanding, which is subject to being critically wrong 😄. I think I'm missing some things here.

I was looking at Reserved Concurrency here:

https://docs.aws.amazon.com/lambda/latest/operatorguide/reserved-concurrency.html

nemreid:
Nate has raised some excellent points, and I think I will need to revisit and make configurable the solution proposed in the [original PR](https://github.com/nasa/cumulus/pull/2310) after consulting with some of our users running in production on what they tend to see in terms of concurrency so as to be able to set a reasonable default. Closing this PR and branch.

PR #2324, opened 2021-06-30
[CUMULUS-2460] Get granule execution history from executions endpoint

dopeters:
**Summary:** POST /executions/search-by-granules endpoint

Addresses [CUMULUS-2460: executions endpoint should be able to return a list of workflowNames](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2460)

## Changes

* `POST` /executions/search-by-granules - allows user to return list of all executions associated with a list of granule Ids or a granuleId query
* Consolidated the two different execution endpoint test files into one

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


packages/api/endpoints/executions.js
```diff
@@ -79,6 +86,38 @@ async function del(req, res) {
...
+  const executionCumulusIds = await granulesExecutionsPgModel
+    .getExecutionCumulusIdsFromGranuleCumulusIds(knex, granuleCumulusIds);
+
+  const workflowNames = await executionPgModel
+    .getWorkflowNamesFromExecutionCumulusIds(knex, executionCumulusIds);
+
+  return res.send([...new Set(workflowNames)]);
+}
+
+router.post('/history', validateBulkGranulesRequest, history);
```
markdboyd, line 119: Are these endpoints always returning a set of workflows for a single granule ID or a set of granule IDs? If so I'd have two change requests:

- Change the route from `/history` to `/workflows`, since that's the data we're really returning
- Move the definition of these routes to `endpoints/granules`, since the data is being returned based on granule ID as an input/path parameter. You could end up having this same route where execution ARN is a path parameter, but where it is a granule ID, I think this belongs with the other granule routes

packages/api/endpoints/executions.js
```diff
@@ -2,9 +2,16 @@
...
-const { getKnexClient, ExecutionPgModel } = require('@cumulus/db');
+const {
+  getKnexClient,
+  ExecutionPgModel,
+  GranulesExecutionsPgModel,
+  GranulePgModel,
+} = require('@cumulus/db');
 const Search = require('@cumulus/es-client/search').Search;
 const models = require('../models');
+const { getGranuleIdsForPayload } = require('../lambdas/bulk-operation');
```
markdboyd, line 13: Nit: If we need to re-use this helper here, I'd prefer to move it somewhere like `lib/` rather than requiring it from another lambda file

packages/api/endpoints/executions.js
```diff
@@ -79,6 +86,38 @@ async function del(req, res) {
...
+ * @param {Object} req - express request object
+ * @param {Object} res - express response object
+ * @returns {Promise<Object>} the promise of express response object
+ */
+async function history(req, res) {
+  const { granuleId } = req.params;
+  const payload = req.body;
+  const granuleIds = granuleId ? [granuleId] : await getGranuleIdsForPayload(payload);
+
+  const knex = await getKnexClient({ env: process.env });
```
markdboyd, line 101:
```suggestion
  const knex = await getKnexClient();
```
`env` defaults to `process.env` in `getKnexClient`, so I think this is unnecessary

packages/api/endpoints/executions.js
markdboyd: You also wouldn't need to have moved `validateBulkGranulesRequest` in that case

packages/api/lambdas/bulk-operation.js
```diff
@@ -52,7 +52,10 @@ async function getGranuleIdsForPayload(payload) {
       body.hits.hits.forEach((hit) => {
         granuleIds.push(hit._source.granuleId);
       });
-      if (body.hits.total.value !== granuleIds.length) {
+
+      const totalHits = body.hits.total.value || body.hits.total;
```
markdboyd, line 56: No issue with the change, but what is the cause? Can Elasticsearch return a response in multiple formats?

packages/db/src/models/base.ts
```diff
@@ -96,6 +96,25 @@ class BasePgModel<ItemType, RecordType extends { cumulus_id: number }> {
...
+   *
+   * @param {Knex | Knex.Transaction} knexOrTransaction -
+   *  DB client or transaction
+   * @param {Function} whereFunction -
+   *  An object or any portion of an object of type RecordType
+   * @returns {Promise<Array<number>>} An array of cumulus_ids for the returned records
+   */
+  async getRecordsCumulusIds(
+    knexOrTransaction: Knex | Knex.Transaction,
+    whereFunction: Function
```
markdboyd, line 110: I don't know if I understand the reason for using a function as the argument type vs an object or array or whatever. It's a little more than a nitpick because a function argument can't be typed very easily to prevent misuse

packages/db/src/models/execution.ts
```diff
@@ -34,6 +34,28 @@ class ExecutionPgModel extends BasePgModel<PostgresExecution, PostgresExecutionR
...
+   * single execution cumulus_id or array of exeuction cumulus_ids
+   * @returns {Promise<Array<number>>} An array of workflow names
+   */
+  async getWorkflowNamesFromExecutionCumulusIds(
+    knexOrTrx: Knex | Knex.Transaction,
+    executionCumulusIds: Array<string> | string
+  ): Promise<Array<string>> {
+    const executionCumulusIdsArray
+      = (typeof executionCumulusIds === 'string') ? [executionCumulusIds] : executionCumulusIds;
+    const executionWorkflowNames = await knexOrTrx(this.tableName)
```
markdboyd, line 53:
```suggestion
    const executionResults = await knexOrTrx(this.tableName)
```

Kind of a nitpick. These values are results from the DB query at this point, which are objects containing the workflow name and not just the workflow name

packages/db/src/models/execution.ts
```diff
@@ -34,6 +34,28 @@ class ExecutionPgModel extends BasePgModel<PostgresExecution, PostgresExecutionR
...
+   * @param {Function} executionCumulusIds -
+   * single execution cumulus_id or array of exeuction cumulus_ids
+   * @returns {Promise<Array<number>>} An array of workflow names
+   */
+  async getWorkflowNamesFromExecutionCumulusIds(
+    knexOrTrx: Knex | Knex.Transaction,
+    executionCumulusIds: Array<string> | string
+  ): Promise<Array<string>> {
+    const executionCumulusIdsArray
+      = (typeof executionCumulusIds === 'string') ? [executionCumulusIds] : executionCumulusIds;
```
markdboyd, line 52: If we're sure we have either a string or an array, which we should be because it's Typescript, then you could do

`const executionCumulusIdsArray = [executionCumulusIds].flat();`

packages/api/endpoints/executions.js
```diff
@@ -79,6 +86,38 @@ async function del(req, res) {
...
+
+  const executionPgModel = new ExecutionPgModel();
+  const granulePgModel = new GranulePgModel();
+  const granulesExecutionsPgModel = new GranulesExecutionsPgModel();
+
+  const granuleCumulusIds = await granulePgModel
+    .getRecordsCumulusIds(knex, (builder) => builder.whereIn('granule_id', granuleIds));
+
+  const executionCumulusIds = await granulesExecutionsPgModel
+    .getExecutionCumulusIdsFromGranuleCumulusIds(knex, granuleCumulusIds);
```
markdboyd, line 111: Doing these two things seems like something common enough we might need to do often, so I'd put a helper in `db/lib/granule.ts` to handle just returning execution cumulus IDs for any set of parameters from granules

packages/api/endpoints/executions.js
Jkovarik: Nitx2: Agree, and (and this is entirely not *this* PR, but given we're moving it) consider cleaning up the logic branches as well (e.g factor out  

```
  const uniqueGranuleIds = [...new Set(granuleIds)];
  return uniqueGranuleIds;
  
```
into a method and do something like (but less terse, ternary is for brevity):

 granule.ids.length ? uniqueify(granuleIds) : uniqueify(queryElasticSearhc(query)); in the parent method.

Not nit: we shouldn't be requiring cross /lambdas

packages/api/endpoints/executions.js
dopeters: So this is interesting. (Also, as a note, I just changed this to return the full executions)

@laurenfrederick had originally said that since we're returning executions/info about executions, this should be encapsulated in the /executions endpoint. 

Though I do agree that having granuleId as a path parameter rather than arn could case some confusion. We could maybe nix the GET request to avoid that? Or we could move it into granules, would just want Lauren's input on that.

Jkovarik: +1 to this should be a granule endpoint.

packages/db/src/models/execution.ts
dopeters: Oh neat! Done

packages/api/lambdas/bulk-operation.js
dopeters: Yeah so this was causing an infinite loop when I was testing the query payload. I didn't dig too deep into it, so maybe there's just some sort of ElasticSearch setting that needs to be changed for testing, but yes. The test was returning `body.hits.total` as a number rather than an object. When looking through ES documentation, this seems to be a valid response, though it wasn't clear to me when it would return one over the other (again, could be some setting we have 🤷🏻‍♀️). I figured it was a safe enough change

packages/db/src/models/base.ts
dopeters: This was me trying to make a whereIn query work while having the same basic params as the single record version of this function since there's no way to pass it an object https://knexjs.org/#Builder-whereIn

However, I've now changed it so that we just pass in the columnNames and the values

Jkovarik: Agree, adding a comment here in a few.

packages/api/endpoints/executions.js
dopeters: Agreed, done!

packages/db/src/models/execution.ts
dopeters: Yep, already changed this. Let me know if you think it should be 'executionResults' rather than just 'executions'. (Or if we should just return the value directly since we're no longer transforming it)

packages/db/src/models/base.ts
Jkovarik: So the reason for using the function is to expose the .whereIn method via the builder, and *that* is needed because we always await results and don't expose the raw knex object back. 

A couple of options:

Jkovarik: Off the cuff and hastily reassembled: 

I think the current stance was to attempt to avoid generic abstraction that isn't type safe by making the query logic explicit in a /lib helper, particularly in the feature branch.  That may be the path of least resistance, but perhaps inelegant.   

We could opt to return the search knex object without awaiting it, allowing users to chain.   That would allow for queries like 

`await granulePgModel.search(knex, {}).whereIn('granule_id', ['foo'])`
`await granulePgModel.search(knex, {}).where((builder) => { granuleId: 'foo'})`

of course, that doesn't fix that chaining the .whereIn query like that isn't handled well via return typing, and in some ways just punts the problem to the endpoint logic (and the less logic of that sort in the API the better, given it's lack of typing).   That seems not great. 



Or...

 punt entirely and just expose a base method like 

const query(knexOrTransaction: Knex | Knex.Transaction) { return knexOrTransaction(this.tableName))

Or

We could improve the base model search to allow for something like: 

```
 search(
    params: Partial<RecordType>,
    select: keyof RecordType[],
    whereIn: Partial<RecordType>[]

  ): Promise<RecordType[]> {
    return knexOrTransaction(this.tableName).select(select).where((builder) => {
      builder.where(params);
      if(whereIn) {
        // add lots of wherein clauses or more cleverly type whereIn and do <logic here to support the need> 
      }
.... <snip>
```

that is a bit more charming in that the whereIn could be properly typed both in our method and relative to the query logic. 

:thinking face:

packages/api/tests/endpoints/test-executions.js
```diff
@@ -2,78 +2,266 @@
 
 const test = require('ava');
 const request = require('supertest');
-const awsServices = require('@cumulus/aws-client/services');
+const cryptoRandomString = require('crypto-random-string');
```
Jkovarik, line 5: Was this changeset a git mv from test/endpoints/execution.js plus additions, or were there updates to existing tests?

packages/api/endpoints/executions.js
Jkovarik: Perhaps consider /executions/search?   We're not really getting an execution's history....   I won't hold up on this either, given the arch discussion as I don't have an alternate *better* suggestion at the moment.

packages/db/src/lib/granule.ts
```diff
@@ -45,3 +45,31 @@ export const upsertGranuleWithExecutionJoinRecord = async (
   }
   return [granuleCumulusId];
 };
+
+/**
+ * Upsert a granule and a record in the granules/executions join table.
```
markdboyd, line 50:
```suggestion
 * Get cumulus IDs for all executions associated to a set of granules
```

dopeters: :doh: thanks

packages/api/tests/endpoints/test-executions.js
dopeters: Yeah the former. There were no updates to existing tests.

packages/db/src/models/granules-executions.ts
```diff
@@ -38,6 +38,26 @@ export default class GranulesExecutionsPgModel {
       .merge();
   }
 
+  /**
+   * Get execution_cumulus_id column values from the granule_cumulus_id
+   *
+   * @param {Knex | Knex.Transaction} knexOrTransaction -
+   *  DB client or transaction
+   * @param {Function} granuleCumulusIds -
```
markdboyd, line 46:
```suggestion
   * @param {string | Array<string>} granuleCumulusIds -
```

packages/db/src/models/base.ts
```diff
@@ -96,6 +96,26 @@ class BasePgModel<ItemType, RecordType extends { cumulus_id: number }> {
...
+  /**
+   * Get cumulus_id column value for multiple records in Postgres
+   *
+   * @param {Knex | Knex.Transaction} knexOrTransaction -
+   *  DB client or transaction
+   * @param {Array<string>} columnNames - column names for whereIn query
+   * @param {Knex.QueryCallback} values - record values for whereIn query
+   * @returns {Promise<Array<number>>} An array of cumulus_ids for the returned records
+   */
+  async getRecordsCumulusIds(
```
markdboyd, line 108: We need a unit test for this function

packages/db/src/models/granules-executions.ts
```diff
@@ -38,6 +38,26 @@ export default class GranulesExecutionsPgModel {
...
+  /**
+   * Get execution_cumulus_id column values from the granule_cumulus_id
+   *
+   * @param {Knex | Knex.Transaction} knexOrTransaction -
+   *  DB client or transaction
+   * @param {number | Array<number>} granuleCumulusIds -
+   * single granule_cumulus_id or array of granule_cumulus_ids
+   * @returns {Promise<Array<number>>} An array of execution_cumulus_ids
+   */
+  async getExecutionCumulusIdsFromGranuleCumulusIds(
```
markdboyd, line 50:
```suggestion
  async searchByGranuleCumulusIds(
```

Pretty subjective, up to you whether you want to take this one

markdboyd: Regardless, we need a unit test here as well

packages/db/src/lib/granule.ts
```diff
@@ -45,3 +45,31 @@ export const upsertGranuleWithExecutionJoinRecord = async (
...
+ * @param {Knex.QueryCallback} values - record values for whereIn query
+ * @param {Object} [granulePgModel] - Granule PG model class instance
+ * @param {Object} [granulesExecutionsPgModel]
+ *   Granules/executions PG model class instance
+ * @returns {Promise<number[]>}
+ */
+export const getGranuleExecutionCumulusIds = async (
+  knexOrTransaction: Knex | Knex.Transaction,
+  columnNames: Array<string>,
+  values: Knex.QueryCallback,
```
markdboyd, line 64: Is `Knex.QueryCallback` the right type here? It seems like that type is for something internal to Knex. In general we're passing these `values` to a `whereIn` clause, so it seems like they have to be an array of any values? So would `Array<any>` be better?

packages/db/src/models/base.ts
```diff
@@ -96,6 +96,26 @@ class BasePgModel<ItemType, RecordType extends { cumulus_id: number }> {
...
+    knexOrTransaction: Knex | Knex.Transaction,
+    columnNames: Array<string>,
```
markdboyd, line 110:
```suggestion
columnNames: Array<keyof Partial<RecordType>>
```

This will give us some type safety of ensuring that the column names have to be valid column names for the record type

packages/db/src/lib/granule.ts
```diff
@@ -45,3 +45,31 @@ export const upsertGranuleWithExecutionJoinRecord = async (
...
+ * @param {Array<string>} columnNames - column names for whereIn query
+ * @param {Knex.QueryCallback} values - record values for whereIn query
+ * @param {Object} [granulePgModel] - Granule PG model class instance
+ * @param {Object} [granulesExecutionsPgModel]
+ *   Granules/executions PG model class instance
+ * @returns {Promise<number[]>}
+ */
+export const getGranuleExecutionCumulusIds = async (
+  knexOrTransaction: Knex | Knex.Transaction,
+  columnNames: Array<string>,
```
markdboyd, line 63:
```suggestion
  columnNames: Array<keyof Partial<PostgresGranuleRecord>>,
```

packages/db/src/models/execution.ts
```diff
@@ -34,6 +34,25 @@ class ExecutionPgModel extends BasePgModel<PostgresExecution, PostgresExecutionR
...
+  /**
+   * Get executions from the execution cumulus_id
+   *
+   * @param {Knex | Knex.Transaction} knexOrTrx -
+   *  DB client or transaction
+   * @param {Array<number>} executionCumulusIds -
+   * single execution cumulus_id or array of exeuction cumulus_ids
+   * @returns {Promise<Array<number>>} An array of exeuctions
+   */
+  async getExecutionsFromCumulusIds(
```
markdboyd, line 47:
```suggestion
  async searchByCumulusIds(
```

Very subjective. Take it or leave it

markdboyd: Need a unit test here though

packages/api/endpoints/executions.js
markdboyd: We discussed this in the arch meeting and I was convinced by the argument that what we're really returning here are execution records, thus this is an executions endpoint which is simply filtered by granule IDs.

packages/api/endpoints/executions.js
```diff
@@ -79,6 +85,32 @@ async function del(req, res) {
...
+  const executionCumulusIds = await getGranuleExecutionCumulusIds(knex, ['granule_id'], granuleIds);
+
+  const executions = await executionPgModel
+    .getExecutionsFromCumulusIds(knex, executionCumulusIds);
+
+  return res.send(executions);
+}
+
+router.post('/history', validateBulkGranulesRequest, history);
+router.get('/history/:granuleId', history);
```
markdboyd, line 113: What is the rationale for having a GET and a POST endpoint to both return executions by granule IDs? Why not just always use the POST that can handle both?

packages/api/endpoints/executions.js
markdboyd: I do kind of like the idea of a `/search` endpoint but it's kind of specific to searching by granules, so I don't know if `/search-by-granules` is a bit wordy.  Because I don't think we could make the endpoint logic generic enough to handle any kind of search

packages/api/tests/endpoints/test-executions.js
```diff
@@ -2,78 +2,266 @@
...
+    {
+      execution_cumulus_id: t.context.executionCumulusId2[0],
+      granule_cumulus_id: Number(t.context.granuleCumulusId),
+    },
+  ];
+
+  await Promise.all(
+    t.context.joinRecords.map((joinRecord) =>
+      granulesExecutionsPgModel.create(knex, joinRecord))
+  );
```
markdboyd, line 249: Use the `upsertGranuleWithExecutionJoinRecord` helper instead of having to create the records and join records separately

packages/api/tests/endpoints/executions.js
```diff
@@ -1,234 +0,0 @@
...
-  const response = await request(app)
-    .get('/logs/executions')
-    .set('Accept', 'application/json')
-    .set('Authorization', 'Bearer ThisIsAnInvalidAuthorizationToken')
-    .expect(401);
-
-  assertions.isInvalidAccessTokenResponse(t, response);
-});
-
-test('GET logs returns all executions', async (t) => {
```
markdboyd, line 141: Did this test get preserved in the move?

packages/api/tests/endpoints/test-executions.js
```diff
@@ -170,3 +369,240 @@ test('GET fails if execution is not found', async (t) => {
...
+  const { fakeGranules, fakePGExecutions } = t.context;
+
+  const response = await request(app)
+    .get(`/executions/history/${fakeGranules[0].granuleId}`)
+    .set('Accept', 'application/json')
+    .set('Authorization', `Bearer ${jwtAuthToken}`);
+
+  response.body.forEach((execution, index) => {
+    t.is(execution.arn, fakePGExecutions[index].arn);
+    t.is(execution.workflow_name, fakePGExecutions[index].workflow_name);
```
markdboyd, line 458: Given that we now are returning full executions from the API, is there any reason not to do this?

```suggestion
    t.deepEqual(execution, fakePGExecutions[index]);
```

This comment applies throughout the tests

packages/api/tests/endpoints/test-executions.js
```diff
@@ -170,3 +369,240 @@ test('GET fails if execution is not found', async (t) => {
...
+test('POST /history returns all workflow names when ids array is passed', async (t) => {
+  const { fakeGranules, fakePGExecutions } = t.context;
+
+  const response = await request(app)
+    .post('/executions/history')
+    .send({ ids: [fakeGranules[0].granuleId, fakeGranules[1].granuleId] })
+    .set('Accept', 'application/json')
+    .set('Authorization', `Bearer ${jwtAuthToken}`);
+
+  response.body.forEach((execution, index) => {
```
markdboyd, line 471: Are the same set of executions returned by this test and the previous one for a specific granule ID? If so, how do we know whether the logic for one granule ID vs multiple is working correctly?

packages/api/tests/endpoints/test-executions.js
```diff
@@ -170,3 +369,240 @@ test('GET fails if execution is not found', async (t) => {
...
+    t.is(execution.arn, fakePGExecutions[index].arn);
+    t.is(execution.workflow_name, fakePGExecutions[index].workflow_name);
+  });
+});
+
+test('POST /history returns all workflow names when ids array is passed', async (t) => {
```
markdboyd, line 462:
```suggestion
test('POST /history returns correct executions when ids array is passed', async (t) => {
```

packages/api/tests/endpoints/test-executions.js
```diff
@@ -170,3 +369,240 @@ test('GET fails if execution is not found', async (t) => {
...
+  const response = await request(app)
+    .delete(`/executions/${nonExistantExecution.arn}`)
+    .set('Accept', 'application/json')
+    .set('Authorization', `Bearer ${jwtAuthToken}`)
+    .expect(404);
+
+  t.is(response.body.message, 'No record found');
+});
+
+test('GET /history/:granuleId returns all workflow names associated with the granule', async (t) => {
```
markdboyd, line 448:
```suggestion
test('GET /history/:granuleId returns all executions associated with the granule', async (t) => {
```

packages/api/tests/endpoints/test-executions.js
```diff
@@ -170,3 +369,240 @@ test('GET fails if execution is not found', async (t) => {
...
+    .set('Accept', 'application/json')
+    .set('Authorization', `Bearer ${jwtAuthToken}`);
+
+  response.body.forEach((execution, index) => {
+    t.is(execution.arn, fakePGExecutions[index].arn);
+    t.is(execution.workflow_name, fakePGExecutions[index].workflow_name);
+  });
+});
+
+test.serial('POST /history returns all workflow names when query is passed', async (t) => {
```
markdboyd, line 477:
```suggestion
test.serial('POST /history returns correct executions when query is passed', async (t) => {
```

packages/api/tests/endpoints/test-executions.js
```diff
@@ -170,3 +369,240 @@ test('GET fails if execution is not found', async (t) => {
...
+test.serial('POST /executions/history returns 400 when the Metrics ELK stack is not configured', async (t) => {
+  const expectedIndex = 'my-index';
+  const expectedQuery = { query: 'fake-query' };
+
+  const body = {
+    index: expectedIndex,
+    query: expectedQuery,
+  };
+
+  process.env.METRICS_ES_USER = undefined;
```
markdboyd, line 598:
```suggestion
  const metricsUser = process.env.METRICS_ES_USER ;
  delete process.env.METRICS_ES_USER;
  t.teardown(() => process.env.METRICS_ES_USER = metricsUser);
```

This will prevent this test from potentially interfering with other tests that might depend on the ELK stack

packages/api/tests/endpoints/executions.js
dopeters: Not specifically, but it seemed this was covered within the test `default returns list of executions`

packages/api/tests/endpoints/test-executions.js
dopeters: Adde an additional test execution so that the two tests are returning different sets

packages/api/tests/endpoints/test-executions.js
dopeters: Was having issues with the date fields not being the same type, but I did a refactor to consolidate some of the fakeExecutions v. fakePGExecutions and also filtered out those fields for the comparison, so this should be solved now for all the tests

packages/api/tests/endpoints/test-executions.js
dopeters: Ohh thanks for the tip!

packages/db/src/models/execution.ts
dopeters: Yeah I like the search phrasing for these functions so made those changes throughout

packages/api/endpoints/executions.js
dopeters: Updated this to be /search-by-granules. I think having "search" in there does make more sense. And making it specific to granules is also important as noted. Maybe a little wordy, but I don't think overly so.

packages/db/src/lib/granule.ts
dopeters: Yeah good point. Made some updates to the types

packages/api/endpoints/executions.js
dopeters: Yeah the original thinking was to allow for an easier single granule request, but probably isn't necessary to have both. Removed the GET

packages/db/src/lib/granule.ts
markdboyd: This seems unaddressed?

packages/db/src/lib/granule.ts
markdboyd: This still needs to be addressed

packages/db/src/models/base.ts
markdboyd: This still needs to be addressed

packages/db/src/models/base.ts
```diff
@@ -96,6 +96,26 @@ class BasePgModel<ItemType, RecordType extends { cumulus_id: number }> {
...
+    values: Knex.QueryCallback
```
markdboyd, line 111: Need to update this type

packages/api/tests/endpoints/executions.js
markdboyd: Yeah this test looks unnecessary. Thanks

packages/api/tests/endpoints/test-executions.js
```diff
@@ -170,3 +370,226 @@ test('GET fails if execution is not found', async (t) => {
...
+
+  const response = await request(app)
+    .post('/executions/search-by-granules')
+    .send({ ids: [fakeGranules[0].granuleId, fakeGranules[1].granuleId] })
+    .set('Accept', 'application/json')
+    .set('Authorization', `Bearer ${jwtAuthToken}`);
+
+  t.is(response.body.length, 3);
+
+  response.body.forEach((execution, index) => t.deepEqual(
```
markdboyd, line 452: By using a `forEach` and assuming the index of the returned results matches the index of the order we created the PG records, we're assuming how the results are ordered when returned by the database. Obviously it's working for now, but it could change if we change how results are ordered in our query to search for records.

I think preferable solution is still to iterate but to find the record for comparison doing a `find` on `fakePgExecutions` where the `arn` values match and then comparing those records. This would apply throughout the tests

We don't need to change it for now, just a heads up.

packages/db/src/models/base.ts
dopeters: Whoops, forget to push that

packages/api/tests/endpoints/test-executions.js
dopeters: I was having some issues with the result order and this does future proof more than what I had, so I made that change

packages/db/src/models/base.ts
```diff
@@ -96,6 +96,26 @@ class BasePgModel<ItemType, RecordType extends { cumulus_id: number }> {
     return record.cumulus_id;
   }
 
+  /**
+   * Get cumulus_id column value for multiple records in Postgres
+   *
+   * @param {Knex | Knex.Transaction} knexOrTransaction -
+   *  DB client or transaction
+   * @param {Array<keyof RecordType>,} columnNames - column names for whereIn query
```
markdboyd, line 104:
```suggestion
   * @param {Array<keyof RecordType>} columnNames - column names for whereIn query
```

packages/api/endpoints/executions.js
```diff
@@ -79,6 +85,28 @@ async function del(req, res) {
...
+ * @param {Object} res - express response object
+ * @returns {Promise<Object>} the promise of express response object
+ */
+async function searchByGranules(req, res) {
+  const payload = req.body;
+  const granuleIds = await getGranuleIdsForPayload(payload);
+  const knex = await getKnexClient();
+  const executionPgModel = new ExecutionPgModel();
+
+  const executionCumulusIds = await getGranuleExecutionCumulusIds(knex, ['granule_id'], granuleIds);
```
markdboyd, line 101: Is there any way that the dashboard or whatever is making this call can specify the collection name/version when making this request?

I ask because in the new PostgreSQL database, only the combination of granule_id/collection_cumulus_id makes a granule unique. `granule_id` by itself is not unique because we have a requirement to support the same granule in multiple collections (which was part of the motivation for moving to PostgreSQL). 

So I think we need to update this somehow so that the desired collection for each granule is specified so that we don't end up returning executions for the wrong granule (since it could exist for multiple collections going forward).

dopeters: Thanks for the heads up on this! I made some updates to handle an array of granule objects containing a granuleId and collectionId. The commit with all the important stuff is here: https://github.com/nasa/cumulus/pull/2324/commits/59590e4d094d6b8ea451bdabfe812d925f755772

The biggest changes were
1. Updating `getGranuleExecutionCumulusIds` to get the granule cumulus ids using the granule_id and collection_cumulus_id fields
2. Updating the elasticsearch query to return both granuleid and collection id so that we can pass that array to `getGranuleExecutionCumulusIds` in the right format.

Also updated and added some more tests. Let me know what you think!

packages/api/lib/granules.js
```diff
@@ -225,10 +225,125 @@ async function moveGranule(apiGranule, destinations, distEndpoint, granulesModel
...
+    });
+
+    esGranules.map((granule) => queryGranules.push({
+      granuleId: granule.granuleId,
+      collectionId: granule.collectionId,
+    }));
+  }
+  // Remove duplicate Granule IDs
+  // TODO: could we get unique IDs from the query directly?
+  const uniqueGranules = uniqWith(queryGranules, isEqual);
```
markdboyd, line 337: So is the idea of this function that either we already have an array of objects like `{ granuleId: string, collectionId: string}` OR we get an array of that structure back from ES based on the query?

packages/db/src/lib/granule.ts
```diff
@@ -1,9 +1,12 @@
 import Knex from 'knex';
 
 import { PostgresGranule } from '../types/granule';
+import { CollectionPgModel } from '../models/collection';
 import { GranulePgModel } from '../models/granule';
 import { GranulesExecutionsPgModel } from '../models/granules-executions';
 
+const { deconstructCollectionId } = require('../../../api/lib/utils');
```
markdboyd, line 8:
```suggestion
const { deconstructCollectionId } = require('@cumulus/message/Collections');
```

We want to require from the cumulus message package rather than cross linking to the API package

packages/db/src/lib/granule.ts
```diff
@@ -45,3 +48,39 @@ export const upsertGranuleWithExecutionJoinRecord = async (
...
+  granules: Array<{ collectionId: string, granuleId: string }>,
+  collectionPgModel = new CollectionPgModel(),
+  granulePgModel = new GranulePgModel(),
+  granulesExecutionsPgModel = new GranulesExecutionsPgModel()
+): Promise<Array<number>> => {
+  const granuleCumulusIds: Array<number> = await Promise.all(granules.map(async (granule) => {
+    const { name, version } = deconstructCollectionId(granule.collectionId);
+    return await granulePgModel.getRecordCumulusId(knexOrTransaction, {
+      granule_id: granule.granuleId,
+      collection_cumulus_id: await collectionPgModel.getRecordCumulusId(
```
markdboyd, line 75: Sort of a nit, but it **could** be expensive to run this query for every collection depending on the number of granules. Can we create a hash map to store the `cumulus_id` of collections for each name/version combo and only look it up if we didn't already?

packages/api/lib/granules.js
```diff
@@ -225,10 +225,125 @@ async function moveGranule(apiGranule, destinations, distEndpoint, granulesModel
...
+  }
+
+  // Remove duplicate Granule IDs
+  // TODO: could we get unique IDs from the query directly?
+  const uniqueGranuleIds = [...new Set(granuleIds)];
+  return uniqueGranuleIds;
+}
+
+/**
+ * Return a unique list of granule cumulus IDs based on the provided list or the response from the
```
markdboyd, line 306:
```suggestion
 * Return a unique list of granules based on the provided list or the response from the
```

packages/api/lib/granules.js
```diff
@@ -225,10 +225,125 @@ async function moveGranule(apiGranule, destinations, distEndpoint, granulesModel
...
+ * query to ES using the provided query and index.
+ *
+ * @param {Object} payload
+ * @param {Object} [payload.granules] - Optional list of granules with granuleId and collectionId
+ * @param {Object} [payload.query] - Optional parameter of query to send to ES
+ * @param {string} [payload.index] - Optional parameter of ES index to query.
+ * Must exist if payload.query exists.
+ * @returns {Promise<Array<string>>}
```
markdboyd, line 314:
```suggestion
 * @returns {Promise<Array<Object>>}
```

I think this is the correct return?

packages/api/endpoints/executions.js
```diff
@@ -79,6 +85,29 @@ async function del(req, res) {
...
+ * Get execution history for a single granule or multiple granules
+ *
+ * @param {Object} req - express request object
+ * @param {Object} res - express response object
+ * @returns {Promise<Object>} the promise of express response object
+ */
+async function searchByGranules(req, res) {
+  const payload = req.body;
+  const knex = await getKnexClient();
+  const granules = await getGranulesForPayload(payload, knex);
```
markdboyd, line 98: If I'm understanding correctly, this will accept or return an array of granules like `{ granuleId: string, collectionId: string}`. And thus we have to look up the collection_cumulus_id for that collection in order to find the executions in PG.

So my question is: can this endpoint actually accept an array of granules like `{ granule_id: string, collection_cumulus_id: number}` as **input**? The answer really depends on when this will be merged/released. If it's being released in the next sprint release after merge, then the answer is probably no. But if it's released with the changes where the granule endpoint will be returning granules from PG with those properties, then the answer might be yes

It's probably a "no", but just checking to see if we can save some DB queries

packages/db/src/lib/granule.ts
```diff
@@ -45,3 +48,39 @@ export const upsertGranuleWithExecutionJoinRecord = async (
...
+ * @param {Knex | Knex.Transaction} knexOrTransaction -
+ *  DB client or transaction
+ * @param {Array<Object>} granules - array of granules with collectionId and granuleId
+ * @param {Object} [collectionPgModel] - Collection PG model class instance
+ * @param {Object} [granulePgModel] - Granule PG model class instance
+ * @param {Object} [granulesExecutionsPgModel]
+ *   Granules/executions PG model class instance
+ * @returns {Promise<number[]>}
+ */
+export const getGranuleExecutionCumulusIds = async (
```
markdboyd, line 64:
```suggestion
export const getApiGranuleExecutionCumulusIds = async (
```

This is nitpicky, but if the function is going to except input in the form of a `collectionId`/`granuleId`, which are from the Dynamo/existing API structure, I think it's helpful

packages/api/lib/granules.js
dopeters: Yeah that's right. Similar to `getGranuleIdsForPayload` but it returns the object rather than just the id. Which I'm now realizing means I should update the type (which you've already noted)

packages/db/src/lib/granule.ts
dopeters: It seems like this doesn't exist in the message package? But I can certainly add it there. Originally I was requiring this from the api package, but that led to a dependency loop, so I changed it to match another require in this same package https://github.com/nasa/cumulus/blob/master/packages/db/src/translate/granules.ts#L7

I can update both places though

packages/db/src/lib/granule.ts
dopeters: Yeah that's a really good call

packages/api/endpoints/executions.js
dopeters: Yeah this is a feature that operators have been waiting on since last PI so I think the plan is to release it asap. Otherwise, I do get what you're saying here. Do you think it's worth adding a ticket to follow up once all the RDS stuff is out there, maybe for an IP sprint? We'd probably have to keep it backwards compatible, but we could add additional handling for the collection cumulus id. At the very least, then any request from the dashboard could save those DB queries.

packages/db/src/lib/granule.ts
markdboyd: Sorry we must have only added that in the feature branch but yeah if we can do that here too that would be ideal. Thanks!

packages/api/endpoints/executions.js
markdboyd: I would probably just hold off on a ticket because I recall now that the plan is for us to return the structure of returned API endpoints. There isn't even any planned work afaik to update the granule endpoints to return the data structure from Postgres vs the current Dynamo output structure.

So no work or follow up needed here. Thanks for clarifying

packages/api/lib/request.js
```diff
@@ -33,6 +33,81 @@ async function verifyJwtAuthorization(requestJwtToken) {
...
+  if (payload.granules) {
+    if (!Array.isArray(payload.granules)) {
+      return res.boom.badRequest(`granules should be an array of values, received ${payload.granules}`);
+    }
+
+    if (!payload.query && payload.granules.length === 0) {
+      return res.boom.badRequest('no values provided for granules');
+    }
+
+    payload.granules.forEach((granule) => {
```
markdboyd, line 82: Very nice

packages/db/src/lib/granule.ts
```diff
@@ -45,3 +48,49 @@ export const upsertGranuleWithExecutionJoinRecord = async (
...
+  collectionPgModel = new CollectionPgModel(),
+  granulePgModel = new GranulePgModel(),
+  granulesExecutionsPgModel = new GranulesExecutionsPgModel()
+): Promise<Array<number>> => {
+  const collectionMap: {[key: string]: number} = {};
+
+  const granuleCumulusIds: Array<number> = await Promise.all(granules.map(async (granule) => {
+    const { collectionId } = granule;
+    let collectionCumulusId = collectionMap[collectionId];
+    const { name, version } = deconstructCollectionId(granule.collectionId);
```
markdboyd, line 76: Nit: I think this could could move inside the `if` because it's only necessary if we need to do the PG lookup. But not critical

packages/api/tests/endpoints/test-executions.js
```diff
@@ -170,3 +365,269 @@ test('GET fails if execution is not found', async (t) => {
...
+    .post('/executions/search-by-granules')
+    .set('Accept', 'application/json')
+    .set('Authorization', `Bearer ${jwtAuthToken}`)
+    .send(body)
+    .expect(400);
+
+  t.regex(response.body.message, /Index is required if query is sent/);
+});
+
+test.serial('POST /executions/search-by-granules returns 400 when no IDs or query is provided', async (t) => {
```
markdboyd, line 519:
```suggestion
test.serial('POST /executions/search-by-granules returns 400 when no granules or query is provided', async (t) => {
```

packages/api/tests/endpoints/test-executions.js
```diff
@@ -170,3 +365,269 @@ test('GET fails if execution is not found', async (t) => {
...
+    .post('/executions/search-by-granules')
+    .set('Accept', 'application/json')
+    .set('Authorization', `Bearer ${jwtAuthToken}`)
+    .send(body)
+    .expect(400, /One of granules or query is required/);
+
+  t.regex(response.body.message, /One of granules or query is required/);
+});
+
+test.serial('POST /executions/search-by-granules returns 400 when IDs is not an array', async (t) => {
```
markdboyd, line 536:
```suggestion
test.serial('POST /executions/search-by-granules returns 400 when granules is not an array', async (t) => {
```

packages/api/tests/endpoints/test-executions.js
```diff
@@ -170,3 +365,269 @@ test('GET fails if execution is not found', async (t) => {
...
+    .post('/executions/search-by-granules')
+    .set('Accept', 'application/json')
+    .set('Authorization', `Bearer ${jwtAuthToken}`)
+    .send(body)
+    .expect(400);
+
+  t.regex(response.body.message, /granules should be an array of values/);
+});
+
+test.serial('POST /executions/search-by-granules returns 400 when IDs is an empty array', async (t) => {
```
markdboyd, line 554:
```suggestion
test.serial('POST /executions/search-by-granules returns 400 when granules is an empty array', async (t) => {
```

packages/api/tests/endpoints/test-executions.js
```diff
@@ -170,3 +365,269 @@ test('GET fails if execution is not found', async (t) => {
...
+    .post('/executions/search-by-granules')
+    .set('Accept', 'application/json')
+    .set('Authorization', `Bearer ${jwtAuthToken}`)
+    .send(body)
+    .expect(400);
+
+  t.regex(response.body.message, /no values provided for granules/);
+});
+
+test.serial('POST /executions/search-by-granules returns 400 when granules do not have collectionId', async (t) => {
```
markdboyd, line 572: Awesome work adding these tests

packages/db/tests/lib/test-granule.js
```diff
@@ -12,9 +13,11 @@ const {
   fakeExecutionRecordFactory,
   fakeGranuleRecordFactory,
   upsertGranuleWithExecutionJoinRecord,
+  getApiGranuleExecutionCumulusIds,
 } = require('../../dist');
 
 const { migrationDir } = require('../../../../lambdas/db-migration/dist/lambda');
+const { constructCollectionId, deconstructCollectionId } = require('../../../es-client/node_modules/@cumulus/message/Collections');
```
markdboyd, line 20:
```suggestion
const { constructCollectionId, deconstructCollectionId } = require('@cumulus/message/Collections');
```

packages/db/tests/lib/test-granule.js
```diff
@@ -326,3 +329,142 @@ test('upsertGranuleWithExecutionJoinRecord() succeeds if granulePgModel.upsert()
...
+    granules,
+    collectionPgModel,
+    granulePgModel,
+    granulesExecutionsPgModel
+  );
+
+  t.deepEqual(results.sort(), [executionCumulusId, secondExecutionCumulusId].sort());
+});
+
+test('getApiGranuleExecutionCumulusIds() only queries DB when collection is not in map', async (t) => {
```
markdboyd, line 391: Very nice

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ package to catch or to otherwise handle errors that you may encounter.
 ### Added
 
 - Added user doc describing new features related to the Cumulus dead letter archive.
+- **CUMULUS-2460**
+  - Adds `POST` /executions/search-by-granules for retriving workflow names from a given granule, list of granules, or granule query
```
markdboyd, line 21:
```suggestion
  - Adds `POST` /executions/search-by-granules for retrieving executions from a list of granules or granule query
```

Since the endpoint is returning full executions and not workflow names at this point

packages/api/endpoints/executions.js
```diff
@@ -79,6 +85,29 @@ async function del(req, res) {
...
+
+  const executionPgModel = new ExecutionPgModel();
+
+  const executionCumulusIds = await getApiGranuleExecutionCumulusIds(knex, granules);
+
+  const executions = await executionPgModel
+    .searchByCumulusIds(knex, executionCumulusIds);
+
+  return res.send(executions);
```
markdboyd, line 107: It just occurs to me that we're returning execution records directly out of the PG database here. While there is nothing wrong with that per se, I thought the intention for our current API work was to still return records in the "API" format (basically matching the existing structure of Dynamo records to preserve backwards compatibility with our API). I guess that would be up to @laurenfrederick 

If you do need to do that, then you can take this `translatePostgresExecutionToApiExecution` helper straight out of the feature branch and run it for each execution being returned:

https://github.com/nasa/cumulus/blob/feature/rds-phase-2/packages/db/src/translate/executions.ts#L13

That being said, it also occurs to me that especially when using an ES query to return a bunch of results, there could be many executions returned, so should this endpoint return a limited set of results by default and support paging? Otherwise for a really broad granule query the lambda would probably run out of memory and crash

dopeters: Yeah you're absolutely right - the records should be in the API format. Grabbed that function from the feature branch.

I added some paging functionality to `executionPgModel.searchByCumulusIds` and some corresponding tests. I set the default limit and page to 1 in order to match the other Cumulus API endpoints. Hopefully that's sufficient

packages/api-client/src/executions.ts
```diff
@@ -118,3 +118,36 @@ export const deleteExecution = async (params: {
...
+ * POST /executions/search-by-granules
+ * @param {Object} params             - params
+ * @param {Object} params.body       - body to pass the API lambda
+ * @param {Function} params.callback  - async function to invoke the api lambda
+ *                                      that takes a prefix / user payload.  Defaults
+ *                                      to cumulusApiClient.invokeApifunction to invoke the
+ *                                      api lambda
+ * @returns {Promise<Object>}         - the response from the callback
+ */
+export const searchExecutionsByGranules = async (params: {
```
markdboyd, line 133: Would add a changelog note that this has been added

packages/api/tests/endpoints/test-executions.js
```diff
@@ -146,6 +329,17 @@ test('executions can be filtered by workflow', async (t) => {
   t.is(fakeExecutions[1].arn, results[0].arn);
 });
 
+test('GET executions with asyncOperationId filter returns the correct executions', async (t) => {
+  const response = await request(app)
+    .get('/executions?asyncOperationId=0fe6317a-233c-4f19-a551-f0f76071402f')
```
markdboyd, line 334: I think this is a copy of an existing test, but I don't understand how it works. I don't see where in the setup we ensure that executions exist that reference this async operation

packages/api/endpoints/executions.js
```diff
@@ -79,6 +86,34 @@ async function del(req, res) {
...
+ * @param {Object} req - express request object
+ * @param {Object} res - express response object
+ * @returns {Promise<Object>} the promise of express response object
+ */
+async function searchByGranules(req, res) {
+  const payload = req.body;
+  const knex = await getKnexClient();
+  const granules = await getGranulesForPayload(payload, knex);
+  const { page = 1, limit = 1 } = req.query;
+  const offset = page < 1 ? 0 : (page - 1) * limit;
```
markdboyd, line 101: Is `page` ever less than 1? It defaults to 1 so it's only less than 1 if someone passes in `page=0` as a query parameter I guess. 

Also do these default values mean that by default 1 result is returned per page? It seems like we could up that to 10 or so without compromising performance and require less pagination from the front-end client. Of course I guess the front-end can/will just specify these query parameters

Regardless, I think we need some unit tests to ensure that sending the query params to the endpoint makes the pagination work as expected. I know we already have unit tests to prove that the PG model handles the pagination, but our logic in this endpoint handler to pass on the pagination values needs to be validated.

dopeters: Yeah basically just wanted to have some protection in case the user passes in 0 or something else funky for the page value.

Set the default to 1 to be consistent with other endpoints (or at least what the docs say here about page and limit https://nasa.github.io/cumulus-api/) but can change if wanted

Couldn't figure out how to link to the particular line, but there is a unit for the paging here https://github.com/nasa/cumulus/pull/2324/files#diff-b8a520e8e7976a793a61b3a1df0386af448d81490730b9ec430419e655ef5d53R456 - L456 in api/tests/endpoints/test-executions.js "POST /executions/search-by-granules supports paging". Let me know if I should add any more.

packages/api/tests/endpoints/test-executions.js
dopeters: Correct this is a copy of an existing test. We're creating an execution with a that asyncOperationId on line 119 and then indexing the array of executions. 

```
  await Promise.all(
    fakeExecutions.map((i) =>
      executionModel
        .create(i)
        .then((record) =>
          indexer.indexExecution(esClient, record, t.context.esAlias)))
  );
```

jennyhliu:
For cumulus-dashboard reingest/bulkReingest, we need interfaces like:

- `/executions/workflowNames-by-granules` returns the 'common' workflow names for all the granules
   The payload will be either ids:[granuleId1, granuleId2] or {index: '', query: ''}
- `/executions?granuleId=granuleid1` returns the executions for a granule order by timestamp desc
   This may not be needed for reingest, but I think it's needed for displaying executions for a single granule

packages/api/endpoints/executions.js
markdboyd: That all makes sense. Sorry I somehow missed those unit tests on the first pass

markdboyd (approved):
Great work!

dopeters:
> * `/executions?granuleId=granuleid1` returns the executions for a granule order by timestamp desc
>   This may not be needed for reingest, but I think it's needed for displaying executions for a single granule

Will address this in a follow up PR.

PR #2353, opened 2021-07-08
Add documentation for deploying Cumulus Distribution

npauzenga:

There is a corresponding change to cumulus-template-deploy open for review here: https://github.com/nasa/cumulus-template-deploy/pull/63

This is user-facing documentation for how to configure and deploy the new Cumulus Distribution API. There is also a page on the wiki that should be reviewed as part of this PR:

https://wiki.earthdata.nasa.gov/display/CUMULUS/Cumulus+Distribution+Options

Testing instructions:

1. Follow the directions starting from docs/deployment/README.md#cumulus-instance-sizing to deploy using the new distribution API and Cognito. Use the `cumulus-template-deploy` repo with these changes: https://github.com/nasa/cumulus-template-deploy/pull/63
2. Test ingesting and publishing a granule. The file from CMR should contain links to download the granule files from S3 via the new distribution API. Ping me for help testing this.
3. Re-deploy with the TEA variables set and test that the distribution app is switched to TEA by running the same workflow and looking for the TEA URL in the CMR file.

I'm very open to adding to these docs. I'm working under the assumption that we're mainly concerned with documenting deployment, not so much Cognito configuration (as we don't control that). It feels weird to leave that piece out entirely though so this might be an area that needs input.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


docs/deployment/cumulus-distribution.md
```diff
@@ -0,0 +1,15 @@
...
+
+The Cumulus Distribution API is a set of endpoints that can be used to enable AWS Cognito authentication when downloading data from S3.
+
+## Configuring a Cumulus Distribution deployment
+
+The Cumulus Distribution API is included in the main [Cumulus](https://github.com/nasa/cumulus/tree/master/tf-modules/cumulus_distribution) repo.
+
+To configure a deployment to use Cumulus Distribution, remove or comment the "Thin Egress App Settings" in [the Cumulus Template Deploy](https://github.com/nasa/cumulus-template-deploy/blob/master/cumulus-tf/main.tf) and enable the Cognito settings.
+
+No additional configuration is necessary to deploy the Cumulus Distribution API.
```
jennyhliu, line 15: They need to provide a bucket map file for cumulus_distribution as they do for TEA, and a bucket map file will be put to s3.
https://github.com/nasa/cumulus/tree/master/example/cumulus-tf/cumulus_distribution
In our example, we either take a bucket_map.yaml or bucket_map.yaml.tmpl
https://github.com/nasa/cumulus/blob/master/example/cumulus-tf/cumulus_distribution.tf#L3-L15

Also we need to call out cumulus_distribution module will generate a distribution bucket map cache file, and overwrites the bucket map cache file generated by distribution(TEA) module if any.  The bucket map cache files generated by cumulus_distribution and distribution module should be the same if they are using the same bucket map file.
cc @Jkovarik

npauzenga: Sure, have a look at this:

https://github.com/nasa/cumulus/pull/2353/commits/d29db71cf327614278cce46dff742a56f1d4a701

Parts were taken from `thin-egress-app.md`. Is it missing anything?

docs/deployment/cumulus-distribution.md
```diff
@@ -0,0 +1,37 @@
...
+  "daac-public-data-bucket": "/path/to/this/kind/of/data"
+}
+```
+
+> Note: Cumulus only supports a one-to-one mapping of bucket->Cumulus Distribution path for 'distribution' buckets.
+
+> Note: The bucket map **must include mappings for all of the `protected` and `public` buckets specified in the `buckets` variable in `cumulus-tf/terraform.tfvars`**, otherwise Cumulus may not be able to determine the correct distribution URL for ingested files and you may encounter errors.
+
+## Switching from the Thin Egress App to Cumulus Distribution
+
```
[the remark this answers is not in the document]
npauzenga: Ah ok, right. We don't include cumulus_distribution in the template deploy yet. I will make a new `cumulus_distribution.tf.example` and a directory for the bucket maps: `cumulus-tf/cumulus_distribution`.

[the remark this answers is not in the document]
npauzenga: I added the module in the template deploy repo here:

https://github.com/nasa/cumulus-template-deploy/pull/63/files

and updated these docs here:

77209a99bebb9ab34c393a612717e4c6724cd3bb

I didn't call out the bucket mapping explicitly because we have it in the above `S3 Bucket Mapping` section but is there something I should highlight?

This is deploying now so is currently untested. I'll make sure this actually does what I think it does.

docs/deployment/README.md#cumulus-instance-sizing
```diff
@@ -422,9 +422,20 @@ Notes on specific variables:
 
 Consider [the sizing of your Cumulus instance](#cumulus-instance-sizing) when configuring your variables.
 
-#### Configure the Thin Egress App
+#### Choose a distribution API
```
Jkovarik, line 425: Why is this file called README.md#cumulus-instance-sizing  🤔  ?

Jkovarik, line 425: This header level doesn't seem correct.     It's a sub topic under `### Configure and deploy the `cumulus-tf` root module`, but then the new sub sections ` ## Configure the Thin Egress App` and ` ## Configure the Thin Egress App` are at a higher level than the surrounding sections.

docs/deployment/README.md#cumulus-instance-sizing
```diff
@@ -422,9 +422,20 @@ Notes on specific variables:
...
 
-The Thin Egress App is used for Cumulus distribution. Follow the steps [in the documentation](./thin_egress_app) to configure distribution in your `cumulus-tf` deployment.
+Cumulus can be configured to use either the Thin Egress App (TEA) or the Cumulus Distribution API. The default selection is the Thin Egress App if you're using the [Deployment Template](https://github.com/nasa/cumulus-template-deploy).
+
+**IMPORTANT** If you already have a deployment using the TEA distribution and want to switch to Cumulus Distribution, there will be an API Gateway change. This means that there will be downtime while you update your CloudFront endpoint to use
+the new API gateway.
+
+## Configure the Thin Egress App
```
Jkovarik, line 432: see above comment re: headings

docs/deployment/README.md#cumulus-instance-sizing
```diff
@@ -422,9 +422,20 @@ Notes on specific variables:
 
 Consider [the sizing of your Cumulus instance](#cumulus-instance-sizing) when configuring your variables.
 
-#### Configure the Thin Egress App
+#### Choose a distribution API
 
-The Thin Egress App is used for Cumulus distribution. Follow the steps [in the documentation](./thin_egress_app) to configure distribution in your `cumulus-tf` deployment.
+Cumulus can be configured to use either the Thin Egress App (TEA) or the Cumulus Distribution API. The default selection is the Thin Egress App if you're using the [Deployment Template](https://github.com/nasa/cumulus-template-deploy).
+
+**IMPORTANT** If you already have a deployment using the TEA distribution and want to switch to Cumulus Distribution, there will be an API Gateway change. This means that there will be downtime while you update your CloudFront endpoint to use
```
Jkovarik, line 429: Is this NGAP specific?

docs/deployment/cumulus-distribution.md
```diff
@@ -0,0 +1,41 @@
...
+id: cumulus_distribution
+title: Using the Cumulus Distribution API
+hide_title: false
+---
+
+The Cumulus Distribution API is a set of endpoints that can be used to enable AWS Cognito authentication when downloading data from S3.
+
+## Configuring a Cumulus Distribution deployment
+
+The Cumulus Distribution API is included in the main [Cumulus](https://github.com/nasa/cumulus/tree/master/tf-modules/cumulus_distribution) repo.
```
Jkovarik, line 11: Nit (and we aren't doing this elsewhere), we should probably explicitly call out the `.zip` target in the release documentation or tf-module README.   I've run into a few users not certain what this sort of statement means in terms of 'where do I get it from'.

docs/deployment/cumulus-distribution.md
```diff
@@ -0,0 +1,41 @@
...
+
+To configure a deployment to use Cumulus Distribution, remove or comment the "Thin Egress App Settings" in [the Cumulus Template Deploy](https://github.com/nasa/cumulus-template-deploy/blob/master/cumulus-tf/main.tf) and enable the Cumulus Distribution settings.
```
Jkovarik, line 13: Is there a PR to add these?

Jkovarik: Answer: yes - https://github.com/nasa/cumulus-template-deploy/pull/63

docs/deployment/cumulus-distribution.md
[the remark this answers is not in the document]
jennyhliu: ok.  Also need to add cognito specific variables to `cumulus-template-deploy` and instructions so the deployment for cognito would work.

docs/deployment/README.md#cumulus-instance-sizing
npauzenga: Yup. No idea. I was thinking it was some weird docasaurus convention. I can't find a good reason why though and naming it README.md doesn't seem to break anything so...

36a798437d75720e24b4fb6347ff62a97626afd5

npauzenga: Yup, you're totally right. It seems like it should be:

`### Configure and deploy the cumulus-tf root module`
`#### Choose a distribution API`
`#### Configure the Thin Egress App`
`#### Configure the Cumulus Distribution API (optional)`
`#### Deploy`

Where everything under that main heading is at the same level. Choosing your distribution API is a concrete step on equal footing with configuring one or the other in my mind.

https://github.com/nasa/cumulus/pull/2353/commits/d632a8bd5332b8b7c1ab10e20490ece54154812c

docs/deployment/README.md#cumulus-instance-sizing
npauzenga: 👍 https://github.com/nasa/cumulus/pull/2353/commits/d632a8bd5332b8b7c1ab10e20490ece54154812c

docs/deployment/README.md#cumulus-instance-sizing
npauzenga: Oh interesting. Would it not be true? I was thinking that you'd be creating a new API gateway regardless and doing so would mean _some_ period of time switching from the old TEA gateway to the new one. Am I thinking about this wrong?

docs/deployment/cumulus-distribution.md
npauzenga: How about we add a note here in this deployment docs?

https://github.com/nasa/cumulus/pull/2353/commits/09f40077771b1e3886e58de81a9919a79eb39d99

And then we can also spell it out in the release notes?

docs/deployment/cumulus-distribution.md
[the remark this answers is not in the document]
npauzenga: Yup, you're totally right. I'm adding these things to this repo but haven't tested a full deploy using Cognito from it. I'll commit once it's fully deployed and working.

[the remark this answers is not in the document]
npauzenga: Here's an update to the docs here and I'll add the variable and test in cumulus-template-deploy

https://github.com/nasa/cumulus/pull/2353/commits/1d49b58df622bb0df9d30ef68c7637ab1abcc1a4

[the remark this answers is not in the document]
jennyhliu: Also the csdap_* variables https://github.com/nasa/cumulus/blob/master/example/cumulus-tf/terraform.tfvars.example#L64

npauzenga:
Made a few updates here. Tested and it deploys the Cumulus Distribution API. This is definitely worth someone else running through it to make sure I'm not making assumptions.

docs/deployment/cumulus-distribution.md
```diff
@@ -0,0 +1,55 @@
...
+ 3. Uncomment the Cumulus Distribution outputs in [outputs.tf](https://github.com/nasa/cumulus-template-deploy/blob/master/cumulus-tf/outputs.tf).
+ 4. Set the Cumulus Distribution variables in your `terraform.tfvars` (which is based on [terraform.tfvars.example](https://github.com/nasa/cumulus-template-deploy/blob/master/cumulus-tf/terraform.tfvars.example)). These include:
+
+  - cumulus_distribution_url
+  - csdap_client_id
+  - csdap_client_password
+  - csdap_host_url
+  - any others under the `Cumulus Distribution Variables` section in the example
+
+  5. Rename `cumulus-template-deploy/cumulus-tf/cumulus_distribution.tf.example` to `cumulus-template-deploy/cumulus-tf/cumulus_distribution.tf`.
```
npauzenga, line 28: There's a lot of manual intervention here to switch between distribution APIs. You have to comment/delete a bunch of things and add a few others. We switch the `cumulus_distribution` module based on a conditional but that isn't as straightforward in all cases. It's almost certainly doable but it's a question of effort given how often people will make the switch. Open to conversation here as we could very probably be more elegant here.

tf-modules/cumulus/variables.tf
```diff
@@ -544,3 +544,9 @@ variable "ecs_custom_sg_ids" {
   type = list(string)
   default = []
 }
+
+variable "deploy_cumulus_distribution" {
+  description = "If true, does not deploy the TEA distribution API"
+  type        = bool
+  default     = true
```
jennyhliu, line 551: We want to default it to false if TEA is the default, so the current user won't need to pass/set this variable.

npauzenga: Yeah hmm that's right. From a user perspective it's `false` as the template defaults to TEA. I was thinking `true` because example defaults to cumulus distribution but that shouldn't matter. I see your point.

https://github.com/nasa/cumulus/pull/2353/commits/3e00220523bad590aa8b1e923bf0507b183e4726

jennyhliu (approved):
Great work!

PR #2355, opened 2021-07-08
remove db-migration-tf module

markdboyd:
This is unnecessary since `tf-modules/data-persistence` now deploys `lambdas/db-migration`, as implemented in https://github.com/nasa/cumulus/pull/2176/files


npauzenga (approved):
Looks good @markdboyd!

PR #2356, opened 2021-07-08
CUMULUS-2548: Fix basic auth for sync from HTTP(s) providers

markdboyd:

Addresses [CUMULUS-2548: HttpProviderClient.sync() auth credentials are not being forwarded during redirect](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2548)

## Changes

* Fixed `@cumulus/ingest/HttpProviderClient.sync` to properly handle basic auth when redirecting to a different host

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
  - Manually created collection/provider/rule to ingest from data set requiring authentication (https://gpm1.gesdisc.eosdis.nasa.gov/data/GPM_L3/GPM_3IMERGHH.06/2020/001/) and confirmed that ingest was successful  
- [ ] Integration tests


packages/ingest/test/test-HttpsProviderClient.js
```diff
@@ -212,3 +236,33 @@ test('HttpsProviderClient supports basic auth with redirects for sync', async (t
     await recursivelyDeleteS3Bucket(destinationBucket);
   }
 });
+
+test('HttpsProviderClient supports basic auth for sync with redirect to different host/port', async (t) => {
```
markdboyd, line 240: So this unit test covers the gap which our logic wasn't handling where during a redirect to a site with a different host/port, basic auth credentials were not forwarded, so the request would fail with a 401 error. This behavior comes from `got` internally:

https://github.com/sindresorhus/got/blob/main/source/core/index.ts#L744

This scenario is exactly what happens when trying to download a file from a DAAC HTTP provider: you request a file from their host (e.g. lpdaac.com/file.hdf) and you get redirected to Earthdata login (e.g. urs.earthdata.nasa.gov) to provide credentials before you are redirected back to download the file. For example you will see the redirects in action if you visit this URL in an incognito window:

https://gpm1.gesdisc.eosdis.nasa.gov/data/GPM_L3/GPM_3IMERGHH.06/2020/001/3B-HHR.MS.MRG.3IMERG.20200101-S223000-E225959.1350.V06B.HDF5

Given the above, I also wonder if we need an integration test. Because the `got` behavior to disable credentials on redirects to different hosts is a recent change that probably broke our sync with basic auth functionality and we didn't catch it because our unit tests didn't test that scenario. With this new unit test, in theory we're covered, but I'm still wondering. However, it would be pretty tricky to set up an integration test that enforces a basic auth redirect to a different host and it's yet another test to maintain.

packages/ingest/src/HttpProviderClient.js
```diff
@@ -62,14 +64,35 @@ class HttpProviderClient {
 
     if (this.username) this.gotOptions.username = this.username;
     if (this.password) this.gotOptions.password = this.password;
+
+    const RedirectHandler = {
+      // Need to use named function and not fat arrow
+      // expression so that we can use the bound value
+      // of "this"
+      handleBeforeRedirect(options, response) {
+        if (redirectCodes.has(response.statusCode)) {
```
markdboyd, line 73: The question is: is this secure? `got` is intentionally disabling the forwarding of credentials when redirecting to a different host, which is good for security. Otherwise you are forwarding your basic auth credentials to a site that may or may not need them and could in theory be a bad actor that intercepts them.

The only way to keep the forwarding of credentials on redirect is maybe to add a new provider type, `basic-auth`, and another provider field of `redirectAuthHost`, so then you can explicitly set what host(s) should be used to allow auth redirects and avoid sending credentials to possibly malicious sites. Thoughts?

npauzenga: So you have a request coming from sync-granule which includes a Basic Auth header because it's a workflow task and we can't prompt for login or anything like that. It redirects to EDL but the basic auth header has been dropped, then EDL responds with a 401 or it redirects back to the distribution API which responds with a 401 (doesn't matter I guess). Is that basically the problem?

A whitelist with our auth providers (currently EDL and Cognito) is one way to go but that does seem like something that could be forgotten about if another auth provider is added. I'm not sure how easily catchable that would be but that might be the best solution. I can't imagine adding auth providers is something that's going to happen regularly and maybe it would be possible to write a test that would fail if you add a new one and don't whitelist it here.

Is there any way to avoid the redirect altogether or is that a major change (seems like yes)? If we separated the steps where one request makes the auth call and another subsequent request either includes the cookie or uses the existing session to download the files?

packages/ingest/src/HttpProviderClient.js
```diff
@@ -62,14 +64,35 @@ class HttpProviderClient {
...
   }
 
   async downloadTLSCertificate() {
     if (!this.certificateUri || this.certificate !== undefined) return;
     try {
       const s3Params = parseS3Uri(this.certificateUri);
       this.certificate = await getTextObject(s3Params.Bucket, s3Params.Key);
-      this.gotOptions.ca = this.certificate;
+      this.gotOptions.https = this.gotOptions.https || {};
+      this.gotOptions.https.certificateAuthority = this.certificate;
```
npauzenga, line 95: What was this change about?

packages/ingest/test/test-HttpsProviderClient.js
npauzenga: Yeah, this is tricky. My opinion is that we should _not_ have an integration test for this. If a third-party library introduces a breaking change, ideally we would have tests to catch it but I'm not sure we want to write tests that explicitly test the behavior of those libraries. That sounds like what an integration test would be:

`Redirects to protected files work if auth is provided` (or something)

Before your change above, if we had that test it would have felt to me like testing the library (e.g. redirects include auth headers) pretty directly. Now that we have a custom handler it's a little different but ...is it? That handler is tested here at the unit level. Actually handling the redirect still seems like what an integration test would test and that doesn't seem like something we would want to include.

packages/ingest/test/test-HttpsProviderClient.js
```diff
@@ -212,3 +236,33 @@ test('HttpsProviderClient supports basic auth with redirects for sync', async (t
...
+  const destinationBucket = randomString();
+  const destinationKey = 'syncedFile.hdf';
+  try {
+    await s3().createBucket({ Bucket: destinationBucket }).promise();
+    await httpsProviderClient.sync({
+      fileRemotePath: protectedFile2,
+      destinationBucket,
+      destinationKey,
+    });
+    t.truthy(fileExists(destinationBucket, destinationKey));
```
npauzenga, line 259: Isn't `fileExists` async? It seems like it wasn't `awaited` before but... shouldn't it be?

npauzenga (changes requested):

packages/ingest/src/HttpProviderClient.js
markdboyd: This was just to fix a warning from `got` because the `ca` option has been deprecated in favor of `https.certificateAuthority`

packages/ingest/test/test-HttpsProviderClient.js
markdboyd: You are probably right. And the tests are only passing because they're doing `truthy` and an unresolved Promise is `truthy`. But we should add the await and update to `t.true`. Great catch

packages/ingest/src/HttpProviderClient.js
markdboyd: >Is that basically the problem?

Yes, that's the problem

>A whitelist with our auth providers (currently EDL and Cognito) is one way to go but that does seem like something that could be forgotten about if another auth provider is added

I'm a little worried about whitelisting only specific auth providers because it just seems likely that either one of the DAACs or our non-NASA users will need to download files from a provider protected by some other auth service. So I'm guess I'm leaning towards making it explicit as a field on the provider.

> Is there any way to avoid the redirect altogether or is that a major change (seems like yes)? If we separated the steps where one request makes the auth call and another subsequent request either includes the cookie or uses the existing session to download the files?

I don't see how we can avoid the redirect whether we separate the request into two steps or not. The HTTP provider itself is what dictates the file being protected by an auth service and so we can't avoid it forcing a redirect on download

You make a good point that I should ensure that a cookie session is set/respected via unit testing, but I believe it will be. However, for the initial request before a session cookie is established, I'm not sure I understand the benefit of separating into two requests vs one

markdboyd: I decided to implement a `basicAuthRedirectHost` field on provider and only forward credentials on redirect if the host for the redirect matches `basicAuthRedirectHost`

Jkovarik: The technical approach seems sound.   The security concern here, in theory, is exposure of EDL credentials via a compromised provider host redirecting to <somewhere else>.... this seems to cover that well enough. 

A couple of thoughts: 

---
> I'm a little worried about whitelisting only specific auth providers because it just seems likely that either one of the DAACs or our non-NASA users will need to download files from a provider protected by some other auth service. So I'm guess I'm leaning towards making it explicit as a field on the provider.

This could avoid adding another key to the provider object if it's a global configuration setting *or* a workflow configuration option for task components that might require it.      Something to consider. 
---

- More generally -- is there a reason not to explicitly fail the request on an unexpected redirect?      Perhaps more to the point, is there generally a good reason other than auth that a http provider *should* redirect other than internally?

markdboyd: >This could avoid adding another key to the provider object if it's a global configuration setting or a workflow configuration option for task components that might require it. Something to consider.

It seems like making it a workflow config is less flexible though - because then it is the same allowed redirect for any invocation of that workflow, whereas a provider field means it can be different for any provider used with that workflow. Absent another benefit to using the workflow config approach, I'm inclined to stick with adding the provider property.

> More generally -- is there a reason not to explicitly fail the request on an unexpected redirect? Perhaps more to the point, is there generally a good reason other than auth that a http provider should redirect other than to an alternate location on the same host?

Can we safely make those kind of assumptions? I mean what if a DAAC for some reason is moving their holdings to a new domain and putting in 301 redirects to get people to the new data location?

Jkovarik: > Can we safely make those kind of assumptions? I mean what if a DAAC for some reason is moving their holdings to a new domain and putting in 301 redirects to get people to the new data location?

I suppose not given the current implementation, although if we make the 'allowed redirect' a *list* that would work, even if it's probably one of my making in the above comment.  

So, yes, we could write that such that it could be adaptable to that situation.     Is it needed/value added?      Up for debate.     We might want to consider a list regardless of our current implementation though.

markdboyd: So we could just generally have an `allowedRedirects` property and deny any redirects outside of that? And forward credentials to any hosts in that list?

Jkovarik: Yes, that's what I was suggesting.    The only hiccup there is if redirects are prevalent in the wild and/or having to explicitly define them places an absurd burden on the end user. 

If you don't hate this idea we should probably whiteboard it briefly

If for some reason we decide we don't like rejecting redirects wholesale at a minimum I think we should consider `allowedRedirects` as a list instead of a single string for reasons of flexibility in the data model/situations we haven't anticipated.

CHANGELOG.md
```diff
@@ -115,6 +125,10 @@ behavior
 - Fixed bug where `cmr_custom_host` variable was not properly forwarded into `archive`, `ingest`, and `sqs-message-remover` modules from `cumulus` module
 - Fixed bug where `parse-pdr` set a granule's provider to the entire provider record when a `NODE_NAME`
   is present. Expected behavior consistent with other tasks is to set the provider name in that field.
+- **CUMULUS-2548**
+  - Fixed `@cumulus/ingest/HttpProviderClient.sync` to
```
Jkovarik, line 129: There should probably be a 'changed' note about the new redirect schema/config updates.

packages/db/src/translate/providers.ts
```diff
@@ -41,6 +41,9 @@ export const translateApiProviderToPostgresProvider = async (
     port: record.port,
     host: record.host,
     protocol: record.protocol,
+    allowed_redirects: record.allowedRedirects
```
Jkovarik, line 44: It shouldn't block this PR, but if providers is in the feature branch, we should either make a point to sort it in the feature under 2548 *or* write a ticket to do so for the phase 2 work.

packages/db/src/types/provider.ts
```diff
@@ -27,8 +28,9 @@ export interface PostgresProvider {
  * postgres for reading.  It differs from the PostgresProvider interface in that it types
  * the autogenerated/required fields in the Postgres database as required
  */
-export interface PostgresProviderRecord extends PostgresProvider {
+export interface PostgresProviderRecord extends Omit<PostgresProvider, 'allowed_redirects'> {
```
Jkovarik, line 31: 👍🏻

packages/ingest/src/HttpProviderClient.js
```diff
@@ -62,14 +64,35 @@ class HttpProviderClient {
 
     if (this.username) this.gotOptions.username = this.username;
     if (this.password) this.gotOptions.password = this.password;
+
+    const RedirectHandler = {
+      // Need to use named function and not fat arrow
+      // expression so that we can use the bound value
+      // of "this"
```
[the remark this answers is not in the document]
Jkovarik: You know generally I'm not a fan of expository comments as we should be writing code that's relatively self-documenting via explicit idioms and maintainable/testable-over-clever structures absent performance concerns, etc. 

In this case the reason for using this isn't tightly coupled to the code in a casual read-through, so the comment is *probably* warranted. 

The alternate is we have a test that more-or-less explicitly covers this and contains the/a comment/commit comment so if an update occurs it fails and it doesn't eat an hour of engineering time to recall 'why'.

packages/ingest/src/HttpProviderClient.js
Jkovarik: We talked over the approach in the 21 Jul Arch meeting with the assembled team.  The updates made were considered an improvement on the existing (failed) behavior.

Jkovarik (changes requested):
Looks good - I'm still reviewing the http test updates, flushing comments as there was a (minor) requested change (CHANGELOG notes), should have this wrapped up shortly.

packages/db/src/translate/providers.ts
markdboyd: What do we need to sort on the feature branch?

packages/ingest/src/HttpProviderClient.js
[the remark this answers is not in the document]
markdboyd: I guess I'm not wedded to having a comment, but it is a departure from most of our code to not use fat arrow functions (`() =>`). However, there's a very specific reason in that we need to use a named function so that the lexical value of `this` is preserved

In terms of unit test vs a comment, the unit test to "prove" that it is working is really any one of the unit tests I've added to prove that redirects are properly handled. So I don't think another specific unit test is needed, but I could be wrong

If people feel strongly I should remove the comment, let me know

packages/db/src/translate/providers.ts
Jkovarik: Sorry! 😅 the postgres->api translation to make sure we're not dropping the value/returning the wrong key.

markdboyd: Ah okay. Yeah I can do a follow up PR

CHANGELOG.md
markdboyd: Is this sufficient?

https://github.com/nasa/cumulus/pull/2356/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR19-R20

CHANGELOG.md
```diff
@@ -6,6 +6,28 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Notable changes
+
+- `@cumulus/sync-granule` task should now properly handle
+syncing files from HTTP/HTTPS providers where basic auth is
+required and involves a redirect to a different host (e.g.
+downloading files protected by Earthdata login)
```
npauzenga, line 14:
```suggestion
downloading files protected by Earthdata Login)
```

npauzenga: Branding 😄

npauzenga: I'm not sure if we want to call it out here but there's a migration required. Maybe that's just highlighted in the release notes?

packages/api/models/schemas.js
```diff
@@ -620,6 +620,13 @@ module.exports.provider = {
       title: 'Host',
       type: 'string',
     },
+    allowedRedirects: {
+      title: 'Allowed hosts for redirect when retrieving data. If a redirect outside of this list is encountered, data retrival will fail.',
+      type: 'array',
+      items: {
+        type: 'string',
+      },
+    },
```
npauzenga, line 629: So when a user creates a new Provider via the dashboard, they will be able to add these? Or is this obscured from users because they don't really need to specify and we know the current hosts?

What happens with existing providers? How do they get `allowedRedirects`?

This might be easier as a quick chat.

packages/api/tests/models/test-providers-model.js
```diff
@@ -162,6 +162,19 @@ test('Providers.create() allows creation of a provider without a globalConnectio
   t.is(fetchedProvider.globalConnectionLimit, undefined);
 });
 
+test('Providers.create() allows creation of a provider with allowedRedirects', async (t) => {
+  const { providersModel } = t.context;
+
+  const provider = fakeProviderFactory();
+  provider.allowedRedirects = ['foo.com'];
```
npauzenga, line 169: Is this the format we're expecting? Just the domain name? Mainly asking because of the bug I just fixed where we're expecting a string but it's unclear what that string represents. In that case ARN vs. full URL.

packages/aws-client/tests/test-S3.js
```diff
@@ -377,3 +379,33 @@ test('getObjectSize() returns the size of an object', async (t) => {
 
   t.is(objectSize, 4);
 });
+
+test('streamS3Upload() uploads contents of stream to S3', async (t) => {
```
npauzenga, line 383: Should we have tests here for the error/reject paths?

lambdas/db-migration/src/migrations/20210715171725_provider_add_allowed_redirects.ts
```diff
@@ -0,0 +1,16 @@
+import * as Knex from 'knex';
+
+export const up = async (knex: Knex): Promise<void> =>
+  await knex.schema.table('providers', (table) => {
+    table
+      .jsonb('allowed_redirects')
```
npauzenga, line 6: Does Knex not allow array types? It seems like we could do a `text[]` and save the trouble of stringifying/decoding but there might be a Knex reason for not doing that.

npauzenga: The internet tells me it's possible with something like this https://stackoverflow.com/questions/50118196/how-to-insert-array-data-type-using-knex-and-potsgres

but there might be reasons to not do it that way. Open to pushback.

packages/db/src/translate/providers.ts
```diff
@@ -41,6 +41,9 @@ export const translateApiProviderToPostgresProvider = async (
...
+      ? JSON.stringify(record.allowedRedirects)
```
npauzenga, line 45: See above. Storing this as an array would save steps like this but 🤷

npauzenga (changes requested):
Looks great @markdboyd! I had a couple questions but nothing critical. 

What is the testing plan for this? Everything looks great and it seems like the tests you have cover it but it would be cool to see it solve the actual problem. Do you have a SIT or some other environment on MAPP that we could use to test before releasing? Might not be critical. Additional fix versions are free 😄

CHANGELOG.md
markdboyd: What is the required migration?

packages/api/tests/models/test-providers-model.js
markdboyd: Yeah that's a very good question. In this case we are expecting just the domain name, or the domain name with a port (`foo.com:80`) if we're being redirected to a different port than the initial request

lambdas/db-migration/src/migrations/20210715171725_provider_add_allowed_redirects.ts
markdboyd: I had absolutely no idea that this existed. I'll look into it. Good find

packages/api/models/schemas.js
markdboyd: Checked with the OX team and it seems like the provider edit form int he dashboard automatically reads the Dynamo schema and should add a field for editing it. However, I'm not sure how that's going to handle the `allowedRedirects` field which is an array type, so I'm gonna test it myself.

CHANGELOG.md
markdboyd: 8bb83f8e44757bd4fc11e273fdc1092e394b94d7

@npauzenga I'm not quite sure what the required migration is though

packages/aws-client/tests/test-S3.js
markdboyd: 520a30525fe721fbae304eb0e99318adee9dfe03

lambdas/db-migration/src/migrations/20210715171725_provider_add_allowed_redirects.ts
markdboyd: b3a173b7afaf8df2689665f914d2e14bcb0355fd
7c494fe0c4751150f450c933708f734939b7fb2f

CHANGELOG.md
npauzenga: I guess it doesn't really matter. I meant the database migration: https://github.com/nasa/cumulus/pull/2356/files#diff-fa9a90bd6d6fb2969591fbf613f3e59e091d0e3b79afa889aea77ac7f6523aee

But those are just run on bootstrap without user intervention? So maybe we don't need to explicitly call these out when they come up?

markdboyd: Yup, the data-persistence module automatically runs the db-migration module for the users, so they don't have to run it manually

https://github.com/nasa/cumulus/blob/master/tf-modules/data-persistence/postgres_migration.tf

packages/db/src/translate/providers.ts
npauzenga: Resolved here: https://github.com/nasa/cumulus/commit/b3a173b7afaf8df2689665f914d2e14bcb0355fd

npauzenga (approved):
Great work @markdboyd! It looks like everything is addressed. The couple post-merge checks/additions seem to be:

1. Check `allowedRedirects` on the dashboard and make sure the array type is supported.
2. Address this in a follow-up PR: https://github.com/nasa/cumulus/pull/2356#discussion_r677689534

packages/api/models/schemas.js
markdboyd: Screen Shot 2021-07-28 at 11 55 16 AM

The dashboard actually handled it great by default

packages/ingest/test/test-HttpsProviderClient.js
```diff
@@ -73,11 +123,64 @@ test.beforeEach(async (t) => {
     port: t.context.server.sslPort,
     certificateUri: `s3://${t.context.configBucket}/certificate.pem`,
   });
+
+  nock(fakeDataServerHost)
```
Jkovarik, line 127: 👍🏻  nice use of knock.

Jkovarik (approved):
You've addressed all my concerns/comments.  Nice work!

PR #2359, opened 2021-07-09
improve ingestGranuleQueueSpec error handling

markdboyd:


example/spec/parallel/ingestGranule/ingestGranuleQueueSpec.js
```diff
@@ -258,7 +256,7 @@ describe('The S3 Ingest Granules workflow', () => {
   });
 
   it('triggers a execution record being added to DynamoDB', async () => {
-    if (beforeAllError) throw SetupError;
+    if (beforeAllError) fail(beforeAllError);
```
npauzenga, line 259: Yup, much clearer to fail the test rather than throwing 👍. I'm curious why it's necessary though. Would line 254:

```
beforeEach(() => {
    if (beforeAllError) fail(beforeAllError);
  });
```
cause the suite to fail but not an individual test? So each `it` block would still run and maybe pass if it didn't require anything from the `beforeAll`? Or something?

npauzenga: I've seen this pattern of failing the test explicitly if `beforeAll` fails but I just haven't paid attention that that `beforeEach`.

npauzenga (approved):

example/spec/parallel/ingestGranule/ingestGranuleQueueSpec.js
markdboyd: Yeah the `beforeEach` doesn't seem to actually do anything. So it's kinda pointless

PR #2365, opened 2021-07-13
missing cmr_custom_host variable in cumulus/ingest submodule

YangusKhan:
## Changes

* When trying to set the `cmr_custom_host` variable in a new terraform deployment of cumulus, there is a missing line inside the cumulus/ingest submodule for propagating that variable out to ingest tasks.

## PR Checklist

None of these should be necessary?

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


markdboyd:
Thanks for this @YangusKhan!

PR #2366, opened 2021-07-13
Increase difference in expected and tested backoff interval.

flamingbear:
Previously this test rounded the returned interval to the nearest integer and tested that against expected backoffs [2, 4, 8].  With this rounding, an interval of 4.51 would not be accepted as a successful test.  This change ensures that if the backoff happens within 1 second of the expected time, the test is successful. e.g. times of [1, 3, 7] as well as [3, 5, 9] would both be accepted as a passed test.

Addresses [CUMULUS-2463: but actually integration test failure](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2463)

## Changes

* Changes acceptance criteria for backoff test timing.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [X] Integration tests


example/spec/parallel/retryConfiguration/retryConfigurationSpec.js
```diff
@@ -118,7 +118,10 @@ describe('When a task is configured', () => {
   describe('with a backoff', () => {
     it('and it fails, Cumulus retries it with the configured backoff time', () => {
       const retryIntervals = getRetryIntervals(retryFailLambdaExecutions);
-      expect(retryIntervals).toEqual([2, 4, 8]);
+      const expectedIntervals = [2, 4, 8];
+      expectedIntervals.forEach((expected, idx) => {
+        expect(Math.abs(expected - retryIntervals[idx])).toBeLessThanOrEqual(1);
```
markdboyd, line 123: Do we need to allow for intervals of less than 2,4,8? I thought we just needed to allow for intervals of 3,5,9 because of rounding up?

flamingbear: I was just guessing that if it could round up incorrectly a fast start on one would cause the rest to also be fast.  I just switched to allow a second on either side of expected. Let me know if you only want later and I'll remove the Math.abs (I think)

markdboyd: It's a nit, so I'll just approve

markdboyd (approved):

example/spec/parallel/retryConfiguration/retryConfigurationSpec.js
flamingbear: It's not hard, I can fix it (just have to do two expects.). I'll do that when I finish my current PR

PR #2369, opened 2021-07-14
Fix cmr_custom_host integration

markdboyd:
Brings in user contribution from https://github.com/nasa/cumulus/pull/2365 and adds additional necessary fixes

## Changes

* Add missing line to `tf-modules/cumulus/ingest.tf` for passing `cmr_custom_host` to the `ingest` module
* Add missing line to `tf-modules/cumulus/sqs-message-remover.tf` for passing `cmr_custom_host` to the `sqs-message-remover` module
* Add missing line to `tf-modules/cumulus/archive.tf` for passing `cmr_custom_host` to the `archive` module

Whoever reviews this PR should manually pull this PR, set a value for `cmr_custom_host` in `cumulus-tf/terraform.tfvars`, then verify that a `CMR_HOST` environment variable shows up in the following Lambdas:

- `PostToCmr`
- `ApiEndpoints`
- `sqsMessageRemover`

## PR Checklist

- [x] Update CHANGELOG


tf-modules/cumulus/variables.tf
```diff
@@ -182,7 +182,7 @@ variable "bucket_map_key" {
 variable "cmr_custom_host" {
   description = "Custom host to use for CMR requests"
   type        = string
-  default     = ""
+  default     = null
```
markdboyd, line 185: This variable is used to set various Lambda environment variables. And it turns out that in Terraform, if the variable used to set an environment variable is `null`, it just excludes it altogether, which is preferable to including and setting it to an empty string

nemreid (approved):
Confirmed that the value gets passed through to target locations in ad-hoc testing.

PR #2370, opened 2021-07-14
CUMULUS-2327: API reserved concurrency

nemreid:

Follow-up from https://github.com/nasa/cumulus/pull/2314
Addresses [CUMULUS-2327: API reserved concurrency](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2327)

## Changes

- Added reserved concurrency setting to the Cumulus API lambda function.
- Added relevant tfvars to the archive and cumulus terraform modules.

## PR Checklist

- [x] Update CHANGELOG


CHANGELOG.md
```diff
@@ -13,10 +13,19 @@ does not match the expected response (200 for most requests and 202 for a few re
 trigger async operations). Previously the helpers in this package would return the response
 regardless of the status code, so you may need to update any code using helpers from this
 package to catch or to otherwise handle errors that you may encounter.
+- The Cumulus API Lambda function has now been configured with reserved concurrency to ensure
+availability in a high-concurrency environment. However, this also caps max concurrency which
+may result in throttling errors if trying to reach the Cumulus API multiple times in a short
+period. Reserved concurrency can be configured with the `archive_api_reserved_concurrency`
+terraform variable on the Cumulus module and increased if you are seeing throttling erorrs.
+The default reserved concurrency value is 8.
```
npauzenga, line 21: It looks like it's set to 5 below?

npauzenga: Oh I see. The default for `example` is 5 and 8 otherwise. Seems fine. Do any users deploy from example? Any reason we would want these to be the same?

npauzenga (commented):
I think this looks good @nemreid! How should we test/verify that this is doing what we want? _Is this something we should test_? Or are we looking at this more as an option that we're providing that _might_ work to alleviate problems?

CHANGELOG.md
nemreid: > Any reason we would want these to be the same?

I was thinking of lowering the example value more, as each account in our sandbox env will deduct from our account wide limit. 8 is a useful default for users, but we don't want to have 30 deployments all reserving 8 concurrency as that would reserve 24% of our account-wide limit.

> How should we test/verify that this is doing what we want?

What I did was deploy it and check on the AWS console that the lambda has the expected value under _Configuration > Concurrency_, which it did.

npauzenga: Verified. Thanks for the info! I deployed to my stack and the `ApiEndpoints` shows a reserved concurrency of 2. Nice work!

npauzenga (approved):
Nice @nemreid!

PR #2371, opened 2021-07-14
Fix dead letter processing

markdboyd:
Fixes the dead letter Lambda to return a structured output so that the write to PG of the completed AsyncOperation should succeed

Update integration test to ensure that async operation is created

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


packages/api/lambdas/process-s3-dead-letter-archive.js
```diff
@@ -64,6 +71,11 @@ async function processDeadLetterArchive({
     }
   } while (listObjectsResponse.IsTruncated);
   /* eslint-enable no-await-in-loop */
+  // Lambda run as an async operation must have a return
+  return {
```
markdboyd, line 75: Without an actual output returned from the Lambda, I think I was seeing the PG write fail on schema validation. Though now I feel like I need to double check this assertion

packages/api/lambdas/process-s3-dead-letter-archive.js
```diff
@@ -47,13 +49,18 @@ async function processDeadLetterArchive({
           return deadLetterObject.Key;
```
nemreid, line 49: > Without an actual output returned from the Lambda, I think I was seeing the PG write fail on schema validation. Though now I feel like I need to double check this assertion

I don't quite understand how this would be the case when the PG write happens on line 48 in this try/catch block. Can you elaborate?

markdboyd: So after re-testing this, the actual failure isn't the writing of the execution/granule/PDR records from `writeRecords`. It's in the async operation when we go to update the dead letter async operation with the `output` of the operation:

```
(node:1) UnhandledPromiseRejectionWarning: error: update "async_operations" set "status" = $1, "output" = $2, "updated_at" = to_timestamp(1629213124.621) where "id" = $3 - invalid input syntax for type json
```

The `output` field is the only JSON type field on an async operation.

CHANGELOG.md
```diff
@@ -11,12 +11,23 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
 
 ### Added
 
 - **CUMULUS-2592**
   - Adds logging when messages fail to be added to queue
 
+### Changed
+
+- Updated `processDeadLetterArchive` Lambda to return an object where
+`processingSucceededKeys` is an array of the S3 keys for succesfully
```
botanical, line 25:
```suggestion
`processingSucceededKeys` is an array of the S3 keys for successfully
```

packages/api-client/src/asyncOperations.ts
```diff
@@ -19,14 +19,15 @@ export const getAsyncOperation = async (params: {
...
-  return await callback({
+  const response = await callback({
     prefix,
     payload: {
       httpMethod: 'GET',
       resource: '/{proxy+}',
       path: `/asyncOperations/${asyncOperationId}`,
     },
   });
+  return JSON.parse(response.body);
```
botanical, line 30: Consider updating the function's description in the comment above for the return value

packages/api/tests/lambdas/test-process-s3-dead-letter-archive.js
```diff
@@ -60,6 +64,16 @@ test('processDeadLetterArchive calls writeRecords for each dead letter Cumulus m
...
+  t.deepEqual(
+    {
+      processingSucceededKeys: output.processingSucceededKeys.sort(),
+      processingFailedKeys: [],
+    },
+    {
+      processingSucceededKeys: t.context.messageKeys.sort(),
+      processingFailedKeys: [],
+    }
+  );
```
botanical, line 76: Cool! I didn't know about `sort()`

packages/db/package.json
```diff
@@ -40,7 +40,8 @@
     "lodash": "^4.17.20",
     "pg": "^8.3.0",
     "snake-camel": "^1.0.6",
-    "uuid": "8.3.1"
+    "uuid": "8.3.1",
+    "crypto-random-string": "^3.2.0"
```
botanical, line 44: this is a nit but can we alphabetize this?

packages/db/tests/lib/test-execution.js
```diff
@@ -565,7 +565,7 @@ test('getWorkflowNameIntersectFromGranuleIds() returns correct values for single
 
   const results = await getWorkflowNameIntersectFromGranuleIds(knex, [granuleCumulusId]);
 
-  t.deepEqual(results.sort(), ['fakeWorkflow', 'fakeWorkflow2']);
+  t.deepEqual(results.sort(), ['fakeWorkflow', 'fakeWorkflow2'].sort());
```
botanical, line 568: Why do we need to sort the second array with `fakeWorkflow`?

botanical (approved):
Had a couple nit comments/suggestions and a question, other than that looks good to me

packages/db/tests/lib/test-execution.js
markdboyd: As I look at this again, I don't know why, since the second array is already sorted alphabetically it should match `results.sort()`, so I'll remove this

CHANGELOG.md
markdboyd: adb15f409ab5b6c9b3464b01e1b9f9320582ad8a

packages/api-client/src/asyncOperations.ts
markdboyd: eef1d9785f65c5981222a1846193cfa4a40df51e

packages/db/package.json
markdboyd: b572abd2a7995d7b31228c566415288225d62778

packages/db/tests/lib/test-execution.js
markdboyd: 5b338537e6e3ebc6ee30bf5c5abdef736ec7a227

PR #2373, opened 2021-07-15
CUMULUS-2373 - Replay archived S3 messages

botanical:

Addresses [CUMULUS-2373: Replay archived S3 messages](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2373)

## Changes
  - Added `replayArchivedS3Messages` lambda to replay archived incoming SQS
    messages from S3.
  - Added `/replayArchivedS3Messages` endpoint to trigger an async operation for
    the `replayArchivedS3Messages` lambda.
  - Added unit tests and integration tests for new endpoint and lambda.
  - Added `getS3PrefixForArchivedMessage` to `ingest/sqs` package to get prefix
    for an archived message.
  - Added new `async_operation` type `Archived S3 Messages Replay`.
  - Added documentation for new endpoint
## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


packages/api/lambdas/replay-archived-s3-messages.js
```diff
@@ -0,0 +1,79 @@
...
+    Bucket: bucket,
+    Prefix: prefix,
+    ContinuationToken: continuationToken,
+    Delimiter: '/',
+  };
+  logger.debug(`Params for listS3Keys bucket: ${bucket}, prefix: ${prefix}`);
+  const archivedMessages = [];
+
+  /* eslint-disable no-await-in-loop */
+  do {
```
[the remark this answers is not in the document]
markdboyd: We should also add a unit test for this function to prove that even if one message errors (because it's not a JSON file), the other messages are returned

example/spec/parallel/testAPI/sqsRuleSpec.js
```diff
@@ -222,7 +224,7 @@ describe('The SQS rule', () => {
     });
 
     afterAll(async () => {
-      const key = getS3KeyForArchivedMessage(config.stackName, messageId);
+      const key = getS3KeyForArchivedMessage(config.stackName, messageId, queueName);
```
markdboyd, line 227: It seems like we could assign `key = getS3KeyForArchivedMessage(config.stackName, messageId, queueName);` in the `beforeAll` and not have to call this function twice

packages/api-client/src/replayArchivedS3Messages.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @param {Object} params              - params
+ * @param {string} params.prefix       - the prefix configured for the stack
+ * @param {string} params.queueName    - name of queue to queue archived messages
+ * @param {Function} params.callback   - async function to invoke the api lambda
+ *                                       that takes a prefix / user payload.  Defaults
+ *                                       to cumulusApiClient.invokeApi
+ * @returns {Promise<Object>}          - promise that resolves to the output
+ *                                       of the API lambda
+ */
+export const postReplayArchivedMessages = async (params: {
```
markdboyd, line 16:
```suggestion
export const replaySqsMessages = async (params: {
```

I would also move it to the `replays` file in here depending on if we end up moving the endpoint

example/spec/parallel/sqsReplay/archivedS3MessagesReplaySpec.js
```diff
@@ -0,0 +1,103 @@
...
+    }).promise();
+  });
+
+  it('starts an AsyncOperation and returns an AsyncOperation ID when a valid SQS replay request is made', async () => {
+    if (beforeAllFailed) fail('beforeAll() failed');
+    const apiRequestBody = {
+      type: 'sqs',
+      queueName: queueName,
+    };
+    const response = await invokeApi({
```
markdboyd, line 70: Should use the `postReplayArchivedMessages` helper from the `api-client`

example/spec/parallel/sqsReplay/archivedS3MessagesReplaySpec.js
```diff
@@ -0,0 +1,103 @@
...
+      const retrievedMessage = await receiveSQSMessages(queueUrl, sqsOptions);
+      const key = getS3KeyForArchivedMessage(stackName, sqsMessage.MessageId, queueName);
+
+      await s3PutObject({
+        Bucket: config.bucket,
+        Key: key,
+        Body: retrievedMessage[0].Body,
+      });
+    } catch (error) {
+      beforeAllFailed = true;
```
markdboyd, line 53:
```suggestion
      beforeAllFailed = error;
```

example/spec/parallel/sqsReplay/archivedS3MessagesReplaySpec.js
```diff
@@ -0,0 +1,103 @@
...
+const { receiveSQSMessages, sendSQSMessage } = require('@cumulus/aws-client/SQS');
+const { getS3KeyForArchivedMessage } = require('@cumulus/ingest/sqs');
+const { s3PutObject } = require('@cumulus/aws-client/S3');
+const {
+  createTimestampedTestId,
+  loadConfig,
+} = require('../../helpers/testUtils');
+
+let asyncOperationId;
+let beforeAllFailed = false;
```
markdboyd, line 17:
```suggestion
let beforeAllFailed;
```

example/spec/parallel/sqsReplay/archivedS3MessagesReplaySpec.js
```diff
@@ -0,0 +1,103 @@
...
+        body: JSON.stringify(apiRequestBody),
+      },
+      expectedStatusCode: 202,
+    });
+    asyncOperationId = JSON.parse(response.body).asyncOperationId;
+    expect(asyncOperationId).toBeDefined();
+  });
+
+  it('updates the async operation results with a list of replayed messages', async () => {
+    if (beforeAllFailed) fail('beforeAll() failed');
```
markdboyd, line 88:
```suggestion
    if (beforeAllFailed) fail(beforeAllFailed);
```

Failing with the explicit error from beforeAll makes it more obvious in the logs what happened

example/spec/parallel/sqsReplay/archivedS3MessagesReplaySpec.js
```diff
@@ -0,0 +1,103 @@
...
+    await delay(100 * 1000);
```
markdboyd, line 89: We shouldn't be using arbitrary delays in our tests because they could always be unreliable. Are we just waiting for the async operation to complete?

If so, let's use the `waitForAsyncOperationStatus` helper from the `integration-tests` package

example/spec/parallel/sqsReplay/archivedS3MessagesReplaySpec.js
```diff
@@ -0,0 +1,103 @@
...
+let beforeAllFailed = false;
+let config;
+let queueName;
+let queueUrl;
+let stackName;
+let testName;
+
+// The test setup entails creating SQS messages that will be archived in S3
+describe('The replay archived S3 messages API endpoint', () => {
+  const invalidMessage = JSON.stringify({ testdata: randomString() });
```
markdboyd, line 26: Is this message actually invalid somehow? If not, we should update the variable name

markdboyd (changes requested):
Great work putting this together! I have some feedback both on overall approach and lower level details

lambdas/db-migration/src/migrations/20201014105030_create_async_operations_table.ts
```diff
@@ -16,6 +16,7 @@ export const up = async (knex: Knex): Promise<void> =>
     table
       .enum('operation_type',
         [
+          'Archived S3 Messages Replay',
```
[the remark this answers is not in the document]
botanical: Ah okay, that makes a lot of sense. Thanks for explaining! And yes, I agree that `S3 messages` is an implementation detail and `SQS replay` is a more suitable name

packages/api/tests/lambdas/test-replay-archived-s3-messages.js
```diff
@@ -0,0 +1,79 @@
...
+const { s3PutObject } = require('@cumulus/aws-client/S3');
+const { getS3KeyForArchivedMessage } = require('@cumulus/ingest/sqs');
+
+const { randomString } = require('@cumulus/common/test-utils');
+
+const {
+  replayArchivedMessages,
+} = require('../../lambdas/replay-archived-s3-messages');
+
+// Copied from @cumulus/api/lib/testUtils to avoid circular dependency
```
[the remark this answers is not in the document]
botanical: Ah yeah this might just be a bad copy paste on my behalf.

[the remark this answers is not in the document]
botanical: [`b207047` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/b207047f571ed752a1c8e221e82cea6dec114154)

packages/api/tests/lambdas/test-replay-archived-s3-messages.js
```diff
@@ -0,0 +1,79 @@
...
+    Key: key1,
+    Body: message1.Body,
+  });
+});
+
+test.after(async (t) => {
+  await S3.recursivelyDeleteS3Bucket(t.context.system_bucket);
+});
+
+test('replayArchivedMessages queues messages to SQS for each archived message', async (t) => {
```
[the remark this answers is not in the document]
botanical: [`bc10940` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/bc10940d1fcc1dd9c7f46a62254c482e2cd1fe3d)

packages/aws-client/src/SQS.ts
```diff
@@ -165,7 +167,7 @@ export const deleteSQSMessage = improveStackTrace(
  *                               if the queue exists
  */
 export const sqsQueueExists = async (queue: string) => {
-  const QueueName = queue.split('/').pop();
+  const QueueName = getQueueNameFromUrl(queue);
```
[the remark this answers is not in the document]
botanical: [`7bbb675` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/7bbb675c711bede824c874fdfc0562ca93afcb71)

packages/api/endpoints/replayArchivedS3Messages.js
```diff
@@ -0,0 +1,50 @@
...
+  const systemBucket = process.env.system_bucket;
+
+  const payload = req.body;
+  logger.debug(`Payload is ${JSON.stringify(payload)}`);
+
+  if (!payload.type) {
+    return res.boom.badRequest('replay type is required');
+  }
+
+  if (payload.type === 'sqs' && !payload.queueName) {
```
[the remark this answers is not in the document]
nemreid: I'm cool with doing whatever, as long we document it as a breaking change in the changelog and update associated user docs.

Note that `payload.type` in the current replays endpoint logic becomes pretty redundant if we move the endpoint to a specific `/kinesis` sub-path.

example/spec/parallel/sqsReplay/archivedS3MessagesReplaySpec.js
botanical: Is it indeed not invalid, nice catch. I will update it

example/spec/parallel/testAPI/sqsRuleSpec.js
botanical: [`9cdb9d2` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/9cdb9d2ca4368869d4f4c37ef224b06788f074cf)

packages/api/lambdas/replay-archived-s3-messages.js
```diff
@@ -0,0 +1,79 @@
...
+      try {
+        await sendSQSMessage(queueUrl, message);
+        logger.info(`Successfully sent message to queue URL ${queueUrl}`);
+        replayedMessages.push(message);
+      } catch (error) {
+        logger.error(`Could not send message to queue. Message: ${JSON.stringify(message)}, Queue: ${queueUrl}`, error);
+        throw error;
+      }
+    },
+    {
```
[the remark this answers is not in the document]
botanical: [`1cd1adc` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/1cd1adc0f360abf13607d262680fdb73d1a0d6a1)

packages/api/lambdas/replay-archived-s3-messages.js
```diff
@@ -0,0 +1,79 @@
...
+const { getJsonS3Object } = require('@cumulus/aws-client/S3');
+const { s3 } = require('@cumulus/aws-client/services');
+const { getQueueUrlByName, sendSQSMessage } = require('@cumulus/aws-client/SQS');
+const { getS3PrefixForArchivedMessage } = require('@cumulus/ingest/sqs');
+const Logger = require('@cumulus/logger');
+
+const logger = new Logger({ sender: '@cumulus/replay-archived-messages' });
+
+// Get messages from S3 using queueName
+const getArchivedMessagesFromQueue = async (queueName) => {
```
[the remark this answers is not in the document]
botanical: [`b95a2ba` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/b95a2ba6643b9ae21ebbbdaa402cebc445612456)

packages/api/lambdas/replay-archived-s3-messages.js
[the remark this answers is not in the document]
botanical: [`1cd1adc` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/1cd1adc0f360abf13607d262680fdb73d1a0d6a1)

tf-modules/archive/replay_archived_s3_messages.tf
```diff
@@ -0,0 +1,97 @@
...
+  timeout          = 300
+  memory_size      = 512
+
+  environment {
+    variables = {
+      system_bucket   = var.system_bucket
+      stackName       = var.prefix
+      dbHeartBeat     = var.rds_connection_heartbeat
+      databaseCredentialSecretArn    = var.rds_user_access_secret_arn
+      RDS_DEPLOYMENT_CUMULUS_VERSION = "9.0.0"
```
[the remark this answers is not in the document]
botanical: [`9324e86` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/9324e86418ad8ee86ce04308c48994c0bcc727bd)

packages/ingest/src/sqs.ts
```diff
@@ -29,7 +35,13 @@ export async function archiveSqsMessageToS3(queueUrl:string, message: SQSMessage
     throw error;
   }
 
-  const key = getS3KeyForArchivedMessage(stackName, message.MessageId);
+  const queueName = getQueueNameFromUrl(queueUrl);
+
+  if (!queueName) {
```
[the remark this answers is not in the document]
botanical: [`40be14d` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/40be14dec998b1b36b83a9dcd251013b21b16e60)

packages/ingest/src/sqs.ts
```diff
@@ -49,12 +61,19 @@ export async function archiveSqsMessageToS3(queueUrl:string, message: SQSMessage
...
  * @returns {undefined}
  */
-export async function deleteArchivedMessageFromS3(messageId: string) {
+export async function deleteArchivedMessageFromS3(messageId: string, queueUrl: string) {
   const bucket = envUtils.getRequiredEnvVar('system_bucket', process.env);
   const stackName = envUtils.getRequiredEnvVar('stackName', process.env);
-  const key = getS3KeyForArchivedMessage(stackName, messageId);
+  const queueName = getQueueNameFromUrl(queueUrl);
+
+  if (!queueName) {
```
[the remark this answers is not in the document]
botanical: [`40be14d` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/40be14dec998b1b36b83a9dcd251013b21b16e60)

packages/ingest/src/sqs.ts
```diff
@@ -49,12 +61,19 @@ export async function archiveSqsMessageToS3(queueUrl:string, message: SQSMessage
  * Deletes archived SQS Message from S3
  *
  * @param {Object} messageId - SQS message ID
+ * @param {Object} queueUrl  - SQS queue URL
```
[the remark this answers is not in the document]
botanical: [`b006086` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/b0060868f380fbbdde6041263472cec0e111b88a)

packages/ingest/test/test-sqs.js
```diff
@@ -63,9 +65,9 @@ test.serial('archiveSqsMessageToS3 archives an SQS message', async (t) => {
     queues.queueUrl,
     { numOfMessages: 1, visibilityTimeout: 5 }
   );
-  const key = getS3KeyForArchivedMessage(process.env.stackName, message.MessageId);
+  const key = getS3KeyForArchivedMessage(stackName, message.MessageId, queues.queueName);
 
-  await Promise.all(messages.map((m) => archiveSqsMessageToS3(queueUrl, m)));
+  await Promise.all(messages.map((m) => archiveSqsMessageToS3(queues.queueUrl, m)));
```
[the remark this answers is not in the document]
botanical: Almost, [`b006086` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/b0060868f380fbbdde6041263472cec0e111b88a)

example/spec/parallel/sqsReplay/archivedS3MessagesReplaySpec.js
botanical: [`56e8350` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/56e8350de448f43d732221f1fcdc4704bda19576)
thank you, this was a good call out

example/spec/parallel/sqsReplay/archivedS3MessagesReplaySpec.js
botanical: [`fb46578` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/fb46578669e6a77074682bc3a76421fcaf580135)

example/spec/parallel/sqsReplay/archivedS3MessagesReplaySpec.js
botanical: [`fb46578` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/fb46578669e6a77074682bc3a76421fcaf580135)

packages/api/lambdas/replay-archived-s3-messages.js
```diff
@@ -0,0 +1,79 @@
...
+  /* eslint-enable no-await-in-loop */
+  return archivedMessages;
+};
+
+async function replayArchivedMessages(event) {
+  const replayedMessages = [];
+  const queueName = event.queueName;
+  const queueUrl = await getQueueUrlByName(queueName);
+  const messagesToReplay = await getArchivedMessagesFromQueue(event.queueName);
+  await pMap(
```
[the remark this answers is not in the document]
botanical: Hmm I don't quite see how that works. I tried to implement it as you suggested but perhaps my implementation is wrong 🤔

packages/api/lambdas/replay-archived-s3-messages.js
```diff
@@ -0,0 +1,79 @@
...
+    messagesToReplay,
+    async (message) => {
+      try {
+        await sendSQSMessage(queueUrl, message);
```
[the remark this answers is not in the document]
botanical: I made the implementation change here https://github.com/nasa/cumulus/pull/2396  but my main concern is that the archived messages won't conform to the type defined here https://github.com/nasa/cumulus/pull/2396/files#diff-56edb8ddc2b0588b9a8b2f0bb5f33759f187a5bd46e30725d6e79cbe6ed46056R185-R188

```
type sendMessageBatchRequestEntry = {
  Id: string,
  MessageBody: string,
};
```
[type SendMessageBatchRequestEntry](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessageBatchRequestEntry.html)

compared to [type Message](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_Message.html)
which has a `Body` and `MessageId`

CHANGELOG.md
```diff
@@ -6,29 +6,23 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
-### Notable changes
```
markdboyd, line 9: This looks like a bad merge, these CHANGELOG entries should not be deleted

lambdas/db-migration/src/migrations/20210727104740_alter_async_operations_table_operation_type.ts
```diff
@@ -0,0 +1,47 @@
...
+    'Reconciliation Report',
+    'Migration Count Report',
+    'Data Migration',
+    'SQS Replay',
+  ]));
+};
+
+export const down = async (knex: Knex): Promise<void> => {
+  await knex.raw(
+    formatAlterTableEnumSql('myTableName', 'myColumnName', [
```
markdboyd, line 35:
```suggestion
    formatAlterTableEnumSql('async_operations', 'operation_type', [
```

packages/api/lambdas/replay-sqs-messages.js
```diff
@@ -0,0 +1,77 @@
...
+      return sqsMessage;
+    }
+  ));
+  archivedMessages.map((message) => {
+    if (message.status === 'fulfilled') {
+      validMessages.push(message.value);
+    }
+    return validMessages;
+  });
+  return validMessages;
```
markdboyd, line 42:
```suggestion
  const validMessages = archivedMessages
    .filter((message) => message.status === 'fulfilled')
    .map((message) => message.value);
  return validMessages;
```

I might change the variable name of `archivedMessage` to `archiveRequests` and `validMessages` to `archivedMessages` for clarity

packages/api/tests/endpoints/test-replays.js
```diff
@@ -19,6 +21,30 @@ const AccessToken = require('../../models/access-tokens');
...
 let jwtAuthToken;
 
+/**
+ * create a source queue
+ *
+ * @param {string} queueNamePrefix - prefix of the queue name
+ * @param {string} visibilityTimeout - visibility timeout for queue messages
+ * @returns {Object} - {queueUrl: <url>} queue created
+ */
+async function createSqsQueues(
```
markdboyd, line 31: Nit: can't we use this helper from `lib/testUtils`?

packages/api/lambdas/replay-sqs-messages.js
```diff
@@ -0,0 +1,77 @@
...
+const getArchivedMessagesFromQueue = async (queueName) => {
+  let continuationToken;
+  const validMessages = [];
+  const bucket = envUtils.getRequiredEnvVar('system_bucket', process.env);
+  const stackName = envUtils.getRequiredEnvVar('stackName', process.env);
+  const prefix = getS3PrefixForArchivedMessage(stackName, queueName);
+  const params = {
+    Bucket: bucket,
+    Prefix: prefix,
+    ContinuationToken: continuationToken,
```
markdboyd, line 23: I don't think you need this parameter anymore

packages/api/tests/endpoints/test-replays.js
```diff
@@ -170,3 +195,75 @@ test.serial('request to /replays endpoint returns 503 if starting ECS task throw
...
+  try {
+    const response = await request(app)
+      .post('/replays/sqs')
+      .set('Accept', 'application/json')
+      .set('Authorization', `Bearer ${jwtAuthToken}`)
+      .send(body)
+      .expect(202);
+
+    // expect a returned async operation ID
+    t.truthy(response.body.asyncOperationId);
```
markdboyd, line 218: Rather than testing for truthiness we should test against the value that your `asyncOperationStartStub` is returning

packages/api/tests/endpoints/test-replays.js
```diff
@@ -170,3 +195,75 @@ test.serial('request to /replays endpoint returns 503 if starting ECS task throw
     asyncOperationStartStub.restore();
   }
 });
+
+test.serial('POST /replays/sqs starts an async-operation with specified payload', async (t) => {
+  const { queues } = t.context;
+  const asyncOperationStartStub = sinon.stub(asyncOperations, 'startAsyncOperation').returns(
+    new Promise((resolve) => resolve({ id: randomId('asyncOperationId') }))
+  );
+  asyncOperationStartStub.resetHistory();
```
markdboyd, line 204: We don't need to call `resetHistory` in each test if each stub is specific to that test, which seems to be the case for all these tests

packages/api/endpoints/replays.js
```diff
@@ -43,6 +47,39 @@ async function startKinesisReplayAsyncOperation(req, res) {
...
+  const payload = req.body;
+  logger.debug(`Payload is ${JSON.stringify(payload)}`);
+
+  if (!payload.queueName) {
+    return res.boom.badRequest('queueName is required for SQS messages replay');
+  }
+
+  const queueUrl = await getQueueUrlByName(payload.queueName);
+  if (!(await sqsQueueExists(queueUrl))) {
+    logger.error(`Could not find queue ${queueUrl}. Unable to process message.`);
```
markdboyd, line 64:
```suggestion
    return res.boom.badRequest(`Could not find queue ${queueUrl}. Unable to process message.`);
```

We should just fail the request as bad at this point

packages/api/lambdas/replay-sqs-messages.js
```diff
@@ -0,0 +1,77 @@
...
+const { getJsonS3Object, listS3ObjectsV2 } = require('@cumulus/aws-client/S3');
+const { getQueueUrlByName, sendSQSMessage } = require('@cumulus/aws-client/SQS');
+const { getS3PrefixForArchivedMessage } = require('@cumulus/ingest/sqs');
+const Logger = require('@cumulus/logger');
+
+const logger = new Logger({ sender: '@cumulus/replay-sqs-messages' });
+
+// Get messages from S3 using queueName
+const getArchivedMessagesFromQueue = async (queueName) => {
+  let continuationToken;
```
markdboyd, line 15:
```suggestion
```

packages/api/tests/lambdas/test-replay-sqs-messages.js
```diff
@@ -0,0 +1,102 @@
...
+  const {
+    numberOfMessagesAvailable,
+    numberOfMessagesNotVisible,
+  } = await getSqsQueueMessageCounts(queueUrl);
+  t.is(numberOfMessagesAvailable, 1);
+  t.is(numberOfMessagesNotVisible, 0);
+  t.deepEqual(replay, expected);
+});
+
+test('replaySqsMessages queues only valid messages to SQS', async (t) => {
```
markdboyd, line 67: I think this is a better test of `getArchivedMessagesFromQueue`, because I think you're proving that a non-JSON message won't get queued. But the logic that does the JSON parsing of the S3 object is in `getArchivedMessagesFromQueue`. So if this test were failing, that's where the fix should be.

packages/api/tests/lambdas/test-sqs-message-consumer.js
```diff
@@ -82,6 +81,9 @@ async function cleanupRulesAndQueues(rules, queues) {
     rules.map((rule) => rulesModel.delete(rule))
   );
 
+  // Delete queueName for each object in list
+  queues.forEach((q) => delete q.queueName);
```
markdboyd, line 85: Why is this necessary?

packages/types/api/replaySqsMessages.d.ts
```diff
@@ -0,0 +1,4 @@
+export interface ReplaySqsMessagesPayload {
+  queueName: string,
+  type?: string,
```
markdboyd, line 3: This isn't part of the payload anymore

tf-modules/archive/replay_sqs_messages.tf
```diff
@@ -0,0 +1,95 @@
...
+  handler          = "index.handler"
+  runtime          = "nodejs12.x"
+  timeout          = 300
+  memory_size      = 512
+
+  environment {
+    variables = {
+      system_bucket                = var.system_bucket
+      stackName                    = var.prefix
+      databaseCredentialSecretArn  = var.rds_user_access_secret_arn
```
markdboyd, line 73: Does this lambda talk to RDS directly? If not, then this environment variable isn't necessary

tf-modules/archive/replay_sqs_messages.tf
```diff
@@ -0,0 +1,95 @@
...
+      "logs:PutLogEvents"
+    ]
+    resources = ["*"]
+  }
+  statement {
+    actions = [
+      "secretsmanager:GetSecretValue"
+    ]
+    resources = [var.rds_user_access_secret_arn]
+  }
```
markdboyd, line 35: This is only necessary if the Lambda is talking to RDS directly, which I don't believe it is

CHANGELOG.md
botanical: [`10b7d34` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/10b7d34b479562248e4d6c138680fcb327939b6a)

lambdas/db-migration/src/migrations/20210727104740_alter_async_operations_table_operation_type.ts
botanical: [`7c2012f` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/7c2012f0247380cca6631c5a8d769d853d978d72)

packages/api/tests/lambdas/test-sqs-message-consumer.js
botanical: This is necessary because `createSqsQueues` returns `queueName` and, with the code that follows, if not removed, will throw an error because it won't be able to find a queue of `queueName` to delete:
```
  const queueUrls = queues.reduce(
    (accumulator, currentValue) => accumulator.concat(Object.values(currentValue)), []
  );

  await Promise.all(
    queueUrls.map((queueUrl) => SQS.deleteQueue(queueUrl))
  );
``

File: packages/api/tests/endpoints/test-replays.js (hunk 27 above)
Comment (botanical) [reply packages/api/tests/endpoints/test-replays.js]: [`030ed97` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/030ed97f35a8892454a2f46e229a87b2bb447203#diff-df2af8d19c0a84c9334f9bf37ffe602f1e7af349aee3c218576a9d3c6cb4590eR17)

File: packages/api/lambdas/replay-sqs-messages.js (hunk 28 above)
Comment (botanical) [reply packages/api/lambdas/replay-sqs-messages.js]: [`030ed97` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/030ed97f35a8892454a2f46e229a87b2bb447203#diff-50b67b19453ae7cd068a961f63d9775c6e52caadcad45e3cc0f33552900d8e3dL15)

File: packages/api/lambdas/replay-sqs-messages.js (hunk 32 above)
Comment (botanical) [reply packages/api/lambdas/replay-sqs-messages.js]: [`030ed97` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/030ed97f35a8892454a2f46e229a87b2bb447203#diff-50b67b19453ae7cd068a961f63d9775c6e52caadcad45e3cc0f33552900d8e3dL15)

File: packages/api/tests/endpoints/test-replays.js (hunk 29 above)
Comment (botanical) [reply packages/api/tests/endpoints/test-replays.js]: [`030ed97` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/030ed97f35a8892454a2f46e229a87b2bb447203#diff-df2af8d19c0a84c9334f9bf37ffe602f1e7af349aee3c218576a9d3c6cb4590eL218)

File: packages/api/tests/endpoints/test-replays.js (hunk 30 above)
Comment (botanical) [reply packages/api/tests/endpoints/test-replays.js]: Oh, okay yeah that makes sense [`030ed97` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/030ed97f35a8892454a2f46e229a87b2bb447203#diff-df2af8d19c0a84c9334f9bf37ffe602f1e7af349aee3c218576a9d3c6cb4590eL239)

File: packages/api/endpoints/replays.js (hunk 31 above)
Comment (botanical) [reply packages/api/endpoints/replays.js]: This is how I ended up changing it :[`fe6605f` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/fe6605f1129d0dfe2524419e4a4e2738c2d4abe8)
that point of the code was never reached in the case of a non existent queue since getQueueUrlByName calls `sqs().getQueueUrl()` which will actually do the throwing

File: packages/api/tests/lambdas/test-replay-sqs-messages.js (hunk 33 above)
Comment (botanical) [reply packages/api/tests/lambdas/test-replay-sqs-messages.js]: Good point [`030ed97` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/030ed97f35a8892454a2f46e229a87b2bb447203#diff-6d4399c4bcca5a24555c1881c0ceb23928bb79136abba53f19f2fc254b0d976eL67)

File: packages/types/api/replaySqsMessages.d.ts (hunk 35 above)
Comment (botanical) [reply packages/types/api/replaySqsMessages.d.ts]: [`030ed97` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/030ed97f35a8892454a2f46e229a87b2bb447203#diff-72236dd0282583674134919c52d1834ec33487006430075c7ec21c9908b37182L3)

File: tf-modules/archive/replay_sqs_messages.tf (hunk 37 above)
Comment (botanical) [reply tf-modules/archive/replay_sqs_messages.tf]: [`030ed97` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/030ed97f35a8892454a2f46e229a87b2bb447203#diff-7b93858158e237529239ab394d5f5c6428fd1d94d04af74213f3bad9f3606ea5L30)

File: tf-modules/archive/replay_sqs_messages.tf (hunk 36 above)
Comment (botanical) [reply tf-modules/archive/replay_sqs_messages.tf]: [`030ed97` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/030ed97f35a8892454a2f46e229a87b2bb447203#diff-7b93858158e237529239ab394d5f5c6428fd1d94d04af74213f3bad9f3606ea5L73)

File: packages/api/lambdas/replay-sqs-messages.js (hunk 26 above)
Comment (botanical) [reply packages/api/lambdas/replay-sqs-messages.js]: Ooh I didn't realize I could use map like how you've suggested! [`9cc6812` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/9cc6812cfe17335ca582246687deaed58fe6a3f0)

File: packages/api/tests/lambdas/test-replay-sqs-messages.js (hunk 33 above)
Comment (markdboyd) [reply packages/api/tests/lambdas/test-replay-sqs-messages.js]: So you updated the test name but you didn't really change the test logic. We should update the test so that it calls and is testing `getArchivedMessagesFromQueue`  directly, not via `replaySqsMessages`

File: CHANGELOG.md [hunk 38]
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
+### Notable changes
```
markdboyd, line 8: Huh. This is weird. This note should be in the CHANGELOG on `master`, but it is not. I suspect it got removed somehow by a bad merge on someone else's PR. Regardless, thanks for re-adding it

packages/api/tests/lambdas/test-sqs-message-consumer.js
markdboyd: OK. I'm still not sure if the code is working even if you delete `q.queueName`. `deleteQueue` expects a queue URL as a single argument, whereas right now, I think it might be receiving a object where `queueUrl` is a property. 

Regardless I think we can get rid of the `forEach` and `reduce` by doing this:

```
  await Promise.all(
    queues.map((queue) => SQS.deleteQueue(queue.queueUrl))
  );
```

The `reduce` was just trying to prepare an array of just queue URLs, but that's totally unnecessary since we can read the object property for `queueUrl` and pass it in to `deleteQueue`

botanical: `queueUrls` is creating an array of the `queueUrl`s and `deadLetterQueueUrl`s, so ` queueUrls.map((queueUrl) => SQS.deleteQueue(queueUrl))` is correctly deleting each queue with the current implementation

packages/api/tests/lambdas/test-replay-sqs-messages.js
botanical: I think I misunderstood. Should be updated now [`442ed82` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/442ed821f55ec235fa6fac74264716f11c966844)

packages/api/tests/lambdas/test-sqs-message-consumer.js
markdboyd: I see. It's because we want to delete both the dead letter queue and regular queue. In that case I still think this code is unnecessarily confusing and would prefer something like:

```
  await Promise.all(
    queues.map(async (queue) => {
      await SQS.deleteQueue(queue.queueUrl);
      await SQS.deleteQueue(queue.deadLetterQueueUrl);
    })
  );
```

which is a lot easier to understand and removes the need for the `forEach` or `reduce`

markdboyd (approved):

packages/api/tests/lambdas/test-sqs-message-consumer.js
botanical: [`c9a4bfe` (#2373)](https://github.com/nasa/cumulus/pull/2373/commits/c9a4bfe0055910a944af43fff5bf4051bb5857c9)

PR #2384, opened 2021-07-22
Update integration tests

botanical:

## Changes

* This PR updates the beforeFailed of an integration test as well as updates the logic to `shouldCatchPolicyError` in `snsRuleSpec` to account for the chance that there are other policies attached to a stack's messageConsumer lambda

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [x] Integration tests


example/spec/parallel/testAPI/snsRuleSpec.js
```diff
@@ -253,19 +276,20 @@ describe('The SNS-type rule', () => {
...
     });
 
     it('adds the new policy and subscription', async () => {
+      if (beforeAllFailed) fail(beforeAllFailed);
       const { Policy } = await lambda().getPolicy({ FunctionName: consumerName }).promise();
       const { Statement } = JSON.parse(Policy);
-      expect(Statement.length).toEqual(1);
-      expect(Statement[0].Sid).toEqual(`${ruleName}Permission`);
-      expect(await getNumberOfTopicSubscriptions(newTopicArn)).toBe(1);
+      expect(Statement.some((s) => s.Sid === expectedStatementId));
```
markdboyd, line 292:
```suggestion
      expect(Statement.some((s) => s.Sid === expectedStatementId)).toBeTrue();
```

I think we want to make an assertion here

packages/api/models/rules.js
```diff
@@ -478,6 +478,7 @@ class Rule extends Manager {
   async deleteSnsTrigger(item) {
     // If event source mapping is shared by other rules, don't delete it
     if (await this.isEventSourceMappingShared(item, 'arn')) {
+      log.info('This rule shares an event source mapping, it will not be deleted.');
```
markdboyd, line 481:
```suggestion
      log.info(`Event source mapping ${arn} is shared by multiple rules, so it will not be deleted.`);
```

example/spec/parallel/testAPI/snsRuleSpec.js
botanical: Ah, yes.

markdboyd (approved):

PR #2390, opened 2021-07-27
[CUMULUS-2626] Fix Execution lookup in PDR migration

npauzenga:
**Summary:** 

https://bugs.earthdata.nasa.gov/browse/CUMULUS-2626

The issue: A Dynamo PDR's `execution` field is a full URL, not just an ARN. A Postgres Execution's `arn` is just an ARN, not a full URL. Attempting to compare the two doesn't work. This was causing the PG Execution lookup to fail (flagged below).

Executions have a `url` field that _does_ represent the full URL and is unique across all records.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


lambdas/data-migration2/src/pdrs.ts
```diff
@@ -69,39 +69,39 @@ export const migratePdrRecord = async (
...
-    { name: dynamoRecord.provider }
+    { name: dynamoPDR.provider }
   );
 
-  const executionCumulusId = dynamoRecord.execution
+  const executionCumulusId = dynamoPDR.execution
     ? await executionPgModel.getRecordCumulusId(
       knex,
-      { arn: dynamoRecord.execution }
+      { url: dynamoPDR.execution }
```
npauzenga, line 78: This is the actual fix.

lambdas/data-migration2/src/pdrs.ts
```diff
@@ -28,18 +28,18 @@ const { deconstructCollectionId } = require('@cumulus/api/lib/utils');
 /**
  * Migrate PDR record from Dynamo to RDS.
  *
- * @param {AWS.DynamoDB.DocumentClient.AttributeMap} dynamoRecord
- *   Record from DynamoDB
+ * @param {AWS.DynamoDB.DocumentClient.AttributeMap} dynamoPDR
```
npauzenga, line 31: Just trying to be more explicit in the naming.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- **CUMULUS-2626**
+  - Update [PDR migration](https://github.com/nasa/cumulus/blob/master/lambdas/data-migration2/src/pdrs.ts) to correctly find Executions by a PDR's `execution` field
```
markdboyd, line 12:
```suggestion
  - Update [PDR migration](https://github.com/nasa/cumulus/blob/master/lambdas/data-migration2/src/pdrs.ts) to correctly find Executions by a Dynamo PDR's `execution` field
```

🐜

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -143,7 +151,7 @@ test.serial('migratePdrRecord correctly migrates PDR record', async (t) => {
   const testPdr = generateTestPdr({
     collectionId: buildCollectionId(testCollection.name, testCollection.version),
     provider: testProvider.name,
-    execution: execution.arn,
+    execution: executionURL,
```
markdboyd, line 154: So it was passing before because we were creating the dynamo record's `execution` field using the `arn` from the created PG record. Thus the lookup for an `arn` using the dynamo `execution` worked, even though in reality it never would. Just making sure I understand.

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -132,7 +132,15 @@ test.serial('migratePdrRecord correctly migrates PDR record', async (t) => {
     testProvider,
   } = t.context;
   const executionPgModel = new ExecutionPgModel();
-  const execution = fakeExecutionRecordFactory();
+  const executionArn = 'arn:aws:states:us-east-1:12345:execution:test-IngestGranule:587f47b1-88b4-4f74-bd51-525085943fd3';
+  const executionURL = `https://console.aws.amazon.com/states/home?region=us-east-1#/executions/details/${executionArn}`;
```
markdboyd, line 136: Not required, but we do have a `getExecutionUrlFromArn` helper in `packages/message/Executions` which might ensure some greater test stability

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -132,7 +132,15 @@ test.serial('migratePdrRecord correctly migrates PDR record', async (t) => {
...
+
+  const execution = fakeExecutionRecordFactory(
```
markdboyd, line 138: It's not required, however, to avoid hopefully making the same mistakes in future unit tests, I think we should update `fakeExecutionRecordFactory` to produce a realistic value for the `arn` and `url` properties.

markdboyd (changes requested):
Great work! I have one pretty minor change request and a couple suggestions

lambdas/data-migration2/src/pdrs.ts
markdboyd: The only other thought I have here is that the PDR migration will fail if there is a `dynamoPdr.execution` but we can't find a matching record. That is perfectly logical, but I notice in the granules migration we have updated the logic to not fail granules migration if the execution cannot be found

Since execution isn't a required field for PDRs, we could do the same thing here. It's not a bug and not something we have to change, just thinking out loud. Maybe it's easier to limit the scope here for now and leave it as is

lambdas/data-migration2/tests/test-pdrs.js
npauzenga: Right. We took the `arn` that `fakeExecutionRecordFactory` produced and basically searched for executions by `pgExecution.arn === pgExecution.arn`, which works fine in test.  

I think the misunderstanding was that a PDR's `execution` field is NOT an ARN. It's a full URL. So comparing that to the Execution's `arn` field is a problem.

lambdas/data-migration2/src/pdrs.ts
npauzenga: I could do this in a follow-up PR. It seems worthwhile, even if it's just to match behavior in other migrations. But yeah, it might be best to not hold this up.

lambdas/data-migration2/tests/test-pdrs.js
npauzenga: Yup, good call: https://github.com/nasa/cumulus/pull/2390/commits/9d773f64b4d1d2aadb781e9b94759c70c67f3153

lambdas/data-migration2/tests/test-pdrs.js
npauzenga: Yeah, updating the factories when something like this comes up is a great idea. https://github.com/nasa/cumulus/pull/2390/commits/9d773f64b4d1d2aadb781e9b94759c70c67f3153

lambdas/data-migration2/tests/test-pdrs.js
```diff
@@ -25,6 +25,7 @@ const {
 } = require('@cumulus/aws-client/S3');
 const { dynamodbDocClient } = require('@cumulus/aws-client/services');
 const { RecordAlreadyMigrated, PostgresUpdateFailed } = require('@cumulus/errors');
+const { getExecutionUrlFromArn } = require('@cumulus/message/Executions');
```
markdboyd, line 28: We have it imported here but I'm not sure if we need it here since the factory is handling this for us now? Regardless linting will fail if it's not needed here

lambdas/data-migration2/package.json
```diff
@@ -39,6 +39,7 @@
     "pg": "^8.3.0"
   },
   "devDependencies": {
-    "@cumulus/test-data": "9.3.0"
+    "@cumulus/test-data": "9.3.0",
+    "@cumulus/message": "9.3.0"
```
markdboyd, line 43: We may not need this import since the factory code handles the integration with cumulus/message, so we don't actually use this package here

lambdas/data-migration2/tests/test-pdrs.js
npauzenga: https://github.com/nasa/cumulus/pull/2390/commits/c1426a9975c50a4d15b75941a87e91e5cf4f9a7e

lambdas/data-migration2/package.json
npauzenga: https://github.com/nasa/cumulus/pull/2390/commits/c1426a9975c50a4d15b75941a87e91e5cf4f9a7e

PR #2393, opened 2021-07-28
move kinesis replays endpoint to explicit subpath

nemreid:

## Changes

*   - Moved kinesis replay functionality from `/replays` to `/replays/kinesis`. Due to the use of explicit pathing to infer the type of the replay request, the `type` parameter is no longer required in the payload of a POST request.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/packages/api-client/src/replays.ts
+++ b/packages/api-client/src/replays.ts
@@ -29,7 +29,7 @@ export const postKinesisReplays = async (params: {
       headers: {
         'Content-Type': 'application/json',
       },
-      path: '/replays',
+      path: '/replays/kinesis',
       body: JSON.stringify(payload),
     },
     expectedStatusCode: 202,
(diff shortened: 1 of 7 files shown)


packages/api-client/src/replays.ts
```diff
@@ -29,7 +29,7 @@ export const postKinesisReplays = async (params: {
       headers: {
```
botanical, line 29: Do we still expect the payload above to be MigrationCountsPayload? I think we can create a new type and add `kinesisStream`

botanical:
Do we need to update `routes.js` as well?

botanical:
Based off our discussion, I've stopped reviewing this PR and we can continue this work at some other point

nemreid:
Closing in favor of rescope under CUMULUS-2632

PR #2395, opened 2021-07-29
Release 8.1.2

markdboyd:


botanical (changes requested):
I think you forgot to add the changes mentioned here https://github.com/nasa/cumulus/blob/master/CHANGELOG.md?plain=1#L175 👀

example/cumulus-tf/variables.tf
```diff
@@ -441,3 +441,8 @@ variable "default_tags" {
   }
 }
 
+variable "cmr_custom_host" {
```
botanical, line 444: I think we can reference this in main.tf now

lambdas/sqs-message-remover/variables.tf
```diff
@@ -23,7 +23,7 @@ variable "lambda_processing_role_arn" {
 # Optional
 
 variable "cmr_custom_host" {
-  description = "Custom host to use for CMR requests"
+  description = "Custom protocol and host to use for CMR requests (e.g. http://cmr-host.com)"
   type        = string
   default     = ""
```
botanical, line 28: default should be null, right?

tf-modules/archive/variables.tf
```diff
@@ -147,7 +147,7 @@ variable "buckets" {
 }
 
 variable "cmr_custom_host" {
-  description = "Custom host to use for CMR requests"
+  description = "Custom protocol and host to use for CMR requests (e.g. http://cmr-host.com)"
   type        = string
   default     = ""
```
botanical, line 152: Should this be null? Can we also reference this in `tf-modules/cumulus/archive.tf` now?

tf-modules/ingest/variables.tf
```diff
@@ -17,7 +17,7 @@ variable "cmr_environment" {
 }
 
 variable "cmr_custom_host" {
-  description = "Custom host to use for CMR requests"
+  description = "Custom protocol and host to use for CMR requests (e.g. http://cmr-host.com)"
   type        = string
   default     = ""
```
botanical, line 22:
```suggestion
  default     = null
```

botanical: Is there a reason to not update `ingest.tf`?

lambdas/sqs-message-remover/variables.tf
botanical: Is there a reason to not update `sqs-message-remover.tf`?

markdboyd: Hmm. That should have came in when I cherry-picked the commit. Good catch. Let me take a look.

example/cumulus-tf/variables.tf
```diff
@@ -441,3 +441,14 @@ variable "default_tags" {
...
+  description = "Custom protocol and host to use for CMR requests (e.g. http://cmr-host.com)"
+  type        = string
+  default     = null
+}
+
+variable "cmr_custom_host" {
```
botanical, line 450: I think this is a duplicate

botanical (approved):
Made one comment about a duplicate variable but other than that, looks good

PR #2397, opened 2021-08-02
Update deployment docs

markdboyd:
Deployment doc updates to go with https://github.com/nasa/cumulus-template-deploy/pull/64/files


laurenfrederick (changes requested):
I would keep the look up ECS-optimized AMI and CMA deploy steps in and put them at the bottom as deprecated with the template-deploy changes in case people deployed with an earlier version of template-deploy and need to go back to that documentation later. I know it'll be in an earlier version of the docs, but it might make it easier to have that there in case people need to go back to it.

markdboyd:
@laurenfrederick I put back on the sections on the AMI and CMA but clearly marked them as deprecated with a note linking to the template-deploy changes. Let me know what you think

laurenfrederick (approved):

PR #2400, opened 2021-08-03
udpate variable names for clarity

markdboyd:


npauzenga (approved):
Looks great @markdboyd thanks for clarifying.

One other thing we _could_ do here is rename `getDbCount` to `getESCount` (or similar).

PR #2412, opened 2021-08-12
Remove logs from ES mapping

markdboyd:
As of #2089, Cumulus no longer writes `logs` records to Elasticsearch, so there should be no need to maintain a `logs` records in our Elasticsearch mappings. It is likely that this change was just missed in the original PR.

## Changes

* Removed `logs` from `mappings.json` for Elasticsearch

## Testing

To test whether this change will have an adverse impact on Elasticsearch indexes that may still have "logs" records (from back when Cumulus still wrote these records), I have manually created some "logs" records in my CI stack's Elasticsearch index. So we can see if integration tests which query the API/Elasticsearch still pass with this change deployed

Update, 8/16 - I can confirm that even after deploying these changes the "logs" records are still discoverable via Kibana for CI stack

## PR Checklist

- [x] Update CHANGELOG
- [x] Integration tests


flamingbear (approved):
Looks good, nice clean up.

PR #2422, opened 2021-08-20
[CUMULUS-2510] Publish SNS messages for executions

botanical:

Addresses [CUMULUS-2510: Publish to reporting SNS topics from RDS](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2510)

## Changes

* This is the first part of a four part change to remove DynamoDB as the source for publishing SNS messages to various SNS topics. 
* This first part removes the Executions Table DynamoDB stream and manually publishes SNS messages to the report_executions_topic from the `sfEventSqsToDbRecords` lambda and from the executions `DELETE` endpoint
* Note that the SNS message for DELETE is just to maintain what we currently do and that changes to this will be made in https://bugs.earthdata.nasa.gov/browse/CUMULUS-2667 

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


packages/api/lambdas/sf-event-sqs-to-db-records/write-execution.js
```diff
@@ -130,6 +130,13 @@ const writeExecutionToDynamoAndES = async (params) => {
   }
 };
 
+const publishExecutionSnsMessage = async (record) => {
```
markdboyd, line 133: This function is used outside of just this lambda, so it should live in a helper file somewhere. Perhaps in `packages/api/lib`?

packages/api/lambdas/sf-event-sqs-to-db-records/write-execution.js
```diff
@@ -130,6 +130,13 @@ const writeExecutionToDynamoAndES = async (params) => {
...
+  const topicArn = process.env.execution_sns_topic_arn;
+  const message = DynamoDB.Converter.marshall(record);
```
markdboyd, line 135: So we don't need this. This function would be converting a Javascript object into a DynamoDB style record, as per https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/DynamoDB/Converter.html#marshall-property. 

However, what Cloud metrics is expecting is a regular Javascript object representation of a record. 

That's why we're calling `unwrap` in the existing `publish-executions` code: to unwrap a DynamoDB style object into a JS object:

https://github.com/nasa/cumulus/blob/master/packages/api/lambdas/publish-executions.js#L17

You can also see that this is the expected behavior from the unit tests:

https://github.com/nasa/cumulus/blob/master/packages/api/tests/lambdas/test-publish-executions.js#L41

packages/api/lambdas/sf-event-sqs-to-db-records/write-execution.js
```diff
@@ -130,6 +130,13 @@ const writeExecutionToDynamoAndES = async (params) => {
   }
 };
 
+const publishExecutionSnsMessage = async (record) => {
+  const topicArn = process.env.execution_sns_topic_arn;
```
markdboyd, line 134: We should use `@cumulus/common/env.getRequiredEnvVar` to get this env var, because it will throw a specific error if the environment variable is missing, which is more helpful than the confusing error we might see from SNS if we try to publish when this variable is undefined

packages/api/lambdas/sf-event-sqs-to-db-records/write-execution.js
botanical: That is a really good point, I will relocate it!

example/spec/parallel/testAPI/snsExecutionSpec.js
```diff
@@ -0,0 +1,52 @@
...
+  afterAll(async () => {
+    await deleteS3Object(config.bucket, executionKey);
+  });
+
+  it('for a deleted execution', async () => {
+    await waitForCompletedExecution(workflowExecutionArn);
+    await deleteExecution({ prefix: config.stackName, executionArn: workflowExecutionArn });
+
+    const executionName = workflowExecutionArn.split(':').pop();
+    executionKey = `${config.stackName}/test-output/${executionName}.output`;
```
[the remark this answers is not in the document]
markdboyd: Upon further investigation, it may be the case that we never have been sending information about deleted executions over to cloud metrics in the first place. 

Consider our publish-executions Lambda:

https://github.com/nasa/cumulus/blob/master/packages/api/lambdas/publish-executions.js#L17

It is just unmarshalling the `NewImage` and sending that in the SNS topic. `NewImage` would only exist for new or updated records.

https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_StreamRecord.html

I guess that all makes sense given that we never previously had a way to delete executions until we added that endpoint. But we would need to discuss with cloud metrics how/if to add support for this before we could do anything, since it would require a breaking change on them. We might want to make sure we have a ticket or architecture topic for it though

packages/api/lambdas/sf-event-sqs-to-db-records/write-execution.js
```diff
@@ -155,8 +158,11 @@ const writeExecution = async ({
       executionModel,
       esClient,
     });
-    return executionCumulusId;
+    await publishExecutionSnsMessage(executionApiRecord);
```
markdboyd, line 161: So this approach isn't quite the same as what we have now with the DynamoDb streams and Lambdas publishing to the SNS topic. This approach would publish the `executionApiRecord` that we're generating from the Cumulus message to the SNS topic. However, the DynamoDb stream approach publishes the record **as created/updated in DynamoDb** to the SNS topic.

Because of conditional logic in our Dynamo/PG write methods, there may be subtle differences between these two things. For example, when an execution has already been saved to the DB as `completed` and then we process a delayed event in this Lambda where the status of the execution is `running`, the `executionApiRecord` would for the delayed event would have a status of `running`, but the record **would not** get updated from `completed` to `running` in DynamoDB. So sending over the `executionApiRecord` would be inaccurate.

We may need to have an impromptu architecture meeting on the best way to handle this

markdboyd: We could possibly modify the PG upsert to return all the properties of the updated record, not just the `cumulus_id`. Then we could translate the PG record to the API structure and send that to the SNS topic. It also isolates us from dependency on DynamoDB, making it easier to remove later

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-write-execution.js
```diff
@@ -489,3 +521,24 @@ test.serial('writeExecution() correctly sets both original_payload and final_pay
...
+
+  const { Messages } = await sqs().receiveMessage({ QueueUrl, WaitTimeSeconds: 10 }).promise();
+
+  t.is(Messages.length, 1);
+
+  const snsMessage = JSON.parse(Messages[0].Body);
+  const executionRecord = JSON.parse(snsMessage.Message);
+
+  t.deepEqual(executionRecord.arn, executionArn);
+  t.deepEqual(executionRecord.status, cumulusMessage.meta.status);
```
markdboyd, line 543: You should use `t.is` when comparing specific properties unless the value of those properties are objects/arrays. 

But really I think we should do a `deepEqual` compare between the `executionRecord` and a translation of PG -> API of the record in PG, which will verify that we are sending over the exact expected record

packages/api/tests/lib/writeRecords/test-write-execution.js
```diff
@@ -514,3 +547,27 @@ test('writeExecutionRecordFromApi() saves execution to Dynamo and RDS with same
...
+  t.is(Messages.length, 1);
+
+  const snsMessage = JSON.parse(Messages[0].Body);
+  const executionRecord = JSON.parse(snsMessage.Message);
+  const executionApiRecord = generateExecutionApiRecordFromMessage(cumulusMessage, Date.now());
+  const omitFields = ['updatedAt', 'timestamp'];
+
+  t.is(executionRecord.arn, executionArn);
+  t.is(executionRecord.status, cumulusMessage.meta.status);
+  t.deepEqual(omit(executionRecord, omitFields), omit(executionApiRecord, omitFields));
```
markdboyd, line 572: We want to prove that the record that is sent to SNS represents the record **after** being written to the database, not before, whereas comparing against `executionApiRecord` is comparing to the record before it's written. Instead, we should get the record out of Dynamo that was written and compare that with what was sent to SNS. Then we also don't want to omit any fields, because we want to prove that what went to SNS was exactly what we expected

packages/db/src/models/execution.ts
```diff
@@ -13,6 +13,15 @@ class ExecutionPgModel extends BasePgModel<PostgresExecution, PostgresExecutionR
     });
   }
 
+  async create(
+    knexOrTransaction: Knex | Knex.Transaction,
+    execution: PostgresExecution
+  ) {
+    return await knexOrTransaction(this.tableName)
+      .insert(execution)
+      .returning('*');
```
markdboyd, line 22: Same sugestion from the other PR:

There's nothing wrong with this, but since we're going to need to do the same thing for every data type, I think we should update the base model create to have an optional argument for returningFields, which could be a string or array of strings. Then from this method (and for all the other data type overrides we're going to do) we could do:

`return super.get(knexOrTransaction, item, '*')`

packages/api/lib/writeRecords/write-execution.js
```diff
@@ -117,15 +119,17 @@ const _writeExecutionRecord = ({
...
+  const translatedExecution = await translatePostgresExecutionToApiExecution(response, trx);
+  logger.info(`Successfully wrote execution ${postgresRecord.arn} to PostgreSQL with cumulus_id ${response.cumulus_id}`);
   await writeExecutionToDynamoAndES({
     dynamoRecord,
     executionModel,
     updatedAt,
     esClient,
   });
-  return executionCumulusId;
+  await publishExecutionSnsMessage(translatedExecution);
```
markdboyd, line 131: I'm trying to decide if we want to do this inside the transaction, because it means that if the publishing to SNS were to throw an error for some reason, then the PG write would be rolled back (but Dynamo and ES) are not. 

I'm leaning towards not doing it in the transaction and instead calling this from `writeExecutionRecordFromMessage` after `_writeExecutionRecord`. 

Either way we go, `publishExecutionSnsMessage` doesn't get reached unless all the writes succeed, but the difference is that not having it in the transaction won't complicate the rollback of the transaction any further. But it might be good to get other team opinions on this question

CHANGELOG.md
```diff
@@ -97,7 +104,17 @@ of response and not the raw API endpoint response
   - **CUMULUS-2634**
     - Changed `sfEventSqsToDbRecords` Lambda to use new upsert helpers for executions, granules, and PDRs
     to ensure out-of-order writes are handled correctly when writing to Elasticsearch
-    
+  - **CUMULUS-2510**
+    - Removed `stream_enabled` and `stream_view_type` from `executions_table` TF
+      definition.
+    - Removed `aws_lambda_event_source_mapping` TF definition on executions DynamoDB table.
```
markdboyd, line 110: These should go under a `Removed` header

CHANGELOG.md
```diff
@@ -97,7 +104,17 @@ of response and not the raw API endpoint response
...
+    - Updated `@cumulus/sfEventSqsToDbRecords/write-execution` to publish SNS
+      messages after a successful write to Postgres, DynamoDB, and ES.
+    - Updated `@cumulus/api/endpoints/executions` to publish an SNS message
+      after a successful execution DELETE.
+    - Updated functions `create` and `upsert` in the `db` model for Executions
+      to return an array of objects containing values of the created or updated
+      columns.
```
markdboyd, line 117:
```suggestion
      to return an array of objects containing all the columns of the created or updated records.
```

markdboyd (changes requested):
Great work! A few requested changes and a few questions

packages/api/tests/endpoints/test-executions.js
```diff
@@ -179,13 +180,14 @@ test.before(async (t) => {
...
-    const executionCumulusIds = await t.context.executionPgModel.create(
+    const [pgExecution] = await t.context.executionPgModel.create(
       t.context.knex,
       executionPgRecord
     );
+    const executionCumulusIds = pgExecution.cumulus_id;
     const dynamoRecord = await executionModel.create(execution);
     await indexer.indexExecution(esClient, dynamoRecord, process.env.ES_INDEX);
-    return { ...executionPgRecord, cumulus_id: executionCumulusIds[0] };
+    return { ...executionPgRecord, cumulus_id: executionCumulusIds };
```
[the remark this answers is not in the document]
botanical: That's a great point

packages/api/tests/endpoints/test-executions.js
```diff
@@ -522,6 +562,13 @@ test.serial('DELETE deletes an execution', async (t) => {
       arn
     )
   );
+
+  const { Messages } = await sqs().receiveMessage({
```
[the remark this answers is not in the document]
botanical: Hmm I was under the impression that I would maintain what we currently do (which is sending messages with undefined)

packages/api/lib/writeRecords/write-execution.js
botanical: I think that's a compelling reason to move it outside of the transaction

CHANGELOG.md
botanical: [`bd0112b` (#2422)](https://github.com/nasa/cumulus/pull/2422/commits/bd0112b50cc0fda564ec0be0a84a830659daf1ea)

CHANGELOG.md
botanical: https://github.com/nasa/cumulus/pull/2422/commits/bd0112b50cc0fda564ec0be0a84a830659daf1ea

packages/api/lib/writeRecords/write-execution.js
botanical: [`bd0112b` (#2422)](https://github.com/nasa/cumulus/pull/2422/commits/bd0112b50cc0fda564ec0be0a84a830659daf1ea)

packages/db/src/models/execution.ts
botanical: [`bd0112b` (#2422)](https://github.com/nasa/cumulus/pull/2422/commits/bd0112b50cc0fda564ec0be0a84a830659daf1ea)

packages/api/tests/lib/writeRecords/test-write-execution.js
botanical: [`bd0112b` (#2422)](https://github.com/nasa/cumulus/pull/2422/commits/bd0112b50cc0fda564ec0be0a84a830659daf1ea)

packages/api/tests/endpoints/test-executions.js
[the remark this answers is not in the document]
botanical: [`bd0112b` (#2422)](https://github.com/nasa/cumulus/pull/2422/commits/bd0112b50cc0fda564ec0be0a84a830659daf1ea)

packages/api/tests/endpoints/test-executions.js
```diff
@@ -179,13 +180,14 @@ test.before(async (t) => {
       omitExecution,
       t.context.knex
     );
-    const executionCumulusIds = await t.context.executionPgModel.create(
+    const [pgExecution] = await t.context.executionPgModel.create(
       t.context.knex,
       executionPgRecord
     );
+    const executionCumulusIds = pgExecution.cumulus_id;
```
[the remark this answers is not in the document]
botanical: [`bd0112b` (#2422)](https://github.com/nasa/cumulus/pull/2422/commits/bd0112b50cc0fda564ec0be0a84a830659daf1ea)

packages/api/lib/writeRecords/write-execution.js
```diff
@@ -117,15 +119,17 @@ const _writeExecutionRecord = ({
   esClient,
 }) => knex.transaction(async (trx) => {
   logger.info(`About to write execution ${postgresRecord.arn} to PostgreSQL`);
-  const [executionCumulusId] = await executionPgModel.upsert(trx, postgresRecord);
-  logger.info(`Successfully wrote execution ${postgresRecord.arn} to PostgreSQL with cumulus_id ${executionCumulusId}`);
+  const [response] = await executionPgModel.upsert(trx, postgresRecord);
```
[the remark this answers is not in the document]
botanical: [`fb82e2d` (#2422)](https://github.com/nasa/cumulus/pull/2422/commits/fb82e2db5dc2045363bd773e6b55db6999d5c32b)

packages/api/tests/endpoints/test-executions.js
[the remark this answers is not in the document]
markdboyd: Hmm yeah I guess this unit test is just verifying existing behavior, not changing anything, so we're good

packages/api/lib/writeRecords/write-execution.js
markdboyd: Confirmed with the team that we don't want to publish to SNS in the transaction

packages/api/endpoints/executions.js
```diff
@@ -106,6 +107,7 @@ async function del(req, res) {
         index: process.env.ES_INDEX,
         ignore: [404],
       });
+      await publishExecutionSnsMessage({});
```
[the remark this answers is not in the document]
markdboyd: So I think that preserving existing behavior would be to send `undefined` on execution deletion (so no arguments to `publishExecutionSnsMessage`), since `publish-executions` is sending `record.dynamodb.NewImage`, which will be undefined for an event where a record was deleted. 

https://github.com/nasa/cumulus/blob/master/packages/api/lambdas/publish-executions.js#L18

I'm also not sure how much this matters, as to whether we need to preserve the existing behavior, because I'm betting it doesn't matter to Cloud Metrics whether they receive an `undefined` message or not. But I'll defer to @nemreid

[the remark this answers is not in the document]
markdboyd: I'll leave it up to you whether you want to keep this @botanical , but if we do keep it then we shouldn't be sending any arguments to `publishExecutionSnsMessage`

packages/db/src/models/execution.ts
```diff
@@ -13,6 +13,13 @@ class ExecutionPgModel extends BasePgModel<PostgresExecution, PostgresExecutionR
     });
   }
 
+  create(
+    knexOrTransaction: Knex | Knex.Transaction,
+    item: PostgresExecution
+  ) {
+    return super.create(knexOrTransaction, item, '*');
```
npauzenga, line 20: Thanks for making this change across the other types. I was initially concerned because `collectionPgModel.create()` was updated with this change before the others but as long as we're consistent I think it's great. It might be worth making the same change to `providerPgModel.create()` and `asynOperationPgModel.create()` in this PR? It doesn't look like it's explicitly part of the scope here, it would just be for consistency. Thoughts?

Also, just curious. The return of this method is an array. e.g.

```
const [executionPgRecord] = await executionPgModel.create(
    knex,
    execution
  );
```
Could we instead find a way to just return a single record so that we don't have to index into an array every time? I think our `create` logic would only ever create one record as we have it. Eh?

markdboyd: I would agree that those changes would be out of scope here. I'm also not 100% sure if we even need them for providers and async operations.

We needed them here because we wanted to be able to return the entire record on PG write, so that we can use it as the basis for the SNS message and we don't have to immediately look it up after write (causing another database query). 

But we don't send providers/async operations to any SNS topics, so I don't see any obvious need for those changes. 

The intent of the refactor was also to preserve behavior of existing models precisely so we could avoid other breaking changes, so absent a compelling reason to change them I'd be inclined to leave them as is.

npauzenga: Yup, totally on board with limiting the scope on this PR. Especially because the change I'm suggesting is really just for consistency. I _do_ like avoiding the array return type though if we're going to override the base model's `create` method anyway. It feels like that could change without too much fuss.

It does also seem like it would make sense to have all of the `create` methods return the same type (in a separate PR) but you're right, there aren't instances I can find where we're looking up a record immediately after `create` _except for the test helpers_.

 On the one hand this is why we extend a `Base` model, so that we can have e.g. `Collection` implement a different `create` method. On the other hand I think it would be really easy to make this same change to all of the models (or really, just the base model's `create` at that point). It looks like it wouldn't even break any logic. It would just require test updates.

markdboyd: Yeah I'm not against the change in practice, but it would require refactoring a bunch of tests at least and possibly other code I'm not thinking of, so just lots of lines of code. But happy to see that as a separate PR if anyone feels so inclined.

CHANGELOG.md
```diff
@@ -104,6 +110,13 @@ of response and not the raw API endpoint response
     - Changed `sfEventSqsToDbRecords` Lambda to use new upsert helpers for executions, granules, and PDRs
     to ensure out-of-order writes are handled correctly when writing to Elasticsearch
   - **CUMULUS-2510**
+    - Updated `@cumulus/sfEventSqsToDbRecords/write-execution` to publish SNS
+      messages after a successful write to Postgres, DynamoDB, and ES.
+    - Updated `@cumulus/api/endpoints/executions` to publish an SNS message
```
markdboyd, line 115: This is no longer true

botanical: Ah good catch

packages/db/src/models/execution.ts
botanical: @npauzenga How do you feel if I made changes to `create` to not return an array in a separate PR? Updating the function return type has major implications for lines of code outside of what my PR handles and I would rather keep those changes in a separate PR.

CHANGELOG.md
botanical: [`36eef2c` (#2422)](https://github.com/nasa/cumulus/pull/2422/commits/36eef2c1080a9bdefcb81119a18d230ca78266d5)

packages/db/src/models/execution.ts
markdboyd: To be clear, I'm on board with @npauzenga's suggestion to update all of our models to return `*` fields from the `create` method, but we all agree that should be done in a separate PR

I'm less convinced that we can/should refactor `create` to return an object vs an array. I can see the minor convenience of not needing to destructure the return value, but is there any value beyond that? It would also require refactoring a lot of affected code. 

The other thing about both of these suggestions is that neither one is really required for us to move forward and at some point we need to start limiting the scope of our work in order to complete RDS phase 2 work in a timely fashion.

npauzenga: Totally on board with not doing any of it in this PR to limit scope and get it out faster. It also doesn't have to be something you handle @botanical. It's really easy for me to pop into someone else's PR and request something that's tangential but not totally relevant to the work 😄

packages/api/tests/lib/writeRecords/test-write-execution.js
```diff
@@ -514,3 +547,27 @@ test('writeExecutionRecordFromApi() saves execution to Dynamo and RDS with same
...
+
+  t.is(Messages.length, 1);
+
+  const snsMessage = JSON.parse(Messages[0].Body);
+  const executionRecord = JSON.parse(snsMessage.Message);
+  const dynamoRecord = await executionModel.get({ arn: executionArn });
+
+  t.is(executionRecord.arn, executionArn);
+  t.is(executionRecord.status, cumulusMessage.meta.status);
+  t.deepEqual(omit(executionRecord, 'timestamp'), omit(dynamoRecord, 'timestamp'));
```
markdboyd, line 572: There should be no need to `omit` when comparing against the Dynamo record

markdboyd (approved):

packages/api/tests/lib/writeRecords/test-write-execution.js
botanical: For posterity, we talked about this offline and timestamp should be the same across PG and dynamo records, and we need to update how this is implemented

PR #2426, opened 2021-08-24
CUMULUS-2317: Update reconciliation reports to read from PostgreSQL

markdboyd:

Addresses [CUMULUS-2317: Update reconciliation reports to read from PostgreSQL instead of DynamoDB](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2317)

## Changes

- Added `@cumulus/db/getFilesAndGranuleInfoQuery()` to build a query for searching file
records in PostgreSQL and return specified granule information for each file
- Added `@cumulus/db/QuerySearchClient` library to handle sequentially fetching and paging
through results for an arbitrary PostgreSQL query
- Added `insert` method to all `@cumulus/db` models to handle inserting multiple records into
the database at once
- Changed reconciliation reports to read file records from PostgreSQL instead of DynamoDB

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests - Blocked, waiting on POST /granules endpoint so that I can update the reconciliation report integration tests


example/spec/parallel/createReconciliationReport/CreateReconciliationReportSpec.js
```diff
@@ -806,13 +817,19 @@ describe('When there are granule differences and granule reconciliation is run',
 
   afterAll(async () => {
     const activeCollectionId = constructCollectionId(extraCumulusCollection.name, extraCumulusCollection.version);
+
     console.log(`update database state back for  ${publishedGranuleId}, ${activeCollectionId}`);
-    await granuleModel.update({ granuleId: publishedGranuleId }, { files: granuleBeforeUpdate.files });
+    // await granuleModel.update({ granuleId: publishedGranuleId }, { files: granuleBeforeUpdate.files });
```
botanical, line 822:
```suggestion

```

packages/db/tests/lib/test-granule.js
```diff
@@ -1,6 +1,7 @@
 const test = require('ava');
 const sinon = require('sinon');
 const cryptoRandomString = require('crypto-random-string');
+const orderBy = require('lodash/orderBy');
```
botanical, line 4: 😮 nifty

packages/api/lambdas/create-reconciliation-report.js
```diff
@@ -712,8 +717,19 @@ async function processRequest(params) {
   };
   await reconciliationReportModel.create(reportRecord);
 
+  const env = params.env ? params.env : process.env;
+  const knex = await getKnexClient(env);
+  const concurrency = process.env.CONCURRENCY || 3;
```
botanical, line 722: Could `params.env` also have a `CONCURRENCY` variable?

packages/api/lambdas/internal-reconciliation-report.js
```diff
@@ -226,43 +243,78 @@ async function reportForGranulesByCollectionId(collectionId, recReportParams) {
...
+  );
 
   let okCount = 0;
   const withConflicts = [];
   const onlyInEs = [];
   const onlyInDb = [];
   const fieldsIgnored = ['timestamp', 'updatedAt'];
-
+  // "dataType" and "version" fields do not exist in the PostgreSQL database
+  // granules table which is now the source of truth
```
botanical, line 276: 👍

packages/api/lambdas/internal-reconciliation-report.js
```diff
@@ -226,43 +243,78 @@ async function reportForGranulesByCollectionId(collectionId, recReportParams) {
...
+  const esFieldsIgnored = [...fieldsIgnored, 'dataType', 'version'];
   const granuleFields = ['granuleId', 'collectionId', 'provider', 'createdAt', 'updatedAt'];
-  let [nextEsItem, nextDbItem] = await Promise.all([esGranulesIterator.peek(), dbGranulesIterator.peek()]); // eslint-disable-line max-len
+
+  let [nextEsItem, nextDbItem] = await Promise.all([esGranulesIterator.peek(), pgGranulesSearchClient.peek()]); // eslint-disable-line max-len
+
+  const translateDbResultToApiGranule = (dbResult) =>
```
botanical, line 282: Should this not be an async function?

packages/api/lambdas/internal-reconciliation-report.js
```diff
@@ -226,43 +243,78 @@ async function reportForGranulesByCollectionId(collectionId, recReportParams) {
...
+    translatePostgresGranuleToApiGranule({
```
botanical, line 283: Does this not need to be awaited?

packages/api/tests/lambdas/test-create-reconciliation-report.js
```diff
@@ -1590,6 +1717,8 @@ test.serial('reconciliationReportForGranuleFiles reports discrepancy of granule
 });
 
 test.serial('reconciliationReportForGranuleFiles does not fail if no distribution endpoint is defined', async (t) => {
+  process.env.DISTRIBUTION_ENDPOINT = 'https://example.com/';
```
botanical, line 1720: Shouldn't this be removed? The unit test is testing that it doesn't fail if a distribution endpoint isn't defined

packages/api/tests/lambdas/test-internal-reconciliation-report.js
```diff
@@ -160,17 +171,41 @@ test.serial('internalRecReportForCollections reports discrepancy of collection h
...
+  } = t.context;
+
+  // Create collection in PG/ES
   const collectionId = constructCollectionId(randomId('name'), randomId('version'));
-  const provider = randomId('provider');
 
-  const matchingGrans = range(10).map(() => fakeGranuleFactoryV2({ collectionId, provider }));
-  const additionalMatchingGrans = range(10).map(() => fakeGranuleFactoryV2({ provider }));
-  const extraDbGrans = range(2).map(() => fakeGranuleFactoryV2({ collectionId, provider }));
+  // const provider = randomId('provider');
```
botanical, line 184:
```suggestion
```

packages/db/src/lib/execution.ts
```diff
@@ -13,22 +13,30 @@ export interface arnRecord {
...
- * @param {Knex | Knex.Transaction} knexOrTransaction
+ * @param {Object} params
+ * @param {Knex | Knex.Transaction} params.knexOrTransaction
  *   Knex client for reading from RDS database
- * @param {number} granuleCumulusId - The primary ID for a Granule
- * @param {number} limit - limit to number of executions to query
+ * @param {Array<string>} params.executionColumns - Columns to return from executions table
+ * @param {number} params.granuleCumulusId - The primary ID for a Granule
+ * @param {number} [params.limit] - limit to number of executions to query
  * @returns {Promise<arnRecord[]>} - Array of arn objects with the most recent first.
```
botanical, line 25: I think this needs to be updated to include case where more than the default 'arn' column is specified

packages/db/src/lib/file.ts
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @param {Object} params
+ * @param {Knex} params.knex - Knex client object
+ * @param {Partial<PostgresFileRecord>} params.searchParams
+ *   Query search parameters for files table
+ * @param {Array<string>} params.sortColumns
+ *   Columns to sort results by
+ * @param {Array<string>} params.granuleColumns
+ *   Columns to return from granules table
+ * @param {number} [params.limit] - Limit on number of results to return. Optional.
```
botanical, line 19:
```suggestion
 * @param {number} params.limit - Limit on number of results to return. Optional.
```
Does the `[]` signify that it's optional? Not sure if it's necessary

botanical: Or perhaps all other params need them too?

packages/db/tests/lib/test-QuerySearchClient.js
```diff
@@ -0,0 +1,256 @@
...
+    knex,
+    searchParams: { bucket },
+    sortColumns: ['cumulus_id'],
+    granuleColumns: ['granule_id'],
+  });
+  const fileSearchClient = new QuerySearchClient(
+    query,
+    1
+  );
+  t.truthy(
```
botanical, line 165: Nit but maybe avoid `truthy` and update the test statement because `peek()` doesn't technically return `true`

packages/db/tests/lib/test-QuerySearchClient.js
```diff
@@ -0,0 +1,256 @@
...
+    sortColumns: ['cumulus_id'],
+    granuleColumns: ['granule_id'],
+  });
+  const fileSearchClient = new QuerySearchClient(
+    query,
+    1
+  );
+  t.falsy(
+    await fileSearchClient.peek()
+  );
```
botanical, line 218:
```suggestion
  t.is(
    await fileSearchClient.peek(),
    undefined
  );
```

packages/db/tests/lib/test-QuerySearchClient.js
```diff
@@ -0,0 +1,256 @@
...
+  t.is(queryOffsetSpy.callCount, 2);
+  t.is(queryOffsetSpy.getCall(0).args[0], 0);
+  t.is(queryOffsetSpy.getCall(1).args[0], 1);
+  t.is(queryLimitSpy.callCount, 2);
+  t.is(queryLimitSpy.getCall(0).args[0], 1);
+  t.is(queryLimitSpy.getCall(1).args[0], 1);
+});
+
+test('QuerySearchClient.peek() correctly returns false if next record does not exist in fetched results', async (t) => {
```
botanical, line 203: Consider updating this to something like:
```suggestion
test('QuerySearchClient.peek() correctly returns undefined if next record does not exist in fetched results', async (t) => {
```

botanical:
I'd consider updating these docs https://github.com/nasa/cumulus/blob/master/website/versioned_docs/version-v3.0.0/features/inventory-reports.md so that `onlyInDynamoDb` is changed to `onlyInDb`

example/spec/parallel/createReconciliationReport/CreateReconciliationReportSpec.js
```diff
@@ -350,6 +345,22 @@ describe('When there are granule differences and granule reconciliation is run',
...
+      extraGranuleInDb = {
+        granuleId: randomString(),
+        collectionId,
+        status: 'completed',
+        files: [extraFileInDb],
+      };
+      await createGranule({
+        prefix: config.stackName,
+        body: extraGranuleInDb,
+      });
```
botanical, line 362: Should we delete this granule at the end of the test?

packages/db/src/lib/collection.ts
```diff
@@ -0,0 +1,18 @@
+import Knex from 'knex';
+
+import { tableNames } from '../tables';
+
```
botanical, line 4: 🐜 JSDoc?

packages/db/src/lib/collection.ts
```diff
@@ -0,0 +1,18 @@
...
+) => {
+  const {
+    collections: collectionsTable,
+    granules: granulesTable,
+  } = tableNames;
+  return knex(collectionsTable)
+    .select(`${collectionsTable}.*`)
+    .innerJoin(granulesTable, `${collectionsTable}.cumulus_id`, `${granulesTable}.collection_cumulus_id`)
+    .whereIn(`${granulesTable}.granule_id`, granuleIds)
+    .groupBy(`${collectionsTable}.cumulus_id`);
```
botanical, line 17: Is there an advantage in not awaiting and then returning the result?

packages/db/src/lib/file.ts
```diff
@@ -0,0 +1,49 @@
...
+      tableNames.granules,
+      `${tableNames.files}.granule_cumulus_id`,
+      `${tableNames.granules}.cumulus_id`
+    )
+    .where(searchParams)
+    .orderBy(sortColumns);
+  if (limit) {
+    query.limit(limit);
+  }
+  return query;
```
botanical, line 48: Do you not need to await `query`?

packages/db/tests/lib/test-collection.js
```diff
@@ -0,0 +1,89 @@
+const test = require('ava');
```
botanical, line 1: When should we omit `use strict`?

markdboyd:
@botanical 

> I'd consider updating these docs https://github.com/nasa/cumulus/blob/master/website/versioned_docs/version-v3.0.0/features/inventory-reports.md so that onlyInDynamoDb is changed to onlyInDb

I don't think we want to update versioned docs for a previous version of Cumulus. Because if you were to deploy and use that version of Cumulus, it would still generate reconciliation reports with a key of `onlyInDynamoDb`. So we only want the docs to be updated for versions of Cumulus going forward.

PR #2431, opened 2021-08-26
CUMULUS-2575 Create granule via API

flamingbear:
**Summary:** Adds endpoint to create granule via API

Addresses [CUMULUS-2575: create a granule via Core API](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2575)

## Changes

* Adds endpoint at create function to create a granule in cumulus via API

* Renames api/lambda/sf-event-sqs-to-db-records/write-granules.js to api/lib/writeRecords/write-granules.js

* refactors write-granules.js to remove the generateGranuleRecord to create a
  postgres record and now generates a dynamoGranuleRecord and calls
  translateApiGranuleToPostgresGranule

* Separates out concerns for creating granules and storing granules.

## PR Checklist

- [X] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


packages/api/lib/writeRecords/write-granules.js
```diff
@@ -384,102 +279,185 @@ const _generateFilesFromGranule = async ({
...
+      productionDateTime,
+      lastUpdateDateTime,
+    };
+
+    const collectionNameVersion = deconstructCollectionId(collectionId);
+    const collectionCumulusId = await getCollectionCumulusId(collectionNameVersion, knex);
 
-    return granuleModel.storeGranuleFromCumulusMessage({
+    let executionCumulusId;
+    if (execution !== undefined) {
```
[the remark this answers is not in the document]
flamingbear: belts and suspenders? I'm always afraid of other truthy values. I can change it.  Under it I'm using  `if (executionCumulusId === undefined) {`. Should that be !executionCumulusId also?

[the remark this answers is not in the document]
flamingbear: cecd2497d5

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -384,102 +279,185 @@ const _generateFilesFromGranule = async ({
...
+    });
+    if (result && result.status === 'rejected') {
+      const theError = new Error(result.reason);
+      log.error('Failed to _writeGranule', theError);
+      throw theError;
+    }
+    return `Wrote Granule ${granule.granuleId}`;
+  } catch (thrownError) {
+    log.error('Failed to write granule', thrownError);
+    log.debug(`thrownError.message: ${thrownError.message}`);
```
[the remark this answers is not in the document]
flamingbear: Nope, that was actually me debugging and forgetting to take it out.

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -536,38 +507,65 @@ const writeGranules = async ({
...
-      knex,
-      granuleModel,
-    })
+    async (granule) => {
+      // compute granule specific data.
+      const files = await _generateFilesFromGranule({ granule, provider });
+      const timeToArchive = getGranuleTimeToArchive(granule);
+      const timeToPreprocess = getGranuleTimeToPreprocess(granule);
+      const productVolume = getGranuleProductVolume(files);
+      const now = Date.now(); // yank me
```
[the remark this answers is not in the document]
flamingbear: That was a note to myself for keeping the times synchronous between the two records, and isn't true anymore.   I'll take out the comment, and use the same version in both calls below. 6d9e5384e6

packages/api/package.json
```diff
@@ -22,6 +22,7 @@
     "serve-oauth": "LOCALSTACK_HOST=localhost LOCAL_ES_HOST=localhost NODE_ENV=test cumulus-api serve",
     "serve-remote": "cumulus-api serve",
     "serve": "LOCALSTACK_HOST=localhost LOCAL_ES_HOST=localhost NODE_ENV=test FAKE_AUTH=true cumulus-api serve",
+    "serve:debug": "LOCALSTACK_HOST=localhost LOCAL_ES_HOST=localhost NODE_ENV=test FAKE_AUTH=true node --inspect-brk ./bin/cli.js serve",
```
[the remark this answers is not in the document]
flamingbear: Yeah just a convenience I added.

packages/db/src/test-utils.ts
```diff
@@ -51,7 +52,7 @@ export const destroyLocalTestDb = async ({
   testDbName: string
 }) => {
   knex.destroy();
-  await deleteTestDatabase(knexAdmin, testDbName);
+  await pRetry(async () => await deleteTestDatabase(knexAdmin, testDbName));
```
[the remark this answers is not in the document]
flamingbear: I was getting intermittent unit test errors that said another user was using the database I wanted to delete. I think it was a race condition.

```

  models › rule-model › after.always hook

  Rejected promise returned by test. Reason:

  error (DatabaseError) @Error {
    code: '55006',
    column: undefined,
    constraint: undefined,
    dataType: undefined,
    detail: 'There is 1 other session using the database.',
    file: 'dbcommands.c',
    hint: undefined,
    internalPosition: undefined,
    internalQuery: undefined,
    length: 160,
    line: '876',
    position: undefined,
    routine: 'dropdb',
    schema: undefined,
    severity: 'ERROR',
    table: undefined,
    where: undefined,
    message: 'drop database if exists "rule_edb686cd12" - database "rule_edb686cd12" is being accessed by other users',
  }

```

packages/integration-tests/Granules.js
```diff
@@ -75,6 +77,17 @@ const getGranuleWithStatus = async (params = {}) =>
     }
   );
 
+const buildRandomizedGranule = (overrides = {}) => ({
```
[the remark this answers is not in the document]
flamingbear: Yeah, we don't, I wasn't sure it was ok to pull it from there because I saw buildRandomizedCollections in the @cumulus/integration-test package https://github.com/nasa/cumulus/blob/master/packages/integration-tests/Collections.js#L80

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -742,3 +632,278 @@ test.serial('writeGranules() stores error on granule if any file fails', async (
...
+  const {
+    knex,
+    granule,
+  } = t.context;
+
+  const error = await t.throwsAsync(
+    writeGranuleFromApi({ ...granule, collectionId: undefined }, knex)
+  );
+  // That's an ugly error, but that's what's coming out.
+  t.is(error.message, 'Cannot read property \'split\' of undefined');
```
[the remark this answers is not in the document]
flamingbear: It should probably be fixed to throw and invalidinput or similar here. https://github.com/nasa/cumulus/blob/master/packages/message/src/Collections.ts#L40
but I was afraid to change it to see what else was going to break.  I can try.

packages/api/tests/endpoints/test-granules.js
```diff
@@ -1064,7 +1064,10 @@ test.serial('When a move granule request fails to move a file correctly, it reco
         (dest) => updatedFiles[0].fileName.match(dest.regex)
       ).bucket);
 
-      t.true(updatedFiles[1].key.startsWith(`${destinationFilepath}/${granuleFileName}`));
+      t.true(
+        updatedFiles[1].key.startsWith(`${destinationFilepath}/${granuleFileName}`),
+        `updatedFile[1] ${updatedFiles[1].key}, did not start with ${destinationFilepath}/${granuleFileName}`
+      );
```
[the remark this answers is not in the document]
flamingbear: Doesn't look like it, I just ran it >100x without an error. But this just adds logging info if it happens again.

example/spec/parallel/testAPI/granuleSpec.js
```diff
@@ -0,0 +1,100 @@
...
+        expect(response.statusCode).toBe(200);
+        const { result } = JSON.parse(response.body);
+        expect(result).toBe(`Wrote Granule ${granuleId}`);
+      }
+    });
+
+    it('Errors creating a bad granule.', async () => {
+      const name = randomId('name');
+      const version = randomId('version');
+      const randomGranuleRecord = buildRandomizedGranule({
```
kkelly51, line 82: Just checking, this fails because the files are not provided in the create step?

packages/db/src/test-utils.ts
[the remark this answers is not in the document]
markdboyd: Just curious - did you have a pgAdmin session open against your database when you tried to delete it?

packages/integration-tests/Granules.js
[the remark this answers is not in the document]
markdboyd: Yeah, we're not exactly being consistent. We can leave this

packages/db/src/test-utils.ts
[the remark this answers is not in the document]
flamingbear: I don't think so. But it's possible.  But on an immediate rerun it would work.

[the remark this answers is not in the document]
flamingbear: Oh, I wouldn't have been connected to the localApi random DB.

packages/api/models/granules.js
```diff
@@ -293,72 +294,89 @@ class Granule extends Manager {
...
-    if (!granule.granuleId) throw new CumulusModelError(`Could not create granule record, invalid granuleId: ${granule.granuleId}`);
+    if (!granule.granuleId) {
+      throw new CumulusModelError(`Could not create granule record, invalid granuleId: ${granule.granuleId}`);
+    }
 
     if (!collectionId) {
       throw new CumulusModelError('collection required to generate a granule record');
     }
 
+    if (!files || files.length === 0) {
```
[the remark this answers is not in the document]
flamingbear: Interesting, I assumed it was a requirement from what I was seeing.  Long story, not short. 


on origin/master: If files is undefined it creates the postgres record correctly because it calls.
`_generateFilesFromGranule` which provides an empty files array as a default.
https://github.com/nasa/cumulus/blob/master/packages/api/lambdas/sf-event-sqs-to-db-records/write-granules.js#L370

But inside _writeGranules when it stores the dynamoDB record it calls `storeGranuleFromCumulusMessage` and in turn `generateGranuleRecord` calls `buildDatabaseFiles` directly with an undefined files.
https://github.com/nasa/cumulus/blob/master/packages/api/lib/FileUtils.js#L129

in the test it's mocked to just resolve as the input files. 
    buildDatabaseFiles: (params) => Promise.resolve(params.files),

But in reality a call to buildDatabaseFiles with an undefined files will throw an `UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'map' of undefined`

I *think* that will propagate up making the trx transaction fail, and keeping either of the granule records from being written.

So, I probably guessed wrong when I assumed this was the expected behavior? I don't know if it makes sense to have a granule without a file, but can make this work either way.

example/spec/parallel/testAPI/granuleSpec.js
flamingbear: It fails because the granule is associated with a collectionId that doesn't exist in the database.

packages/api/models/granules.js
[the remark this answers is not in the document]
Jkovarik: Hmm, I'm wondering if that's just a bug if we're calling it that way.  The most salient user of that method is this

```
const translateGranule = async (
  granule,
  fileUtils = FileUtils
) => {
  if (isNil(granule.files)) return granule;

  return {
    ...granule,
    files: await fileUtils.buildDatabaseFiles({
      s3: awsClients.s3(),
      files: granule.files,
    }),
  };
};```

which has a catch re: taht value being nil.

File: example/spec/parallel/testAPI/granuleSpec.js [hunk 11]
```diff
@@ -0,0 +1,100 @@
...
+        });
+        granuleId = randomGranuleRecord.granuleId;
+        const response = await createGranule({
+          prefix,
+          body: randomGranuleRecord,
+        });
+
+        expect(response.statusCode).toBe(200);
+        const { result } = JSON.parse(response.body);
+        expect(result).toBe(`Wrote Granule ${granuleId}`);
```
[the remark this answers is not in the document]
flamingbear: yes, most definitely.

packages/api/tests/models/granules/test-granule-generateGranuleRecord.js
```diff
@@ -141,21 +120,23 @@ test('generateGranuleRecord() builds a failed granule record', async (t) => {
...
-    workflowStatus: 'failed',
+    workflowStatus,
+    status: getGranuleStatus(workflowStatus, granule),
     error,
   });
 
   t.deepEqual(
     record.files,
-    granule.files.map((file) => granuleFileToRecord(file, provider))
+    granule.files
```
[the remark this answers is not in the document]
flamingbear: Before, they were being built by buildDatabaseFiles inside the call, and now they are just being passed along. (For granules from messages the files pass to this still goes through the same function so no change in what's being stored).  I suppose for the API they have some freedom to do what they want.  I can take out this check since it's not doing much if you think that's misleading?

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -742,3 +632,278 @@ test.serial('writeGranules() stores error on granule if any file fails', async (
...
+    allfiles.push(fakeFileFactory());
+  }
+  const validFileCount = allfiles.length - invalidFiles.length;
+
+  await writeGranuleFromApi({ ...granule, files: allfiles }, knex);
+
+  t.false(await filePgModel.exists(knex, { key: invalidFiles[0].key }));
+  t.false(await filePgModel.exists(knex, { key: invalidFiles[1].key }));
+
+  const fileRecords = await filePgModel.search(knex, {});
```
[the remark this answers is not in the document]
flamingbear: To be fair, I copied this test, so I'll fix it in both places. 64346bf61b

packages/integration-tests/Granules.js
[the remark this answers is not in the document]
flamingbear: I'm fine yanking it if you think it's better to pull it from inside the api/lib/testUtils.  I don't love duplicate code. But also just wasn't sure what the team wants to do with those.

packages/api/lib/FileUtils.js
```diff
@@ -126,10 +126,15 @@ const cleanDatabaseFile = (file) =>
     removeNilProperties,
   ])(file);
 
-const buildDatabaseFiles = async ({ s3, providerURL, files }) =>
-  await Promise.all(
+const buildDatabaseFiles = async ({ s3, providerURL, files }) => {
```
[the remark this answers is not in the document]
flamingbear: Yup, that's fine now. e6b96203df

packages/api/models/granules.js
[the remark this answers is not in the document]
flamingbear: fef362ebef

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -384,102 +279,185 @@ const _generateFilesFromGranule = async ({
...
+    const result = await _writeGranule({
+      postgisGranuleRecord,
+      dynamoGranuleRecord,
+      collectionCumulusId,
+      executionCumulusId,
+      knex,
+      granuleModel,
+
+    });
+    if (result && result.status === 'rejected') {
```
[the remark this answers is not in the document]
flamingbear: Hmmm, no idea.  Nope, that's just bad refactoring.  I thought that was coming out of _writeGranule ,but is from Promise.allSettled() (for the writeGranuleFromMessage code.)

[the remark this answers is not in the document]
flamingbear: 2eca64456a

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -742,3 +632,278 @@ test.serial('writeGranules() stores error on granule if any file fails', async (
...
+test.serial('writeGranuleFromApi() throws for a granule with no granuleId provided', async (t) => {
+  const {
+    knex,
+    granule,
+  } = t.context;
+
+  const error = await t.throwsAsync(
+    writeGranuleFromApi({ ...granule, granuleId: undefined }, knex)
+  );
+  t.is(error.message, 'Could not create granule record, invalid granuleId: undefined');
```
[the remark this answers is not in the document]
flamingbear: Cool, TIL. 031c596a28

packages/api/tests/lib/writeRecords/test-write-granules.js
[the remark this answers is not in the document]
flamingbear: ca9543500a

packages/api/models/granules.js
```diff
@@ -705,19 +684,43 @@ class Granule extends Manager {
...
+            timeToArchive,
+            timeToPreprocess,
+            productVolume,
+            duration,
+            status,
+            processingTimeInfo,
+            queryFields,
+          });
+          return this.storeGranule(granuleRecord);
+        } catch (writeError) {
```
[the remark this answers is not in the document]
flamingbear: I don't think so.  We're swallowing the error, not just logging and rethrowing.  It morphed to this from https://github.com/nasa/cumulus/blob/master/packages/api/models/granules.js#L720
Which I could morph back. 
But I would think that would prevent all granules from being written if I took out the catch. 
Yea if I take that out, this fails
``` 
 models › granules › granule-reporting › storeGranulesFromCumulusMessage() handles failing and succcessful granules independently
```

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -560,15 +451,14 @@ test.serial('writeGranules() does not persist records to Dynamo or Postgres if D
     granuleId,
   } = t.context;
 
-  const fakeGranuleModel = {
-    storeGranuleFromCumulusMessage: () => {
-      throw new Error('Granules dynamo error');
-    },
-    describeGranuleExecution: () => Promise.resolve({}),
+  const fakeGranuleModel = new Granule();
```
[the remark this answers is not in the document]
flamingbear: Nope, I added that so that some of the other methods were available to the test. 
Looking closer, I can add just a method for generateGranule that returns a valid granule. 
742e99e361

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -742,3 +632,278 @@ test.serial('writeGranules() stores error on granule if any file fails', async (
...
+
+test.serial('writeGranuleFromApi() does not persist records to Dynamo or Postgres if Dynamo write fails', async (t) => {
+  const {
+    granule,
+    granuleId,
+    granuleModel,
+    knex,
+  } = t.context;
+
+  const fakeGranuleModel = new Granule();
```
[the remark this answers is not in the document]
flamingbear: 3281ba8eaa

example/spec/parallel/testAPI/granuleSpec.js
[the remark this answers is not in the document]
flamingbear: 8b82322272

packages/api/models/granules.js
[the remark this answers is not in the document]
markdboyd: Ah okay, I misunderstood. This is fine then

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -742,3 +632,278 @@ test.serial('writeGranules() stores error on granule if any file fails', async (
...
+      insert: () => {
+        throw new Error('Granules Postgres error');
+      },
+    });
+    return cb(fakeTrx);
+  };
+  const trxStub = sinon.stub(knex, 'transaction').callsFake(fakeTrxCallback);
+  t.teardown(() => trxStub.restore());
+
+  const error = await t.throwsAsync(writeGranuleFromApi(granule, knex));
```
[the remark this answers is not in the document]
flamingbear: to be fair, I copied it from an existing test. https://github.com/nasa/cumulus/blob/master/packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-write-granules.js#L588-L624

But let me see if I can clean that up too.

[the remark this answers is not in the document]
flamingbear: Maybe?  8e0aa42c2d

packages/api/endpoints/granules.js
```diff
@@ -55,6 +56,29 @@ async function list(req, res) {
...
 }
 
+/**
+ * Create new granule
+ *
+ * @param {Object} req - express request object
+ * @param {Object} res - express response object
+ * @returns {Promise<Object>} promise of an express response object.
+ */
+async function create(req, res) {
```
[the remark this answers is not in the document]
flamingbear: 8d3957ef3e

flamingbear:
@markdboyd this is ready for some questions answered and re-review if the questions are fine.

packages/api/tests/models/granules/test-granule-generateGranuleRecord.js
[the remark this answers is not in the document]
markdboyd: No, I think it's fine, I mostly just wanted to check my understanding. I like it much better now, in fact

packages/integration-tests/Granules.js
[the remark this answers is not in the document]
markdboyd: Yeah we could pull from api/lib/testUtils but I don't love that either. I don't love treating our `api` package as the source of any helpers. So the duplication here doesn't bother me too much

example/spec/parallel/testAPI/granuleSpec.js
```diff
@@ -0,0 +1,114 @@
...
+    });
+
+    it('can discover the granule in the API.', async () => {
+      const granule = await getGranule({
+        prefix,
+        granuleId,
+      });
+      expect(granule.files[0].size).toEqual(jasmine.any(Number));
+      delete granule.files[0].size;
+      granule.execution = undefined;
```
markdboyd, line 89: Why are these two lines necessary?

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -384,102 +281,183 @@ const _generateFilesFromGranule = async ({
...
   knex,
-  error,
-  executionUrl,
-  processingTimeInfo,
-  providerCumulusId,
-  pdrCumulusId,
-  granuleModel,
-  updatedAt = Date.now(),
+  granuleModel = new Granule(),
+  granulePgModel = new GranulePgModel(),
```
markdboyd, line 301:
```suggestion
  granuleModel,
  granulePgModel,
```

The parameters are necessary but the initialization is not (I don't believe) since everything that calls `_writeGranule` has already initialized these models

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -207,66 +137,33 @@ const getGranuleCumulusIdFromQueryResultOrLookup = async ({
...
+const _writePostgresGranuleViaTransaction = async ({
+  granuleRecord,
   collectionCumulusId,
-  providerCumulusId,
   executionCumulusId,
-  pdrCumulusId,
   trx,
-  updatedAt,
-  files,
+  granulePgModel = new GranulePgModel(),
```
markdboyd, line 158:
```suggestion
  granulePgModel,
```

Initialization is unnecessary since it already should be initialized by calling code

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -742,3 +635,291 @@ test.serial('writeGranules() stores error on granule if any file fails', async (
...
+
+  const result = await writeGranuleFromApi({ ...granule }, knex);
+
+  t.is(result, `Wrote Granule ${granuleId}`);
+
+  t.true(await granuleModel.exists({ granuleId }));
+  t.true(await granulePgModel.exists(knex, { granule_id: granuleId }));
+});
+
+test.serial('writeGranuleFromApi() writes a granule without an execution to postgres and dynamo.', async (t) => {
```
markdboyd, line 704:
```suggestion
test.serial('writeGranuleFromApi() writes a granule without an execution to PostgreSQL and DynamoDB.', async (t) => {
```

Definitely a nitpick

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -79,6 +79,7 @@ test.beforeEach(async (t) => {
   t.context.granule = fakeGranuleFactoryV2({
     files: t.context.files,
     granuleId: t.context.granuleId,
+    collectionId: `${t.context.collection.name}___${t.context.collection.version}`,
```
markdboyd, line 82:
```suggestion
    collectionId: constructCollectionId(t.context.collection.name, t.context.collection.version),
```

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -742,3 +635,291 @@ test.serial('writeGranules() stores error on granule if any file fails', async (
...
+    granuleModel,
+    granulePgModel,
+  } = t.context;
+
+  const result = await writeGranuleFromApi({ ...granule }, knex);
+
+  t.is(result, `Wrote Granule ${granuleId}`);
+
+  t.true(await granuleModel.exists({ granuleId }));
+  t.true(await granulePgModel.exists(knex, { granule_id: granuleId }));
```
markdboyd, line 701: Everywhere we are doing a granulePgModel.get or exists, we should probably get in the habit of doing it like this (which you are doing below)

```
  const granuleRecord = await granulePgModel.get(
    knex,
    {
      granule_id: granuleId,
    }
  );
```

The reason why is that in PG, only the combo of granule_id + collection_cumulus_id makes a granule unique, so otherwise the `get` or `exists` could return true/truthy for the `granule_id` if we had bad test setup even when our logic didn't create the record

packages/api/tests/models/granules/test-granule-reporting.js
```diff
@@ -467,10 +479,13 @@ test('storeGranulesFromCumulusMessage() handles failing and succcessful granules
 
   const granule1 = fakeGranuleFactoryV2({
     files: [fakeFileFactory({ bucket })],
+    status: 'completed',
   });
-  // Missing files should cause failure to write
+
+  // If both the message's workflow status and the granule status it will fail validation.
```
markdboyd, line 485:
```suggestion
  // If both the message's workflow status and the granule status are undefined, the granule will fail validation.
```

packages/api/tests/endpoints/test-granules.js
```diff
@@ -1481,3 +1484,29 @@ test.serial('PUT with action move returns failure if more than one granule file
 
   filesExistingStub.restore();
 });
+
+test.serial('POST creates new granule in dynamoDB and postgres', async (t) => {
```
markdboyd, line 1488: I think we also should have two basic tests to show that that the granule doesn't appear in Dynamo if the PG write fails and vice versa

markdboyd: The mocking in the collections endpoint and the unit tests might be a helpful guide:

https://github.com/nasa/cumulus/blob/master/packages/api/endpoints/collections.js#L110
https://github.com/nasa/cumulus/blob/master/packages/api/tests/endpoints/collections/create-collection.js#L375-L429

flamingbear: Outside of the unit tests that already exist for that?

example/spec/parallel/testAPI/granuleSpec.js
flamingbear: L88 because I don't know the size of the file when I create it, oh I guess I do because the body is just 'testbody' but size it isn't part of the input granule. But I would either have to add a size key to the randomGranuleRecord, or delete it from the returned value.
L89 I didn't want to set up an execution, so created a granule without one.  But it gets stripped out of the return value, so I set it to undefined explicitly because it's in the test randomGranuleRecord. 
Both are just to use the expectation on the next line

packages/api/tests/endpoints/test-granules.js
markdboyd: I suppose given that the endpoint just calls `writeGranuleFromApi`, which does test this logic, we don't need to. It is possible that our endpoint logic could somehow change and we would want to be sure that the endpoint still behaves as expected, but YAGNI for now

packages/api/tests/lib/writeRecords/test-write-granules.js
flamingbear: Pick it!

packages/api/lib/writeRecords/write-granules.js
flamingbear: Yeah that's probably just belts and suspenders.  We can add it back for all of these after the fact when we need them.

packages/api/tests/lib/writeRecords/test-write-granules.js
flamingbear: yup have to now that it has guards. d65a289996, and lint fix to go with a6322f9a88

packages/api/tests/endpoints/test-granules.js
flamingbear: There's a bit more to the endpoint than there was before now that we lifted the finding of collectionCumulusId out of the original.  I added other tests. But the endpoint itself, only uses PG granule models, and all of the logic for writing DynamoDB vs PG is encapsulated in writeGranuleFromApi so I think those cases are handled at the moment.  I can take out the test context because it's not being used in testing.

flamingbear:
@markdboyd Think I've addressed all additional feedback.

packages/integration-tests/Granules.js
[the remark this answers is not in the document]
flamingbear: Well, Jenny followed me down this road and added some for executions.  I'm not a fan and I'm going to pull from testUtils in my next follow-on PR I think to get rid of this.

packages/api/endpoints/granules.js
```diff
@@ -55,6 +61,48 @@ async function list(req, res) {
...
+    knex = await getKnexClient(),
+    granulePgModel = new GranulePgModel(),
+  } = req.testContext || {};
+
+  let collectionCumulusId;
+  const granule = req.body;
+
+  try {
+    const collectionNameVersion = deconstructCollectionId(granule.collectionId);
+    collectionCumulusId = await getCollectionCumulusId(collectionNameVersion, knex);
```
markdboyd, line 82: Should we pre-emptively throw an error if `collectionCumulusId` is undefined somehow, since it's required for the PG write anyway?

packages/api/endpoints/granules.js
```diff
@@ -55,6 +61,48 @@ async function list(req, res) {
...
+    if (await granulePgModel.exists(knex, {
```
markdboyd, line 83: This just occurred to me: why is your work checking PosgreSQL instead of DynamoDB to see if the granule already exists. The code on the `master` branch is still RDS phase 1 where DynamoDB is the source of truth. It's not until RDS phase 2, which has our work in a feature branch, where we would need to check PostgreSQL instead of DynamoDB.

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -384,102 +280,182 @@ const _generateFilesFromGranule = async ({
...
-  await _writeGranuleFiles({
-    files,
-    granuleCumulusId,
-    granule,
-    workflowError: error,
-    workflowStatus,
-    knex,
-    granuleModel,
-  });
+    const postgresGranuleRecord = await translateApiGranuleToPostgresGranule(
```
markdboyd, line 436: So I just realized that this translate function is going to look up the `collection_cumulus_id` anyway. So at minimum that's two lookups for it: one in the endpoint logic and one in this function. And furthermore, the `collectionCumulusId` that we're passing into `_writeGranule` is completely unnecessary - it's not used in any write logic, just a log statement. The same information is already available in `postgresGranuleRecord.collection_cumulus_id`

But this may all be irrelevant since I think your endpoint should be checking DynamoDB and not PostgreSQL for an existing record

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -698,11 +603,15 @@ test.serial('writeGranules() writes all valid files if any non-valid file fails'
   t.false(await filePgModel.exists(knex, { key: invalidFiles[0].key }));
   t.false(await filePgModel.exists(knex, { key: invalidFiles[1].key }));
 
-  const fileRecords = await filePgModel.search(knex, {});
+  const granuleCumulusId = await granulePgModel.getRecordCumulusId(
```
markdboyd, line 606: This method on the granule model should also force you to pass in the minimal required fields to identify a granule record uniquely in PG, but I'm not gonna put that on you in this PR

packages/db/src/models/granule.ts
```diff
@@ -31,6 +39,49 @@ export default class GranulePgModel extends BasePgModel<PostgresGranule, Postgre
       .del();
   }
 
+  /**
+   * Checks if a granule is present in Postgres
```
markdboyd, line 43:
```suggestion
   * Checks if a granule is present in PostgreSQL
```

packages/db/src/models/granule.ts
```diff
@@ -31,6 +39,49 @@ export default class GranulePgModel extends BasePgModel<PostgresGranule, Postgre
...
+    } catch (error) {
+      if (error instanceof RecordDoesNotExist) {
+        return false;
+      }
+      throw error;
+    }
+  }
+
+  /**
+   * Fetches a single granule from Postgres
```
markdboyd, line 66:
```suggestion
   * Fetches a single granule from PostgreSQL
```

packages/db/src/models/granule.ts
```diff
@@ -31,6 +39,49 @@ export default class GranulePgModel extends BasePgModel<PostgresGranule, Postgre
...
+   * @param {PostgresGranuleUniqueColumns | RecordSelect} params - An object
+   *         of PostgresGranuleUniqueColumns or RecordSelect
+   * @returns {Promise<PostgresGranuleRecord>} The returned record
+   */
+  get(
+    knexOrTransaction: Knex | Knex.Transaction,
+    params: PostgresGranuleUniqueColumns | RecordSelect
+  ): Promise<PostgresGranuleRecord> {
+    if (!isRecordSelect(params)) {
+      if (!(params.granule_id && params.collection_cumulus_id)) {
```
markdboyd, line 78: You are a hero for doing this

packages/db/src/models/granule.ts
```diff
@@ -1,13 +1,21 @@
...
 
 import { BasePgModel } from './base';
 import { GranulesExecutionsPgModel } from './granules-executions';
 import { translateDateToUTC } from '../lib/timestamp';
 
+interface RecordSelect {
+  cumulus_id: number
+}
+
+function isRecordSelect(param: RecordSelect | PostgresGranuleUniqueColumns): param is RecordSelect {
```
markdboyd, line 15: I love this

CHANGELOG.md
```diff
@@ -37,6 +40,11 @@ for objects that could not be processed
 - Updated async operations to handle writing records to the databases
 when output of the operation is `undefined`
 
+- **CUMULUS-2575**
```
markdboyd, line 43: We need to note the changes to `get` and `exists` on the PG granule model. They're great changes, we just need to advertise them

Also, the change to `deconstructCollectionId` in the `message` package

And the new `createGranule` helper in the `api-client` package

example/spec/parallel/testAPI/granuleSpec.js
markdboyd: For this first issue, can we just add a `size` key to `granuleFile` around line 44 since we do know the size of the body?

For the second issue, can we remove `  execution: undefined,` from `randomGranuleRecord`? Omitting the property in the return from `randomGranuleRecord` would have the same effect of returing a granule that has no execution, but there wouldn't be an explicit `execution` property set that we have to compare against

packages/api/endpoints/granules.js
flamingbear: no since you don't need it anymore in this fxn.

packages/api/lib/writeRecords/write-granules.js
flamingbear: Yeah, I will pull out the collectionCumulusId from writeGranuleFromApi and downwards when I switch to dynamoDB.

packages/db/src/models/granule.ts
flamingbear: I'm a full-stack overflow developer. Actually not true, I found this in the typescript docs.

example/spec/parallel/testAPI/granuleSpec.js
flamingbear: Yup, Had done (most of) that in the update granule PR,  Can pull it in here. 5eb6f6836d

CHANGELOG.md
flamingbear: 8850faba9a

packages/api/tests/lib/writeRecords/test-write-granules.js
flamingbear: probably getRecordsCumulusId too...

packages/api/endpoints/granules.js
flamingbear: faae13024c

CHANGELOG.md
```diff
@@ -16,6 +16,19 @@ of response and not the raw API endpoint response
...
+  - Adds `POST /granules` API endpoint to create a granule
+  - Updates `@cumulus/db/src/model/granules` functions `get` and `exists` to
+    enforce parameter checking so that requests include either (granule\_id
+    and collection\_cumulus\_id) or (cumulus\_id) to prevent incorrect results.
+  - `@cumulus/message/src/Collections.deconstructCollectionId` has been
+    modified to throw a descriptive error if the input `collectionId` is
+    undefined rather than `TypeError: Cannot read property 'split' of
+    undefined`. This function has also been updated to throw descriptive errors
+    if an incorrectly formated collectionId is input.
```
markdboyd, line 28: These two notes should be under the "Changed" header

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -536,38 +501,66 @@ const writeGranules = async ({
...
+
+      const postgresGranuleRecord = await translateApiGranuleToPostgresGranule(
+        dynamoGranuleRecord,
+        knex
+      );
+
+      return _writeGranule({
+        postgresGranuleRecord,
+        dynamoGranuleRecord,
+        collectionCumulusId,
```
markdboyd, line 543: Unnnecessary now, I believe

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -737,7 +646,328 @@ test.serial('writeGranules() stores error on granule if any file fails', async (
...
+    granule,
+  } = t.context;
+  const execution = `execution${cryptoRandomString({ length: 5 })}`;
+  await t.throwsAsync(
+    writeGranuleFromApi({ ...granule, execution }, knex),
+    { message: `Could not find execution in PostgreSQL database with url ${execution}` }
+  );
+});
+
+test.serial('writeGranuleFromApi() saves granule records to Dynamo and Postgres with same timestamps.', async (t) => {
```
markdboyd, line 775: This test coverage is 💯

markdboyd (approved):
I have two more minor things that I'd like to see addressed but going to pre-emptively approve this.

This is really really good work with some super solid refactoring. The changes to the granule PG model and the write-granules logic in particular are 👨‍🍳 💋

packages/api/lib/writeRecords/write-granules.js
flamingbear: Yeah, I missed that, and that should propagate up also, to not being needed in the sf-event-sqs-to-db-records call to writeGranulesFromMessage.  Which makes sense since it's just the same as writeGranulesFromApi.

flamingbear: 6e6a9eb5d1

CHANGELOG.md
flamingbear: 82d8c679fc

CHANGELOG.md
```diff
@@ -37,6 +40,20 @@ for objects that could not be processed
...
+    that will not affect granules created in the normal way.
+  - Updates `@cumulus/db/src/model/granules` functions `get` and `exists` to
+    enforce parameter checking so that requests include either (granule\_id
+    and collection\_cumulus\_id) or (cumulus\_id) to prevent incorrect results.
+  - `@cumulus/message/src/Collections.deconstructCollectionId` has been
+    modified to throw a descriptive error if the input `collectionId` is
+    undefined rather than `TypeError: Cannot read property 'split' of
+    undefined`. This function has also been updated to throw descriptive errors
+    if an incorrectly formated collectionId is input.
+  - Adds helper `createGranule` to `@cumulus/api-client`
```
markdboyd, line 55: This one is in the wrong place now. Not a big deal though

flamingbear: ugh.  I thought you counted wrong.
It's alright something failed, probably lint. I'll fix it

flamingbear: 05f2c65969

PR #2432, opened 2021-08-28
[CUMULUS-2510] Publish SNS messages for collections

botanical:

Addresses [CUMULUS-2510: Publish SNS messages for collections](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2510)

## Changes

* Added the new function `publishCollectionSnsMessage` in `@cumulus/api` to
      publish SNS messages to the report collections topic.
* Updated functions `create` and `upsert` in the `db` model for Collections
      to return an array of objects containing values of the created or updated
      columns.
* Updated `@cumulus/api/endpoints/collections` to publish an SNS message
      after a successful collection delete, update (PUT), create (POST).

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


example/spec/parallel/testAPI/collectionSpec.js
```diff
@@ -46,13 +52,40 @@ describe('Collections API', () => {
...
+  });
+
+  it('updating a collection publishes a record to the collection reporting SNS topic', async () => {
+    if (beforeAllFailed) {
+      fail('beforeAll() failed');
+    } else {
+      await apiTestUtils.updateCollection({
+        prefix: config.stackName,
+        collection,
+        updateParams: { files: [] },
```
markdboyd, line 69: I'm pretty surprised that this update works. I think `collection.files` are required by the collection write code:

https://github.com/nasa/cumulus/blob/feature/rds-phase-2/packages/api/models/collections.js#L60

So I'm a bit confused how this is passing if you can shed any light

example/spec/parallel/testAPI/collectionSpec.js
```diff
@@ -46,13 +52,40 @@ describe('Collections API', () => {
...
+      expect(message.event).toEqual('Create');
+      expect(omit(message.record, ['createdAt', 'updatedAt'])).toEqual(collection);
+    }
+  });
+
+  it('updating a collection publishes a record to the collection reporting SNS topic', async () => {
+    if (beforeAllFailed) {
+      fail('beforeAll() failed');
+    } else {
+      await apiTestUtils.updateCollection({
```
markdboyd, line 66: We should use `updateCollection` from the `api-client` package. The `apiTestUtils` in the integration-tests package will/should be ultimately deprecated

markdboyd: We should also update the test to use `createCollection` from the `api-client`

packages/api/endpoints/collections.js
```diff
@@ -131,13 +132,15 @@ async function post(req, res) {
 
     try {
       await knex.transaction(async (trx) => {
-        await collectionPgModel.create(trx, dbRecord);
+        const [pgCollection] = await collectionPgModel.create(trx, dbRecord);
```
markdboyd, line 135: Great work 👏

packages/api/lib/publishSnsMessageUtils.js
```diff
@@ -0,0 +1,26 @@
...
+const publishCollectionSnsMessage = async (record, event) => {
+  let messageToPublish;
+  const topicArn = envUtils.getRequiredEnvVar('collection_sns_topic_arn', process.env);
+
+  if (event === 'Delete') {
+    const deleteRecord = { name: record.name, version: record.version };
+    messageToPublish = { event, record: deleteRecord, deletedAt: Date.now() };
+  } else if (event === 'Create' || event === 'Update') {
+    messageToPublish = { event, record };
+  }
```
markdboyd, line 19: This is a nitpick, but I slightly prefer this implementation of determining what collection SNS message to send:

https://github.com/nasa/cumulus/blob/feature/rds-phase-2/packages/api/lambdas/publish-collections.js#L11-L20

It provides a default case if some how the event type didn't match any expected value, but I also just think it's a bit easier to read. I would vote to copy that switch style logic into a function named `createCollectionSnsMessage` or somethign and use it in this function.

packages/api/tests/endpoints/collections/create-collection.js
```diff
@@ -180,9 +211,21 @@ test('POST creates a new collection in all data stores', async (t) => {
...
+    QueueUrl: t.context.QueueUrl,
+    WaitTimeSeconds: 10,
+  }).promise();
+
+  t.is(Messages.length, 1);
+
+  const message = JSON.parse(JSON.parse(Messages[0].Body).Message);
+  const omitFields = ['createdAt', 'updatedAt'];
+  t.is(message.event, 'Create');
+  t.deepEqual(omit(message.record, omitFields), omit(newCollection, omitFields));
```
markdboyd, line 225: I think we want to compare against `fetchedDynamoRecord`, not `newCollection`. The intent here is to prove that what we send to the SNS topic represents the data after we have written to the database, not before. By comparing against `newCollection`, we're testing against the record before it is written to the database. These two things may often be the same, but not always

packages/api/tests/endpoints/collections/create-collection.js
```diff
@@ -150,7 +181,7 @@ test('POST with invalid authorization scheme returns an invalid token response',
   assertions.isInvalidAuthorizationResponse(t, res);
 });
 
-test('POST creates a new collection in all data stores', async (t) => {
+test.serial('POST creates a new collection in all data stores and publishes an SNS message', async (t) => {
```
markdboyd, line 184: Why are all the tests `serial` now?

packages/api/tests/endpoints/collections/delete-collection.js
```diff
@@ -184,9 +219,24 @@ test('Deleting a collection removes it from all data stores', async (t) => {
...
+  const { Messages } = await sqs().receiveMessage({
+    QueueUrl: t.context.QueueUrl,
+    WaitTimeSeconds: 10,
+  }).promise();
+
+  t.is(Messages.length, 1);
+
+  const message = JSON.parse(JSON.parse(Messages[0].Body).Message);
+  t.is(message.event, 'Delete');
+  t.truthy(message.deletedAt);
```
markdboyd, line 232:
```suggestion
  t.true(Date.now() > message.deletedAt);
```

Slightly prefer this because I think it demonstrates that `message.deletedAt` is a timestamp

packages/api/tests/endpoints/collections/delete-collection.js
```diff
@@ -144,7 +179,7 @@ test('Attempting to delete a collection with an invalid access token returns an
 
 test.todo('Attempting to delete a collection with an unauthorized user returns an unauthorized response');
 
-test('Deleting a collection removes it from all data stores', async (t) => {
+test.serial('Deleting a collection removes it from all data stores and publishes an SNS message', async (t) => {
```
markdboyd, line 182: Why are all the tests `serial` now?

packages/api/tests/endpoints/collections/update-collection.js
```diff
@@ -187,9 +224,21 @@ test('PUT replaces an existing collection', async (t) => {
...
+    QueueUrl: t.context.QueueUrl,
+    WaitTimeSeconds: 10,
+  }).promise();
+
+  t.is(Messages.length, 1);
+
+  const message = JSON.parse(JSON.parse(Messages[0].Body).Message);
+  const omitFields = ['createdAt', 'updatedAt'];
+  t.is(message.event, 'Update');
+  t.deepEqual(omit(message.record, omitFields), omit(updatedCollection, omitFields));
```
markdboyd, line 238: Again, I think we should compare against `actualCollection` and not `updatedCollection` given that what we want to prove is that the record in the SNS message matches what was written to the database

packages/api/tests/endpoints/collections/update-collection.js
```diff
@@ -117,7 +154,7 @@ test('CUMULUS-912 PUT with pathParameters and with an invalid access token retur
 
 test.todo('CUMULUS-912 PUT with pathParameters and with an unauthorized user returns an unauthorized response');
 
-test('PUT replaces an existing collection', async (t) => {
+test.serial('PUT replaces an existing collection and sends an SNS message', async (t) => {
```
markdboyd, line 157: Do these tests need to be `serial`? If the SNS topic and queue that are used per test have randomized names, then there shouldn't be test collisions

packages/api/tests/endpoints/collections/update-collection.js
```diff
@@ -280,9 +329,20 @@ test('PUT creates a new record in RDS if one does not exist', async (t) => {
...
+    QueueUrl: t.context.QueueUrl,
+    WaitTimeSeconds: 10,
+  }).promise();
+
+  t.is(Messages.length, 1);
+
+  const message = JSON.parse(JSON.parse(Messages[0].Body).Message);
+  const omitFields = ['createdAt', 'updatedAt'];
+  t.is(message.event, 'Update');
+  t.deepEqual(omit(message.record, omitFields), omit(updatedCollection, omitFields));
```
markdboyd, line 342:
```suggestion
  t.deepEqual(omit(message.record, omitFields), omit(fetchedDynamoRecord, omitFields));
```

packages/db/src/models/collection.ts
```diff
@@ -11,6 +11,14 @@ class CollectionPgModel extends BasePgModel<PostgresCollection, PostgresCollecti
       tableName: tableNames.collections,
     });
   }
+  async create(
+    knexOrTransaction: Knex | Knex.Transaction,
+    item: PostgresCollection
+  ) {
+    return await knexOrTransaction(this.tableName)
```
markdboyd, line 18: There's nothing wrong with this, but since we're going to need to do the same thing for every data type, I think we should update the base model `create` to have an optional argument for `returningFields`, which could be a string or array of strings. Then from this method (and for all the other data type overrides we're going to do) we could do:

```
  return super.get(knexOrTransaction, item, '*')
```

CHANGELOG.md
```diff
@@ -97,7 +102,16 @@ of response and not the raw API endpoint response
...
+  - **CUMULUS-2510**
+    - Removed `stream_enabled` and `stream_view_type` from `collections_table`
+      TF definition.
+    - Removed `aws_lambda_event_source_mapping` TF definition on collections
+      DynamoDB table.
+    - Updated `@cumulus/api/endpoints/collections` to publish an SNS message
+      after a successful collection delete, update (PUT), create (POST).
+    - Updated functions `create` and `upsert` in the `db` model for Collections
+      to return an array of objects containing values of the created or updated
+      columns.
```
markdboyd, line 114:
```suggestion
      to return an array of objects containing all columns for the created or updated records
```

I believe `returning` returns whatever columns you specify whether they were changed or not:

https://knexjs.org/#Builder-returning

So `*` would always return all columns

CHANGELOG.md
```diff
@@ -97,7 +102,16 @@ of response and not the raw API endpoint response
   - **CUMULUS-2634**
     - Changed `sfEventSqsToDbRecords` Lambda to use new upsert helpers for executions, granules, and PDRs
     to ensure out-of-order writes are handled correctly when writing to Elasticsearch
-    
+  - **CUMULUS-2510**
+    - Removed `stream_enabled` and `stream_view_type` from `collections_table`
+      TF definition.
+    - Removed `aws_lambda_event_source_mapping` TF definition on collections
+      DynamoDB table.
```
markdboyd, line 109: We usually create another `### Removed` header to list the deletions. You also deleted the `publishCollections` Lambda itself

CHANGELOG.md
```diff
@@ -23,6 +23,11 @@ of response and not the raw API endpoint response
       - `@cumulus/es-client/indexer.upsertExecution` to upsert an execution
       - `@cumulus/es-client/indexer.upsertPdr` to upsert a PDR
       - `@cumulus/es-client/indexer.upsertGranule` to upsert a granule
+  - **CUMULUS-2510**
+    - Added `collection_sns_topic_arn` environment variable to
+      `api.tf`.
```
markdboyd, line 28:
```suggestion
    - Added `collection_sns_topic_arn` environment variable to
      `PrivateApiLambda` and `ApiEndpoints` lambdas
```

The more important thing to note is what user facing things were changed, not what files were edited necessarily

markdboyd (changes requested):
This is great work! My comments are mostly around testing and a few minor changes

example/spec/parallel/testAPI/collectionSpec.js
botanical: Hm interesting. If it were required, then wouldn't the file value here be invalid? https://github.com/nasa/cumulus/blob/feature/rds-phase-2/packages/api/lib/testUtils.js#L257 

```
  collection.files.forEach((file) => checkRegex(file.regex, file.sampleFileName));

``` wouldn't this just be undefined if file were `[]`?

File: packages/api/lib/testUtils.js [hunk 16]
```diff
@@ -432,7 +432,7 @@ const createCollectionTestRecords = async (context, collectionParams) => {
   await collectionModel.create(originalCollection);
   const [collectionCumulusId] = await collectionPgModel.create(testKnex, insertPgRecord);
   const originalPgRecord = await collectionPgModel.get(
-    testKnex, { cumulus_id: collectionCumulusId }
+    testKnex, { cumulus_id: collectionCumulusId.cumulus_id }
```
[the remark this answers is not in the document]
botanical: [`84b2d25` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/84b2d25af943ef4258b05f7b430694c83e51fa61)

example/spec/parallel/testAPI/collectionSpec.js
botanical: [`84b2d25` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/84b2d25af943ef4258b05f7b430694c83e51fa61)

packages/api/tests/endpoints/collections/create-collection.js
botanical: Good point

botanical: [`889e8be` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/889e8be4a4de6901e7e3a1adb756dfbe44fd1c00)

CHANGELOG.md
botanical: [`754f871` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/754f871e6b2a2231b24c1dc75a24f09bf623fabd)

CHANGELOG.md
botanical: Ah, I think that was what I was trying to articulate but your suggestion is much better [`754f871` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/754f871e6b2a2231b24c1dc75a24f09bf623fabd)

packages/api/tests/endpoints/collections/update-collection.js
botanical: [`754f871` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/754f871e6b2a2231b24c1dc75a24f09bf623fabd)

packages/api/tests/endpoints/collections/delete-collection.js
botanical: [`754f871` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/754f871e6b2a2231b24c1dc75a24f09bf623fabd)

packages/db/src/models/collection.ts
botanical: https://github.com/nasa/cumulus/pull/2432/commits/9ef092e76bb4d8842b815f4af0173dd7c0a61a2c
🤔

example/spec/parallel/testAPI/collectionSpec.js
```diff
@@ -46,13 +52,43 @@ describe('Collections API', () => {
...
+      });
+
+      await expectAsync(waitForObjectToExist({
+        bucket: config.bucket,
+        key: recordUpdatedKey,
+      })).toBeResolved();
+      const savedEvent = await getJsonS3Object(config.bucket, recordUpdatedKey);
+      const message = JSON.parse(savedEvent.Records[0].Sns.Message);
+      expect(message.event).toEqual('Update');
+      expect(omit(message.record, ['createdAt', 'updatedAt', 'files'])).toEqual(omit(updatedCollection, ['createdAt', 'updatedAt', 'files']));
```
markdboyd, line 82: Why are we omitting `files`?

example/spec/parallel/testAPI/collectionSpec.js
```diff
@@ -46,13 +52,43 @@ describe('Collections API', () => {
...
+      expect(message.record.reportToEms).toEqual(false);
```
markdboyd, line 83: We probably want to check that `reportToEms` is true after creating the record so that we can prove that the record was definitely updated

packages/api-client/src/collections.ts
```diff
@@ -35,6 +35,40 @@ export const createCollection = async (params: {
...
+ *
+ * @param {Object} params              - params
+ * @param {string} params.prefix       - the prefix configured for the stack
+ * @param {Object} params.collection   - collection object to update in the database
+ * @param {Function} params.callback   - async function to invoke the api lambda
+ *                                     that takes a prefix / user payload.  Defaults
+ *                                     to cumulusApiClient.invokeApi
+ * @returns {Promise<Object>}          - the response from the callback
+ */
+export const updateCollection = async (params: {
```
markdboyd, line 49: Great job adding this, just note it in the changelog

example/spec/parallel/testAPI/collectionSpec.js
botanical: Ah sorry that was residual that I didn't clean up from my first round of implementing this test

packages/api-client/src/collections.ts
botanical: [`8d76b1c` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/8d76b1c48461cee69ebff33bd04b3d57dc3c0796)

example/spec/parallel/testAPI/collectionSpec.js
botanical: [`8d76b1c` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/8d76b1c48461cee69ebff33bd04b3d57dc3c0796)

example/spec/parallel/testAPI/collectionSpec.js
botanical: [`8d76b1c` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/8d76b1c48461cee69ebff33bd04b3d57dc3c0796)

example/spec/parallel/testAPI/collectionSpec.js
```diff
@@ -46,13 +52,45 @@ describe('Collections API', () => {
         bucket: config.bucket,
         key: recordCreatedKey,
       })).toBeResolved();
+      const savedEvent = await getJsonS3Object(config.bucket, recordCreatedKey);
+      const message = JSON.parse(savedEvent.Records[0].Sns.Message);
+      expect(message.event).toEqual('Create');
+      expect(omit(message.record, ['createdAt', 'updatedAt'])).toEqual(omit(collection, ['createdAt', 'updatedAt']));
```
markdboyd, line 58: Comparing against `collection` is comparing against the record before it was written, whereas what we want to prove is that the written record was sent to the SNS message.

So I think we should do a `getCollection` and use that to compare with the SNS message

example/spec/parallel/testAPI/collectionSpec.js
```diff
@@ -46,13 +52,45 @@ describe('Collections API', () => {
...
+
+      await expectAsync(waitForObjectToExist({
+        bucket: config.bucket,
+        key: recordUpdatedKey,
+      })).toBeResolved();
+      const savedEvent = await getJsonS3Object(config.bucket, recordUpdatedKey);
+      const message = JSON.parse(savedEvent.Records[0].Sns.Message);
+      expect(message.event).toEqual('Update');
+      expect(omit(message.record, ['createdAt', 'updatedAt'])).toEqual(omit(updatedCollection, ['createdAt', 'updatedAt']));
+      expect(originalCollection.reportToEms).toEqual(true);
```
markdboyd, line 84: We should move this up to be a comparison against the record retrieved from the API (see comment above), rather than against the fixture we sent to the API

example/spec/parallel/testAPI/collectionSpec.js
markdboyd: We also then don't need to omit the timestamps, which makes the comparison more meaningful.

We should do the same thing below, do a `getCollection` after the update and compare against that

packages/api-client/tests/test-collections.js
```diff
@@ -101,3 +124,31 @@ test('getCollections calls the callback with the expected object and returns the
...
   t.deepEqual(result, { foo: 'bar' });
 });
+
+test('updateCollection calls the callback with the expected object', async (t) => {
+  const collection = fakeCollectionFactory();
+  await t.context.collectionModel.create(collection);
+  const updatedCollection = {
+    ...collection,
+    reportToEms: false,
+  };
```
markdboyd, line 134: You really don't need to create the record in the database for this test to succeed. Nor do you need to create a complex object by overriding properties of `collection`. We just need to prove that whatever object is send in the function arguments is received in the request body

packages/api/lib/publishSnsMessageUtils.js
```diff
@@ -0,0 +1,36 @@
...
+const { envUtils } = require('@cumulus/common');
+
+const Logger = require('@cumulus/logger');
+
+const logger = new Logger({ sender: '@cumulus/publishSnsMessageUtils' });
+
+const constructSnsMessage = (record, event) => {
+  switch (event) {
+  case 'Create': return { event, record };
+  case 'Update': return { event, record };
```
markdboyd, line 13:
```suggestion
  case 'Create': 
  case 'Update':
    return { event, record };
```

This is one of the nice features of switch statements

packages/api/tests/endpoints/collections/update-collection.js
markdboyd: I was very wrong

packages/db/src/models/base.ts
```diff
@@ -170,15 +170,18 @@ class BasePgModel<ItemType, RecordType extends BaseRecord> {
    *
    * @param {Knex | Knex.Transaction} knexOrTransaction - DB client or transaction
    * @param {ItemType} item - A record to insert into the DB
-   * @returns {Promise<number[]>} List of IDs of the inserted records
+   * @param {string | Array<Ssring>} returningFields - A string or array of strings
```
markdboyd, line 173:
```suggestion
   * @param {string | Array<string>} returningFields - A string or array of strings
```

packages/db/src/models/base.ts
```diff
@@ -170,15 +170,18 @@ class BasePgModel<ItemType, RecordType extends BaseRecord> {
...
+   *   of columns to return. Defaults to 'cumulus_id'.
+   * @returns {Promise<number[] | Object[]>} List of IDs of the inserted records
    */
   async create(
     knexOrTransaction: Knex | Knex.Transaction,
-    item: ItemType
-  ): Promise<number[]> {
+    item: ItemType,
+    returningFields: string | string[] = 'cumulus_id'
+  ): Promise<number[] | Object[]> {
```
markdboyd, line 181: So looking at the [Knex docs for `returning`](https://knexjs.org/#Builder-returning), it says:

>When passed in a string, makes the SQL result be reported as an array of values from the specified column. When passed in an array of strings, makes the SQL result be reported as an array of objects, each containing a single property for each of the specified columns.

So actually we really don't have a guarantee about what the value would be if only a single column is passed - it would depend on what column is passed in. 

So I would make the return type here and in the JSdoc `Promise<unknown[] | Object[]>`

packages/api/tests/endpoints/collections/create-collection.js
```diff
@@ -180,9 +211,21 @@ test('POST creates a new collection in all data stores', async (t) => {
...
+    QueueUrl: t.context.QueueUrl,
+    WaitTimeSeconds: 10,
+  }).promise();
+
+  t.is(Messages.length, 1);
+
+  const message = JSON.parse(JSON.parse(Messages[0].Body).Message);
+  const omitFields = ['createdAt', 'updatedAt'];
+  t.is(message.event, 'Create');
+  t.deepEqual(omit(message.record, omitFields), omit(fetchedDynamoRecord, omitFields));
```
markdboyd, line 225: There is no need to `omit` if we're comparing against the fetched record from dynamo which has the timestamps

packages/api/tests/endpoints/collections/update-collection.js
```diff
@@ -187,9 +224,21 @@ test('PUT replaces an existing collection', async (t) => {
...
+    QueueUrl: t.context.QueueUrl,
+    WaitTimeSeconds: 10,
+  }).promise();
+
+  t.is(Messages.length, 1);
+
+  const message = JSON.parse(JSON.parse(Messages[0].Body).Message);
+  const omitFields = ['createdAt', 'updatedAt'];
+  t.is(message.event, 'Update');
+  t.deepEqual(omit(message.record, omitFields), omit(actualCollection, omitFields));
```
markdboyd, line 238: There is no need to `omit` if we're comparing against the fetched record from dynamo which has the timestamps

packages/api/tests/endpoints/collections/update-collection.js
```diff
@@ -280,9 +329,20 @@ test('PUT creates a new record in RDS if one does not exist', async (t) => {
...
+    QueueUrl: t.context.QueueUrl,
+    WaitTimeSeconds: 10,
+  }).promise();
+
+  t.is(Messages.length, 1);
+
+  const message = JSON.parse(JSON.parse(Messages[0].Body).Message);
+  const omitFields = ['createdAt', 'updatedAt'];
+  t.is(message.event, 'Update');
+  t.deepEqual(omit(message.record, omitFields), omit(fetchedDynamoRecord, omitFields));
```
markdboyd, line 342: There is no need to `omit` if we're comparing against the fetched record from dynamo which has the timestamps

packages/db/src/models/base.ts
botanical: Nice catch

packages/api/tests/endpoints/collections/update-collection.js
botanical: [`bdd6661` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/bdd66615c308abb672a3d4ca00c6eb92cb5eef48#diff-b87875b83c4639dd8008bd26f239a6580ba66ed6ab9441aac8c86575d8b7ab73R339)

packages/api/tests/endpoints/collections/update-collection.js
botanical: [`bdd6661` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/bdd66615c308abb672a3d4ca00c6eb92cb5eef48#diff-b87875b83c4639dd8008bd26f239a6580ba66ed6ab9441aac8c86575d8b7ab73R236)

packages/api/tests/endpoints/collections/create-collection.js
botanical: [`bdd6661` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/bdd66615c308abb672a3d4ca00c6eb92cb5eef48#diff-3e17073a9180eeafda94cabe8eaef8263b9978a4e105ba063d56185e48916081R223)

packages/db/src/models/base.ts
botanical: [`bdd6661` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/bdd66615c308abb672a3d4ca00c6eb92cb5eef48#diff-3827859fa5de2f6a67108f80b2a4ec150451909cb0ef78e58ec517789311da7fR175) Thoughts on the return description?

example/spec/parallel/testAPI/collectionSpec.js
markdboyd: We also want to compare against the collection from the API here. Again, otherwise . So we want to move this logic into this block:

```
     const originalCollectionFromApi = await getCollection({
        prefix: config.stackName,
        collectionName: originalCollection.name,
        collectionVersion: originalCollection.version,
      });
      expect(originalCollectionFromApi.reportToEms).toEqual(true);
```

Then you can re-use `originalCollectionFromApi` below. We also don't want to omit the timestamps in the comparison

example/spec/parallel/testAPI/collectionSpec.js
markdboyd: We don't need/want to omit properties if they're coming back from the database, because it gives us an even more precise test

packages/api-client/tests/test-collections.js
```diff
@@ -1,12 +1,35 @@
...
+
 const collectionsApi = require('../collections');
+const { fakeCollectionFactory } = require('../../api/lib/testUtils');
+const { Collection } = require('../../api/models');
+const { randomId, randomString } = require('../../common/test-utils');
 
-test.before((t) => {
+test.before(async (t) => {
+  process.env.stackName = randomString();
+  process.env.system_bucket = randomString();
```
markdboyd, line 17: You don't need this now

packages/api-client/tests/test-collections.js
```diff
@@ -1,12 +1,35 @@
...
   t.context.testPrefix = 'unitTestStack';
   t.context.collectionName = 'testCollection';
   t.context.collectionVersion = 1;
+
+  // create a fake bucket
+  await createBucket(process.env.system_bucket);
+
+  process.env.CollectionsTable = randomId('collection');
+  t.context.collectionModel = new Collection();
+  await t.context.collectionModel.createTable();
```
markdboyd, line 27: You don't need this now

packages/api-client/tests/test-collections.js
```diff
@@ -1,12 +1,35 @@
...
+});
+
+test.after.always(async (t) => {
```
markdboyd, line 30: You don't need this now

packages/db/src/models/base.ts
```diff
@@ -170,15 +170,19 @@ class BasePgModel<ItemType, RecordType extends BaseRecord> {
    *
    * @param {Knex | Knex.Transaction} knexOrTransaction - DB client or transaction
    * @param {ItemType} item - A record to insert into the DB
-   * @returns {Promise<number[]>} List of IDs of the inserted records
+   * @param {string | Array<string>} returningFields - A string or array of strings
+   *   of columns to return. Defaults to 'cumulus_id'.
+   * @returns {Promise<unknown[] | Object[]>} Returns an array of objects or an
+   *   array of values from the specified column from returningFields.
```
markdboyd, line 176:
```suggestion
   *   array of values from the specified column(s) from returningFields.
```

packages/db/src/models/base.ts
markdboyd: One minor suggestion otherwise it looks good

packages/db/src/models/base.ts
botanical: [`eb916aa` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/eb916aa52b4a766b18602366fbc3550eb2deb9ae)

packages/api-client/tests/test-collections.js
botanical: [`eb916aa` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/eb916aa52b4a766b18602366fbc3550eb2deb9ae)

packages/api-client/tests/test-collections.js
botanical: [`eb916aa` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/eb916aa52b4a766b18602366fbc3550eb2deb9ae)

packages/api-client/tests/test-collections.js
botanical: [`eb916aa` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/eb916aa52b4a766b18602366fbc3550eb2deb9ae)

example/spec/parallel/testAPI/collectionSpec.js
markdboyd: We are still omitting properties in the comparison, which is unnecessary if we're comparing against the record fetched from the API

example/spec/parallel/testAPI/collectionSpec.js
markdboyd: We are still omitting properties in the comparison, which is unnecessary if we're comparing against the record fetched from the API

example/spec/parallel/testAPI/collectionSpec.js
```diff
@@ -46,13 +52,56 @@ describe('Collections API', () => {
...
+  it('updating a collection publishes a record to the collection reporting SNS topic', async () => {
+    if (beforeAllFailed) {
+      fail('beforeAll() failed');
+    } else {
+      const originalCollection = collection;
+      const updatedCollection = {
+        ...originalCollection,
+        reportToEms: false,
+      };
+      const originalCollectionFromApi = await getCollection({
```
markdboyd, line 71: We should move at least this request into the above test for the creation SNS message so that we can compare against that rather than comparing against the `collection` fixture

example/spec/parallel/testAPI/collectionSpec.js
botanical: [`68b2fb6` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/68b2fb64fb360c103d4a24de28286ad68b4c75fb)

Sorry to have missed this the first time you mentioned it!

example/spec/parallel/testAPI/collectionSpec.js
botanical: [`68b2fb6` (#2432)](https://github.com/nasa/cumulus/pull/2432/commits/68b2fb64fb360c103d4a24de28286ad68b4c75fb)

markdboyd (approved):

PR #2440, opened 2021-09-02
Cumulus 2644 Update serve.js and serveUtils.js

jaortega527:
**Summary:** Updated serve.js and serveUtils.js to include postgres when seeding/resetting via serveUtils

Addresses [CUMULUS-2644: Update LocalAPI to support Postgres seeding for development and testing](https://bugs.earthdata.nasa.gov/browse/CUMULUS-XXX)

## Changes

* Updated `addXXX` methods to include writing to Postgres tables
* Added `resetPostgresDb` method to support resetting between integration test suite runs

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


packages/api/bin/serveUtils.js
```diff
@@ -3,70 +3,185 @@
...
+  translateApiPdrToPostgresPdr,
+  translateApiRuleToPostgresRule,
+  getKnexClient,
+  localStackConnectionEnv,
+  envParams,
+  createTestDatabase
+} = require('@cumulus/db');
+const { log } = require('console');
+
+const migrationDir = 'node_modules/@cumulus/db/dist/migrations';
```
markdboyd, line 31: Why is this not part of the require out of `@cumulus/db`?

jaortega527: Great question. You are 100% correct.

jaortega527: it's been updated, and pushed

packages/api/bin/serveUtils.js
```diff
@@ -1,72 +1,183 @@
...
+  try {
+    await createTestDatabase(knexAdmin, 'postgres', localStackConnectionEnv.PG_USER);
+  } catch (error) {
+    log(`Skipping Postgres DB creation because ${error}`);
+  }
+
+  await knex.migrate.latest();
+
+  await erasePostgresTables(knex);
+}
```
jaortega527, line 50: Created a `resetPostgresDb` because between each suite of integration tests we reset the dynamoDb tables and ES without restarting the API. This method simply empties each table and ensures the latest schema is applied. It will create the database if it doesn't exist (ie first run).

packages/api/bin/serveUtils.js
```diff
@@ -1,72 +1,183 @@
...
-        .then((collection) =>
-          indexer.indexCollection(es.client, collection, es.index)))
+    collections.map(async (c) => {
+      const dynamoRecord = await collectionModel.create(c);
+      await indexer.indexCollection(es.client, dynamoRecord, es.index);
+      const dbRecord = await translateApiCollectionToPostgresCollection(c);
+      const collectionPgModel = new CollectionPgModel();
+      await collectionPgModel.create(knex, dbRecord);
+    })
   );
```
jaortega527, line 70: Updated each `addXXX` method to create a postgres record using the appropriate `translateApiXxxToPostgresXxx()` and then calling `xxxPgModel.create()`

packages/api/bin/serveUtils.js
```diff
@@ -1,72 +1,183 @@
...
+  return await executions.reduce((p, e) => p
+    .then(async () => await executionModel.create(e))
+    .then((dynamoRecord) => {
+      indexer.indexExecution(es.client, dynamoRecord, es.index);
+      return translateApiExecutionToPostgresExecution(dynamoRecord, knex)
+        .then(async (dbRecord) => {
+          const executionPgModel = new ExecutionPgModel();
+          return await executionPgModel.create(knex, dbRecord);
+        });
+    }), starterPromise);
```
jaortega527, line 160: `Executions` is a little different since it has a parent/child relationship with itself records cannot be asynchronously created without having some records fail to be inserted sometimes. There may be a more elegant way to accomplish this, but currently am using a starter promise with `reduce()` to force records to be inserted in the order they are provided in the `executionsFixture.json` seed file

packages/api/bin/serveUtils.js
```diff
@@ -82,12 +193,57 @@ async function addReconciliationReports(reconciliationReports) {
...
+      const dbRecord = {
+        granule_cumulus_id: granuleCumulusId,
+        execution_cumulus_id: executionCumulusId,
+      };
+
+      await granulesExecutionsPgModel.create(knex, dbRecord);
+    })
+  );
+}
+
```
jaortega527, line 238: `granules_executions` is also a little different in that it required a lookup of the granule and execution Ids prior to record creation. Also, since there is not equivalent dynamoDb table it is Postgres only.

packages/api/bin/serve.js
```diff
@@ -427,4 +427,5 @@ module.exports = {
   serveApi,
   serveDistributionApi,
   resetTables,
+  erasePostgresTables,
 };
```
jaortega527, line 431: Exported `erasePostgresTables` to support `resetPostgresDb` in `serveUtils.js`

packages/api/bin/serveUtils.js
```diff
@@ -1,72 +1,183 @@
...
-    collections.map((c) =>
-      collectionModel
-        .create(c)
-        .then((collection) =>
-          indexer.indexCollection(es.client, collection, es.index)))
+    collections.map(async (c) => {
+      const dynamoRecord = await collectionModel.create(c);
+      await indexer.indexCollection(es.client, dynamoRecord, es.index);
+      const dbRecord = await translateApiCollectionToPostgresCollection(c);
+      const collectionPgModel = new CollectionPgModel();
```
markdboyd, line 67: I would instantiate this model outside of the Promise.all, maybe after line 60, rather than doing per call

packages/api/bin/serveUtils.js
```diff
@@ -1,72 +1,183 @@
...
   return await Promise.all(
-    granules.map((c) =>
-      granuleModel
-        .create(c)
-        .then((granule) => indexer.indexGranule(es.client, granule, es.index)))
+    granules.map(async (g) => {
+      const dynamoRecord = await granuleModel.create(g);
+      await indexer.indexGranule(es.client, dynamoRecord, es.index);
+      const dbRecord = await translateApiGranuleToPostgresGranule(dynamoRecord, knex);
+      const granulePgModel = new GranulePgModel();
```
markdboyd, line 88: Same comment - move the instantiation of the model above the Promise.all call

packages/api/bin/serveUtils.js
jaortega527: Great catch. Moved and pushed.

packages/api/bin/serveUtils.js
jaortega527: Great catch. Moved and pushed.

markdboyd: Same comment for all of the models

jaortega527: Also addressed for all remaining methods.

packages/api/bin/serveUtils.js
```diff
@@ -82,12 +193,57 @@ async function addReconciliationReports(reconciliationReports) {
...
+      ...envParams,
+      ...localStackConnectionEnv,
+    },
+  });
+
+  const collectionPgModel = new CollectionPgModel();
+  return await Promise.all(
+    granulesExecutions.map(async (ge) => {
+      // Fetch the Collection ID
+      const [name, version] = ge.granule.collectionId.split('___');
```
markdboyd, line 208: I would use the `deconstructCollectionId` helper from the `message` package

packages/api/bin/serveUtils.js
markdboyd: Can you point me to the fixture that has the parent/child executions? I'm not totally following

packages/api/bin/serveUtils.js
```diff
@@ -82,12 +193,57 @@ async function addReconciliationReports(reconciliationReports) {
   );
 }
 
+async function addGranulesExecutions(granulesExecutions) {
```
markdboyd, line 196: Can you help me understand what is calling this? I'm wondering if instead we can create the granuleExecution records as we create the granules or the execution records themselves

CHANGELOG.md
```diff
@@ -27,6 +27,10 @@ of response and not the raw API endpoint response
   from the RDS-Phase-2 feature branch in support of CUMULUS-2644.
   - Pulled `erasePostgresTables` method in `serve.js` implemented as part of CUMULUS-2644,
   and CUMULUS-2306 from the RDS-Phase-2 feature branch in support of CUMULUS-2644
+  - Moved `migration` directory from the `db-migration-lambda` to the `db` package and
+  updated unit test references to migrationDir to be pulled from `@cumulus/db`
+  - Updated `addXXX` methods to include writing to Postgres tables and added
+  `resetPostgresDb` method to support resetting between integration test suite runs
```
markdboyd, line 33:
```suggestion
  - Updated  serveUtils to write records to PostgreSQL tables
  - Added `resetPostgresDb` method to support resetting between integration test suite runs
```

Some of these changelog notes under the `Added` header are additions and others are changes and should be under the `Changed` header

jaortega527: Moved the  `Updated  serveUtils to write records to PostgreSQL tables` and `Updated  serveUtils to write records to PostgreSQL tables` to the changed section and split `Added `resetPostgresDb` method to support resetting between integration test suite runs` and left it in `Changed`.

packages/api/bin/serveUtils.js
jaortega527: Sure. On the dashboard integration tests we have a custom cypress plugin called `seedEverything`. Its purpose is to reset the dynamoDb and now Postgres tables between each suite of tests. We seed the database by emptying the tables and then reseeding each table in turn. We do seed granules and executions separately prior to calling `addGranulesExecutions`. We have a granulesFixture, executionsFixture, and I added granulesExecutionsFixture. I think you're asking if I could create the granules record, create the execution record, and then create the granules_executions record using the ID generated rather than having to look them up in a separate method. I believe I _could_. Truthfully, my goal was to have as little impact on existing integration tests as I could so I left the granulesFixture and executionsFixture exactly as they were and just added the granulesExecutionsFixture creating relationships  between the already existing granules and executions.

packages/api/bin/serveUtils.js
jaortega527: Sure. [Here ya go.](https://github.com/nasa/cumulus-dashboard/blob/CUMULUS-2459/cypress/fixtures/seeds/executionsFixture.json#L105). For dynamo, the order of the records didn't matter, but for Postgres, with the foreign key (parent_cumulus_id) on executions the insert order matters. Since our fixture creates some of these relationships, and I was looking to avoid making changes to existing seed data.

packages/api/bin/serveUtils.js
jaortega527: Updated to use `deconstructCollectionId`.

CHANGELOG.md
```diff
@@ -37,6 +38,11 @@ for objects that could not be processed
 - Updated async operations to handle writing records to the databases
 when output of the operation is `undefined`
 
+- **CUMULUS-2644**
+  - Moved `migration` directory from the `db-migration-lambda` to the `db` package and
+  updated unit test references to migrationDir to be pulled from `@cumulus/db`
+  - Updated `addXXX` methods to include writing to Postgres tables
```
markdboyd, line 44:
```suggestion
  - Updated `@cumulus/api/bin/serveUtils` to write records to PostgreSQL tables
```

packages/api/bin/serveUtils.js
markdboyd: So even with this approach, we are relying on the order of the data in the fixture in order to make sure that we insert the parent records before the child records, correct?

If so, that seems a little brittle. I think it might be simpler/more reliable to first `.filter()` the executions from the fixture for records that have no `parentArn` (so possible parents) and create those via Promise.all. Then `filter()` the executions which have `parentArn` and create those

packages/api/bin/serveUtils.js
markdboyd: So I think you can instead do:

- create executions first
- create granules, and as you're creating them, look up the execution PG record where `url` of the PG record matches the `granule.execution` value from the fixture. Then you can call `upsertGranuleWithExecutionJoinRecord` to write both the granule and granule execution records at the same time

https://github.com/nasa/cumulus/blob/master/packages/db/src/lib/granule.ts#L21

Then there should be no need to do a separate write or fixture for granulesExecutinos

packages/api/bin/serveUtils.js
jaortega527: That's right. Fair enough. I can add that. Is it possible for a child execution to also be a parent? Schema-wise it's possible. I'm just not sure if it's possible from the business logic side. If that's the case then this change will reduce brittleness, but not remove it entirely. If that scenario is possible/likely then I feel like I'm in the position of having to determine all the dependencies and ensure records are inserted in the correct order.

packages/api/bin/serveUtils.js
jaortega527: This was addressed as suggested.

packages/api/bin/serveUtils.js
jaortega527: The relationship of executions does have an execution that is both a parent and a child. Splitting the executions into "has parent" and "has no parent" still resulted in records being inserted before their parent occasionally, but unpredictably. I ended up adding some logic to sort the array based on the following rules:
1. if two executions have no parents then they are considered equal and placed anywhere in the array
2. if  execution a has no parent and execution b does have a parent then a must go before b
3. if execution a's arn is equal to execution b's parentArn then a must go before b
4. else a can go after b

I still forced the inserts to run synchronously, but not it doesn't matter how the executions are ordered in the fixture.

packages/api/bin/serveUtils.js
```diff
@@ -1,72 +1,197 @@
...
-    granules.map((c) =>
-      granuleModel
-        .create(c)
-        .then((granule) => indexer.indexGranule(es.client, granule, es.index)))
+    granules.map(async (g) => {
+      const dynamoRecord = await granuleModel.create(g);
+      await indexer.indexGranule(es.client, dynamoRecord, es.index);
+      const dbRecord = await translateApiGranuleToPostgresGranule(dynamoRecord, knex);
+
+      const executionPgModel = new ExecutionPgModel();
```
markdboyd, line 88: Let's pull the instantiation of the model up above the Promise.all

packages/api/bin/serveUtils.js
```diff
@@ -1,72 +1,197 @@
...
+
+    return 1;
+  });
+
+  // Since executions has a parent/child relationship with itself
+  // a fake promise is used with reduce() to force records to be created
+  // synchronously
+  const executionPgModel = new ExecutionPgModel();
+  const starterPromise = Promise.resolve(null);
+  return await executions.reduce((p, e) => p
```
markdboyd, line 168: Is the whole intent of this logic just to make sure that we create all the records serially rather than in parallel to ensure we don't get creation errors?

If so, this is a very clever solution, but I think it would be cleaner to use [`p-each-series`](There is no need to `omit` if we're comparing against the fetched record from dynamo which has the timestamps), which we already have as a dependency of the `api`

packages/api/bin/serveUtils.js
markdboyd: Got it. Great explanation. Thanks Anthony

packages/api/bin/serveUtils.js
jaortega527: Right. That is the only purpose for that logic. Interesting. i've never used p-each-series. Looking at it now.

packages/api/bin/serveUtils.js
jaortega527: Gah! Thanks for catching. Moved and pushed.

packages/api/bin/serveUtils.js
jaortega527: Wow. That's much cleaner. Updated and pushed.

markdboyd (approved):
Great work!

PR #2441, opened 2021-09-03
Update Granule with API

flamingbear:
**Summary:**. Add endpoint to update granule via API

Addresses [CUMULUS-2576: Update a granule via API](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2576)

## Changes

* Adds endpoint/granule function update that allows a user to update an existing granule.
* Adds tests to support above

## PR Checklist

- [X] Update CHANGELOG
- [X] Unit tests
- [X] Integration tests


example/spec/parallel/testAPI/granuleSpec.js
```diff
@@ -20,6 +21,7 @@ describe('The Granules API', () => {
   let config;
   let collection;
   let collectionId;
+  let discoveredGranule;
   let granuleId;
   let prefix;
   let granuleFile;
```
jennyhliu, line 27: Do you plan to get rid of buildRandomizedGranule method?

example/spec/parallel/testAPI/granuleSpec.js
```diff
@@ -81,11 +83,31 @@ describe('The Granules API', () => {
...
+      });
+
+      expect(response.statusCode).toBe(200);
+      const updatedGranuleFromApi = await getGranule({
+        prefix,
+        granuleId: modifiedGranule.granuleId,
+      });
+      updatedGranuleFromApi.execution = undefined;
+      expect(updatedGranuleFromApi).toEqual(jasmine.objectContaining(modifiedGranule));
     });
```
jennyhliu, line 112: suggest to add a test to `search` the granule to make sure the granule is added to `es`

packages/api-client/src/granules.ts
```diff
@@ -411,6 +411,36 @@ export const createGranule = async (params: {
...
+  callback?: InvokeApiFunction
+}): Promise<ApiGatewayLambdaHttpProxyResponse> => {
+  const { prefix, body, callback = invokeApi } = params;
+
+  return await callback({
+    prefix,
+    payload: {
+      httpMethod: 'PUT',
+      resource: '/{proxy+}',
+      path: '/granules',
```
jennyhliu, line 437: The path is different from other `update` endpoints, e.g. updateProvider, updateCollection, updateExecution etc., which as `providerId`, `collection name version` or `executionArn` as the parameter.

packages/api/endpoints/granules.js
```diff
@@ -461,6 +502,7 @@ router.get('/:granuleName', get);
 router.get('/', list);
 router.post('/', create);
 router.put('/:granuleName', put);
+router.put('/', update);
```
jennyhliu, line 505: same question about the `path`

example/spec/parallel/testAPI/granuleSpec.js
flamingbear: I did and then forgot. thanks.

packages/api-client/src/granules.ts
flamingbear: There already exists a `PUT /granules/:granuleName` So I left this as just a plain PUT with the granule in the json data. I thought it was cleaner to leave it out instead of trying to determine which type of put it should be (plus I left it off the POST for create)

packages/api/tests/endpoints/test-granules.js
```diff
@@ -1510,7 +1510,7 @@ test.serial('PUT with action move returns failure if more than one granule file
   filesExistingStub.restore();
 });
 
-test.serial('POST creates new granule in dynamoDB and postgres', async (t) => {
+test.serial('create (POST) creates new granule in dynamoDB and postgres', async (t) => {
   const newGranule = fakeGranuleFactoryV2({
     collectionId: t.context.collectionId,
     execution: undefined,
```
jennyhliu, line 1516: Do we want to test the case when `execution` is defined? For both `create` and `update`.

example/spec/parallel/testAPI/granuleSpec.js
flamingbear: 0bd7337

example/spec/parallel/testAPI/granuleSpec.js
flamingbear: 0bd7337

packages/api/tests/endpoints/test-granules.js
flamingbear: fcf8a2363f

packages/api-client/src/granules.ts
flamingbear: 809a81b1fe

packages/api/endpoints/granules.js
flamingbear: 809a81b1fe

packages/api/endpoints/granules.js
```diff
@@ -206,7 +248,6 @@ async function put(req, res) {
       status: 'SUCCESS',
     });
   }
-
   return res.boom.badRequest('Action is not supported. Choices are "applyWorkflow", "move", "reingest", or "removeFromCmr"');
```
Jkovarik, line 251: This feels like it might be incorrect now, given <none> is an option.  

I grok how/why we got to specifying no action in the body results in an update, but perhaps something like would be more correct: 

```suggestion
  return res.boom.badRequest('Action is not supported. Choices are "applyWorkflow", "move", "reingest", "removeFromCmr" or specify no "action" to update an existing granule');
```

packages/api-client/src/granules.ts
```diff
@@ -411,6 +411,36 @@ export const createGranule = async (params: {
   });
 };
 
+/**
+ * Update granule in cumulus.
+ * PUT /granules
```
jennyhliu, line 416:
```suggestion
 * PUT /granules/{granuleId}
```

packages/api/endpoints/granules.js
Jkovarik: It's admittedly a use-edge case, but users can and do things like put 'update' in the action field after misreading docs.     This shouldn't hold up the PR, it's a take/leave suggestion only.

example/spec/parallel/testAPI/granuleSpec.js
```diff
@@ -80,19 +90,52 @@ describe('The Granules API', () => {
...
+    it('can discover the granule directly via the API.', async () => {
+      discoveredGranule = await getGranule({
         prefix,
         granuleId,
       });
-      expect(granule).toEqual(jasmine.objectContaining(randomGranuleRecord));
+      expect(discoveredGranule).toEqual(jasmine.objectContaining(randomGranuleRecord));
+    });
+
+    it('can search the granule the API.', async () => {
```
jennyhliu, line 101:
```suggestion
    it('can search the granule via the API.', async () => {
```

packages/api/endpoints/granules.js
flamingbear: Ok, I can see where you're coming from.

packages/api/endpoints/granules.js
```diff
@@ -109,7 +152,8 @@ async function put(req, res) {
   const action = body.action;
 
   if (!action) {
-    return res.boom.badRequest('Action is missing');
+    req.body.granuleId = granuleId;
```
jennyhliu, line 155: For updating a granule, req.body.granuleId should already exist and matches with the req.params.granuleName.  We should check that.

packages/api/endpoints/granules.js
```diff
@@ -95,9 +95,52 @@ const create = async (req, res) => {
...
+    knex = await getKnexClient(),
+  } = req.testContext || {};
+  const body = req.body || {};
+
+  let existingGranule;
+  try {
+    existingGranule = await granuleModel.get({ granuleId: body.granuleId });
+  } catch (error) {
+    if (error instanceof RecordDoesNotExist) {
+      return res.boom.badRequest(`No granule found to update for ${JSON.stringify(body)}`);
```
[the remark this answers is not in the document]
flamingbear: This may be a result of debugging I was doing.  But definitely makes sense to only return the body.granuleId since that's all that was used to determine it doesn't have a record.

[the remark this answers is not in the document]
flamingbear: 61d890eaa6

packages/api/endpoints/granules.js
flamingbear: Agreed the granuleId should already exist. Are you saying we should verify that the body payload matches the url path param (granuleName) and throw an error if it doesn't match?  Because I agree if that's what you're saying.  I was thinking you always wanted to update the granule you PUT to, but if your body doesn't match, you probably don't want to post it.

jennyhliu: Yeah, we can throw an error if the body.granuleId doesn't match the req.params.granuleName

packages/api/tests/endpoints/test-granules.js
```diff
@@ -444,16 +468,16 @@ test.serial('PUT fails if action is not supported', async (t) => {
   t.true(message.includes('Action is not supported'));
 });
 
-test.serial('PUT fails if action is not provided', async (t) => {
+test.serial('PUT updates granule if action is not provided', async (t) => {
   const response = await request(app)
     .put(`/granules/${t.context.fakeGranules[0].granuleId}`)
     .set('Accept', 'application/json')
     .set('Authorization', `Bearer ${jwtAuthToken}`)
```
jennyhliu, line 475: Do you have to send a granule object here?

packages/api/tests/endpoints/test-granules.js
```diff
@@ -1585,3 +1642,138 @@ test.serial('POST return bad request if a granule is submitted with a bad collec
...
+  );
+
+  t.deepEqual(JSON.parse(modifiedResponse.text), {
+    message: `Successfully updated granule with Granule Id: ${newGranule.granuleId}`,
+  });
+
+  t.is(fetchedDynamoRecord.status, 'failed');
+  t.deepEqual(fetchedDynamoRecord.error, { some: 'error' });
+  t.is(new Date(fetchedDynamoRecord.timestamp).valueOf(), now);
+  t.is(new Date(fetchedDynamoRecord.createdAt).valueOf(), now);
```
jennyhliu, line 1715: Why `createAt` is the same as `now`?  Is there race condition here?  Also fetchedDynamoRecord.createdAt is a number, no need to convert

packages/api/tests/endpoints/test-granules.js
```diff
@@ -1585,3 +1642,138 @@ test.serial('POST return bad request if a granule is submitted with a bad collec
...
+  t.is(new Date(fetchedPostgresRecord.created_at).valueOf(), now);
+});
+
+test.serial('update (PUT) returns an updated granule with associated execution', async (t) => {
+  const now = Date.now();
+  const newGranule = fakeGranuleFactoryV2({
+    collectionId: t.context.collectionId,
+    createdAt: now,
+    timestamp: now,
+    execution: t.context.executionUrl,
```
jennyhliu, line 1728: Can you move `execution: t.context.executionUrl,` to `modifiedGranule`?  Then we can check if fetchedDynamoRecord and fetchedPostgresRecord have execution info.

packages/types/api/executions.d.ts
```diff
@@ -1,5 +1,10 @@
 export type ExecutionRecordStatus = 'completed' | 'failed' | 'running' | 'unknown';
 
+export type ExecutionProcessingTimes = {
```
jennyhliu, line 3: ExecutionProcessingTimes type is not used for execution.  Should we move it to granules.d.ts?

packages/api/endpoints/granules.js
flamingbear: 5818774029

packages/api/tests/endpoints/test-granules.js
flamingbear: If I don't then the test for granuleName === granuleId fails. Do you think there should be another test for that?  PUT granule/granuleName with no object I think it will fail with `input :granuleName (gran0313) must match body's granuleId (undefined)`

packages/api/tests/endpoints/test-granules.js
flamingbear: No reason, I can subtract some time off of the now. 6004df870b

packages/types/api/executions.d.ts
flamingbear: I just pulled it directly from the rds-phase-2 branch.  I assume they had a reason for wanting it in executions.d.ts?

packages/api/tests/endpoints/test-granules.js
flamingbear: It's spread into the modified granule.  I can add a test for it in the dynamo record, but it isn't returned in the PostgreSQL record.
Are you saying update the modifiedGranule to have a different execution?

flamingbear: Ah, take it out of the first granule and move it to the modified and make sure it's updated in both?

packages/types/api/executions.d.ts
jennyhliu: ok

packages/api/tests/endpoints/test-granules.js
jennyhliu: yeap, to make it simple. just move it to the modified granule
I thought PostgreSQL record has execution info.

packages/api/tests/endpoints/test-granules.js
jennyhliu: we could keep it as a error case.

packages/api/tests/endpoints/test-granules.js
flamingbear: 3dd344a6e2

flamingbear: 3dd344a6e2

packages/api/tests/endpoints/test-granules.js
flamingbear: db2daaa7a2

packages/api/tests/endpoints/test-granules.js
jennyhliu: 👍

jennyhliu (approved):
Great work!

PR #2443, opened 2021-09-03
Merge master into feature branch

markdboyd:


botanical (changes requested):
There are a couple tests failing with 
```
migration failed with error: Invalid migration: index.js must have both an up and down function
```
which I've been able to reproduce locally 🤔

packages/api/tests/lib/test-granule-delete.js
```diff
@@ -341,8 +366,14 @@ test.serial('deleteGranuleAndFiles() will not delete granule or S3 files if the
     { message: 'Dynamo delete failed' }
   );
 
-  // granule should still exist in DynamoDB and PostgreSQL
```
botanical, line 344: This is a nit but I prefer this comment

botanical: I think there are a few others like this as well

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -390,197 +285,242 @@ const _generateFilesFromGranule = async ({
...
-      queryFields,
-      granuleModel,
+    await writeGranuleToDynamoAndEs({
+      dynamoGranuleRecord,
       esClient,
-      updatedAt,
-      files,
+      granuleModel,
     });
+    // return granuleModel.storeGranule(dynamoGranuleRecord);
```
botanical, line 344: I believe this should be uncommented 
```suggestion
     return granuleModel.storeGranule(dynamoGranuleRecord);
```

botanical: Looking into this further, it looks like this can be removed altogether because of the storing of the granule occurs in `writeGranuleToDynamoAndEs`

Jkovarik (approved):

botanical (approved):
Thanks for addressing my feedback! Looks good to me

PR #2451, opened 2021-09-10
Add KMS decrypt permissions for ECS cluster instance role

markdboyd:
We ran into an issue on MAAP when running `sync-granule` as an ECS activity in a Step Function where it was failing to decrypt provider credentials around here in the code:

https://github.com/nasa/cumulus/blob/master/packages/ingest/src/util.ts#L9

Because that code swallows the error encountered from KMS.decrypt, it was difficult to identify the cause. 

However, I noticed that the `lambda-processing` role used by `sync-granule` when run **as a Lambda** has permissions to decrypt the KMS key.

https://github.com/nasa/cumulus/blob/master/tf-modules/cumulus/iam.tf#L142

But the ECS cluster instance role, which is used when `sync-granule` is run as an ECS activity, doesn't have permissions to decrypt the KMS key:

https://github.com/nasa/cumulus/blob/master/tf-modules/cumulus/ecs_cluster.tf#L9

So this PR adds that KMS decrypt permissions for the ECS cluster instance role and updates the code to log the error if/when this happens.

### DIFF
--- a/packages/ingest/src/util.ts
+++ b/packages/ingest/src/util.ts
@@ -4,7 +4,7 @@ import { S3KeyPairProvider } from '@cumulus/common/key-pair-provider';
 
 import Logger from '@cumulus/logger';
 
-const logger = new Logger({ sender: '@cumulus/ingest/utils' });
+const logger = new Logger({ sender: '@cumulus/ingest/util' });
 
 export const decrypt = async (
   ciphertext: string
--- a/packages/ingest/src/util.ts
+++ b/packages/ingest/src/util.ts
@@ -2,12 +2,17 @@ import mime from 'mime-types';
 import * as KMS from '@cumulus/aws-client/KMS';
 import { S3KeyPairProvider } from '@cumulus/common/key-pair-provider';
 
+import Logger from '@cumulus/logger';
+
+const logger = new Logger({ sender: '@cumulus/ingest/utils' });
+
 export const decrypt = async (
   ciphertext: string
 ): Promise<string | undefined> => {
   try {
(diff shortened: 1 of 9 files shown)


nemreid (approved):

packages/ingest/src/util.ts
```diff
@@ -2,12 +2,17 @@ import mime from 'mime-types';
 import * as KMS from '@cumulus/aws-client/KMS';
 import { S3KeyPairProvider } from '@cumulus/common/key-pair-provider';
 
+import Logger from '@cumulus/logger';
+
+const logger = new Logger({ sender: '@cumulus/ingest/utils' });
```
botanical, line 7:
```suggestion
const logger = new Logger({ sender: '@cumulus/ingest/util' });
```

tf-modules/cumulus/ecs_cluster.tf
```diff
@@ -138,6 +138,11 @@ data "aws_iam_policy_document" "ecs_cluster_instance_policy" {
       var.rds_user_access_secret_arn
     ]
   }
+
+  statement {
+    actions   = ["kms:Decrypt"]
+    resources = [module.archive.provider_kms_key_arn]
+  }
```
botanical, line 145: Is it not important to note this in the changelog?

botanical (approved):
Had two comments but looks good otherwise

tf-modules/cumulus/ecs_cluster.tf
markdboyd: It is. Will do

PR #2455, opened 2021-09-14
CUMULUS-2583

dopeters:

## Changes

* Updates to QueueGranules and granule write logic to avoid race conditions of updateGranule api call

## PR Checklist

- [ ] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


example/cumulus-tf/queue_granules_workflow.asl.json
```diff
@@ -0,0 +1,66 @@
...
+      "Parameters": {
+        "cma": {
+          "event.$": "$",
+          "task_config": {
+            "queueUrl": "{$.meta.queueUrl}",
+            "provider": "{$.meta.provider}",
+            "internalBucket": "{$.meta.buckets.internal.name}",
+            "stackName": "{$.meta.stack}",
+            "granuleIngestWorkflow": "${ingest_granule_workflow_name}",
+            "childWorkflowMeta": { "staticValue": "aStaticValue", "interpolatedValueStackName": "{$.meta.stack}" },
```
nemreid, line 15: You don't need this line if you aren't testing for it in the spec test, which it doesn't look like you are.

example/spec/parallel/queueGranules/QueueGranulesSpec.js
```diff
@@ -0,0 +1,204 @@
...
+      ),
+    ]);
+  });
+
+  it('completes execution with success status', () => {
+    expect(workflowExecution.status).toEqual('completed');
+  });
+
+  describe('the QueueGranules Lambda function', () => {
+    it('sets granule status to queued', async () => {
```
nemreid, line 172: nice 💯

packages/api/tests/models/granules/test-granule-reporting.js
```diff
@@ -253,6 +267,96 @@ test('_storeGranuleRecord() will allow a running status to replace a failed stat
...
+    {
+      ...granule,
+      status: 'running',
+      createdAt: updateTime,
+      execution: 'new-execution-url',
+    }
+  );
+});
+
+test('_storeGranuleRecord() will allow a queued status to replace a running status for a new execution', async (t) => {
```
nemreid, line 330: Is there a test for something like `_storeGranuleRecord() will not allow a queued status to replace any other status for same execution`? I guess since we don't know the executionArn now at time of status update to queued, that may still be an issue/hard to determine...

tasks/queue-granules/index.js
```diff
@@ -81,7 +81,22 @@ async function queueGranules(event) {
...
+        (queuedGranule) => granulesApi.updateGranule({
+          prefix: event.config.stackName,
+          body: {
+            collectionId: constructCollectionId(
+              queuedGranule.dataType,
+              queuedGranule.version
+            ),
+            granuleId: queuedGranule.granuleId,
+            status: 'queued',
+          },
```
nemreid, line 95: Since this body does not include execution information, do we actually have any situation where we send a `queued` update with execution info?

nemreid (commented):
A lot of this is really nice, I do have a concern about a remaining possible edge case..

packages/api/tests/models/granules/test-granule-reporting.js
```diff
@@ -253,6 +267,96 @@ test('_storeGranuleRecord() will allow a running status to replace a failed stat
   );
 });
 
+test('_storeGranuleRecord() will allow a completed status to replace a queued status for a new execution', async (t) => {
```
markdboyd, line 270: This test is written correctly according to the test name, but I don't think this is a valid test case. The expected case I think is that a "completed" status should be able to replace a "queued" status for the same execution, not a new one.

packages/db/tests/models/test-granule-model.js
```diff
@@ -402,6 +402,41 @@ test('GranulePgModel.upsert() will allow a running status to replace a non-runni
   t.is(record.status, 'running');
 });
 
+test('GranulePgModel.upsert() will allow a queued status to replace a non-queued status for different execution', async (t) => {
```
markdboyd, line 405: For completeness, I think we would also want the unit tests to prove that a "completed" status can replace a "queued" status and that a "running' status can replace a "queued" status

packages/db/tests/models/test-granule-model.js
```diff
@@ -402,6 +402,41 @@ test('GranulePgModel.upsert() will allow a running status to replace a non-runni
...
+    t.context.knex,
+    fakeExecutionRecordFactory({ status: 'running' })
+  );
+
+  const updatedGranule = {
+    ...granule,
+    status: 'queued',
+  };
+
+  await granulePgModel.upsert(knex, updatedGranule, newExecutionCumulusId);
```
markdboyd, line 431: Nit: can we use `upsertGranuleWithExecutionJoinRecord` for consistency?

```suggestion
  await upsertGranuleWithExecutionJoinRecord(knex, granule, newExecutionCumulusId);
```

packages/db/src/translate/granules.ts
```diff
@@ -46,7 +46,7 @@ export const translateApiGranuleToPostgresGranule = async (
         { name: dynamoRecord.pdrName }
       )
       : undefined,
-    provider_cumulus_id: dynamoRecord.provider ? await providerPgModel.getRecordCumulusId(
+    provider_cumulus_id: (dynamoRecord.provider && dynamoRecord.status !== 'queued') ? await providerPgModel.getRecordCumulusId(
```
markdboyd, line 49: I'm not sure that I understand what this logic is doing

packages/api/tests/endpoints/test-granules.js
```diff
@@ -1912,37 +1912,20 @@ test.serial('update (PUT) returns bad request when the path param granuleName do
...
-
-  t.is(response.status, 200);
-  t.deepEqual(JSON.parse(response.text), {
-    message: `Successfully updated granule with Granule Id: ${runningGranuleId}`,
-  });
-});
-
-test.serial('update (PUT) does not set status to queued if not running', async (t) => {
-  const granuleId = t.context.fakeGranules[0].granuleId;
+test.serial('update (PUT) can set granule status to queued', async (t) => {
```
markdboyd, line 1915: I think we also need/want a unit test to prove that the PUT endpoint can create a new granule with a status of "queued", in addition to updating a granule to have a status of queued

example/spec/serial/DiscoverGranulesHttpSpec.js
```diff
@@ -113,23 +113,14 @@ describe('The Discover Granules workflow with http Protocol', () => {
...
-        await deleteGranule({
-          prefix: config.stackName,
-          granuleId: granule.granuleId,
-        });
+        await waitForGranuleAndDelete(
+          config.stackName,
+          granule.granuleId,
+          'completed',
+          {
+            retries: 5,
```
markdboyd, line 121: Just curious if we know why the retry behavior needed to be updated here

packages/api/endpoints/granules.js
```diff
@@ -110,36 +109,23 @@ const update = async (req, res) => {
...
-      {
-        retries,
-      }
-    );
+    existingGranule = await granuleModel.get({ granuleId: body.granuleId });
   } catch (error) {
-    if (error instanceof RecordDoesNotExist) {
-      return res.boom.notFound(`No granule found to update for ${body.granuleId}`);
+    if (body.status === 'queued') {
+      existingGranule = {};
```
markdboyd, line 118: Is this logic here basically to handle the fact that we might be creating a new "queued" granule via this endpoint, so we might hit this catch but don't want to return an error from the endpoint?

markdboyd: If so, then I think we're possibly confusing what a PUT endpoint should do. A PUT should create a new resource or update an existing one, meaning it shouldn't be relying on a check like this to see if a record already exists

https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT

Now, implementing the endpoint that way may require some refactoring of our write logic, but it might be worth trying because it's ultimately a lot cleaner and easier to understand, and also compliant with how PUT endpoints generally work.

markdboyd (changes requested):
Looks good! I have a few questions about the approach and some requests for unit test changes

packages/db/tests/models/test-granule-model.js
dopeters: Since we're testing the `upsert` method here, shouldn't we call it directly?

markdboyd: Yeah you're right. My bad.

packages/db/src/translate/granules.ts
dopeters: I was getting some failures in the QueueGranules step related to the provider record not being found for the newly created queued granules

packages/api/endpoints/granules.js
dopeters: That makes a lot of sense and was actually a really easy change. I kept in the try/catch so that if there is some other issue retrieving the record, it will still throw an error, but it'll never throw the 'record does not exist error'

example/spec/serial/DiscoverGranulesHttpSpec.js
dopeters: Nope, removed it.

packages/db/src/translate/granules.ts
dopeters: And got rid of this, not needed

packages/api/tests/endpoints/test-granules.js
markdboyd: It's a little clearer to not conflate PUT/update

packages/api/tests/endpoints/test-granules.js
```diff
@@ -1912,37 +1912,38 @@ test.serial('update (PUT) returns bad request when the path param granuleName do
...
   t.is(response.status, 200);
   t.deepEqual(JSON.parse(response.text), {
-    message: `Successfully updated granule with Granule Id: ${runningGranuleId}`,
+    message: `Successfully updated granule with Granule Id: ${granuleId}`,
   });
 });
 
-test.serial('update (PUT) does not set status to queued if not running', async (t) => {
-  const granuleId = t.context.fakeGranules[0].granuleId;
+test.serial('update (PUT) can create a new granule with status queued', async (t) => {
```
markdboyd, line 1932:
```suggestion
test.serial('PUT can create a new granule with status queued', async (t) => {
```

packages/db/tests/models/test-granule-model.js
```diff
@@ -402,6 +402,93 @@ test('GranulePgModel.upsert() will allow a running status to replace a non-runni
   t.is(record.status, 'running');
 });
 
+test('GranulePgModel.upsert() will allow a queued status to replace a non-queued status for the same execution', async (t) => {
```
markdboyd, line 405: I don't think this behavior is what we want. We only want `queued` to replace a non-queued status if it is a different execution or it is a new record. We may be more protected from this now with how you've refactored queue-granules, but our write logic still shouldn't allow it

packages/api/endpoints/granules.js
markdboyd: Just to be clear, I don't think we should need/have the `granuleModel.get` logic in this endpoint at all. A PUT endpoint should accept a full body representing a granule to be written, so there should be no need to look anything up from the database before doing the write.

Is there some reason that wouldn't be feasible here?

packages/api/tests/models/granules/test-granule-reporting.js
```diff
@@ -253,6 +267,96 @@ test('_storeGranuleRecord() will allow a running status to replace a failed stat
...
+    {
+      ...granule,
+      status: 'completed',
+      createdAt: updateTime,
+      execution: 'new-execution-url',
+    }
+  );
+});
+
+test('_storeGranuleRecord() will allow a running status to replace a queued status for a new execution', async (t) => {
```
markdboyd, line 300: This test case isn't bad to have, but I think we at least want another unit test to prove that a running status can replace a queued status for the same execution, since that is the typically expected case

packages/api/tests/models/granules/test-granule-reporting.js
markdboyd: I think we also want a test to prove that a queued status cannot replace a running/completed status for the same execution

packages/db/tests/models/test-granule-model.js
markdboyd: I'm assuming this will require edits to the conditional write logic for the PG model. And we may need to do the same thing for the Dynamo write logic

packages/api/tests/models/granules/test-granule-reporting.js
nemreid: @markdboyd It's fair enough to test, but I am concerned that we won't see that side of things being used in `queue-granules` because it's not sending any execution ID with the status update.

nemreid: Perhaps we should be avoiding the race condition in the update logic and indeed do the update after the call to `enqueue...` so that we have and are able to include the executionArn in the update.

markdboyd: Regardless of how we write `queue-granules`, I think we need to be sure that our core database write logic does the correct thing, especially since we now have an endpoint that people can call at any time when the database could be in any state

packages/api/endpoints/granules.js
markdboyd: Discussed offline and changing this would require larger refactoring to this endpoint logic which isn't essential for this PR, so resolving this comment

markdboyd (approved):

tasks/queue-granules/tests/index.js
```diff
@@ -709,3 +709,30 @@ test.serial('If a childWorkflowMeta is provided, it is passed through to the mes
...
+  const updateGranuleMock = sinon.spy(({ body }) => body.createdAt);
+  const enqueueGranuleIngestMessageMock = sinon.spy(() => Date.now());
+
+  const testMocks = {
+    updateGranuleMock,
+    enqueueGranuleIngestMessageMock,
+  };
+
+  await queueGranules(event, testMocks);
+  t.assert(updateGranuleMock.returnValues[0] <= enqueueGranuleIngestMessageMock.returnValues[0]);
```
nemreid, line 737: I think we actually want

```suggestion
  t.assert(updateGranuleMock.returnValues[0] < enqueueGranuleIngestMessageMock.returnValues[0]);
```

because if they somehow equaled each other the overwrite from e.g. `running` to `queued` would be allowed.

PR #2456, opened 2021-09-15
Cumulus-2594 Making CMR Oauth information available over API

kkelly51:
Addresses [CUMULUS-2594: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2594)

Added cmr_oauth_provider info to /instanceMeta endpoint to be displayed on the dashboard.

### DIFF
--- a/packages/api/endpoints/instance-meta.js
+++ b/packages/api/endpoints/instance-meta.js
@@ -14,6 +14,7 @@ function instanceMetadata(req, res) {
     cmr: {
       provider: process.env.cmr_provider,
       environment: process.env.CMR_ENVIRONMENT || 'UAT',
+      oauth_provider: process.env.cmr_oauth_provider || '',
     },
     cumulus: {
       stackName: process.env.stackName,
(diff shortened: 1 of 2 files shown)


packages/api/endpoints/instance-meta.js
```diff
@@ -14,6 +14,7 @@ function instanceMetadata(req, res) {
     cmr: {
       provider: process.env.cmr_provider,
       environment: process.env.CMR_ENVIRONMENT || 'UAT',
+      oauth_provider: process.env.cmr_oauth_provider || '',
```
kkelly51, line 17: maybe I should make "Earthdata" the OR because that is technically the default

flamingbear: No, I just slacked you somthing like this:  I’m not sure it should switch.  This metadata endpoint probably shouldn’t even return ‘UAT’ if process.env.CMR_ENVIRONMENT is not set.
 This endpoint should represent how cumulus is actually configured, and the defaults should flow to here, if not, we don't want to lie to the user.

flamingbear (approved):
Looks good, don't update the default value.

PR #2457, opened 2021-09-15
CUMULUS-2676: control concurrency for DiscoverGranules duplicate checking (needs review)

nemreid:

Addresses [CUMULUS-2676](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2676)

## Changes

- **CUMULUS-2676**
  - The `discover-granules` task has been updated to limit concurrency on checks to identify and skip already ingested granules in order to prevent issues with the private API endpoint and Lambda API request and concurrency limits. This may increase the task's overall runtime when large numbers of granules are discovered. If you are facing Lambda timeout errors with this task, we recommend converting your `discover-granules` task to an ECS activity.

- Changelog note for CUMULUS-2583

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Notable Changes
+
+- **CUMULUS-2583**
+  - The `queue-granules` task now updates granule status to `queued` when a granule is queued. In order to prevent issues with the private API endpoint and Lambda API request and concurrency limits, this may increase the task's overall runtime when large numbers of granules are discovered. If you are facing Lambda timeout errors with this task, we recommend converting your `queue-granules` task to an ECS activity.
+- **CUMULUS-2676**
+  - The `discover-granules` task has been updated to limit concurrency on checks to identify and skip already ingested granules in order to prevent issues with the private API endpoint and Lambda API request and concurrency limits. This may increase the task's overall runtime when large numbers of granules are discovered. If you are facing Lambda timeout errors with this task, we recommend converting your `discover-granules` task to an ECS activity.
+
 ### Added
 
 - **CUMULUS-2576**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Notable Changes
 
 - **CUMULUS-2583**
-  - The `queue-granules` task now updates granule status to `queued` when a granule is queued. In order to prevent issues with the private API endpoint and Lambda API request and concurrency limits, this functionality runs with limited concurrency, which may increase the task's overall runtime when large numbers of granules are being queued. If you are facing Lambda timeout errors with this task, we recommend converting your `queue-granules` task to an ECS activity.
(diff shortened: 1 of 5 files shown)


tasks/queue-granules/index.js
```diff
@@ -64,6 +65,7 @@ async function queueGranules(event) {
       });
       if (executionArn) {
         const queuedGranule = {
+          collectionId: constructCollectionId(granule.dataType, granule.version),
```
nemreid, line 68: Fix for the issue where `updateGranule` cannot do a PG write of a non-existent granule without a `collectionId`

tasks/discover-granules/index.js
```diff
@@ -231,12 +233,16 @@ const checkGranuleHasNoDuplicate = async (granuleId, duplicateHandling) => {
...
  */
-const filterDuplicates = async (granuleIds, duplicateHandling) => {
-  const keysPromises = granuleIds.map((key) =>
-    checkGranuleHasNoDuplicate(key, duplicateHandling));
+const filterDuplicates = async (granuleIds, duplicateHandling, concurrency) => {
+  const keysPromises = await pMap(
+    granuleIds,
+    (key) => checkGranuleHasNoDuplicate(key, duplicateHandling),
+    { concurrency }
+  );
```
nemreid, line 246: Wrap in pMap with default concurrency of 3.

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Notable Changes
+
+- **CUMULUS-2583**
```
markdboyd, line 11: These changelog notes are 💯

tasks/discover-granules/schemas/config.json
```diff
@@ -90,6 +90,11 @@
       "description": "flag to set DiscoverGranules duplicate handling behavior",
       "default": "replace",
       "type": "string"
+    },
+    "concurrency": {
+      "description": "concurrency used when querying Cumulus API for already discovered granules",
+      "default": 3,
```
markdboyd, line 96: Didn't we find out that the CMA schema enforcement actually doesn't set default values when none are provided? I think we need to apply this default in code and not in the schema definition

markdboyd: Oh, I see that you already are enforcing it in the code. So do we need this default in the schema?

nemreid: I put in a default because the other values in this config also have a default, presumably just to document what those values are.

tasks/discover-granules/index.js
```diff
@@ -231,12 +233,16 @@ const checkGranuleHasNoDuplicate = async (granuleId, duplicateHandling) => {
...
  * @param {string} duplicateHandling - flag that defines this function's behavior (see description)
+ * @param {number} concurrency - pMap concurrency
  *
  * @returns {Array.string} returns granuleIds parameter with applicable duplciates removed
  */
-const filterDuplicates = async (granuleIds, duplicateHandling) => {
-  const keysPromises = granuleIds.map((key) =>
-    checkGranuleHasNoDuplicate(key, duplicateHandling));
+const filterDuplicates = async (granuleIds, duplicateHandling, concurrency) => {
+  const keysPromises = await pMap(
```
markdboyd, line 241: Hmm - does this make sense with the line below that does a `Promise.all`? Previously we were doing a `map` on granuleIds to prepare an array of promises, then awaiting them all with `Promise.all`

But now we're explicitly awaiting the pMap operation and the mapped promises, so I think the Promise.all below is unnecessary? Am I missing something?

nemreid: I don't think you are missing something. Will push a change.

nemreid: https://github.com/nemreid/cumulus/commit/c4e5faa5cfd6182086a7e45b7f1d99f32ac1b5cf

PR #2483, opened 2021-09-27
Improve timing of eraseDynamoTables in serve.js for Providers

jaortega527:
**Summary:** Improve timing of eraseDynamoTables in serve.js for Providers. Intended to prevent intermittent dashboard integration test failures caused by an empty provider dynamo table after a re-seed.

## Changes

* Moved the call to `providerModel.deleteProviders()` out of the `Promise.all()` to ensure provider records are deleted prior to reseeding.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


jennyhliu:
I believe the granule deletion failure caused the `seedEverything` in dashboard to fail.
`npm run seed-database
DeletePublishedGranule: You cannot delete a granule that is published to CMR. Remove it from CMR first
    at Granule.delete (/Users/jhliu/cumulus-dashboard/node_modules/@cumulus/api/models/granules.js:555:13)
    at /Users/jhliu/cumulus-dashboard/node_modules/@cumulus/api/models/granules.js:567:18
    at Array.map (<anonymous>)
    at Granule.deleteGranules (/Users/jhliu/cumulus-dashboard/node_modules/@cumulus/api/models/granules.js:566:45)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Promise.all (index 3)
    at async eraseDynamoTables (/Users/jhliu/cumulus-dashboard/node_modules/@cumulus/api/bin/serve.js:377:5)
    at async eraseDataStack (/Users/jhliu/cumulus-dashboard/node_modules/@cumulus/api/bin/serve.js:399:3)
    at async Promise.all (index 0)
    at async Promise.all (index 0)
{"level":"info","message":"Created esIndex localrun-es","sender":"@cumulus/es-client/indexer","timestamp":"2021-09-28T13:43:20.299Z"}
`

jennyhliu:
We can update `eraseDynamoTables` to either
- find a way to delete published granule from table, or
- move `granulesModel.deleteGranules()` outside of Promise.all, catch and log the granule deletion error.  It's possible the granules won't get completely refreshed though.

jaortega527:
Absolutely. I wasn't intending on addressing that error as part of this PR, but am more than happy to. 

It's something that has been a part of the seeding since upgrading the version of the API from 8.x.x to 9.x.x. Since we manage the seeding of the dynamo and now the postgres tables through `serveUtils.js` and `serve.js` everything is run through our custom validation. That error is in a couple of places, but is thrown when the `published` property for a granule is set to `true` and an attempt to delete it is made. Since our seeding fixtures have a granule with published set to `true` that error is thrown. I don't believe moving the call around will make a difference with regards to that error. Since it isn't truly published, but only flagged as published we could also edit the `published` value prior to reseeding to prevent that error from being thrown at all. 

As of right now, the error is thrown but the reseed continues and completes.

jennyhliu:
`SeedEverything` fails when the granule error occurs, so the providers don't get populated, this is my theory. 
If we call `deletedGranules` separately and catch the error, unless new granules are added during the test which maybe not get deleted, everything else will be fine.

packages/api/bin/serve.js
```diff
@@ -374,9 +374,9 @@ async function eraseDynamoTables(stackName, systemBucket) {
 
   try {
     await rulesModel.deleteRules();
+    await providerModel.deleteProviders();
     await Promise.all([
```
jennyhliu, line 378: I suggest instead of moving the `await providerModel.deleteProviders()` outside,
```suggestion
    await Promise.allSettled([
```

jennyhliu (approved):

PR #2532, opened 2021-10-20
CUMULUS-2683:configure a default chunk size for multi-part uploads across the system

jennyhliu:

Addresses [CUMULUS-2683: configure a default chunk size for multi-part uploads across the system](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2683)

## Changes

* Added `default_s3_multipart_chunksize_mb` setting to the `MoveGranules` lambda function.
* Added `default_s3_multipart_chunksize_mb` tfvars to the archive and cumulus terraform modules.
* Added optional parameter `chunkSize` to `@cumulus/aws-client/S3.moveObject` and `@cumulus/aws-client/S3.multipartCopyObject` to set the chunk size of the S3 multipart uploads.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


packages/aws-client/src/S3.ts
```diff
@@ -968,7 +968,8 @@ export const multipartCopyObject = async (
     sourceObject?: AWS.S3.HeadObjectOutput,
     ACL?: AWS.S3.ObjectCannedACL,
     copyTags?: boolean,
-    copyMetadata?: boolean
+    copyMetadata?: boolean,
+    maxChunkSize?: number
```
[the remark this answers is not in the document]
flamingbear: would it be terrible to rename the parameter to chunkSize instead of maxChunkSize?    It's technically the max size for the chunks we build but it's not the maximum possible size.

packages/aws-client/src/S3.ts
```diff
@@ -1001,7 +1003,7 @@ export const multipartCopyObject = async (
       throw new Error(`Unable to determine size of s3://${sourceBucket}/${sourceKey}`);
     }
 
-    const chunks = S3MultipartUploads.createMultipartChunks(objectSize);
+    const chunks = S3MultipartUploads.createMultipartChunks(objectSize, maxChunkSize);
```
flamingbear, line 1006: I wonder if we should check that we haven't gone over the maximum number of chunks?  "A multipart upload requires that a single file is uploaded in not more than 10,000 distinct parts. You must be sure that the chunksize that you set balances the part file size and the number of parts."   If that array is longer than 10000 we could throw an error?

packages/aws-client/src/S3.ts
```diff
@@ -1063,7 +1065,8 @@ export const moveObject = async (
     destinationBucket: string,
     destinationKey: string,
     ACL?: AWS.S3.ObjectCannedACL,
-    copyTags?: boolean
+    copyTags?: boolean,
+    maxChunkSize?: number
```
flamingbear, line 1069: Yeah I think we should name this chunkSize unless you can give me a good reason not to?

packages/aws-client/tests/S3/test-multipartCopyObject.js
```diff
@@ -89,6 +90,33 @@ test('multipartCopyObject() copies a file between buckets', async (t) => {
   t.truthy(etag, 'Missing etag in copy response');
 });
 
+test('multipartCopyObject() fails when the maxChunkSize is smaller than the minimum allowed object size', async (t) => {
```
flamingbear, line 93: should we have a test for a chunksize too large?  or too many parts?

packages/aws-client/tests/S3/test-multipartCopyObject.js
```diff
@@ -107,6 +135,7 @@ test("multipartCopyObject() sets the object's ACL", async (t) => {
     destinationBucket,
     destinationKey,
     ACL: 'public-read',
+    maxChunkSize: 5 * MB,
```
flamingbear, line 138: do you need this?  Is it testing anything?

flamingbear (changes requested):
Most of this looks good.  I'm not sure how to test that it's being used properly, more than the test that shows, it errors when you have a chunksize that is too small (and perhaps a test for too large)

I do feel you should change the parameter form maxChunkSize to chunkSize or multipartChunkSize though.

packages/aws-client/src/S3.ts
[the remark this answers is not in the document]
jennyhliu: renamed maxChunkSize to chunkSize, added docstring for chunkSize
[1cefbe8](https://github.com/nasa/cumulus/pull/2532/commits/1cefbe8da7441fb4cdf39730eff3d8c8d89dbc2f)
[5aa4517](https://github.com/nasa/cumulus/pull/2532/commits/5aa4517a8f70ea68154374cbce2e0c086a799f79)

packages/aws-client/src/S3.ts
jennyhliu: Since the minimal chunk size is 5MB, I tried to create 55,000MB size file for testing, and got 502 error from localstack, and I think 50GB file is too big for testing.

packages/aws-client/tests/S3/test-multipartCopyObject.js
jennyhliu: chunkSize is 5 MB to 5 GB.  we will need a file > 5GB to test the maximum chunk size.

packages/aws-client/tests/S3/test-multipartCopyObject.js
jennyhliu: removed

packages/aws-client/tests/S3/test-multipartCopyObject.js
flamingbear: ok.  yeah guess we don't really need that then.

packages/aws-client/src/S3.ts
flamingbear: yup I didn't do that math.

flamingbear: And if the default value is 250MBs, that would be a 2.5TB file before it was broken.  I wonder if AWS would have a good enough error message for us to know we broke the limits

jennyhliu: My concern is that aws limit could change, and we don't want to hardcode the limit.

flamingbear: good point. I think this is good to go. I'll approve

flamingbear (approved):

flamingbear:
Looks good!  thanks

PR #2542, opened 2021-10-27
CUMULUS-2388: Update task schemas for consistency

markdboyd:


Jkovarik (approved):
👍🏻   I did a last once-over on the code and validated the integration test runs looked as expected in AWS.      Nice work!

PR #2565, opened 2021-11-10
[CUMULUS-2663] Add unit test for writeGranuleFromApi

botanical:

Addresses [CUMULUS-2663: Update PUT /granules endpoint to post SNS messages and write to Elasticsearch](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2663)

## Changes

* Update unit test to check that writeGranuleFromApi stores granule in ES

## PR Checklist

- [ ] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


markdboyd (approved):
Normally I would say that we need to explicitly verify that the PUT /granules endpoint itself has the correct behavior, otherwise we're relying on the fact that the endpoint is written to use `writeGranuleFromApi`, which could change. 

However, it looks like we already have that test coverage:

https://github.com/nasa/cumulus/blob/feature/rds-phase-2/packages/api/tests/endpoints/test-granules.js#L1748-L1945

So I think we're good to here, just wanted to mention that. Great work!

PR #2568, opened 2021-11-15
CUMULUS-2694: Remove dynamo only writes

markdboyd:

Addresses [CUMULUS-2694: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2694)

## Changes

- Updated database write logic in `sfEventSqsToDbRccords` to log message if Cumulus
workflow message is from pre-RDS deployment but still attempt parallel writing to DynamoDB
and PostgreSQL
- Updated database write logic in `sfEventSqsToDbRccords` to throw error if requirements to write execution to PostgreSQL cannot be met
- Removed `@cumulus/api/models/granules.storeGranulesFromCumulusMessage()` method

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests


packages/api/lib/writeRecords/utils.js
```diff
@@ -40,9 +40,7 @@ const isPostRDSDeploymentExecution = (cumulusMessage) => {
   }
 };
 
-const isFailedLookupError = (error) =>
-  error instanceof InvalidArgument
-  || error instanceof RecordDoesNotExist;
+const isFailedLookupError = (error) => error instanceof RecordDoesNotExist;
```
botanical, line 43: Why did you remove `error instanceof InvalidArgument`?

packages/api/lib/writeRecords/utils.js
```diff
@@ -139,12 +140,13 @@ const getMessageProviderCumulusId = async (
   try {
     const providerId = getMessageProviderId(cumulusMessage);
     if (isNil(providerId)) {
-      throw new InvalidArgument('Could not find provider ID in message');
+      log.info('Could not find provider ID in message');
```
botanical, line 143:
```suggestion
      log.info('Could not find provider ID in message, skipping');
```

botanical (approved):
Couple comments but they aren't really blocking. Looks good to me

packages/api/lib/writeRecords/utils.js
markdboyd: So this is all cleanup that is mostly unrelated to the goal of the ticket. That being said, the previous implementation of `getAsyncOperationCumulusId` looked like

```
const getAsyncOperationCumulusId = async (
  asyncOperationId,
  knex,
  asyncOperationPgModel = new AsyncOperationPgModel()
) => {
  try {
    if (isNil(asyncOperationId)) {
      throw new InvalidArgument('There is no async operation ID to lookup on the message, skipping');
    }
    return await asyncOperationPgModel.getRecordCumulusId(
      knex,
      {
        id: asyncOperationId,
      }
    );
  } catch (error) {
    if (isFailedLookupError(error)) {
      log.info(error);
      return undefined;
    }
    throw error;
  }
};
```

The problem with this is that when we did `log.info(error)`, it included a nasty stack trace and the log entry made it look like something had failed, even when it hadn't. This is confusing because the whole point of this helper is to look an object in the database which is optional - you don't **have to have an async operation to write an execution**. I've run into this confusion several times when debugging for myself or with others. 

And really, throwing and catching the `InvalidArgument` error was just a cheap way to return `undefined`. So instead of that, I decided to just explicitly return `undefined` when there is no information to lookup and then to do `log.info(error.message)` if any error is caught from the lookup, which avoids the ugly and possibly confusing stack trace. 

Make sense?

PR #2574, opened 2021-11-17
[CUMULUS-2744] Add associated granules to execution status

dopeters:

Addresses [CUMULUS-2744: Fix GranuleID on execution page for execution with multiple Granules](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2744)

## Changes

* Finds associated granules for execution status that is retrieved from the Step Function API
  * Previously, this was only added to execution status for executions retrieved from the database. Reworked some of the logic to work for both

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests


kkelly51 (approved):
👍

PR #2587, opened 2021-11-29
CUMULUS-2661: Update executions PUT endpoint to send SNS message

markdboyd:

Addresses [CUMULUS-2661: Update executions PUT endpoint to send SNS message](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2661)

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/packages/api/tests/endpoints/test-executions.js
+++ b/packages/api/tests/endpoints/test-executions.js
@@ -122,7 +122,9 @@ test.before(async (t) => {
   await accessTokenModel.createTable();
 
   jwtAuthToken = await createFakeJwtAuthToken({ accessTokenModel, username });
-  const { knex, knexAdmin } = await generateLocalTestDb(testDbName, migrationDir);
+  const { knex, knexAdmin } = await generateLocalTestDb(testDbName, migrationDir, {
+    KNEX_DEBUG: 'true',
+  });
   t.context.knex = knex;
   t.context.knexAdmin = knexAdmin;
   process.env = {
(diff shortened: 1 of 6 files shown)


packages/api/tests/endpoints/test-executions.js
```diff
@@ -1660,3 +1678,55 @@ test.serial('PUT /executions with non-existing parentArn still updates the execu
...
 });
+
+test.serial('PUT /executions publishes message to SNS topic', async (t) => {
+  const {
+    executionPgModel,
+    knex,
+    QueueUrl,
+  } = t.context;
+
+  const newExecution = fakeExecutionFactoryV2({
```
botanical, line 1689: I'd consider using the fakeExecutionRecordFactory function

markdboyd: I don't think that will work. The POST /executions endpoint calls `writeExecutionRecordFromApi`:

https://github.com/nasa/cumulus/blob/feature/rds-phase-2/packages/api/endpoints/executions.js#L61

And that function expects the incoming record to be a Dynamo record, which is then translated to the PostgreSQL record:

https://github.com/nasa/cumulus/blob/feature/rds-phase-2/packages/api/lib/writeRecords/write-execution.js#L216-L228

So we can't use `fakeExecutionRecordFactory` because it would be sending a PostgreSQL-style record to the endpoint, not a Dynamo/API style record as it expects

botanical (approved):
Looks good to me!

PR #2589, opened 2021-11-30
[CUMULUS-2662] Remove call to addToLocalES and add unit tests for POST granules endpoint

botanical:

Addresses [CUMULUS-2662: Update POST /granules endpoint to post SNS messages and write to Elasticsearch](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2662)

## Changes

* Removed call to addToLocalES since we default write to ES
* Added unit tests to check that POST writes to ES

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


botanical:
I don't believe there is a need to add unit tests that SNS messages are published within test-granules.js because it is unit tested here (since the call to publish SNS messages is made in `_writeGranule()`):https://github.com/nasa/cumulus/blob/05599383accae4c5453d3aacd348017f1310f333/packages/api/tests/lib/writeRecords/test-write-granules.js#L1559 but I am open to adding a unit test

packages/api/endpoints/granules.js
```diff
@@ -88,9 +88,6 @@ const create = async (req, res) => {
 
   try {
     await createGranuleFromApi(granule, knex, esClient);
-    if (inTestMode()) {
```
markdboyd, line 91: Can we check to see if `addToLocalES` is used elsewhere and if not delete it altogether? I don't want to keep unnecessary code around

packages/api/tests/endpoints/test-granules.js
```diff
@@ -1726,13 +1730,16 @@ test.serial('create (POST) creates new granule with associated execution in dyna
       collection_cumulus_id: t.context.collectionCumulusId,
     }
   );
-
+  const fetchedESRecord = await t.context.esGranulesClient.get(
```
markdboyd, line 1733: Need to update the test title here as well

packages/api/tests/endpoints/test-granules.js
```diff
@@ -1669,7 +1669,7 @@ test.serial('PUT with action move returns failure if more than one granule file
   filesExistingStub.restore();
 });
 
-test.serial('create (POST) creates new granule without an execution in dynamoDB and postgres', async (t) => {
+test.serial('create (POST) creates new granule without an execution in DynamoDB, Postgres, and Elasticsearch', async (t) => {
```
markdboyd, line 1672:
```suggestion
test.serial('create (POST) creates new granule without an execution in DynamoDB, PostgreSQL, and Elasticsearch', async (t) => {
```

markdboyd:
> I don't believe there is a need to add unit tests that SNS messages are published within test-granules.js because it is unit tested here (since the call to publish SNS messages is made in _writeGranule()):

I'm not sure that I agree. While we do have those unit tests for `_writeGranule` as you say, the fact that the POST /granules endpoint logic is using that function is an implementation detail which could change. Thus, we want to test the POST /granules endpoint itself to ensure that an SNS message is sent when it is invoked. Then we can be sure that even if the inner details of the endpoint logic change, it still behaves as expected

packages/api/tests/endpoints/test-granules.js
botanical: [`9e7f401` (#2589)](https://github.com/nasa/cumulus/pull/2589/commits/9e7f401d7f28b92ea71722f414f9a290fa430faf)

packages/api/tests/endpoints/test-granules.js
botanical: [`9e7f401` (#2589)](https://github.com/nasa/cumulus/pull/2589/commits/9e7f401d7f28b92ea71722f414f9a290fa430faf)

packages/api/tests/endpoints/test-granules.js
```diff
@@ -1742,6 +1742,26 @@ test.serial('create (POST) creates new granule with associated execution in dyna
   t.is(fetchedESRecord.granuleId, newGranule.granuleId);
 });
 
+test.serial('create (POST) publishes an SNS message upon successful granule creation', async (t) => {
+  const newGranule = fakeGranuleFactoryV2({
```
botanical, line 1746: I would have used fakeGranuleRecordFactory but this fails with a bad request error since we are reading from DynamoDB and not PG here https://github.com/nasa/cumulus/blob/feature/rds-phase-2/packages/api/endpoints/granules.js#L81-L87 

A ticket was already created to address updating this https://bugs.earthdata.nasa.gov/browse/CUMULUS-2759

botanical: Not sure if I should wait on that ticket or just leave this unit test as is

markdboyd: I think that you can leave it as is

packages/api/endpoints/granules.js
botanical: [`f2eef77` (#2589)](https://github.com/nasa/cumulus/pull/2589/commits/f2eef77cba67a1c09d5fbd38139fb7d850aa7d7e)

markdboyd (approved):

PR #2592, opened 2021-12-04
CUMULUS-2774: Update constructGranuleSnsMessage and constructCollectionSnsMessage

botanical:

Addresses [CUMULUS-2774: Update constructGranuleSnsMessage and constructCollectionSnsMessage to throw error if eventType is not provided](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2774)

## Changes

* ...

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


CHANGELOG.md
```diff
@@ -346,6 +346,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     - Updated POST `/executions` endpoint to publish SNS message of created record to executions SNS topic
   - **CUMULUS-2661**
     - Updated PUT `/executions/<arn>` endpoint to publish SNS message of updated record to executions SNS topic
+  - **CUMULUS-2774**
+    - Updated `constructGranuleSnsMessage` and `constructCollectionSnsMessage`
+      to throw error if `eventType` is invalid or undefined.
+    - Updated `writeGranuleFromApi` to make `snsEventType` default to `Create`.
```
botanical, line 352: I'm unsure about this

packages/api/lib/publishSnsMessageUtils.js
```diff
@@ -19,7 +19,7 @@ const constructCollectionSnsMessage = (record, eventType) => {
     },
     deletedAt: Date.now(),
   };
-  default: return {};
+  default: throw new Error('constructCollectionSnsMessage requires a valid eventType of \'Create\', \'Update\', or \'Delete\'');
```
markdboyd, line 22:
```suggestion
  default: throw new Error(`Invalid eventType: ${eventType}`);
```

I would suggest that the error specify the invalid eventType that was sent rather than enumerating the currently valid options, because those could change and the user may way to know what invalid value was received by the code

packages/api/lib/publishSnsMessageUtils.js
```diff
@@ -33,7 +33,7 @@ const constructGranuleSnsMessage = (record, eventType) => {
     record,
     deletedAt: Date.now(),
   };
-  default: return {};
+  default: throw new Error('constructGranuleSnsMessage requires a valid eventType of \'Create\', \'Update\', or \'Delete\'');
```
markdboyd, line 36:
```suggestion
  default: throw new Error(`Invalid eventType: ${eventType}`);
```

I would suggest that the error specify the invalid eventType that was sent rather than enumerating the currently valid options, because those could change and the user may way to know what invalid value was received by the code

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -497,7 +497,7 @@ const writeGranuleFromApi = async (
   },
   knex,
   esClient,
-  snsEventType
+  snsEventType = 'Create'
```
markdboyd, line 500: Why are we adding a default value? I feel like we want to make sure that people specify the correct value when calling this function so not having a default is preferable, but I could be wrong

botanical: Yeah I was unsure about adding a default. I thought it was wise so we don't have to update every call to `writeGranuleFromApi` but I can definitely see how that's not super compelling.

botanical:
Thanks for the feedback and review @markdboyd! I believe I've addressed it all and my PR is ready for another review.

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -463,7 +463,7 @@ const _writeGranule = async ({
  * @param {Object} [params.granulePgModel] - only for testing.
  * @param {Knex} knex - knex Client
  * @param {Object} esClient - Elasticsearch client
- * @param {string} snsEventType - SNS Event Type
+ * @param {string} snsEventType - SNS Event Type, defaults to 'Create'
```
markdboyd, line 466: This is no longer accurate

CHANGELOG.md
markdboyd: This line is no longer accurate

packages/api/lib/writeRecords/write-granules.js
botanical: [`a18bcde` (#2592)](https://github.com/nasa/cumulus/pull/2592/commits/a18bcde9747312de1761d9f2ba649978d9570dc4)

markdboyd (approved):

PR #2599, opened 2021-12-09
Example deployment updates/improvements

markdboyd:
These are some quality of life/convenience improvements for our deployment, in particular to make VPC/subnet variables unnecessary when we're deploying to NGAP since we can look them up by expected tag names

### DIFF
--- a/example/cumulus-tf/main.tf
+++ b/example/cumulus-tf/main.tf
@@ -40,6 +40,7 @@ locals {
   public_bucket_names             = [for k, v in var.buckets : v.name if v.type == "public"]
   rds_security_group              = lookup(data.terraform_remote_state.data_persistence.outputs, "rds_security_group", "")
   rds_credentials_secret_arn      = lookup(data.terraform_remote_state.data_persistence.outputs, "database_credentials_secret_arn", "")
+  subnet_ids                      = length(var.lambda_subnet_ids) > 0 ? var.lambda_subnet_ids : data.aws_subnet_ids.subnet_ids[0].ids
 }
 
 data "aws_caller_identity" "current" {}
@@ -67,7 +68,6 @@ data "aws_ecr_repository" "async_operation" {
   name = "async_operations"
 }
 
-
 module "cumulus" {
   source = "../../tf-modules/cumulus"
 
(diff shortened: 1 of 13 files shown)


example/data-persistence-tf/vpc.tf
```diff
@@ -0,0 +1,16 @@
+data "aws_vpc" "application_vpc" {
+  count = var.vpc_id == null ? 1 : 0
```
botanical, line 2: Why do we want the count to be 1 when `vpc_id` is null?

botanical: I understand it now so I'm going to resolve this conversation

example/cumulus-tf/main.tf
```diff
@@ -93,7 +95,7 @@ module "cumulus" {
   async_operation_image = "${data.aws_ecr_repository.async_operation.repository_url}:${var.async_operation_image_version}"
 
   ecs_cluster_instance_image_id   = data.aws_ssm_parameter.ecs_image_id.value
-  ecs_cluster_instance_subnet_ids = length(var.ecs_cluster_instance_subnet_ids) == 0 ? var.lambda_subnet_ids : var.ecs_cluster_instance_subnet_ids
+  ecs_cluster_instance_subnet_ids = length(var.ecs_cluster_instance_subnet_ids) == 0 ? local.subnet_ids : var.ecs_cluster_instance_subnet_ids
```
botanical, line 98: This can also be `lambda_subnet_ids` since you set that on line 89 https://github.com/nasa/cumulus/pull/2599/files#diff-192847ae3f8b97c1193324513f1a0e04d4f474bcf998b911a8a34370931c01c7R89

example/rds-cluster-tf/variables.tf
```diff
@@ -17,11 +26,13 @@ variable "region" {
 variable "vpc_id" {
   description = "VPC ID for the Cumulus Deployment"
   type        = string
+  default = null
```
botanical, line 29: 🐜 definitely a nit but can we align this with the lines above it?

botanical (approved):
Couple 🐜 comments, pulled down and tested and looks good to me

PR #2638, opened 2021-12-31
Fixed the wrong link to the eslintrc

BiswaViraj:
**Summary:** Summary of changes
Updated the link to the **eslintrc.js** config file in the docs.
Addresses [CUMULUS-XX: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-XXX)

Fixes #2538 
## Changes
Updated the file to the use the correct Link
* Detailed list or prose of changes
* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Ad-hoc testing - Deploy changes and test manually
- [ ] Integration tests

PR #2648, opened 2022-01-05
Fix eslint docs

markdboyd:
Bring #2638 into master


botanical (approved):
I think you can set the skip changelog var in CI and then it will pass. Looks good to me

PR #2649, opened 2022-01-05
CUMULUS-1985

markdboyd:
Bring #2614 into master


botanical (approved):
I think you need to skip changelog in CI but looks good oterhwise

PR #2653, opened 2022-01-11
Cumulus 2783

markdboyd:

Addresses [CUMULUS-2783: ECS Cluster Autoscaling Alarm fails to execute action](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2783)

## Changes

* Fix autoscaling group "scale out" policy to set a lower bound, which is required for autoscaling to occur
* Remove duplicate cloudwatch metric/alarms on MemoryReservation for scale in/scale out policies. I cannot find any AWS docs to state this with certainty - but it seems to be the case that AWS will only use one cloudwatch metric for your autoscaling policy, so having multiple defined in Terraform is pointless. You'll notice that all of our deployed autoscaling policies in the AWS console are only using CPUReservation as ther metric despite also having the metric for MemoryReservation defined in Terraform.
* Remove unnecessary and misconfigured Cloudwatch alarms on ECS services created for our integration tests. These alarms were configured to be in "alarm" anytime MemoryUtlization > 1 percent, which will almost always be true. Also, these alarms aren't used for anything and were just copied from https://nasa.github.io/cumulus/docs/data-cookbooks/run-tasks-in-lambda-or-docker#example-replacing-aws-lambda-with-a-docker-container-run-on-ecs
* Fixed docs for https://nasa.github.io/cumulus/docs/data-cookbooks/run-tasks-in-lambda-or-docker#example-replacing-aws-lambda-with-a-docker-container-run-on-ecs to have a valid alarm configuration

## Testing

Notes on testing can be found in the ticket

## PR Checklist

- [x] Update CHANGELOG


flamingbear (approved):
Nice. We ❤️ expanded scope.  but looks great.

PR #2672, opened 2022-01-19
CUMULUS-2756: granule remained in qeueud status after IngestGranule failed

jaortega527:
**Summary:** Set granule status to `failed` if the granule write fails due to an issue with updated granule content

Addresses [CUMULUS-2756: granule remained in queued status after IngestGranule failed](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2756)

## Changes

* Updated `_writeGranule()` to catch failed granule writes, log the failure and then attempt to set the status of the granule to `failed` if it already exists to prevent a failure from allowing the granule to get "stuck" in a non-failed status.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


packages/api/lib/writeRecords/write-granules.js
```diff
@@ -299,15 +299,61 @@ const _writeGranule = async ({
...
+    const postgresGranuleExists
+      = await granulePgModel.exists(
+        knex,
+        {
+          granule_id: postgresGranuleRecord.granule_id,
+          collection_cumulus_id: postgresGranuleRecord.collection_cumulus_id,
+        }
+      );
+
+    if (dynamoGranuleExists && postgresGranuleExists) {
```
markdboyd, line 328: So my understanding of this code is that if an any error is caught from attempting to write a granule, then we update the granule status to "failed". However, that also could be misleading. An error could be thrown (due to database connectivity or anything) when trying to write a granule as "running" that is currently "queued", in which case immediately updating the granule to "failed" would be misleading because the granule workflow is not really in a terminal state yet.

Also, in general, this catch block could be (and most often probably will be) reached because of database connectivity issues. In that case, I'm not sure if it makes sense to immediately to try and do more database queries to update the granule.

It might be worth having an architecture discussion here about this

jaortega527: Fair enough. I'll add a topic to discuss this afternoon.

packages/api/tests/lib/writeRecords/test-utils.js
```diff
@@ -248,3 +249,11 @@ test('getExecutionCumulusId returns undefined if url cannot be found.', async (t
   const { knex } = t.context;
   t.is(await getExecutionCumulusId('https://example.com/still/not/in/db', knex), undefined);
 });
+
+test.only('isStatusFinalState returns true if status only if status is "completed" or "failed"', (t) => {
```
markdboyd, line 253: Remove `.only`, but **great job** adding units for this

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -194,16 +292,14 @@ const _writePostgresGranuleViaTransaction = async ({
...
   status,
   knex,
-  granuleModel = new Granule(),
-  granulePgModel = new GranulePgModel(),
 }) => {
   let fileRecords = [];
-
-  if (status !== 'running' && status !== 'queued') {
+  const { granuleId } = granule;
+  if (isStatusFinalState(status)) {
```
markdboyd, line 302: Very, very nice. Much better

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -176,6 +177,103 @@ const _writePostgresGranuleViaTransaction = async ({
...
+        }
+      );
+      // delete the execution field as well
+      await granuleModel.update(
+        { granuleId },
+        {
+          status,
+          error: errorObject,
+        },
+        ['execution']
```
markdboyd, line 237: So this is relevant in the case of writing granules as "queued", but i don't think it is relevant if we're writing them as "failed". When we're updating a granule as "queued", it's going to have an execution, so we want to delete the `execution` value in Dynamo if there is one. But when updating a granule as "failed", we want to preserve whatever value there is for `execution`. 

This function should probably just accept an optional array parameter named `apiGranuleDeleteFields` which could be set to `['execution']` in `updateGranuleStatusToQueued`

jaortega527: ah. got it. making that change now

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -176,6 +177,103 @@ const _writePostgresGranuleViaTransaction = async ({
...
+ * @param {string} params.status - new status for granule
+ * @returns {Promise}
+ * @throws {Error}
+ */
+async function updateGranuleStatus({
+  granule,
+  knex,
+  status,
+  error = '',
+  errorCause = '',
```
markdboyd, line 196:
```suggestion
  error,
  errorCause,
```

Empty strings and undefined are both falsy, so I think these initializations are unnecessary

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -299,15 +373,33 @@ const _writeGranule = async ({
...
+      });
+      return granuleModel.storeGranule(dynamoGranuleRecord);
     });
-    return granuleModel.storeGranule(dynamoGranuleRecord);
-  });
+  } catch (thrownError) {
+    log.error(`Write Granule failed: ${JSON.stringify(thrownError)}`);
+
+    if (isStatusFinalState(postgresGranuleRecord.status)
+      && isStatusFinalState(dynamoGranuleRecord.status)
```
markdboyd, line 391: These should always be the same, so I'm inclined to say we just need `isStatusFinalState(dynamoGranuleRecord.status)` for now. Once Phase2 is delivered, it would switch to `isStatusFinalState(postgresGranuleRecord.status)`

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -299,15 +373,33 @@ const _writeGranule = async ({
...
+      && thrownError.name === 'SchemaValidationError') {
+      const originalError = dynamoGranuleRecord.error || undefined;
```
markdboyd, line 393:
```suggestion
      const originalError = dynamoGranuleRecord.error;
```

It should automatically evaluate to `undefined` if  the`error`property is missing

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -317,12 +409,12 @@ const _writeGranule = async ({
...
   log.info('Successfully wrote granule %j to DynamoDB', dynamoGranuleRecord);
 
-  const { files, granuleId, status, error } = dynamoGranuleRecord;
+  const { files, status, error } = dynamoGranuleRecord;
 
   await _writeGranuleFiles({
     files,
     granuleCumulusId,
-    granuleId,
+    granule: dynamoGranuleRecord,
```
markdboyd, line 418: If you're going to do this, which is fine, then you can remove this above

`const { files, status, error } = dynamoGranuleRecord;`

and just destructure those properties that inside `_writeGranuleFiles({`

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -536,6 +540,61 @@ test.serial('writeGranulesFromMessage() writes a granule and marks as failed if
   t.deepEqual(pgGranule.error.Error, 'Failed writing files to PostgreSQL.');
 });
 
+test.serial('_writeGranules attempts to mark granule as failed if a SchemaValidationException occurs when a granule is in a final state', async (t) => {
```
markdboyd, line 543: Great job on this test

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -1064,3 +1125,119 @@ test.serial('updateGranuleStatusToQueued() throws error if record does not exist
...
+    knex,
+    collectionCumulusId,
+    granule,
+    granuleId,
+    granuleModel,
+    granulePgModel,
+  } = t.context;
+  await writeGranuleFromApi({ ...granule }, knex);
+  const dynamoRecord = await granuleModel.get({ granuleId });
+  const postgresRecord = await granulePgModel.get(
```
markdboyd, line 1140: We should verify that the status of these records is not `failed`, so we can be sure that `updateGranuleStatusToFailed` is what is setting that status

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -1064,3 +1125,119 @@ test.serial('updateGranuleStatusToQueued() throws error if record does not exist
...
+});
+
+test.serial('updateGranuleStatusToFailed() throws error if record does not exist in pg', async (t) => {
+  const {
+    knex,
+    granule,
+    granuleId,
+  } = t.context;
+
+  await writeGranuleFromApi({ ...granule }, knex);
```
markdboyd, line 1165: I think this is unnecessary in this test

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -1064,3 +1125,119 @@ test.serial('updateGranuleStatusToQueued() throws error if record does not exist
...
+    knex,
+    collectionCumulusId,
+    granule,
+    granuleId,
+    granuleModel,
+    granulePgModel,
+  } = t.context;
+  await writeGranuleFromApi({ ...granule }, knex);
+  const dynamoRecord = await granuleModel.get({ granuleId });
+  const postgresRecord = await granulePgModel.get(
```
markdboyd, line 1193: We should verify that the status of these records is not `completed` so we can be sure that `updateGranuleStatus` is what is doing the updates

markdboyd (changes requested):
Great, great work on the refactor! Just a few minor comments and requested changes

packages/api/lib/writeRecords/write-granules.js
jaortega527: gotcha

packages/api/lib/writeRecords/write-granules.js
jaortega527: good catch. done.

packages/api/tests/lib/writeRecords/test-write-granules.js
jaortega527: done.

packages/api/tests/lib/writeRecords/test-write-granules.js
jaortega527: another good catch.

packages/api/tests/lib/writeRecords/test-write-granules.js
jaortega527: great catch. thanks.

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -1064,3 +1125,118 @@ test.serial('updateGranuleStatusToQueued() throws error if record does not exist
...
+test.serial('updateGranuleStatusToFailed() updates granule status in the database', async (t) => {
+  const {
+    knex,
+    collectionCumulusId,
+    granule,
+    granuleId,
+    granuleModel,
+    granulePgModel,
+  } = t.context;
+  await writeGranuleFromApi({ ...granule }, knex);
```
markdboyd, line 1138:
```suggestion
  granule.status = 'running';
  await writeGranuleFromApi({ ...granule }, knex);
```

CHANGELOG.md
```diff
@@ -138,6 +138,8 @@ all Cumulus workflow tasks
     found in the metadata file.  The granule's temporal information is
     returned so that both beginningDateTime and endingDateTime are set to the
     discovered singleDateTimeValue.
+- **CUMULUS-2756**
+  - Updated `_writeGranule()` in `write-granules.js` to catch failed granule writes, log the failure and then attempt to set the status of the granule to `failed` if it already exists to prevent a failure from allowing the granule to get "stuck" in a non-failed status.
```
markdboyd, line 142:
```suggestion
  - Updated `_writeGranule()` in `write-granules.js` to catch failed granule writes due to schema validation, log the failure and then attempt to set the status of the granule to `failed` if it already exists to prevent a failure from allowing the granule to get "stuck" in a non-failed status.
```

markdboyd (approved):

PR #2684, opened 2022-01-26
[CUMULUS-2814] Remove dynamo logic from rules DELETE

botanical:

Addresses [CUMULUS-2814: Remove dynamo logic from rules DELETE](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2814)

## Changes

- Remove DynamoDB logic from rules `DELETE` endpoint
- Move event resources deletion logic from `rulesModel` to `rulesHelper`
  -  `delete` is now `deleteRuleResources`
  - `deleteKinesisEventSources`
    - this is called by `updateRuleTrigger` so it can't be deleted from the rules model
  - `deleteSnsTrigger`
    - this is called by `updateRuleTrigger` so it can't be deleted from the rules model
  - `isEventSourceMappingShared`
    - this is used by `deleteKinesisEventSources` in the rules model so it can't be deleted from the rules model yet
- Added unit tests 

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


botanical:
This work is blocked on this PR https://github.com/nasa/cumulus/pull/2694

packages/api/endpoints/rules.js
```diff
@@ -221,34 +220,16 @@ async function del(req, res) {
...
     });
-  } catch (error) {
-    // Delete is idempotent, so there may not be a DynamoDB
-    // record to recreate
-    if (apiRule) {
-      await ruleModel.create(apiRule);
-    }
-    throw error;
-  }
+    if (rule) await deleteRuleResources(knex, rule);
```
botanical, line 231: This is necessary in the case that the record exists in Elasticsearch but not PG. Though, in the case that the rule record only exists in Elasticsearch, can we still retrieve values we need to delete rule resources (name, log_event_arn, arn, etc) ?

packages/api/lib/rulesHelpers.js
```diff
@@ -113,13 +122,157 @@ async function queueMessageForRule(rule, eventObject, eventSource) {
...
+    await CloudwatchEvents.deleteTarget(targetId, name);
+    await CloudwatchEvents.deleteEvent(name);
+    break;
+  }
+  case 'kinesis': {
+    await deleteKinesisEventSources(knex, rule);
+    break;
+  }
+  case 'sns': {
+    if (rule.enabled === true) {
```
jaortega527, line 253: This is such a nit pick, but do we need `rule.enabled` to be strictly `true` such that `if (rule.enabled)` can't be used?

packages/api/lib/rulesHelpers.js
```diff
@@ -113,13 +122,157 @@ async function queueMessageForRule(rule, eventObject, eventSource) {
...
+      await deleteSnsTrigger(knex, rule);
+    }
+    break;
+  }
+  case 'sqs':
```
jaortega527, line 258: Can `case 'sqs':` be removed or is it here to be explicit documentation that we do nothing when the `rule.type` is 'sqs'?

packages/api/lib/rulesHelpers.js
jaortega527: Last question. Why is the check for `rule.enabled` only around the `deleteSnsTrigger` method and not a requirement for all deletions? Maybe you answered this yesterday. I was trying to absorb as much as i could, and maybe missed that.

packages/api/lib/rulesHelpers.js
jaortega527: If we are explicitly documenting that we do nothing for rules of type `sqs` then maybe a test should be added to validate that.

packages/api/lib/rulesHelpers.js
botanical: I think you're right that it doesn't need to strictly check `rule.enabled === true` so I've updated it. [`8c20550` (#2684)](https://github.com/nasa/cumulus/pull/2684/commits/8c2055054f86fc8f751aa4896632215ab27d7950)

I think having the condition outside of the deleteSnsTrigger function helps us avoid a call to that function if the condition isn't met, which I _think_ might be slightly more preferable/efficient. 
As for why we only check for enabled for sns rules, I'm not sure so I will get back to you on that!

jaortega527 (approved):
Very well done. Thanks for taking time to share the PR after scrum! I might not have had the confidence to review otherwise.

PR #2693, opened 2022-02-02
Add events actions to step_policy for the step aws_am_role

abarciauskas-bgse:

When deploying MAAP cumulus stack I got the following error:

```
Error: error creating Step Function State Machine (maap-dit-uw2-IngestAndPublishGranuleWorkflow): AccessDeniedException: 'arn:aws:iam::532XXX:role/maap-dit-uw2-steprole' is not authorized to create managed-rule.
```

Some digging discovered that my steprole was missing some events permissions: https://stackoverflow.com/questions/60612853/nested-step-function-in-a-step-function-unknown-error-not-authorized-to-cr/60623051#60623051

@markdboyd wonders if perhaps these permissions are automatically granted in NGAP. So this change supports non-NGAP environments if that is the case

## Changes

* Added 

```
      "events:PutTargets",
      "events:PutRule",
      "events:DescribeRule",
```

to tf-modules/ingest/iam.tf

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


markdboyd:
@abarciauskas-bgse Yeah it looks like we have these permissions in NGAP by default, so that explains why we haven't seen this. Thanks for the contribution!

PR #2686, opened 2022-01-27
Mhs/cumulus 2811/remove dynamo logic from provider delete

flamingbear:
**Summary:** Summary of changes

Addresses [CUMULUS-2811: remove dynamo from DELETE /provider](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2811)

## Changes

* removes dynamoDB from api endpoint DELETE /provider 
* updates tests.

## PR Checklist

- [ X ] Update CHANGELOG
- [ X ] Unit tests
- [ X ] Ad-hoc testing - Deploy changes and test manually
- [ ] Integration tests

PR #2694, opened 2022-02-02
[CUMULUS-2817] Update collections endpoint to remove DynamoDB logic

botanical:

Addresses [CUMULUS-2817: Update collections endpoint to remove DynamoDB logic](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2817)

## Changes
* Updated DELETE endpoint for collections to remove DynamoDB logic

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/packages/api/tests/endpoints/collections/delete-collection.js
+++ b/packages/api/tests/endpoints/collections/delete-collection.js
@@ -18,6 +18,8 @@ const {
   generateLocalTestDb,
   localStackConnectionEnv,
   migrationDir,
+  fakeRuleRecordFactory,
+  RulePgModel,
 } = require('@cumulus/db');
 const {
   constructCollectionId,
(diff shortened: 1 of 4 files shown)


packages/api/tests/endpoints/collections/delete-collection.js
```diff
@@ -83,7 +82,6 @@ test.before(async (t) => {
 
   t.context.collectionModel = new models.Collection({ tableName: process.env.CollectionsTable });
   await t.context.collectionModel.createTable();
```
flamingbear, line 84: I think, you can delete both os these lines, and the cleanup on L140 if you update the createCollectionTestRecords function to check for a collectionModel before creating a model
```
  if (collectionModel) {
    await collectionModel.create(originalCollection);
  }
```
That way there's no api.model collections in these tests at all.

I think that's better, but you can disagree

flamingbear: I think it's better because it's what I did in my removal PR . #2686 remove dynamo from provider delete.

flamingbear (changes requested):
Great work, with only one suggestion to remove collection model completely from the tests.  Or you can convince me it's fine.

packages/api/tests/endpoints/collections/delete-collection.js
botanical: Thanks for the feedback! I took your suggestions and updated the test setup and the createCollectionTestRecords function [`8f7216b` (#2694)](https://github.com/nasa/cumulus/pull/2694/commits/8f7216bc6b4ffa767d2d9954ef9c1f5b796b391a)

flamingbear (approved):

PR #2704, opened 2022-02-07
Mhs/cumulus 2833/build bug fix

flamingbear:

Addresses build errors with `npm run clean` and `npm run bootstrap`

## Changes
Runs git clean instead of selecting files to clean up 
adds an additional build step in the python-reference-task to build with the dev dependencies.


example/lambdas/python-reference-task/package.json
```diff
@@ -16,7 +16,7 @@
     "clean": "rm -rf dist && rm -rf lib && rm -rf .venv && mkdir dist && mkdir lib",
     "build": "pip install pipenv && PIPENV_VENV_IN_PROJECT=1 pipenv install --dev --deploy --ignore-pipfile",
     "prepare": "npm run build",
-    "package": "npm run clean && ./package.sh $(pwd)",
+    "package": "npm run clean && ./package.sh $(pwd) && npm run build",
```
flamingbear, line 19: I don't have a better idea, and this feels hacky.
But currently `npm clean && npm install && npm run bootstrap` leaves this directory without the dev dependencies to run lint

flamingbear: The second run rebuild with the dev dependencies (see L17).  I could have added the dev deps to the package.sh but then the lambda zips would have the dev dependencies in them and I didn't think that was a great idea. 

Part of this is maybe me not understanding fully how lerna bootstrap works, but when npm package is run it's stripping out the development dependencies and lint can't run

flamingbear: When I looked at the documentation for lerna bootstrap, I wasn't sure where `package` was getting called from, but I see it's from our `npm run bootstrap`, which runs package after the bootstrap, which is deleting the dev dependencies in this directory.  I suppose the other solution would be to update package.sh to install dev dependencies after it zips the lambdas.

flamingbear: 529253fb07 Is that better?  Seems like it.

markdboyd: It seems like there are still some cobwebs in how this is setup, but this works for now

markdboyd (approved):

example/lambdas/python-reference-task/package.json
flamingbear: I agree. 🕸️

PR #2702, opened 2022-02-04
Merge CUMULUS-2840

markdboyd:
Merge in changes from #2685

PR #2721, opened 2022-02-11
cherry-pick CUMULUS-2480 into release 9.9.2

markdboyd:


Jkovarik (approved):
Confirmed changes match https://github.com/nasa/cumulus/pull/2702, int/units should be sufficient to show functionality.

PR #2725, opened 2022-02-18
[RDS phase 2] CUMULUS-2846: Update unit tests

markdboyd:

Addresses [CUMULUS-2846: Fix rules PUT logic](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2846)

## Changes

- Update unit tests

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/packages/api/tests/endpoints/test-rules.js
+++ b/packages/api/tests/endpoints/test-rules.js
@@ -467,13 +467,13 @@ test.serial('post() creates SNS rule with same trigger information in Dynamo/Pos
     pgCollection,
   } = t.context;
 
-  const topic1 = randomId('sns');
+  const topic1 = await awsServices.sns().createTopic({ Name: randomId('topic1_') }).promise();
 
   const rule = fakeRuleFactoryV2({
     state: 'ENABLED',
     rule: {
       type: 'sns',
-      value: topic1,
+      value: topic1.TopicArn,
     },
     collection: {
       name: pgCollection.name,
@@ -490,6 +490,8 @@ test.serial('post() creates SNS rule with same trigger information in Dynamo/Pos
 
   await post(expressRequest, response);
 
+  t.true(response.send.called);
+
   const dynamoRule = await ruleModel.get({ name: rule.name });
   const pgRule = await t.context.rulePgModel
     .get(t.context.testKnex, { name: rule.name });
@@ -504,7 +506,7 @@ test.serial('post() creates SNS rule with same trigger information in Dynamo/Pos
   t.like(dynamoRule, {
     rule: {
       type: 'sns',
-      value: topic1,
+      value: topic1.TopicArn,
       arn: dynamoRule.rule.arn,
     },
   });
@@ -513,7 +515,7 @@ test.serial('post() creates SNS rule with same trigger information in Dynamo/Pos
     {
       rule: {
         type: 'sns',
-        value: topic1,
+        value: topic1.TopicArn,
         arn: dynamoRule.rule.arn,
       },
     }
@@ -523,7 +525,7 @@ test.serial('post() creates SNS rule with same trigger information in Dynamo/Pos
     enabled: true,
     type: 'sns',
     arn: dynamoRule.rule.arn,
-    value: topic1,
+    value: topic1.TopicArn,
   });
 });
 


packages/api/models/rules.js
```diff
@@ -462,18 +477,18 @@ class Rule extends Manager {
         Endpoint: process.env.messageConsumer,
         ReturnSubscriptionArn: true,
       };
-      const r = await awsServices.sns().subscribe(subscriptionParams).promise();
+      const r = await this.SnsClient.subscribe(subscriptionParams).promise();
       subscriptionArn = r.SubscriptionArn;
+      // create permission to invoke lambda
+      const permissionParams = {
```
markdboyd, line 483: Move this back out?

markdboyd: Determine if you can overwrite existing permission

packages/api/tests/endpoints/test-rules.js
```diff
@@ -460,6 +461,243 @@ test('POST creates a rule in all data stores', async (t) => {
   t.like(esRecord, fetchedDynamoRecord);
 });
 
+test.serial('post() creates SNS rule with same trigger information in Dynamo/PostgreSQL/Elasticsearch', async (t) => {
```
Jkovarik, line 464: :thumbsup: :100: :hero:

packages/api/tests/endpoints/test-rules.js
```diff
@@ -1183,16 +1419,21 @@ test.serial('put() creates the same updated Kinesis rule in Dynamo/PostgreSQL/El
...
-  const updatedEsRecord = await t.context.esRulesClient.get(
+  const updatedEsRule = await t.context.esRulesClient.get(
     originalDynamoRule.name
   );
-  t.truthy(updatedRule.rule.arn);
-  t.truthy(updatedRule.rule.logEventArn);
-  t.truthy(updatedPgRule.arn);
-  t.truthy(updatedPgRule.log_event_arn);
+
+  t.truthy(originalDynamoRule.rule.arn);
```
Jkovarik, line 1426: Is this block a duplication of L1393 above?   Unless I'm misunderstanding something I think these can be removed.

packages/api/lib/testUtils.js
```diff
@@ -510,14 +510,15 @@ const createRuleTestRecords = async (context, ruleParams) => {
...
+  const dynamoRuleWithTrigger = await ruleModel.createRuleTrigger(originalRule);
+  const originalDynamoRule = await ruleModel.create(dynamoRuleWithTrigger);
+  const insertPgRecord = await translateApiRuleToPostgresRule(originalDynamoRule, testKnex);
+
   const [ruleCumulusId] = await rulePgModel.create(testKnex, insertPgRecord);
   const originalPgRecord = await rulePgModel.get(
     testKnex, { cumulus_id: ruleCumulusId }
   );
-  await indexRule(esClient, ruleWithTrigger, process.env.ES_INDEX);
+  await indexRule(esClient, originalDynamoRule, process.env.ES_INDEX);
```
Jkovarik, line 521: Interesting, and nice catch/protect against future badness.       To confirm, the reason this didn't blow up the rule units is that in most cases we're not relying on the trigger Arn for the OG record,
(e.g. cases where it isn't supposed to write, or it's supposed to be updated anyway), etc?

Jkovarik (approved):
Really nice work!    A couple of comments, but nothing that should halt merging.    Approved.

packages/api/tests/endpoints/test-granules.js
```diff
@@ -107,7 +107,7 @@ process.env.ExecutionsTable = randomId('executions');
 process.env.CollectionsTable = randomId('collection');
 process.env.GranulesTable = randomId('granules');
 process.env.stackName = randomId('stackname');
-process.env.system_bucket = randomId('system-bucket');
+process.env.system_bucket = randomId('bucket');
```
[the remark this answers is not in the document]
markdboyd: I believe when I made the change I was a wholesale find/replace on what I thought were invalid bucket names, because the newer version of Localstack requires only valid bucket names. But this one probably isn't invalid? I can revert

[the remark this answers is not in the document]
Jkovarik: I don't think it needs reverting, that was mostly curiosity on my part.

packages/api/lib/testUtils.js
markdboyd: I think the reason it didn't matter is there is little if any difference between `dynamoRuleWithTrigger` and `originalDynamoRule`, so using either as the basis for indexing to ES is unlikely to produce any issues. But it seemed perhaps safer to index exactly what got created in ES in case somehow the logic changes.

It could be argued we should go the other way though, since P2 and thus PG is now the basis of truth:

- generating a PG record fixture
- create PG record
- translate PG record -> API
- save API record to dynamo
- save created dynamo record to ES

markdboyd: Scratch that. That'll have to wait for phase 3 because `createRuleTrigger` currently accepts a dynamo rule so it's a pain to make that change without translating from PG to API and back again

packages/api/tests/endpoints/test-rules.js
```diff
@@ -460,6 +461,243 @@ test('POST creates a rule in all data stores', async (t) => {
...
+test.serial('post() creates SNS rule with same trigger information in Dynamo/PostgreSQL/Elasticsearch', async (t) => {
+  const {
+    pgProvider,
+    pgCollection,
+  } = t.context;
+
+  const topic1 = randomId('sns');
+  const fakeSubscriptionArn1 = randomId('subscription');
+
+  const addSnsTriggerStub = sinon.stub(Rule.prototype, 'addSnsTrigger')
```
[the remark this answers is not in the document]
markdboyd: How about I just remove the mocking altogether? It's not necessary with the updated localstack

5a4f8e4d566442e9e669e0247c382b5fa5551b16

packages/api/tests/endpoints/test-rules.js
markdboyd: It was and in fact needed to be updated to test what I want:

88c3502310f14be77f3168d5532e3745dd0e8313

packages/api/tests/endpoints/test-rules.js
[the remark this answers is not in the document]
Jkovarik: Even better.

PR #2728, opened 2022-02-21
[CUMULUS-2813] Remove DynamoDB dependency from rules POST endpoint

botanical:

Addresses [CUMULUS-2813: Remove DynamoDB dependency from rules POST endpoint](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2813)

## Changes
    - Removes creation and deletion of DynamoDB record from API endpoint POST /rules/

### Added
    - Added function `create` in the `db` model for Rules
      to return an array of objects containing all columns of the created record.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


packages/api/lib/rulesHelpers.js
```diff
@@ -492,18 +493,18 @@ async function addRule(item, payload) {
...
     throw error;
   }
   if (!rule.workflow) {
-    error.detail = 'Rule workflow is undefined.';
+    error.message += 'Rule workflow is undefined.';
     throw error;
   }
   if (!rule.rule.type) {
-    error.detail = 'Rule type is undefined.';
+    error.message += 'Rule type is undefined.';
```
botanical, line 507: I changed these because POST doesn't throw the error like PUT does, it returns a response with `error.message` https://github.com/nasa/cumulus/pull/2728/files#diff-907b67e0c8c9667871fad5f88fa710ae9231500c156a692721ff54c84254b7b9R115

packages/api/lib/rulesHelpers.js
```diff
@@ -468,11 +469,10 @@ function updateKinesisRuleArns(ruleItem, ruleArns) {
    */
 async function addRule(item, payload) {
   const name = `${process.env.stackName}-custom-${item.name}`;
-  const state = item.enabled ? 'ENABLED' : 'DISABLED';
   await CloudwatchEvents.putEvent(
     name,
     item.rule.value,
-    state,
+    item.state,
```
botanical, line 475: Added unit test to cover this function https://github.com/nasa/cumulus/pull/2728/files#diff-f5c5b96c6230da09644104acbdca2268e22d32509bc5b275a5b4648a8939584eR409

example/package.json
```diff
@@ -24,7 +24,6 @@
     "load-test": "../node_modules/.bin/jasmine ./spec/loadTest/runScaledtest.js",
     "package": "for x in lambdas/*; do (echo \"packaging $x\" && cd $x && test -e package.json && npm run package); done",
     "parallel-tests": "sh scripts/tests-parallel.sh",
-    "python-lint": "for x in lambdas/*; do cd $x && npm run python-lint && cd -; done",
```
jaortega527, line 27: Was this removed because it isn't called by the root `package.json` -> `lint` script? If so, great clean-up!

packages/api/lib/rulesHelpers.js
jaortega527: 👍🏽

packages/api/lib/rulesHelpers.js
jaortega527: if the `POST` returns a response with `error.message`, Does this cause an issue with `PUT` since this is a change in behavior?

jaortega527: Nevermind. I see the answer to my question in `test-rules.js`

jaortega527 (approved):
Everything looks great. I even tried to get nitpicky there at the end. Great work!

PR #2736, opened 2022-02-24
Port CUMULUS-2845 to master

markdboyd:
Port changes from #2707 to `master` to fix issues with rules being different in DynamoDB and PostgreSQL on POST


botanical (approved):
Looks good to me, thanks for doing this

PR #2874, opened 2022-03-01
cumulus-2809: remove dynamo from provider PUT

flamingbear:
**Summary:** 
Eliminates dynamoDB from the endpoint/provider 'PUT'  command and updates the tests to match

There are no extra features of the api provider model that need to be added as the `create` function just encrypted the password and username before storage, and that is handled in the translateApiProviderToPostgressProvider function so that the PG provider has encrypted username and passwords.

Addresses [CUMULUS-2809: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2809)

## Changes

* Removes all traces of the api/models/provider from the `PUT /provider` endpoint.
* Updates unit tests to rely only on the postgress database  

## PR Checklist

- [X] Update CHANGELOG
- [X] Unit tests
- [ ] Integration tests

### DIFF
--- a/packages/api/endpoints/providers.js
+++ b/packages/api/endpoints/providers.js
@@ -164,8 +164,8 @@ async function put(req, res) {
   const postgresProvider = await translateApiProviderToPostgresProvider(apiProvider);
 
   await createRejectableTransaction(knex, async (trx) => {
-    const updatedPostgresProvider = await providerPgModel.upsert(trx, postgresProvider);
-    record = translatePostgresProviderToApiProvider(updatedPostgresProvider);
+    await providerPgModel.upsert(trx, postgresProvider);
+    record = translatePostgresProviderToApiProvider(postgresProvider);
     await indexProvider(esClient, record, process.env.ES_INDEX);
   });
 
--- a/packages/api/endpoints/providers.js
+++ b/packages/api/endpoints/providers.js
@@ -164,7 +164,7 @@ async function put(req, res) {
   const postgresProvider = await translateApiProviderToPostgresProvider(apiProvider);
 
   await createRejectableTransaction(knex, async (trx) => {
-    const updatedPostgresProvider = await providerPgModel.upsert(trx, postgresProvider);
+    const [updatedPostgresProvider] = await providerPgModel.upsert(trx, postgresProvider);
     record = translatePostgresProviderToApiProvider(updatedPostgresProvider);
     await indexProvider(esClient, record, process.env.ES_INDEX);
   });
--- a/packages/api/endpoints/providers.js
+++ b/packages/api/endpoints/providers.js
@@ -164,8 +164,8 @@ async function put(req, res) {
   const postgresProvider = await translateApiProviderToPostgresProvider(apiProvider);
 
   await createRejectableTransaction(knex, async (trx) => {
-    await providerPgModel.upsert(trx, postgresProvider);
-    record = translatePostgresProviderToApiProvider(postgresProvider);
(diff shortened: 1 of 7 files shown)


packages/api/endpoints/providers.js
```diff
@@ -159,34 +157,17 @@ async function put(req, res) {
...
-  } catch (innerError) {
-    // Revert Dynamo record update if any write fails
-    if (oldProvider) {
-      await providerModel.create(oldProvider);
-    }
-    throw innerError;
-  }
+  await createRejectableTransaction(knex, async (trx) => {
+    await providerPgModel.upsert(trx, postgresProvider);
+    record = translatePostgresProviderToApiProvider(postgresProvider);
```
jennyhliu, line 168: Is this `record` not the same as `apiProvider` after `apiProvider` is going through two translations?  Ideally, we should get the record from postgres after update, and then translate to api record.

jennyhliu: `await providerPgModel.upsert(trx, postgresProvider);` returns a pgRecord, we can use that to translate to api record

flamingbear: so I think that record should be the same as the apiProvider, but was hoping it would be fully valid after the translation loop api -> pg -> api  (that was my thinking). But if the upsert returns a PG record, I would use that in the translate on L168.  That makes the most sense to me.

flamingbear: Is this what you meant? 18d5d66d399626c73c1b1c63d376a22ecae724f3

packages/api/endpoints/providers.js
```diff
@@ -159,34 +157,17 @@ async function put(req, res) {
...
-    });
-  } catch (innerError) {
-    // Revert Dynamo record update if any write fails
-    if (oldProvider) {
-      await providerModel.create(oldProvider);
-    }
-    throw innerError;
-  }
+  await createRejectableTransaction(knex, async (trx) => {
+    const updatedPostgresProvider = await providerPgModel.upsert(trx, postgresProvider);
```
jennyhliu, line 167:
```suggestion
    const [updatedPostgresProvider] = await providerPgModel.upsert(trx, postgresProvider);
```
example:
https://github.com/nasa/cumulus/blob/feature/rds-phase-3/packages/api/lib/writeRecords/write-execution.js#L137

flamingbear: That's what I get for not running my tests. 🤦

packages/api/endpoints/providers.js
flamingbear: Yup, that doesn't work, the upsert only returns the cumulusId.  So I think the question is should we store the user's input `apiRecord` or the record returned from the translatePostgresProviderToApiProvider `record`.  I already stated why I returned the record, is there a reason to just use the `apiRecord`?

markdboyd: >should we store the user's input apiRecord or the record returned from the translatePostgresProviderToApiProvider record

Definitely the latter. There could be subtle differences between `apiRecord` and `record` based on the PG upsert logic (which might only update certain fields - even though it currently doesn't do that), so translating the PG record and writing that to ES is safest

flamingbear: So I ended up modifying the upsert to return the full record and translated that to store into ES.  I assume that's the very best.

flamingbear:
@jennyhliu with some CI problems this is passed.  even if it's blocked on snyk.
would still take an approval.

PR #2883, opened 2022-03-08
CUMULUS-2820: Remove dynamo logic from pdr DELETE endpoint

jennyhliu:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/packages/api/lib/testUtils.js
+++ b/packages/api/lib/testUtils.js
@@ -515,7 +515,6 @@ const createRuleTestRecords = async (context, ruleParams) => {
 const createPdrTestRecords = async (context, pdrParams = {}) => {
   const {
     knex,
-    pdrModel,
     pdrPgModel,
     esClient,
     esPdrsClient,
@@ -537,7 +536,6 @@ const createPdrTestRecords = async (context, pdrParams = {}) => {
   });
 
   const insertPgRecord = await translateApiPdrToPostgresPdr(originalPdr, knex);
-  const originalDynamoPdr = await pdrModel.create(originalPdr);
   const [pdrCumulusId] = await pdrPgModel.create(knex, insertPgRecord);
   const originalPgRecord = await pdrPgModel.get(
     knex, { cumulus_id: pdrCumulusId }
@@ -547,7 +545,6 @@ const createPdrTestRecords = async (context, pdrParams = {}) => {
     originalPdr.pdrName
   );
   return {
-    originalDynamoPdr,
     originalPgRecord,
     originalEsRecord,
   };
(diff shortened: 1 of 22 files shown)


packages/api/tests/endpoints/test-pdrs.js
```diff
@@ -406,48 +379,13 @@ test.serial('DELETE a pdr', async (t) => {
...
-
-  const parsedBody = response.body;
-  t.is(parsedBody.detail, 'Record deleted');
-});
-
-test.serial('DELETE handles the case where the PDR exists in DynamoDb but not in S3', async (t) => {
-  const {
-    originalDynamoPdr,
+    originalPgRecord,
   } = await createPdrTestRecords(t.context);
```
markdboyd, line 385: It looks like `createPdrTestRecords` does stage a file to S3, so I'm not sure that this test proves what it claims to

packages/api/lib/testUtils.js
```diff
@@ -537,7 +536,6 @@ const createPdrTestRecords = async (context, pdrParams = {}) => {
   });
 
   const insertPgRecord = await translateApiPdrToPostgresPdr(originalPdr, knex);
```
markdboyd, line 538: Nitpick: but can we refactor this function to use `fakePdrRecordFactory` so it uses a PG style record instead of translating API -> PG. Then also translating `originalPgRecord` for writing to ES via `translatePostgresToApiPdr`. I'd like to remove our reliance on API style records/helpers as we move away from Dynamo.

But this isn't strictly necessary.

markdboyd (changes requested):
One minor requested change and one other comment. Great work!

packages/api/lib/testUtils.js
jennyhliu: addressed the comments

packages/api/tests/endpoints/test-pdrs.js
```diff
@@ -456,150 +410,47 @@ test.serial('DELETE handles the case where the PDR exists in DynamoDb but not in
...
-
-  const response = await request(app)
-    .delete(`/pdrs/${pdrName}`)
+    .delete(`/pdrs/${originalPgRecord.name}`)
     .set('Accept', 'application/json')
     .set('Authorization', `Bearer ${jwtAuthToken}`);
   t.is(response.status, 200);
 
-  // Check Dynamo and RDS. The PDR should not exist in either.
+  t.false(await t.context.pdrPgModel.exists(t.context.knex, { name: originalPgRecord.name }));
```
markdboyd, line 428: I think we should also verify in this test that the object was deleted from S3

packages/api/lib/testUtils.js
markdboyd: Thanks!

markdboyd (approved):

PR #2912, opened 2022-03-29
[CUMULUS-2870] Update terraform to redeploy ECS service when lambda is modified

botanical:

Addresses [CUMULUS-2870: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2870)

## Changes
Leveraging `aws_lambda_function`'s exported attribute `last_modified`, I added this value to the command list of the `hello_world_service` module to force a state change whenever the lambda is modified, causing terraform to redeploy the ECS service. I recognize that the `cumulus-ecs-task` won't do anything with the `lastModified` parameter since it doesn't expect it. 
(https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function)

* Added `last_modified_date` as output to `hello_world_task` in Terraform `ingest` module.
* Tested manually by:
  * deploying changes (which triggers Terraform to redeploy ECS service anyway), 
  * and then updating the task (tasks/hello-world/index.js) to return a string other than `Hello World`, 
  * packaged new task (`npm run package`)
  * and then ran `terraform apply` (where I saw :
    * `# module.cumulus.module.ingest.aws_lambda_function.hello_world_task will be updated in-place`
    ` ~ last_modified                  = "2022-03-29T18:46:07.000+0000" -> (known after apply)`
  * Finally ran `HelloWorldEcsSpec` to verify that the integration test fails since the lambda now contains a new string. 

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     allow this timeout to be user configurable
 - **CUMULUS-2868**
   - Added `iam:PassRole` permission to `step_policy` in `tf-modules/ingest/iam.tf`
+- **CUMULUS-2870**
+  - Added `last_modified_date` as output to `hello_world_task` in Terraform `ingest` module.
+
+### Changed
+
+- **CUMULUS-2870**
+  - Updated `hello_world_service` module to pass in `lastModified` parameter in command list to trigger a Terraform state change when the `hello_world_task` is modified.
 
 ## [v10.1.1] 2022-03-04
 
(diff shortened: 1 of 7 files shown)


jennyhliu:
1. Do you need to update the [cookbook](https://github.com/nasa/cumulus/blob/master/docs/data-cookbooks/run-tasks-in-lambda-or-docker.md) to include the `--lastModified`?
2. Is the `--lastModified` parameter ignored in the task execution?  Do we need to update [cumulus-ecs-task](https://github.com/nasa/cumulus-ecs-task)?

jennyhliu (approved):

PR #2946, opened 2022-04-29
[CUMULUS-2932] Update sync-granules to take in configurable ACL

botanical:

Addresses [CUMULUS-2932: As a user, I would like the sync granule task to allow disabling or configuration of ACLs through task configuration](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2932)

## Changes
- Updates `SyncGranule` task to include `setOrDisableAcl` function that uses the configuration ACL parameter to set ACL or disable ACL. Defaults to private.
- Updates `@cumulus/sync-granule` `download()` function to take in ACL parameter
- Updates `@cumulus/ingest` `proceed()` function to take in ACL parameter
- Updates `@cumulus/ingest` `addLock()` function to take in an optional ACL parameter
- Updates `SyncGranule` example worfklow config `example/cumulus-tf/sync_granule_workflow.asl.json` to include `ACL` parameter.
- Added integration tests to check that the lambda output meta contains the configured ACL 
- Added relevant unit tests

I am unsure that using `getObjectAcl` in the integration test was useful because a disabled ACL (versus private ACL) both return `FULL_CONTROL` as the granted permission. Definitely open to hearing suggestions on this.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


botanical:
Hello 👋  This is a user contribution PR coming from the MAAP team. We would like to update sync-granule to allow us to disable ACLs but maintain the current behavior by having it default the ACL to private. The external request ticket is https://bugs.earthdata.nasa.gov/browse/CUMULUS-2932 which contains more context. Thanks!

example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js
```diff
@@ -367,6 +367,55 @@ describe('The Sync Granules workflow', () => {
...
+      objectACLs = await Promise.all(files.map(
+        (file) => s3().getObjectAcl({
+          Bucket: file.bucket,
+          Key: file.key,
+        })
+      ));
+    });
+
+    afterAll(async () => {
+      await deleteExecution({ prefix: config.stackName, executionArn: executionArn });
```
npauzenga, line 407: I _think_ we also need to do a couple `deleteS3Object`s here? I haven't actually written a new test that creates S3 objects so I could be wrong and it's not necessary.

example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js
```diff
@@ -367,6 +367,55 @@ describe('The Sync Granules workflow', () => {
...
+
+    it('puts objects with ACL disabled (full control permission)', () => {
+      objectACLs.map((acl) => expect(acl.Grants[0].Permission).toBe('FULL_CONTROL'));
+    });
+
+    it('a disabled ACL configuration value in the lamba output', () => {
+      expect(lambdaOutput.meta.ACL).toBe('disabled');
+    });
+  });
+
```
npauzenga, line 418: If we do allow custom ACLs should we add a test for something other than 'disabled'?

tasks/sync-granule/index.js
```diff
@@ -78,6 +81,23 @@ async function download({
...
+*
+* @param {string} acl
+* @returns {string | undefined }
+*/
+function setOrDisableAcl(acl) {
+  if (acl === 'disabled') {
+    return undefined;
+  }
+  if (acl) {
+    return acl;
```
npauzenga, line 96: Do we want to allow a custom ACL? Is there a use case for that? It's easy enough to do but is there any issue with security/compliance? It feels like it could add some confusion/complexity to support. If there's a want/need for it I could totally be persuaded. I'm just haven't seen it mentioned that users are wanting to add their own ACLs.

npauzenga: It also looks like (according to Hector's comment) users can't do a `public` ACL. Would it be confusing if a user tried to set a `public` ACL and it just didn't work or tossed an error?

npauzenga (changes requested):
Looks great @botanical! I had a couple questions and it sounds like you might make a change based on the conversation in the architecture meeting but otherwise 👍

example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js
botanical: Yeah, I thought that as well! But I _think_ the afterAll in the beginning of the spec deletes the entire folder/path in S3 which handles deleting the objects with it https://github.com/nasa/cumulus/blob/master/example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js#L178

example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js
botanical: I considered doing that as well but I wasn't sure how to actually prove (other than using the lambdaOutput) that provided ACL was used because `getObjectAcl` returns the same `FULL_CONTROL` permission for all the canned ACLs 😓  Do you think I should add one anyway? (This entire comment might be irrelevant as well since I may just want to constrict the ACL setting to be either disabled or `private`)

tasks/sync-granule/index.js
```diff
@@ -78,6 +81,23 @@ async function download({
...
+  }
+  return 'private';
```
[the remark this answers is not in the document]
botanical: Yeah 🤔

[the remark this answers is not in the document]
botanical: Talked to the MAAP team and we're going to stick with updating sync granules only for now, and then consider making a separate PR to disable ACLs altogether @npauzenga

example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js
botanical: I added the delete calls anyway 😄  [`d01be12` (#2946)](https://github.com/nasa/cumulus/pull/2946/commits/d01be12a10703b75204c5f22ea23328cbacb13aa#diff-6b021707aab039821866154177097c5c7c8a23911b14bf7012dc6437528e5882R409-R410)

example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js
botanical: What are you thoughts on this test [`d01be12` (#2946)](https://github.com/nasa/cumulus/pull/2946/commits/d01be12a10703b75204c5f22ea23328cbacb13aa#diff-6b021707aab039821866154177097c5c7c8a23911b14bf7012dc6437528e5882R424) 👀 ?

tasks/sync-granule/index.js
```diff
@@ -88,6 +108,7 @@ async function download({
 function syncGranule(event) {
   const now = Date.now();
   const config = event.config;
+  const ACL = setOrDisableAcl(config.ACL);
```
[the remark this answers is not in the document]
botanical: Yeah I ended up constricting it [`d01be12` (#2946)](https://github.com/nasa/cumulus/pull/2946/commits/d01be12a10703b75204c5f22ea23328cbacb13aa#diff-2f2b6b18a2b5bb11ee7c3b817ca91da90127269bc36fc33ab72291f8b6ba9c2dL95-L97)

docs/tasks.md
```diff
@@ -6,173 +6,194 @@ hide_title: false
...
 
+> **NOTE:** For a detailed description of each task, visit the task's `README.md`. Information on the  input or output of a task is specified in the task's `schemas` directory.
+
 ## Tasks
 
 ### [@cumulus/add-missing-file-checksums](https://github.com/nasa/cumulus/tree/master/tasks/add-missing-file-checksums#readme)
 
 Add checksums to files in S3 which don't have one
 
+- Readme: Check out the [README](https://github.com/nasa/cumulus/tree/master/tasks/add-missing-file-checksums#readme) for additional information.
```
botanical, line 17: Definitely open to changing this if you can think of a better alternative

example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js
npauzenga: Oh cool! I didn't see the `deleteFolder` call. Yeah, I would also suspect that that would do it. 

Probably doesn't hurt to have redundant deletes here. The downside I guess would be if the `deleteFolder` does clean up S3 and then the new delete calls clutter up the logs with errors. Not sure how much I care about that 😄 . Definitely not worth holding anything up.

example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js
npauzenga: Hmm I guess I was thinking there was a way for a user to add a non-canned ACL. Maybe that's not possible though because they'd have to write their own XML and we might not have the facility to accept it? I don't totally know what I'm talking about here.

In any case, if we do decide to allow this, these tests for other canned ACLs make sense.

tasks/sync-granule/index.js
npauzenga: 👍  looks good in https://github.com/nasa/cumulus/pull/2946/commits/d01be12a10703b75204c5f22ea23328cbacb13aa#diff-2f2b6b18a2b5bb11ee7c3b817ca91da90127269bc36fc33ab72291f8b6ba9c2dL95-L97

example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js
```diff
@@ -416,6 +421,59 @@ describe('The Sync Granules workflow', () => {
     });
   });
 
+  describe('when an ACL is provided in the task config and set to the canned ACL authenticated-read', () => {
+    const meta = { ACL: 'authenticated-read' };
```
npauzenga, line 425: So given the change to only allow `private` and `disabled`, I'm not totally sure how this is passing. I think I just don't understand the expected outputs, e.g. `FULL_CONTROL`.

It seems like the new logic in `setOrDisableAcl` should set this to `private`. Does `private` then result in `FULL_CONTROL` and `authenticated-read`? 

I could be misunderstanding what's happening.

example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js
npauzenga: Given the change to only allow `private` or `disabled` I suspect we don't need these new tests?

tasks/sync-granule/index.js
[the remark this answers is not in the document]
npauzenga: Super nit-picky: now that we can't actually set a "custom" ACL, should we rename the function?

docs/tasks.md
npauzenga: What's the context for this change? Not totally following.

npauzenga:
Looks great @botanical! I added a couple comments/questions but nothing major.

docs/tasks.md
botanical: Yeah, the context for this change is to make task information clearer for users to find. This was something I could limit to sync-granule, but thought I would update this entire task.md file. Someone from the MAAP team brought up that READMEs weren't something they knew they could find in our repo, and suggested that the docs be updated to clarify that

example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js
botanical: Yeah I tried to explain when I opened the PR but the somewhat unfortunate thing about the `getObjectAcl` function is that it only returns the permissions and not the actual ACL of the object. And in many (perhaps all) cases for canned ACLs, the permissions look the same (aka FULL_CONTROL) which is probably why this passes. Let me remove this test.

botanical: Ah, so the `aunthenticated-read` ACL will have a `FULL_CONTROL` permission, like you said, but the lamdaOutput will still contain `authenticated-read` since that was what was inputted for `meta.ACL` 🤔  Which is why this test passes. The tricky thing is, it would be kind of hard to determine the accuracy of the test given that both `authenticated-read` and `private` will return `FULL_CONTROL` permissions for `getObjectAcl` 😢  

I can remove this test as well if we feel the unit test sufficiently captures the case that a user tries to input something other than `disabled`

tasks/sync-granule/index.js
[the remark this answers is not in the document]
botanical: Thoughts on `disableOrDefaultAcl`? 🤔

example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js
botanical: This test covers the case where a user inputs a value for ACL that isn't disabled and it supposed to check that the ACL that is set is `private` but the tricky part is, using `getObjectAcl` makes that tough to determine since it only reveals the `Permission` which is generally the same for all canned ACLs. And the lambda output is `authenticated-read` since that is what was provided to this step of the workflow. I can remove the test altogether since it's kind of confusing 😕  Let me know what you think!

npauzenga: Gotchya! Thanks for that. I guess I don't really feel strongly one way or the other. It makes sense to leave these I think.

tasks/sync-granule/index.js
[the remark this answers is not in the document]
npauzenga: That seems clearer to me. Again, super nit-picky.

docs/tasks.md
npauzenga: Got it. Makes sense to me!

example/spec/parallel/syncGranule/SyncGranuleSuccessSpec.js
npauzenga: Cool. Testing the case where someone enters something other than disabled in units and not integration feels right to me I think. Or clarifying what's going on here in some way, but I think you can probably just remove this test.

npauzenga (approved):
Great work @botanical! I'm gonna go ahead and give it the 👍. There are a couple pending comments about renaming a function and possibly removing a test so I'll leave those to your discretion 😄

PR #2961, opened 2022-05-12
Fix changelog entry

npauzenga:
This is a fix due to a merge error on CUMULUS-2903

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Breaking Changes
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Breaking Changes
```
npauzenga, line 9: I'm calling the Node upgrade a breaking change. I think it is because if you upgrade to the Node 14 packages we don't want to guarantee compatibility with Node 12. Does that sound reasonable? This is debatable.

jennyhliu (approved):
👍

PR #2973, opened 2022-05-25
WIP: Bug/cumulus 1 fix post build cleanup

npauzenga:


npauzenga:
This is going to require a bit of a refactor. There doesn't seem to be a set of scopes we can apply to the bootstrap that don't pull in just about everything. 

the issue here is only on new releases and it isn't blocking even then. I will create a ticket and we can prioritize accordingly.

PR #2993, opened 2022-06-14
CUMULUS-2930: Update granules API to allow listing past 10k

nemreid:

Addresses [CUMULUS-2930: Fix 10k limit](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2930)

## Changes

- The `GET /granules` endpoint has a new optional query parameter:
    `searchContext`, which is used to resume listing within the same search
    context. It is provided in every response from the endpoint as
    `meta.searchContext`. The searchContext value must be submitted with every
    consequent API call, and must be fetched from each new response to maintain
    the context.
- Updated `GET /granules` endpoint to leverage ElasticSearch search-after API.
is provided in a request.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


packages/es-client/tests/test-esSearchAfter.js
```diff
@@ -0,0 +1,72 @@
...
+      let { meta, results } = await esSearchAfter.query();
+      t.truthy(meta.searchContext);
+      t.is(results.length, testSearchSize);
+
+      const spy = sinon.spy(esSearchAfter.client, 'search');
+      let calls = 0;
+      /* eslint-disable no-await-in-loop */
+      do {
+        allResults = allResults.concat(results);
+        ({ meta, results } = await esSearchAfter.query(meta.searchContext));
```
jennyhliu, line 59: can we check that 4 granules get back for the first n calls, and the last call gets 1 result back?  Also check calls === 6?

packages/es-client/esScrollSearch.js
```diff
@@ -35,28 +35,37 @@ const {
...
+      // this.params.limit exists for API invocation vs. not for reconciliation usage
+      searchParams.size = Number(this.params.limit
+        ? this.size : (process.env.ES_SCROLL_SIZE || defaultESScrollSize));
       searchParams.scroll = process.env.ES_SCROLL || defaultESScrollDuration;
       response = await this.client.search(searchParams);
       this.scrollId = response.body._scroll_id;
     } else {
       response = await this.client.scroll({
-        scrollId: this.scrollId,
+        scrollId: searchContext || this.scrollId,
```
jennyhliu, line 53: question about `searchContext` vs `scrollId`:
`searchContext` is set to `sort` object in search.js and esSearchAfter.js, and is set to `scrollId` here. If we want to return `searchContext`, should we return JSON.stringify({ scrollId }) instead?
Is there a use case to use query(searchContext)?

nemreid: No, this is left over from the initial scroll-based implementation. It does still need an update to stringify and parse, but I can also take it out since we are not using it. Otherwise however, we could also implement a scroll-based endpoint with these changes so I am mixed on removing it entirely. Scrolls are, according to ES, not suitable for real-time user requests because maintaining the scroll context is expensive. However, unlike search-after, the scroll context is static as opposed to dynamically updated with the index as search-after is so there are some benefits to keeping the option open. Thoughts?

packages/api/endpoints/granules.js
```diff
@@ -69,17 +71,26 @@ function _returnPutGranuleStatus(isNewRecord, granule, res) {
...
 
-  let result = await es.query();
+  let result;
+  if (queryStringParameters.searchContext) {
+    const es = new ESSearchAfter(
+      { queryStringParameters },
+      'granule',
+      process.env.ES_INDEX
+    );
+    result = await es.query(queryStringParameters.searchContext);
```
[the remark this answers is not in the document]
nemreid: You're right, I will make that optimization. 👍

packages/es-client/esScrollSearch.js
jennyhliu: I would suggest to remove the changes from this file for now if we think `search-after` is the preferred way for the end user.

jennyhliu:
curl search command returns string searchContext
curl "https://dz5bdrhmhd.execute-api.us-east-1.amazonaws.com:8000/dev/granules?limit=2&sort_key%5B%5D=-granuleId&collectionId=MOD09GQ___006&status=completed&timestamp__from=1654801304583&timestamp__to=1655406104583&searchContext=%5B%22MOD09GQ.A9804659.1y_tNG.006.4259819584889%22%5D" --header 'Authorization: Bearer token' 
{"meta":{"name":"cumulus-api","stack":"jl-tf","table":"granule","count":2071,"page":1,"searchContext":"[\"MOD09GQ.A9389979.BFftFJ.006.4532584746778\"]"},

But when I search with searchContext, I have to passed in the encoded `["MOD09GQ.A9389979.BFftFJ.006.4532584746778"]` array, not the encoded searchContext string `"[\"MOD09GQ.A9389979.BFftFJ.006.4532584746778\"]"` to have the result back.
curl "https://dz5bdrhmhd.execute-api.us-east-1.amazonaws.com:8000/dev/granules?limit=2&sort_key%5B%5D=-granuleId&collectionId=MOD09GQ___006&status=completed&timestamp__from=1654801304583&timestamp__to=1655406104583&searchContext=%5B%22MOD09GQ.A9804659.1y_tNG.006.4259819584889%22%5D" --header 'Authorization: Bearer token'

packages/es-client/tests/test-esSearchAfter.js
nemreid: Updated with a dynamic set of test conditions based on `numGranules / testSearchSize` and `numGranules % testSearchSize`

packages/api/endpoints/granules.js
```diff
@@ -69,17 +71,25 @@ function _returnPutGranuleStatus(isNewRecord, granule, res) {
...
+  let es;
+  if (queryStringParameters.searchContext) {
+    es = new ESSearchAfter(
+      { queryStringParameters, sortParams: granuleSortParams },
+      'granule',
+      process.env.ES_INDEX
+    );
+  } else {
+    es = new Search(
+      { queryStringParameters, sortParams: granuleSortParams },
```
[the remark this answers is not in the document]
jennyhliu: Cumulus dashboard calls api to get a list of granules for a given time range.  For `Recent` granule, the api call would be:
https://xxx.execute-api.us-east-1.amazonaws.com:8000/dev/granules?limit=50&page=1&sort_key%5B%5D=-timestamp&timestamp__from=1657565642677
Is this query affected by the changes?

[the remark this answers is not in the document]
jennyhliu: When `limit` is set to a number > 100, it still returns 100 granules.
curl "https://xxx.execute-api.us-east-1.amazonaws.com:8000/dev/granules?limit=500&sort_key%5B%5D=-timestamp&timestamp__from=1657565642677&fields=granuleId&searchContext=%5B1657651164741%5D" --header 'Authorization: Bearer xxx'

packages/es-client/esSearchAfter.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+class ESSearchAfter extends Search {
+  /**
+   * Build search params for search-after API.
+   *
+   * @returns {Object} ES search params
+   */
+  _buildSearch() {
+    const params = super._buildSearch();
+    delete params.from;
```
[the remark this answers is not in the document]
nemreid: We need to delete `params.from` (which is always passed in from the dashboard currently) because having `params.from` causes ES to do a normal search (which is capped at 10k) and overrides any `search_after` config. `params.to` is ignored when `from` is not provided, but we could delete it too to be explicit, I guess.

packages/es-client/esDefaults.js
```diff
@@ -1,7 +1,10 @@
 const defaultESScrollSize = 1000;
 const defaultESScrollDuration = '6m';
 
+const granuleSortParams = { sort: [{ timestamp: 'desc' }, { _uid: 'asc' }] };
```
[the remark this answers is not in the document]
nemreid: The `uid` value is important for the edge case when timestamps are identical to ensure we can continue `search_after` in the middle of a duplicate timestamp range, and we don't lose the portion of records with identical time stamps that exceed the page limit.

packages/es-client/esScrollSearch.js
nemreid: All scroll changes removed in https://github.com/nasa/cumulus/pull/2993/commits/92ec3147d06575b9d3f911f98694f115b03cd676

CHANGELOG.md
```diff
@@ -98,7 +98,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 - **CUMULUS-2923**
   - Changed public key setup for SFTP local testing.
-
+- **CUMULUS-2930**
```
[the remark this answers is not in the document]
nemreid: Done in https://github.com/nasa/cumulus/pull/2993/commits/ed7186a6e591065f3e4bc824c6e5735be81f9e67

packages/es-client/esDefaults.js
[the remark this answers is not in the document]
nemreid: Moved the sort params in https://github.com/nasa/cumulus/pull/2993/commits/d4233b24baa062cd9921b551180727db379bb546 to be used as the default in `queries.js` (which effectively only adds the `_uid` tiebreaker since `timestamp: 'desc'` was already the default), so `sort_by` and `order` should now be respected properly again.

CHANGELOG.md
```diff
@@ -126,6 +126,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
   - Updated `move-granule` task to check the optional collection configuration parameter
     `meta.granuleMetadataFileExtension` to determine the granule metadata file.
     If none is specified, the granule CMR metadata or ISO metadata file is used.
+- **CUMULUS-2930**
```
jennyhliu, line 129: There is a new release, so move this up to Unreleased section

packages/es-client/tests/test-esSearchAfter.js
```diff
@@ -0,0 +1,224 @@
...
+    { queryStringParameters: { ...params, searchContext: queryResult.meta.searchContext } },
+    'granule',
+    process.env.ES_INDEX
+  );
+  const secondResult = await secondEs.query();
+  t.is(queryResult.meta.count, 10);
+  t.is(queryResult.results.length, 5);
+  const allResults = firstPage.concat(secondResult.results);
+  t.deepEqual(allResults.map((g) => g.granuleId), sortedGranules.map((g) => g.granuleId));
+});
```
jennyhliu, line 224: These are very good tests!

jennyhliu (approved):
Good  work!

PR #3000, opened 2022-06-17
CUMULUS-2688: Update bulk operations to fetch granules by unique columns

npauzenga:
**Summary:** 

Addresses [CUMULUS-2688](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2688)

This is PR 2/3 for the above ticket. This PR address the changes to bulk operations. When a bulk operation operates on granules it now needs to accept and fetch granules by `granuleId` + `collectionId`, not just `granuleId`. This is because the unique identifiers changed in the Postgres switchover.

## Changes

* Updates `BULK_GRANULE`, `BULK_GRANULE_DELETE`, and `BULK_GRANULE_REINGEST` operations in bulk-operations to support collectionId
* Updates integration tests to use unique identifiers

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased Phase 3
 
+### Breaking Changes
```
npauzenga, line 9: Wondering if there should be migration instructions for this... 🤔.

packages/api/lambdas/bulk-operation.js
```diff
@@ -5,38 +5,47 @@ const Logger = require('@cumulus/logger');
...
 const { updateGranuleStatusToQueued } = require('../lib/writeRecords/write-granules');
-const { getGranuleIdsForPayload } = require('../lib/granules');
+const { getGranulesForPayload } = require('../lib/granules');
 const { reingestGranule, applyWorkflow } = require('../lib/ingest');
 
 const log = new Logger({ sender: '@cumulus/bulk-operation' });
 
 async function applyWorkflowToGranules({
-  granuleIds,
+  granules,
```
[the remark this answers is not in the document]
nemreid: Yeah, I think it should be something better across the stack, esp at the endpoint level if we're renaming there anyway. I just can't decide on a good name. 🤔

CHANGELOG.md
nemreid: Definitely requires an update to https://github.com/nasa/cumulus-api/

nemreid (changes requested):
All the changes look fine, but I'd really like to avoid overloading `granules` even more in this way. Let's discuss regarding another term when you're available.

CHANGELOG.md
npauzenga: Yeah, good call. I'll update that repo once all three tickets for 2688 are in as they all have changes that'll need to be noted. It looks like we might need a new `rds-phase-3` branch there that gets released at the same time as the main cumulus phase 3 release too 🤔

nemreid (approved):

PR #3041, opened 2022-07-13
Release 13.0.1 (#3036)

npauzenga:
This is the 13.0.1 release merge to main. The main branch already has the `moment` upgrade that was the point of this release. This merge should just be the changelog and Cumulus version updates.

This PR sidesteps an issue on the `release-13.0.x` branch. That branch has two commits that do not exist on master. The ACTUAL CHANGES in those commits do exist on master. It appears that there was a squash or something that rolled up two commits from the release branch:

ba43ea3bfdfcacfff052770ed2fd357eb9dadb58
7e7c1fc6eaf780351c9e9709da464127b51c7ff8

 into this single commit on master: 

7f06d2d37b1cf818dc7410707b6e86f1af20a965

I don't think we should update the release-13.0.x branch because one of the commits above was used for the release and is tagged with v13. 

What we could try is merging this PR to master, then opening another for what's left on the `release-13.0.x` branch (which is technically nothing, there are no changes. Only errant commits). There were several merge conflicts that, although they didn't appear directly related to the questionable commits, could impact the state here once resolved via this merge.


jennyhliu (approved):
LGTM

PR #3054, opened 2022-07-20
fix: removing unnecessary code

vhenckel:
Removing unnecessary code

Addresses [[ftpPopulateTestLambda](https://github.com/nasa/cumulus/blob/master/example/lambdas/ftpPopulateTestLambda/index.js)

## Changes

* removing unnecessary else
* adjusting styles
* replacing the map with forEach where it is not necessary to return anything

### DIFF
--- a/example/lambdas/ftpPopulateTestLambda/index.js
+++ b/example/lambdas/ftpPopulateTestLambda/index.js
@@ -1,19 +1,17 @@
 'use strict';
 
-const {
-  randomStringFromRegex,
-} = require('@cumulus/common/test-utils');
+const { randomStringFromRegex } = require('@cumulus/common/test-utils');
 const Logger = require('@cumulus/logger');
 
 const { fetchFakeProviderIp } = require('@cumulus/common/fake-provider');
-// eslint-disable-next-line no-unused-vars
-const { granules } = require('@cumulus/test-data');
 const replace = require('lodash/replace');
 const JSFtp = require('jsftp');
 const path = require('path');
 const fs = require('fs');
 
-const log = new Logger({ sender: '@cumulus/example/lambdas/ftpPopulateTestLambda' });
+const log = new Logger({
+  sender: '@cumulus/example/lambdas/ftpPopulateTestLambda',
+});
 
 const updateAndUploadTestFileToFtpHost = (params) => {
   const {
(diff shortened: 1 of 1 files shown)


example/lambdas/ftpPopulateTestLambda/index.js
```diff
@@ -24,14 +22,11 @@ const updateAndUploadTestFileToFtpHost = (params) => {
     targetReplacementRegex,
     targetReplacementString,
   } = params;
-  let data;
+  let data = fs.readFileSync(file);
```
botanical, line 25: Nice 👍

example/lambdas/ftpPopulateTestLambda/index.js
```diff
@@ -85,24 +78,22 @@ const uploadFtpGranuleDataForDiscovery = async ({
...
-            prefix,
-            targetReplacementRegex: 'MOD09GQ.A2016358.h13v04.006.2016360104606',
-            targetReplacementString: newGranuleId,
-          });
-        } catch (error) {
-          log.info('Error!', error);
-          throw error;
-        }
-        return objectKey;
+    ftpData.forEach(async (file) => {
```
botanical, line 81: Hm I don't believe this is functionally the same since `forEach` returns `undefined`, so we will be awaiting `undefined` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#return_value

On the other hand, `map` doesn't return `undefined` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#return_value 

Do you mind elaborating on this proposed change?

example/lambdas/ftpPopulateTestLambda/index.js
```diff
@@ -130,9 +121,7 @@ const deleteFtpGranule = async (filePath) => {
   });
 };
 
-const deleteFtpGranules = async ({
-  filePaths,
-}) => {
+const deleteFtpGranules = async ({ filePaths }) => {
```
botanical, line 124: 👍

example/lambdas/ftpPopulateTestLambda/index.js
```diff
@@ -1,19 +1,17 @@
...
-const { granules } = require('@cumulus/test-data');
 const replace = require('lodash/replace');
 const JSFtp = require('jsftp');
 const path = require('path');
 const fs = require('fs');
 
-const log = new Logger({ sender: '@cumulus/example/lambdas/ftpPopulateTestLambda' });
+const log = new Logger({
+  sender: '@cumulus/example/lambdas/ftpPopulateTestLambda',
+});
```
botanical, line 14: 👍

botanical (changes requested):
Thanks for the contribution! I've provided some feedback and once you take a look, you can request a re-review 😄

example/lambdas/ftpPopulateTestLambda/index.js
vhenckel: Hello, this change was really a mistake. I have already reverted to the old code.
Thanks for the review.

vhenckel (commented):
The change from map to forEach was indeed a mistake. I've gone back to the old code.
Thanks for the review.

filipgraniczny (approved):
Thank you for correcting the issue found in @botanical's review! Looks good. 🎉

PR #3043, opened 2022-07-14
[CUMULUS-2954] Update to allow Lzards backup task as standalone in workflow

botanical:
**Summary:** Summary of changes

Addresses [CUMULUS-2954: As an integrator, I would like to run the lzards-backup task in a standalone workflow](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2954)

## Changes

* Updated task to run as standalone in a workflow
* Update task to allow collectionId to be provided in payload 

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Ad-hoc testing - Deploy changes and test manually
- [x] Integration tests

PR #3055, opened 2022-07-21
Update node versions on packages that were skipped during initial Node 14 upgrade

npauzenga:
@botanical found a node engine that was not updated to 14 in this PR: https://github.com/nasa/cumulus/pull/3043/files

It looks like any packages that missed the 12.18 update also missed the 14 update.

I don't think this will cause any issues because all of the missed packages were `>=12.0.0` so they should run fine in a v14 environment but double-check me.


botanical (approved):
Pulled down your branch and did a quick search for `"node": ">=12` and `"node":` to verify they were all updated to v14. Thanks for handling this 🙇

PR #3063, opened 2022-07-28
fix: await should not be used redundantly

vhenckel:
**Summary:** await should not be used redundantly

File [lambdas/ftpPopulateTestLambda](https://github.com/nasa/cumulus/blob/master/example/lambdas/ftpPopulateTestLambda/index.js)

## Changes

* removing await which was redundant, async functions always return promises

[Details RSPEC-4326](https://rules.sonarsource.com/typescript/RSPEC-4326)


example/lambdas/ftpPopulateTestLambda/index.js
```diff
@@ -131,10 +131,8 @@ const deleteFtpGranules = async ({ filePaths }) => {
 
 const handler = async (event, _context) => {
   log.info(event);
-  if (event.command === 'delete') {
-    return await deleteFtpGranules(event);
-  }
-  return await uploadFtpGranuleDataForDiscovery(event);
+  if (event.command === 'delete') return deleteFtpGranules(event);
```
filipgraniczny, line 134: 👍

filipgraniczny (approved):
Thank you for your contribution! 🎉  

I will now have to pull this into my branch to pass through CI.

PR #3067, opened 2022-08-02
From vhenckel fix redundantly code

filipgraniczny:
**Summary:** 

User contribution from @vhenckel. 

Addresses no open issue.

## Changes

* Remove redundant `await` statement

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/example/lambdas/ftpPopulateTestLambda/index.js
+++ b/example/lambdas/ftpPopulateTestLambda/index.js
@@ -131,10 +131,8 @@ const deleteFtpGranules = async ({ filePaths }) => {
 
 const handler = async (event, _context) => {
   log.info(event);
-  if (event.command === 'delete') {
-    return await deleteFtpGranules(event);
-  }
-  return await uploadFtpGranuleDataForDiscovery(event);
+  if (event.command === 'delete') return deleteFtpGranules(event);
+  return uploadFtpGranuleDataForDiscovery(event);
 };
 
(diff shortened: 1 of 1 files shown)


example/lambdas/ftpPopulateTestLambda/index.js
```diff
@@ -131,10 +131,8 @@ const deleteFtpGranules = async ({ filePaths }) => {
 
 const handler = async (event, _context) => {
   log.info(event);
-  if (event.command === 'delete') {
-    return await deleteFtpGranules(event);
-  }
-  return await uploadFtpGranuleDataForDiscovery(event);
+  if (event.command === 'delete') return deleteFtpGranules(event);
+  return uploadFtpGranuleDataForDiscovery(event);
```
npauzenga, line 135: I'm not sure if we want this change 🤔. We generally use `return await` in try/catch blocks to preserve error handling (https://jakearchibald.com/2017/await-vs-return-vs-return-await/). This doesn't appear to be the case here so it may be redundant but if it's not a functional improvement or fix I'm not sure if there's value in changing it.

vhenckel: It makes sense Nate.
I was just removing what was redundant but I didn't really think about the whole context.
Thanks for the feedback and the review!

PR #3071, opened 2022-08-08
CUMULUS-3002: upgrade node from 14.19.1-buster to v16.16.0-buster

filipgraniczny:
**Summary:** Updated node from `14.19.3-buster` to `gallium`

Addresses [CUMULUS-3002: [Snyk] Security upgrade node from 14.19.1-buster to gallium or gallium-buster](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3002)

## Changes

* Node upgrade in Bamboo Dockerfile
* All runtimes updated from `nodejs14.x` to `nodejs16.x`

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


npauzenga (changes requested):
The updates to the Terraform look right but we need to update all of our packages as well. See this previous PR where we did the same thing for Node 14.19.1: https://github.com/nasa/cumulus/pull/2917

Also a couple things:

1. What is the significance of the `buster` version and why are we moving away from that with the `gallium` image?
2. We'll need to update the cumulus-ecs-task to use this same Node version. See this previous PR for an example: https://github.com/nasa/cumulus-ecs-task/pull/48 (this needs to be released at the same time as this core change)
3. Testing this is somewhat complicated. You'll need to create a new Cumulus Image from this Bamboo plan https://ci.earthdata.nasa.gov/browse/CUM-CIP and then a new branch plan for this branch that uses the new image. I'll try to dig up some instruction for that and we can walk through it.

npauzenga:
@filipgraniczny I see this is in "Needs Review" but it looks like the comment above still needs addressing? Has this been tested with a Node16 container and have we updated the cumulus-ecs-task?

Also, at this point should we upgrade beyond Node16? I think we'd talked about going straight to 18 but I don't know where we landed. AWS supports 18 now so it might be worth bumping this up: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

filipgraniczny:
This has been tested with `cumulus-ecs-task` but fell victim to the security vulnerability problems last week. There is another ticket currently in the works for the Node 18 upgrade; closing this in favor of that work.

PR #3073, opened 2022-08-09
Cumulus-2926-Ingest load test

vpnguye2:
**Summary:** Ingest Load Test on collections of +200K granules

Addresses [CUMULUS-2926: Write an integration test for Ingesting a Huge number of granules](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2926)

## Changes

* Added ingest load tests to ensure users still able to ingest collections of +200K granules with Cumulus framework when code changes being introduced in a new release or feature.
* Note that the main focus is to test the Cumulus framework, not workflow components. We will not be using DiscoverGranules for this test.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [x] Integration tests


packages/api/endpoints/reconciliation-reports.js
```diff
@@ -79,6 +79,7 @@ async function getReport(req, res) {
     if (Key.endsWith('.json') || Key.endsWith('.csv')) {
       const reportSize = await getObjectSize({ s3: s3(), bucket: Bucket, key: Key });
       // estimated payload size, add extra
+      //  this was an intentional change per Jonathan's PR comment #2976
```
npauzenga, line 82: I don't think we need this comment. It looks like Jonathan was asking if the change on line 86 was intentional in the `main` to `feature/rds` merge https://github.com/nasa/cumulus/pull/2976/files#r889022958

It looks like a whitespace change caused by a different linter configuration.

vpnguye2: I think line 25 in this commit (https://github.com/nasa/cumulus/commit/9be1fa7ea443838ca9c121e1f7ef40ac988a0e24#diff-a116c50394a9d17b72826a6dc87a35d7ca471c46de55657d67d5a16da2e68cb0) which has been pushed down to line 30 now (https://github.com/nasa/cumulus/blob/d044595cdb671d32b5f9b5bbf04540c8e1614686/packages/api/endpoints/reconciliation-reports.js#L30) is what he was talking about. :) 
I had to trace way back because I recall it made more sense (than just a whitespace) when he told me at the time. But I could be wrong tho. So I should still remove this comment right?

npauzenga (changes requested):
Hey @vpnguye2 I'm gonna leave a partial review here of just the `granuleUtils` helpers. You'd mentioned possibly refactoring this to not use a workflow so I'll pause until that's completed. 

Most of the comments are just clarification or moving things around. I am concerned about putting actual files on S3 though and want to be sure we're all in agreement that we need to do that. There may not be a way around it and it gets into what we're trying to test. A recap of the intention here might be useful.

example/spec/helpers/granuleUtils.js
```diff
@@ -144,11 +145,127 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
...
+}
+
+/**
+ * Set up files in the S3 data location for numerous new granules to use for this
+ * test. Use the input payload and granuleCount to determine which files are needed and
+ * return updated input with a list of granules with new ids.
+ *
+ * @param {string} bucket - data bucket
+ * @param {string} inputPayloadJson - input payload as a JSON string
+ * @param {string} granuleCountPerPayload - number of granules per workflow
```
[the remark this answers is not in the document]
vpnguye2: Good catch ! Updated it. Thanks @npauzenga

example/spec/helpers/granuleUtils.js
```diff
@@ -144,11 +145,127 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
...
+  const oldGranuleId = granule2Replicate.granuleId;
+  const newGranuleId = randomStringFromRegex(granuleRegex);
+
+  if (passThrough) {
+    // Setup for Ingest Passthrough
+    granule2Replicate.files = granule2Replicate.files.map(
+      (file) => ({ ...file, bucket: bucket })
+    );
+
+    granule2Replicate.createdAt = Date.now();
```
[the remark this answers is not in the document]
vpnguye2: It was intended to test queueGranules behavior on a different test, but verified that it is not needed here. Hence, it is now removed. 👍

example/spec/helpers/granuleUtils.js
```diff
@@ -144,11 +145,127 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
...
+  } else {
+    await createGranuleFiles(
```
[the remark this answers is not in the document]
vpnguye2: It was there as a side-effect in case of the `ingestLoad_queueGranSFCompleted` test scenario which would carry out the ingestion via official route with the existing Ingest workflow which would require the S3 data staging.

example/spec/helpers/granuleUtils.js
```diff
@@ -144,11 +145,127 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
   ])(baseInputPayload);
 }
 
-const deleteGranules = async (prefix, granules) => {
+async function generatePayloadGranule(passThrough, granule2Replicate, bucket, granuleRegex) {
+  const oldGranuleId = granule2Replicate.granuleId;
+  const newGranuleId = randomStringFromRegex(granuleRegex);
+
+  if (passThrough) {
```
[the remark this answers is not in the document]
vpnguye2: "Ingest Passthrough" was to indicate the bypass of the Ingest workflow. `"It was a sort of dummy task (workflow) to simulate processing" 👍`  to set status from  "queued" to "completed" for the ingested granule without the official route of the Ingest workflow. Good point. Would changing the comment to `Passthrough Ingest` help with clarification?

example/spec/helpers/granuleUtils.js
```diff
@@ -144,11 +145,127 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
   ])(baseInputPayload);
 }
 
-const deleteGranules = async (prefix, granules) => {
+async function generatePayloadGranule(passThrough, granule2Replicate, bucket, granuleRegex) {
```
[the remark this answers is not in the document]
vpnguye2: Updated as suggested.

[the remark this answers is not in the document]
vpnguye2: Added docstring

packages/api/endpoints/reconciliation-reports.js
vpnguye2: Comment is removed.

vpnguye2:
> Hey @vpnguye2 I'm gonna leave a partial review here of just the `granuleUtils` helpers. You'd mentioned possibly refactoring this to not use a workflow so I'll pause until that's completed.
> 
> Most of the comments are just clarification or moving things around. I am concerned about putting actual files on S3 though and want to be sure we're all in agreement that we need to do that. There may not be a way around it and it gets into what we're trying to test. A recap of the intention here might be useful.

These test scripts should still be reviewed as they are potentially going be added to the more comprehensive test suite. Test scripts that are not using workflow will probably be added in separately. Follow-up responses to the feedback are inline. Thanks

npauzenga:
> These test scripts should still be reviewed as they are potentially going be added to the more comprehensive test suite. Test scripts that are not using workflow will probably be added in separately. Follow-up responses to the feedback are inline. Thanks

So what we have here is:

1. A new set of test utilities that set up granule payloads and stage files in S3 for ingest
2. Three new integration tests that use that new test setup and run the existing QueueGranules workflow against the newly staged granules/files.

That's definitely a step in the right direction but I think there's more to do to meet the intention noted in the PR description:
>Note that this is intended to test the Cumulus framework, not workflow components. We will not be using DiscoverGranules for this test.

The catch with these new integration tests is that they are really testing QueueGranules and all of its workflow components, not that our framework can handle a large number of granules/files (or at least not _exclusively_ that our framework can handle a large number of granules).

The QueueGranules task does quite a lot: https://github.com/nasa/cumulus/blob/master/tasks/queue-granules/index.js

I believe that's going to run a full ingest with discovery and tasks like MoveGranules. It's also going to need to set Granules to the `queued` status and likely a few other things that add complexity but that we don't really care about for these tests. It also, I assume, will limit our ability to "tune" these tests by e.g. running things in parallel because we're limited by the workflow.

I'm also a little uneasy about adding 100s of thousands of files to S3 for these tests. It's not super expensive but It's likely fairly slow and, again, not _really_ something we want to test here.

My thinking is that we should do the refactor you'd mentioned where we're creating a new task or workflow using components from QueueGranule to construct a Cumulus message without actually adding files to S3, moving granules, etc. Then use that new task/workflow with the new Cumulus message to test that pieces of the Cumulus framework like writing granules, queueing, etc. work (or don't work) as expected.

I'm looking at this function: https://github.com/nasa/cumulus/blob/master/packages/ingest/src/queue.js#L96.

It's called by `queueGranules` and is responsible for creating the actual message and putting it on the SQS queue. Could we call `enqueueGranuleIngestMessage` to create the message and use that for our tests?

npauzenga:
Just for a bit more context on my post above: 

I don't think running the QueueGranules workflow as-is is _necessarily_ a problem but the overhead that it adds was made clear when I tried to run the test locally and saw ~800 errors because my stack wasn't configured properly. That's what is leading me to think that this test is doing too much and we should try to limit it if at all possible.

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
```diff
@@ -0,0 +1,238 @@
...
+                  granuleId: granule.granuleId,
+                },
+                expectedValues
+              );
+              expect(expectedValues).toContain(record.status);
+            })
+          );
+        });
+
+        it('has expected arns output', async () => {
```
npauzenga, line 211: We talked yesterday about these tests intentionally testing workflow components as opposed to "framework" components like our queueing systems, database writes, etc. (There's still discussion to be had there and adding comments and details about what these tests are doing would be very helpful to track your thought process.)

I think, even if we move forward with these tests as "workflow component tests" we should consider the assertions we're making. These seem to be saying "QueueGranules works". That's cool but we already have tests for QueueGranules. I would think tests for workflow components would explicitly validate things related to scale. e.g. "this workflow does not throw limit exceeded errors for 10,000 granules" or something. I'm not sure which assertions would be useful but that's also a part of this effort.

example/spec/helpers/granuleUtils.js
```diff
@@ -144,11 +145,127 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
...
+    payloadJson = replace(
+      new RegExp('cumulus-test-data/pdrs', 'g'),
+      testDataFolder,
+      inputPayloadJson
+    );
+  } else {
+    payloadJson = inputPayloadJson;
+  }
+  const baseInputPayload = JSON.parse(payloadJson);
+  baseInputPayload.granules[0].dataType += testSuffix;
```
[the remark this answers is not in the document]
vpnguye2: Correct. The suffix is added to make the collection name uniquely specific to that particular payload.

example/spec/helpers/granuleUtils.js
```diff
@@ -144,11 +145,127 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
...
+  );
+  await Promise.all(createS3GranuleFilesPromises);
+  console.log(`===== Uploaded ${granuleCountPerPayload * granule2Replicate.files.length} granule files per payload =====`);
+
+  return baseInputPayload;
+}
+
+async function setupLoadTestGranuleForIngestPassthrough(bucket, inputPayloadJson, granuleCountPerPayload, granuleRegex, testSuffix = '') {
+  const baseInputPayload = JSON.parse(inputPayloadJson);
+  baseInputPayload.granules[0].dataType = `MOD09GQ${testSuffix}`;
```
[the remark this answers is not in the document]
vpnguye2: See the same response for the previous comment.

example/spec/helpers/granuleUtils.js
```diff
@@ -144,11 +145,127 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
...
+    }
+  );
+  await Promise.all(createS3GranuleFilesPromises);
+  console.log(`===== Uploaded ${granuleCountPerPayload * granule2Replicate.files.length} granule files per payload =====`);
+
+  return baseInputPayload;
+}
+
+const cleanupLoadTestGranules = async (stackName, granuleList) => {
+  const granIds = granuleList.map((g) => g.granuleId);
```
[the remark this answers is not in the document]
vpnguye2: this is the format of the granule id inside a payload input of a workflow.

example/spec/loadTest/ingestLoadTest.input.payload.json
```diff
@@ -0,0 +1,25 @@
+{
+  "granules": [
+    {
+      "granuleId": "MOD09GQ.A2016358.h13v04.006.2016360104606",
+      "dataType": "MOD09GQ",
+      "version": "006",
+      "files": [
+        {
+          "path": "cumulus-test-data/pdrs",
```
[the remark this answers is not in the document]
vpnguye2: this is the template that we are currently using in the integration parallel specs

example/spec/helpers/granuleUtils.js
```diff
@@ -144,11 +145,127 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
...
+    JSON.stringify,
+    replace(new RegExp(oldGranuleId, 'g'), newGranuleId),
+    JSON.parse,
+  ])(granule2Replicate);
+}
+
+/**
+ * Set up files in the S3 data location for numerous new granules to use for this
+ * test. Use the input payload and granuleCount to determine which files are needed and
+ * return updated input with a list of granules with new ids.
```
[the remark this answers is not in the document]
vpnguye2: Updated with the commit.

example/spec/helpers/granuleUtils.js
```diff
@@ -144,11 +145,127 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
...
+      baseInputPayload.granules.push(replicatedGranule);
+    }
+  );
+  await Promise.all(createS3GranuleFilesPromises);
+  console.log(`===== Uploaded ${granuleCountPerPayload * granule2Replicate.files.length} granule files per payload =====`);
+
+  return baseInputPayload;
+}
+
+async function setupLoadTestGranuleForIngestPassthrough(bucket, inputPayloadJson, granuleCountPerPayload, granuleRegex, testSuffix = '') {
```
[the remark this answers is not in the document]
vpnguye2: attempting to merge these functions.

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
vpnguye2: Assertion logic has been updated in the latest commit.

example/spec/helpers/granuleUtils.js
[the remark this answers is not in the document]
vpnguye2: Refactoring is in the latest code

CHANGELOG.md
```diff
@@ -41,7 +47,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
   - Updated `GET /granules` endpoint to leverage ElasticSearch search-after API.
     The endpoint will only use search-after when the `searchContext` parameter
     is provided in a request.
-
+
```
npauzenga, line 50: Might need to configure your editor to remove white space on save.

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
```diff
@@ -0,0 +1,253 @@
...
+    );
+
+    console.log(blueConsoleLog(), '\n===== Delete lambdaOutputPayload ExecutionArns =====');
+    queueGranulesChildExecutionArns.forEach((lambdaOutputPayload) => {
+      lambdaOutputPayload.running.forEach((childExecutionArn) => {
+        deleteExecution({
+          prefix: config.stackName,
+          executionArn: childExecutionArn,
+        });
+        // console.log(blueConsoleLog(), `${JSON.stringify(childExecutionArn)}`);
```
npauzenga, line 215: Remove comment

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
```diff
@@ -0,0 +1,253 @@
...
+          providersDir,
+          testSuffix
+        );
+      })
+    );
+  });
+
+  it('writes to database the expected number of granules with status completed', () => {
+    const expectedGranuleCount = granuleCountPerWorkflow * totalWorkflowCount;
+    expect(expectedGranuleCount * granuleCountThreshold < totalGranulesCompleted.length && totalGranulesCompleted.length <= expectedGranuleCount).toBeTruthy();
```
npauzenga, line 250: Can you explain what this expectation is asserting? What's the reasoning behind `granuleCountThreshold`? Why is OK for `totalGranulesCompleted.length <= expectedGranuleCount`? Shouldn't we have a problem if `totalGranulesCompleted` is less than `expectedGranuleCount`?

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
```diff
@@ -0,0 +1,253 @@
...
+          config.stackName,
+          config.bucket,
+          providersDir,
+          testSuffix
+        );
+      })
+    );
+  });
+
+  it('writes to database the expected number of granules with status completed', () => {
```
npauzenga, line 248: We'd talked about adding additional assertions here. I think using QueueGranules might be alright if we're willing to make do with the limitations in using a workflow (e.g. we have to test the workflow to test that the rest of the framework works) but I'm still hesitant about what we're actually testing.

This test seems to say that "QueueGranules successfully wrote the expected number of granules". That might be the best we can do but it feels like we're just asserting that the workflow works. Any failures in DB writes, queuing, etc. would be a side effect and not explicitly tested.

I'd like other opinions here. @Jkovarik? You had thoughts about this during PI planning.

example/spec/helpers/testUtils.js
```diff
@@ -193,6 +193,34 @@ function isValidAsyncOperationId(asyncOperationId) {
   return /[\da-f]{8}(?:-[\da-f]{4}){3}-[\da-f]{12}/.test(asyncOperationId);
 }
 
+function greenConsoleLog() {
+  const Bright = '\u001B[1m';
+  const colorReset = '\u001B[0m';
+  const FgGreen = '\u001B[32m';
+  return `${FgGreen}${Bright}%s${colorReset}`;
+}
```
npauzenga, line 201: Cool!

example/spec/helpers/granuleUtils.js
```diff
@@ -144,11 +146,120 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
...
+ * @param {Object} granuleToReplicate - granule being replicated
+ * @param {string} bucket - S3 data bucket if NOT using QueueGranulesPassthrough workflow logic
+ * @param {string} granuleRegex - regex to generate the new granule id
+ * @returns {Promise<Object>} - a granule JS object with the updated granule ids
+ */
+async function generatePayloadGranule(isPassThroughWorkflow = true, granuleToReplicate, bucket, granuleRegex) {
+  const oldGranuleId = granuleToReplicate.granuleId;
+  const newGranuleId = randomStringFromRegex(granuleRegex);
+
+  if (isPassThroughWorkflow === true) {
```
npauzenga, line 162:
```suggestion
  if (isPassThroughWorkflow) {
```

example/spec/helpers/granuleUtils.js
```diff
@@ -144,11 +146,120 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
...
+  console.log(blueConsoleLog(), `\n===== Create ${granuleCountPerPayload} granules per payload =====`);
+  const createS3GranuleFilesPromises = new Array(granuleCountPerPayload).fill().map(
+    async () => {
+      const replicatedGranule = await generatePayloadGranule(isPassThroughWorkflow, granuleToReplicate, bucket, granuleRegex);
+      baseInputPayload.granules.push(replicatedGranule);
+    }
+  );
+  await Promise.all(createS3GranuleFilesPromises);
+
+  if (isPassThroughWorkflow === false) {
```
npauzenga, line 222:
```suggestion
  if (!isPassThroughWorkflow) {
```

example/spec/helpers/granuleUtils.js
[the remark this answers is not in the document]
npauzenga: That's true and likely a problem for our workflows generally. I'm OK with addressing this in https://bugs.earthdata.nasa.gov/browse/CUMULUS-3015 unless you can find a good way to derive the collectionId here.

example/spec/helpers/granuleUtils.js
```diff
@@ -144,11 +146,120 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
...
+      return false;
+    }
+    if (!asyncOperation.body) {
+      return false;
+    }
+    if (['RUNNER_FAILED', 'TASK_FAILED'].includes(JSON.parse(asyncOperation.body).status)) {
+      throw new Error(`Cleanup failed on delete granules async operation ${JSON.stringify(asyncOperation.body)}`);
+    }
+    return JSON.parse(asyncOperation.body).status === 'SUCCEEDED';
+  }, { interval: 60 * 1000, timeout: 15 * 60 * 1000 });
```
npauzenga, line 253: What's up with the `{ interval: 60 * 1000, timeout: 15 * 60 * 1000 }`?

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
```diff
@@ -0,0 +1,253 @@
...
+const providersDir = './data/providers/s3/';
+const collectionsDir = './data/collections/s3_MOD09GQ_006';
+const granuleRegex = '^MOD09GQ\\.A[\\d]{7}\\.[\\w]{6}\\.006\\.[\\d]{13}$';
+
+// ** Configurable Variables
+
+const granuleCountPerWorkflow = 50; // 450 granules per workflow is the max allowable by the API
+const totalWorkflowCount = 8; // number of workflows to fire off
+
+const granuleCountThreshold = 0.95;
```
npauzenga, line 49: What does this do?

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
```diff
@@ -0,0 +1,253 @@
...
+  }
+  console.log(colorConsoleLog,
+    `\n___ Lambda Output of Workflow ${nthWorkflow} has ${lambdaOutput.payload.running.length}/${granuleCountPerWorkflow} of expected arns ___`);
+
+  const completedGranules = [];
+  const incompletedGranules = [];
+
+  const expectedValues = ['completed'];
+
+  await Promise.all(
```
npauzenga, line 142: If I'm reading this right I don't think you need the `Promise.all`. You're `awaiting` in `map`, which should make this synchronous. I could be misreading though.

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
```diff
@@ -0,0 +1,253 @@
...
+      queueGranulesExecutionArns.forEach((qGranulesExecutionArn) => {
+        deleteExecution({
+          prefix: config.stackName,
+          executionArn: qGranulesExecutionArn,
+        });
+        console.log(blueConsoleLog(), `${JSON.stringify(qGranulesExecutionArn)}`);
+      });
+    }
+
+    await Promise.all(
```
npauzenga, line 230: Same as above. You _may_ not need a `Promise.all` here. It looks like it would be synchronous. Double-check what's going on in `cleanupProviders` though. I don't think it returns a promise but could be wrong.

example/spec/loadTest/ingestLoad_queueGranSFCompleted.js
```diff
@@ -0,0 +1,278 @@
...
+    }
+    console.log(colorConsoleLog,
+      `\n___ Lambda Output of Workflow ${nthWorkflow} has ${lambdaOutput.payload.running.length}/${granuleCountPerWorkflow} of expected arns ___`);
+
+    completedGranules = [];
+    incompletedGranules = [];
+
+    const expectedValues = ['completed'];
+
+    await Promise.all(
```
npauzenga, line 160: Same as above, double check that you need to await promises here.

example/spec/loadTest/ingestLoad_queueGranSFCompleted.js
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    console.log(blueConsoleLog(), `\n===== Clean up S3TestDataFolder: ${testDataFolder} =====`);
+    await deleteFolder(config.bucket, testDataFolder);
+  }
+};
+
+describe('The Granule Ingest Load Test ', () => {
+  beforeAll(async () => {
+    if (!Number.isFinite(Number(granuleCountPerWorkflow)) || granuleCountPerWorkflow > 425) {
```
npauzenga, line 201: We should store `425` as a variable to spell out what it represents.

npauzenga: Also, is `!Number.isFinite(Number(granuleCountPerWorkflow))` the same as `Number(granuleCountPerWorkflow) >= 0`? Are we trying to account for negative numbers?

example/spec/loadTest/ingestLoad_queueGranSFCompleted.js
```diff
@@ -0,0 +1,278 @@
...
+    );
+
+    console.log(blueConsoleLog(), '\n===== Delete lambdaOutputPayload ExecutionArns =====');
+    queueGranulesChildExecutionArns.forEach((lambdaOutputPayload) => {
+      lambdaOutputPayload.running.forEach((childExecutionArn) => {
+        deleteExecution({
+          prefix: config.stackName,
+          executionArn: childExecutionArn,
+        });
+        // console.log(blueConsoleLog(), `${JSON.stringify(childExecutionArn)}`);
```
npauzenga, line 240: Remove comment

CHANGELOG.md
vpnguye2: Updated the editor setting to trim trailing whitespace

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
vpnguye2: Comment removed

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
```diff
@@ -0,0 +1,253 @@
+const fs = require('fs');
```
npauzenga, line 1: let's think about the naming here. In 6 months I'm not sure we're going to know what `ingestLoad_queueGranPassThruSF` means. What about something like `queueGranulePassThroughLoadTest`? Same would apply to the other tests.

npauzenga: The "Pass Through" concept is also potentially confusing if you don't have all of the context so I'm open to other ideas.

vpnguye2: I see your good point. Im open to the name change. Note that the queueGranPassThruSF = QueueGranulesPassthrough Step Functions which is the actual existing workflow name in Core currently, however. 
Same as the queueGranSF being the name of the existing QueueGranules workflow. So the naming kinda indicates the usage of existing workflows. But suggestions are welcome.

example/spec/loadTest/ingestLoad_queueGranSFCompleted.js
vpnguye2: comment removed

example/spec/loadTest/ingestLoad_queueGranSFCompleted.js
vpnguye2: Assigned 425 to a meaningful variable name. 
Number.isFinite checks the number is neither positive infinity, negative Infinity, nor NaN

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
npauzenga: Put another way, what is the question these tests are attempting to answer? How many granules can we ingest using Cumulus before it blows up? Can we answer that?

example/spec/helpers/granuleUtils.js
vpnguye2: These are the Options params for node_models > pWaitFor, as in Number of milliseconds to wait before retrying and automatically
rejecting.

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
vpnguye2: This means to “allow” 5% room for ingestion failure/timeout since we are working with a large volume of granules, taking into account the AWS resource availability at runtime. So long as the system load ingest capability can be validated, the 95% round about number should suffice. This idea was actually "derived" from an existing load test previously written by the one and only Jonathan.

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
vpnguye2: The question these tests are attempting to answer is, Can the system still handle ingestion of a large specified amount of granules after any code change? And they did/should.

example/spec/helpers/granuleUtils.js
vpnguye2: It's always good to be explicit when checking if a variable is equal to a specific value.

example/spec/helpers/granuleUtils.js
vpnguye2: Same as previous response.

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
vpnguye2: Added the comment for this variable.
// Percent granules to count as a successful batch on ingest timeout

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
vpnguye2: It wouldn't work without the Promise.all with my current code layout.

example/spec/loadTest/ingestLoad_queueGranSFCompleted.js
vpnguye2: Validated in my previous response.

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
vpnguye2: Changed script names as suggested.

example/spec/loadTest/queueGranulePassThroughLoadTest.js
```diff
@@ -0,0 +1,253 @@
...
+        process.env.GranulesTable = `${config.stackName}-GranulesTable`;
+        process.env.ExecutionsTable = `${config.stackName}-ExecutionsTable`;
+        process.env.CollectionsTable = `${config.stackName}-CollectionsTable`;
+
+        await pMap(
+          new Array(totalWorkflowCount).fill().map((_, index) => index + 1),
+          async (index) => {
+            await batchGranulesProcessing(index);
+          },
+          { concurrency: 8 }
```
Jkovarik, line 193: This should probably be a configurable variable, or possibly set to totalWorkflowCount?         Some of that depends on the outcome of the workflow/parallelism discussion.

example/spec/loadTest/queueGranulePassThroughLoadTest.js
```diff
@@ -0,0 +1,253 @@
...
+          providersDir,
+          testSuffix
+        );
+      })
+    );
+  });
+
+  it('writes to database the expected number of granules with status completed', () => {
+    const expectedGranuleCount = granuleCountPerWorkflow * totalWorkflowCount;
+    expect(expectedGranuleCount * granuleCountThreshold < totalGranulesCompleted.length && totalGranulesCompleted.length <= expectedGranuleCount).toBeTruthy();
```
Jkovarik, line 250: This test is validating that the granules all completed after some period of time, given some percentage of acceptable?      That seems..... within the parameters of the initial test intent, except we should be representing that *all* of them ingest, or if there's an acceptable failure rate despite retries, I would expect it needs to be much lower than 5% to consider the test a success at this ingest rate.      

There's room to talk about data integrity concerns at load here, but that could probably be acceptably pushed to a downstream epic ticket.     I think the focus of this ticket is looking at tuning/architecture failures, and the data we're providing 

Given we're operating in a multi-datastore environment, I think we need to also evaluate elastic search counts here, and possibly consider validating file records (e.g. granule API return files object) and such.    

For it to be a valid, repeatable test scenario should we document some of this as well?

example/spec/loadTest/queueGranuleCompletedLoadTest.js
```diff
@@ -0,0 +1,277 @@
+const fs = require('fs');
```
Jkovarik, line 1: Is there a need for this test in the context of the ticket?   We should discuss and consider removing this for this PR unless it meets the specific objective of 2926.

example/spec/loadTest/queueGranuleQueuedLoadTest.js
```diff
@@ -0,0 +1,242 @@
+const fs = require('fs');
```
Jkovarik, line 1: What is the intended test target for this test?  We should discuss and consider removing this for this PR unless it meets the specific objective of 2926.

Jkovarik (changes requested):
Did another review.  Thanks for your patience.      This generally looks good, but there are some questions about the approach/intent that should be discussed.      I'm at your disposal whenever we can get that scheduled.

example/spec/loadTest/ingestLoad_queueGranPassThruSF.js
Jkovarik: I think there's a legit discussion here.  

I commented in the review https://github.com/nasa/cumulus/pull/3073#discussion_r975772654 re: it, but I'm not as concerned about some of the data integrity concerns we've uncovered elsewhere as those were logic failures.  We do need to be careful we're sure all records make it in, and not rely on counts, but the purpose of this test should be narrowly constrained to load related concerns.    Happy to discuss further what I mean by that.

example/spec/loadTest/queueGranulePassThroughLoadTest.js
vpnguye2: Removed granuleCountThreshold

example/spec/loadTest/queueGranuleCompletedLoadTest.js
vpnguye2: Removed queueGranuleCompletedLoadTest

example/spec/loadTest/queueGranuleQueuedLoadTest.js
vpnguye2: Removed queueGranuleQueuedLoadTest

example/spec/loadTest/queueGranulePassThroughLoadTest.js
```diff
@@ -0,0 +1,253 @@
...
+const inputPayloadJson = fs.readFileSync(inputPayloadFilename, 'utf8');
+
+const providersDir = './data/providers/s3/';
+const collectionsDir = './data/collections/s3_MOD09GQ_006';
+const granuleRegex = '^MOD09GQ\\.A[\\d]{7}\\.[\\w]{6}\\.006\\.[\\d]{13}$';
+
+// ** Configurable Variables
+
+const granuleThresholdPerWorkflow = 450;
+const granuleCountPerWorkflow = 50; // 450 granules per workflow is the max allowable by the API
```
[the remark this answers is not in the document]
vpnguye2: Object({ code: 'ValidationException', time: Date(Thu Sep 22 2022 09:53:04 GMT-0500 (Central Daylight Time)), requestId: '498762db-3121-4baa-939c-0b3201cac739', statusCode: 400, retryable: false, retryDelay: 16.85302387714025 })
   ValidationException: 1 validation error detected: Value at 'input' failed to satisfy constraint: Member must have size less than or equal to 262144 bytes in UTF-8 encoding
EED/git/cumulus/packages/aws-client/node_modules/aws-sdk/lib/request.js:688:12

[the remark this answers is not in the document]
Jkovarik: Ah, yes, there's an open ticket re: that limit in the OP support epic.      Fair enough, thanks for the clarification.

example/spec/helpers/granuleUtils.js
```diff
@@ -144,11 +146,120 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
...
+  const oldGranuleId = granuleToReplicate.granuleId;
+  const newGranuleId = randomStringFromRegex(granuleRegex);
+
+  if (isPassThroughWorkflow === true) {
+    // Setup for Ingest using QueueGranulesPassthrough workflow logic
+    granuleToReplicate.files = granuleToReplicate.files.map(
+      (file) => ({ ...file, bucket: bucket })
+    );
+  } else {
+    await createGranuleFiles(
```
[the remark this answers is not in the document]
Jkovarik: This concern is OBE by later changes.

example/spec/helpers/testUtils.js
vpnguye2: @Jkovarik I need help implementing the switchable for CI logic. Thanks

Jkovarik: @vpnguye2  I'd be happy to.   The idea is to switch the behavior if a bamboo-run-specifc env var is in place.     It can either be something we set in set-bamboo-env-varaibles or more narrowly in wherever we're going to execute this.   

It'd probably also be ok to punt this concern until we're actually running it in CI, if we have a ticket to run this test (e.g. add it to ... deployment, ????) addressing this in that AC would probably be fine as well. 

Feel free to ping me in a DM/channel if you want to pair on it.

example/spec/loadTest/queueGranulePassThroughLoadTest.js
```diff
@@ -0,0 +1,253 @@
...
+
+  const completedGranules = [];
+  const incompletedGranules = [];
+
+  const expectedValues = ['completed'];
+
+  await Promise.all(
+    inputPayload.granules.map(async (granule) => {
+      try {
+        const record = await waitForApiStatus(
```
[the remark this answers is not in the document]
vpnguye2: Could you help recommend other more sustainable method of verification for granule status here? TIA

PR #3087, opened 2022-08-22
CUMULUS-2792: Remove API granules model

botanical:

Addresses [CUMULUS-2792: Remove API granules model](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2792)

## Changes

* Removed granule model code references 
* Remove granule-csv endpoint

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


example/spec/helpers/granuleUtils.js
```diff
@@ -145,29 +144,20 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
...
-      // and cannot be deleted via the API
-      try {
-        if (granule.published === true) {
-          return await removePublishedGranule({
-            prefix,
-            granuleId: granule.granuleId,
-          });
-        }
-        return await deleteGranule({
+    // Temporary fix to handle granules that are in a bad state
```
charleshuang80, line 149: Some questions here:
1. what was/is the 'Temporary fix', and is it still needed?
2. was the `try...catch` related to the Temporary fix? curious what the reason is that it is no longer needed
3. not important, but I am so curious what the `TODO wat` was all about

botanical: `what was/is the 'Temporary fix', and is it still needed?`
I could be wrong but my understanding is that this temporary fix addresses the need for granules to be unpublished from CMR before being deleted, and that logic is still in place? Though I say I could be wrong because that functionality seems permanent rather than temporary 🤔 

`was the try...catch related to the Temporary fix? curious what the reason is that it is no longer needed`
Yes? The reason the try/catch is no longer needed is because we no longer need to delete the granule manually from DynamoDB

example/spec/helpers/granuleUtils.js
```diff
@@ -145,29 +144,20 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
 }
 
 const deleteGranules = async (prefix, granules) => {
-  process.env.GranulesTable = `${prefix}-GranulesTable`;
-  const granuleModel = new Granule();
-  return await Promise.all(
+  await Promise.all(
```
jennyhliu, line 147: why `return` is removed?

jennyhliu: The granule delete endpoint returns information that can be useful for verification purpose
https://github.com/nasa/cumulus/pull/3093/files

charleshuang80 (approved):

example/spec/helpers/granuleUtils.js
botanical: I remove `returned` because it gives the lint error
```
Unexpected block statement surrounding arrow body; move the returned value immediately after the `=>`  arrow-body-style
```

example/spec/helpers/granuleUtils.js
```diff
@@ -145,29 +144,20 @@ async function setupTestGranuleForIngest(bucket, inputPayloadJson, granuleRegex,
...
-          return await removePublishedGranule({
-            prefix,
-            granuleId: granule.granuleId,
-          });
-        }
-        return await deleteGranule({
+    // Temporary fix to handle granules that are in a bad state
+    // and cannot be deleted via the API
+      if (granule.published === true) {
+        return await removePublishedGranule({
```
jennyhliu, line 152: This shouldn't return, and should continue to `return await deleteGranule(`

jennyhliu (approved):

PR #3100, opened 2022-09-12
CUMULUS-3000:Add data_type(dataType) to collection table and object

jennyhliu:

Addresses [CUMULUS-3000: When adding a new collection to the Cumulus, the dataType object in the collection definition file disappears when submitting for the first time](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3000)

## Changes

* ...

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


Jkovarik (changes requested):
I don't believe we should be re-adding the dataType property to collections.  This was a deliberate removal.

jennyhliu:
Based on the discussion, we will not fix.
cumulus: There is an optional meta field in collection schema.
https://github.com/nasa/cumulus/blob/master/packages/api/models/schemas.js#L211
You can move dataType to there, and  use collection.meta.dataType (or a different name you like).
DAAC: Yes, we can add an optional field under meta . We need to change all of our collection definition then.

PR #3102, opened 2022-09-14
CUMULUS-2557: fix S3 moveObject such that a 0 byte file can be moved

charleshuang80:

Addresses [CUMULUS-2557: fix S3 moveObject such that a 0 byte file can be moved](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2557)

## Changes

* fix S3 moveObject such that a 0 byte file can be moved (and not encounter MalformedXML error because of multipartCopyObject)
* update moveGranules integration tests for the issue
* update comments where relevant

## PR Checklist

- [x] Update CHANGELOG
- [o] Unit tests
- [x] Integration tests


example/spec/parallel/moveGranules/MoveGranulesTagsAndZeroByteFileSpec.js
```diff
@@ -0,0 +1,145 @@
+'use strict';
```
charleshuang80, line 1: Not all of this test is new, I combined new tests with the test file below that is being deleted/renamed with some test setup changes.

packages/aws-client/src/S3.ts
```diff
@@ -1003,6 +1003,13 @@ const uploadPartCopy = async (
  * @param {number} [params.chunkSize] - chunk size of the S3 multipart uploads
  * @returns {Promise.<{ etag: string }>} object containing the ETag of the
  *    destination object
+ *
```
jennyhliu, line 1006: Do we need so many comments here?  It should be enough to just add 'the method is for non-zero size file only'.
The error message  mentioned could change or no longer valid in the future, and can cause confusion.

packages/aws-client/tests/S3/test-multipartCopyObject.js
```diff
@@ -1,5 +1,11 @@
 'use strict';
 
+/*
```
jennyhliu, line 3: I suggest to remove the comments here.  It's not related to the tests.
Add a test for `moveObject` with 0 size file

example/spec/parallel/moveGranules/MoveGranulesTagsAndZeroByteFileSpec.js
jennyhliu: If you do `git mv filename newfilename`, it will preserve the history.
Maybe you can rename it back and checkin, then rename again.

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- **CUMULUS-2557**
+  - Updated `@cumulus/aws-client/S3/moveObject` such that moveGranules operations with zero byte
+    files (0 byte files) will succeed and not throw MalformedXML errors. Updated moveGranules tests
+    in cover such operations, and added comments for the issue.
```
jennyhliu, line 29: remove `Updated moveGranules tests in cover such operations, and added comments for the issue. `

jennyhliu (changes requested):

CHANGELOG.md
jennyhliu: `@cumulus/aws-client/S3/moveObject` could be used for other things, so maybe:
Updated `@cumulus/aws-client/S3/moveObject` to handle zero byte files.

packages/aws-client/src/S3.ts
charleshuang80: I think that the MalformedXML error is a misleading error message from AWS, and that's why I thought it might be helpful to have the information. I also generally prefer a bit more information than less. But I agree the error messages may change, and as long as someone can find the references they can find the history and the error messages encountered, so the comments can be trimmed down.

example/spec/parallel/moveGranules/MoveGranulesTagsAndZeroByteFileSpec.js
charleshuang80: I attempted your suggestion. When I pushed and it was the old file name, it showed here as the one file with all the changes. I used the command to commit and push, and it showed here as 2 different files again even though before the commit it showed as a rename change.

jennyhliu: We can use the original filename.

charleshuang80: But the file is testing more than just Tags, it is also testing the ZeroByteFile. Using the old filename seems incomplete or misleading, doesn't it?

jennyhliu: I think it's better than losing the git history.  We can rename it to a more generic name later

jennyhliu (approved):
👍 Verified that granule with zero-size file can be ingested into the system, and IngestAndPublishGranule (with MoveGranule step) is successful.

PR #3103, opened 2022-09-16
CUMULUS-2943: Bulk delete does not fail if no PG granule

nemreid:

Addresses [CUMULUS-2943: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2943)

## Changes
  - Bulk granule deletion will continue with DDB/ES if the granule does not exist in PG.

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


jennyhliu (changes requested):
I have a granule which is listed in cumulus dashboard, and it's in the DynamoDB table as well.  The granule is missing from the postgres, since the granule detail page has error: `Granule not found`.
When I tried to delete the granule, it returns error:
`1 error(s) occurred: Granule MOD09GQ.A0232712.7yn1sN.006.7189179839261 does not exist or was already deleted`

packages/api/tests/lambdas/test-bulk-granule-delete.js
```diff
@@ -184,3 +184,147 @@ test('bulkGranuleDelete does not fail on published granules if payload.forceRemo
...
+  t.false(await t.context.esGranulesClient.exists(granuleId));
+
+  const s3Buckets = granule.s3Buckets;
+  t.teardown(() => deleteS3Buckets([
+    s3Buckets.protected.name,
+    s3Buckets.public.name,
+  ]));
+});
+
+test.serial('bulkGranuleDelete does not fail if granule is not in PG and payload.forceRemoveFromCmr is true', async (t) => {
```
jennyhliu, line 259: This is for `published` granule, right?

npauzenga:
Closing this PR. The feature is still relevant but the code here is stale. We can make Postgres and Elastic deletes idempotent by removing the `return` statement here: https://github.com/nasa/cumulus/blob/master/packages/api/lambdas/bulk-operation.js#L139 (and testing). The ticket will be updated and a new PR will be required.

PR #3104, opened 2022-09-21
CUMULUS-3051: Propagate ACL value in S3ProviderClient for multipartCopyObject call

botanical:

Addresses [CUMULUS-3051: Propagate ACL value in S3ProviderClient for multipartCopyObject call](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3051)

## Changes

* Update S3ProviderClient.sync() to take ACL parameter

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests

#### Ad-hoc testing - Deploy changes and test manually
It was necessary to create an S3 bucket in a different account since ACL errors don't show when you write objects to from one account to an S3 bucket owned by that same AWS account no matter the configuration of the bucket object ownership (ACLs disabled or enabled).

I tested by creating a bucket in our SIT account and then redeployed my dev Cumulus stack to reference the new SIT bucket. Updated the SIT bucket to have ACLs disabled (necessary to do after deploying rather than before because the deployment will fail since we use `aws_s3_bucket_object` which[ defaults the ACL to `private` ](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_object)). Then I ran SyncGranuleSuccessSpec.js to ensure that the error `AccessControlListNotSupported` appears when the SyncGranule workflow isn't configured to disable ACLs in order to verify that the[ multipartCopyObject call fails since it originally defaults the ACL to `private`. 
](https://github.com/nasa/cumulus/blob/master/packages/ingest/src/S3ProviderClient.ts#L135)
Then I implemented my parameter changes and redeployed, and ran the tests using my SIT bucket as the system bucket, and verified that the SyncGranuleSuccessSpec test succeeds.


packages/ingest/test/test-S3ProviderClient.js
```diff
@@ -158,3 +159,37 @@ test.serial('S3ProviderClient.sync throws an error if the source file does not e
...
+    fileRemotePath: t.context.sourceKey,
+  });
+  const objectACL = await s3().getObjectAcl({
+    Bucket: t.context.targetBucket,
+    Key: targetKey,
+  });
+  t.is(objectACL.Grants[0].Permission, 'FULL_CONTROL');
+});
+
+test.serial('S3ProviderClient.sync syncs a file when ACL parameter is not specified', async (t) => {
```
botanical, line 180: I'm not entirely sold that this test is even necessary

tasks/sync-granule/tests/sync_granule_test.js
```diff
@@ -1062,6 +1062,54 @@ test.serial('download multiple granules from S3 provider to staging directory',
   }
 });
 
+test.serial('download() ingests multiple granules from S3 provider to staging directory with ACL', async (t) => {
+  t.context.event.input.granules = t.context.event_multigran.input.granules;
+  try {
+    await prepareS3DownloadEvent(t);
+    t.context.event.config.ACL = 'disabled';
```
botanical, line 1069: Setting ACL here. Not sure if this test is necessary and maybe it should be combined with the test here https://github.com/nasa/cumulus/pull/3104/files#diff-80939d13722e8a8ef23dfa696e83b3361b1d72ea586b7c54dd5d38cf52ef1fecR1024 ?

botanical:
Suggestion from @kkelly51 - Update docs to list out the canned ACLs that AWS accepts

Jkovarik (commented):
The implementation looks sound, I wasn't able to allocate time to evaluate the deployment *fixed* the issue MAP was reporting, and have a comment on the tests.

tasks/sync-granule/tests/sync_granule_test.js
botanical: Yes, I am suggesting to combine it with `download multiple granules from S3 provider to staging directory` test

tasks/sync-granule/tests/sync_granule_test.js
```diff
@@ -1062,6 +1062,54 @@ test.serial('download multiple granules from S3 provider to staging directory',
...
+            Bucket: t.context.internalBucketName,
+            Key: `${keypath}/${granuleFileName}`,
+          })
+        );
+        // eslint-disable-next-line no-await-in-loop
+        const objectACL = await s3().getObjectAcl({
+          Bucket: t.context.internalBucketName,
+          Key: `${keypath}/${granuleFileName}`,
+        });
+        t.is(objectACL.Grants[0].Permission, 'FULL_CONTROL');
```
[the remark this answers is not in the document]
botanical: https://github.com/nasa/cumulus/pull/3104/commits/be4100f68122486a87fedb4c274723aed075d59e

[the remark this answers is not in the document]
Jkovarik: Looks good, ty!

npauzenga:
@botanical just going through old PRs. Does this need a fresh set of eyes? It looks like there were comments that may need addressing?

botanical:
> @botanical just going through old PRs. Does this need a fresh set of eyes? It looks like there were comments that may need addressing?

@npauzenga Hey Nate! Thanks for following up. I think this does need a fresh set of eyes. I believe I did address all the comments as well. The main thing holding up this review, I believe, is that testing it manually isn't trivial. It would be great to get this reviewed though so we can fix the sync-granule ACL configuration bug.

packages/ingest/src/S3ProviderClient.ts
```diff
@@ -81,6 +82,7 @@ class S3ProviderClient implements ProviderClient {
    *
    * @param {Object} params - the full path to the remote file to be fetched
    * @param {string} params.sourceKey - the full path to the remote file to be fetched
+   * @param {string} params.ACL
```
marchuffnagle, line 85: Specify correct type in JSDocs, and indicate that it's optional with `[]`

```suggestion
   * @param {ObjectCannedACL} [params.ACL]
```

packages/ingest/src/S3ProviderClient.ts
```diff
@@ -89,13 +91,14 @@ class S3ProviderClient implements ProviderClient {
...
     params: {
+      ACL?: ObjectCannedACL,
       bucket?: string,
       destinationBucket: string,
       destinationKey: string,
       fileRemotePath: string,
     }
   ): Promise<{ s3uri: string, etag: string }> {
-    const { fileRemotePath, destinationBucket, destinationKey, bucket } = params;
+    const { fileRemotePath, destinationBucket, destinationKey, bucket, ACL } = params;
```
marchuffnagle, line 101: Maintain backward compatibility.

```suggestion
    const { fileRemotePath, destinationBucket, destinationKey, bucket, ACL = ObjectCannedACL.private } = params;
```

marchuffnagle:
Should we be applying an ACL in this case as well?

https://github.com/nasa/cumulus/blob/CUMULUS-3051/packages/ingest/src/S3ProviderClient.ts#L111

packages/ingest/src/types.ts
```diff
@@ -40,6 +40,7 @@ export interface ProviderClient {
       destinationBucket: string,
       destinationKey: string,
       bucket?: string,
+      ACL?: string,
```
marchuffnagle, line 43:
```suggestion
      ACL?: ObjectCannedACL,
```

packages/ingest/test/test-S3ProviderClient.js
marchuffnagle: It looks like the default ACL before this change was `private`. Should we be verifying that in this test?

packages/ingest/test/test-S3ProviderClient.js
```diff
@@ -161,3 +162,37 @@ test.serial('S3ProviderClient.sync throws an error if the source file does not e
...
+    destinationBucket: t.context.targetBucket,
+    destinationKey: targetKey,
+    fileRemotePath: t.context.sourceKey,
+  });
+  const objectACL = await s3().getObjectAcl({
+    Bucket: t.context.targetBucket,
+    Key: targetKey,
+  });
+  t.is(objectACL.Grants[0].Permission, 'FULL_CONTROL');
+});
```
marchuffnagle, line 198: Please add a test for a zero-length file as well.

tasks/sync-granule/GranuleFetcher.js
```diff
@@ -338,14 +340,15 @@ class GranuleFetcher {
    * @param {Object} file - file to download
    * @param {string} destinationBucket - bucket to put file in
    * @param {string} duplicateHandling - how to handle duplicate files
+   * @param {string | undefined} ACL
```
marchuffnagle, line 343:
```suggestion
   * @param {string} [ACL]
```

marchuffnagle: Given the types in the files up above, this should be `ObjectCannedACL`, but that opens a can of worms that isn't worth getting into in this PR.

tasks/sync-granule/tests/sync_granule_test.js
```diff
@@ -25,6 +25,7 @@ const {
   validateInput,
   validateOutput,
 } = require('@cumulus/common/test-utils');
+const S3ProviderClient = require('../../../packages/ingest/S3ProviderClient');
```
marchuffnagle, line 28:
```suggestion
const S3ProviderClient = require('@cumulus/ingest/S3ProviderClient');
```

marchuffnagle (changes requested):

botanical:
Closing this since it is no longer valid

PR #3121, opened 2022-10-11
Release 13.3.x with master

npauzenga:
This merges the 13.3.2 release back to main. Note that it's a little weird here because 13.3.2 is technically a backport because we did not release everything on main BUT it is also the most forward version. This means that the Cumulus version should be updated in all of our packages to 13.3.2.


CHANGELOG.md
```diff
@@ -69,22 +54,35 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
+
 ### Fixed
 
 - **CUMULUS-2557**
   - Updated `@cumulus/aws-client/S3/moveObject` to handle zero byte files (0 byte files).
-- **CUMULUS-2969**
-  - Updated `@cumulus/api/models/rules.buildPayload` to only include
-    stepFunction name and arn in for the `definition` return value, excluding
-    step function definition and other extraneous step function object
-    key/values that are not used downstream, but were causing rules to exceed internal AWS limits.
```
botanical, line 80: This should be in the `Fixed` section of `Unreleased`, right?

botanical (approved):
I was confused about one thing but everything else looks great so I didn't want to block on my confusion. Approved ✅

PR #3123, opened 2022-10-12
CUMULUS-2915: Update /executions/status/executionArn endpoint to include presigned s3 URL

jennyhliu:

Addresses [CUMULUS-2915: Dashboard errors when ingesting ATL12 v005 data sets (many file granules)](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2915)

## Changes

* Due to the lambda payload limit 6MB, when execution status is too large, it can't returned via api endpoint.  By adding presignedS3Url to the response of the /executions/status/executionArn endpoint, user can download the execution status.
* API PR https://github.com/nasa/cumulus-api/pull/323
* Dashboard https://github.com/nasa/cumulus-dashboard/pull/1079
* Fix ci build error on IngestGranuleFailureSpec.js
const apiMessage = JSON.parse(error.apiMessage)
`
>> An error was thrown in an afterAll
--
03-Nov-2022 19:39:06 | SyntaxError: Unexpected token u in JSON at position 0
03-Nov-2022 19:39:06 | at JSON.parse (<anonymous>)
03-Nov-2022 19:39:06 | at UserContext.<anonymous> /cumulus/example/spec/parallel/ingestGranule/IngestGranuleFailureSpec.js:111:31
`

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


tf-modules/archive/main.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 3.0,!= 3.14.0"
+      version = ">= 3.75.2"
```
jennyhliu, line 5: aws_s3_bucket_lifecycle_configuration feature is only available >= 3.75.2

packages/cmrjs/package.json
```diff
@@ -40,7 +40,7 @@
     "@cumulus/errors": "13.3.2",
     "@cumulus/launchpad-auth": "13.3.2",
     "@cumulus/logger": "13.3.2",
-    "got": "^11.8.1",
+    "got": "^11.8.5",
```
npauzenga, line 43: So I was having trouble with my bootstrap and saw an error similar to what you saw with `keyv`. I eventually got it to work by _only_ upgrading npm though. I reverted the change here on my branch and used `"got": "^11.8.1",`, which worked fine. Might be worth re-trying.

I think "^11.8.1" and "^11.8.5" will also functionally work the same? They'll both pull the latest patch release on the minor version?

packages/api/tests/endpoints/test-execution-status.js
```diff
@@ -495,16 +500,34 @@ test('when execution is no longer in step function API, returns status from data
...
-  const executionStatus = response.body;
-  t.is(executionStatus.error, 'Not Found');
-  t.is(executionStatus.message, `Execution record with identifiers ${JSON.stringify({ arn: expiredMissingExecutionArn })} does not exist.`);
+  const responseBody = response.body;
+  t.is(responseBody.error, 'Not Found');
+  t.is(responseBody.message, `Execution record with identifiers ${JSON.stringify({ arn: expiredMissingExecutionArn })} does not exist.`);
+  t.falsy(response.body.presignedS3Url);
+});
+
+test.serial('when execution status exceeds maximum allowed payload size', async (t) => {
```
botanical, line 520: 👍

packages/api/endpoints/execution-status.js
```diff
@@ -1,10 +1,15 @@
 'use strict';
 
 const router = require('express-promise-router')();
+const moment = require('moment');
 const { getStateMachineArnFromExecutionArn } = require('@cumulus/message/Executions');
 const { pullStepFunctionEvent } = require('@cumulus/message/StepFunctions');
+const S3ObjectStore = require('@cumulus/aws-client/S3ObjectStore');
+const { s3 } = require('@cumulus/aws-client/services');
+const { buildS3Uri } = require('@cumulus/aws-client/S3');
```
botanical, line 9: We can avoid importing `@cumulus/aws-client/services` if we import `s3PutObject` function and pass in `ACL: undefined` to `s3PutObject`
```suggestion
const { buildS3Uri, s3PutObject } = require('@cumulus/aws-client/S3');
```

botanical (approved):
Deployed and tested endpoint and dashboard. Had one comment but otherwise looks good

packages/cmrjs/package.json
jennyhliu: It's the same, and 11.8.5 is the latest version for 11.8.x

PR #3127, opened 2022-10-17
[CUMULUS-3056] Setup for testing Postgres LIST queries

npauzenga:
Ticket: https://bugs.earthdata.nasa.gov/browse/CUMULUS-3056

This PR is a bit of an oddball. The idea is outlined here: https://wiki.earthdata.nasa.gov/display/CUMULUS/Eliminate+ElasticSearch+-+Initial+Research

We're attempting to set up a framework to test ElasticSearch vs. Postgres queries and roughly define how we will make the functional and performance comparisons.

These changes are targeting a new feature branch that will NOT be merged to main at this time as it's really only for testing. If we decide later to go through with the ElasticSearch transition, we will need to take the queries written in a future ticket out of this branch.

Note that the actual Postgres queries are placeholders. They work but they don't provide all of the features ES does (e.g. pagination). The actual queries will be written and tested in the next ticket.


packages/api/tests/endpoints/collections/list-collections.js
```diff
@@ -124,3 +152,40 @@ test.serial('returns list of collections with stats when requested', async (t) =
   t.is(results[0].name, t.context.testCollection.name);
   stub.restore();
 });
+
+// Development only, do not merge
+// Test setup creates collections before EACH test so running these
+// in serial will have different results than running in isolation
```
npauzenga, line 158: Definitely fixable but I didn't want to take the time to refactor the test as this work is only for research purposes and won't be merged.

jennyhliu:
- Do you have a list of queries we have to evaluate?
- For the performance comparison, we probably can't trust the metrics from unit tests, we will want to test against real servers.

npauzenga:
> * Do you have a list of queries we have to evaluate?
> * For the performance comparison, we probably can't trust the metrics from unit tests, we will want to test against real servers.

- Not yet. The PG queries to replicate the existing ES queries will be written in the next ticket: https://bugs.earthdata.nasa.gov/browse/CUMULUS-3058. This ticket is intended to set up some basic tests, configure a couple endpoints to use PG queries, and define how we're going to measure performance + compare results.
- Yup, absolutely. The units here are really just intended to help develop the behavior and write the queries. The section here on performance comparisons outlines what I was thinking: https://wiki.earthdata.nasa.gov/display/CUMULUS/Eliminate+ElasticSearch+-+Initial+Research. Basically a manual approach, at least to start, looking at the Lambda execution time as a baseline metric.

jennyhliu (approved):

PR #3149, opened 2022-11-09
CUMULUS-2797: Remove rules model

botanical:
**Summary:** 

Addresses [CUMULUS-2797: Remove API rules model](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2797)

Removes the API Rules model and associated code.

## Changes

* Removes Rules model
* Removes Rule schema
* Creates rulesHelper for Rule methods
* Updates tests and utilities to use rulesHelper

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


botanical:
Closing this since I unassigned myself this ticket. Feel free to reference the work done so far to whoever picks this work up

npauzenga:
Re-opening this PR. The work here is excellent. No need to replicate on another branch.

packages/api/tests/endpoints/test-rules.js
```diff
@@ -1626,6 +1607,29 @@ test('DELETE deletes a rule', async (t) => {
...
+  const response = await request(app)
+    .delete(`/rules/${originalPgRecord.name}?onlyResources=true`)
+    .set('Accept', 'application/json')
+    .set('Authorization', `Bearer ${jwtAuthToken}`)
+    .expect(200);
+
+  const { message } = response.body;
+
+  t.true(await t.context.rulePgModel.exists(t.context.testKnex, { name: originalPgRecord.name }));
+  t.is(message, 'Record resources (e.g. CloudWatch Events, Kinesis Event Sources) deleted');
```
npauzenga, line 1630: This doesn't actually test that the rule's resources are deleted. My thinking is that this is an endpoint test so verifying that we hit the correct logic block seems appropriate and `deleteRuleResources` is unit tested elsewhere.

packages/api/models/rules.js
```diff
@@ -1,640 +0,0 @@
...
-const CloudwatchEvents = require('@cumulus/aws-client/CloudwatchEvents');
-const log = require('@cumulus/common/log');
-const s3Utils = require('@cumulus/aws-client/S3');
-const workflows = require('@cumulus/common/workflows');
-const { invoke } = require('@cumulus/aws-client/Lambda');
-const SQS = require('@cumulus/aws-client/SQS');
-const { ValidationError } = require('@cumulus/errors');
-
-const Manager = require('./base');
-const { rule: ruleSchema } = require('./schemas');
```
marchuffnagle, line 18: Can this schema be deleted from the `schemas.js` file?

packages/api/lib/rulesHelpers.js
```diff
@@ -104,6 +104,44 @@ function lookupCollectionInEvent(eventObject) {
   });
 }
 
+/*
```
marchuffnagle, line 107:
```suggestion
/**
```

This will allow the TypeScript compiler (and VS Code when viewing files) to use the JSDocs to infer types.

packages/api/lib/rulesHelpers.js
```diff
@@ -104,6 +104,44 @@ function lookupCollectionInEvent(eventObject) {
...
+ * Build payload from rule for lambda invocation
+ *
+ * @param {ApiRule} rule              - API rule
```
marchuffnagle, line 110: Where is this type defined?

packages/api/lib/rulesHelpers.js
```diff
@@ -104,6 +104,44 @@ function lookupCollectionInEvent(eventObject) {
...
+ * @param {object} [cumulusMeta]      - Optional cumulus_meta object
+ * @param {string} [asyncOperationId] - Optional ID for asynchronous operation
+ *
+ * @returns {Object} lambda invocation payload
+ */
+async function buildPayload(rule, cumulusMeta, asyncOperationId) {
+  // makes sure the workflow exists
+  const bucket = process.env.system_bucket;
```
marchuffnagle, line 118: We need to be testing to make sure that these two environment variables are set and doing something if they are. Probably throw an exception?

packages/api/endpoints/rules.js
```diff
@@ -192,6 +192,7 @@ async function del(req, res) {
   } = req.testContext || {};
 
   const name = (req.params.name || '').replace(/%20/g, ' ');
+  const onlyResources = req.query && req.query.onlyResources === 'true';
```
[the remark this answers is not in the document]
npauzenga: This about this. Is exposing this abnormally dangerous?

packages/api/lib/rulesHelpers.js
npauzenga: It's not. All of the types in this file point to ApiRule but should be the API's actual Rule Type. 
https://github.com/nasa/cumulus/pull/3149/commits/d253a57e275ecd9babd588a95a139a3f195e7f44

packages/api/lib/rulesHelpers.js
npauzenga: Hmm I do agree that we should be defensive but I'm not sure you could do much of anything if system_bucket or stackName weren't set. They're required for deploying pretty much all of our modules. Validating that they exist is probably a good idea as I'm not sure where it fails if you don't provide them but I'm not sure that this is the right spot. It feels pretty late in the process if we're gonna catch it. I guess we could catch it here because it's easy but it feels a little random? Just me?

Also, heads up that this function isn't new. It was just moved in the file. We should still obvious make it better.

packages/api/lib/rulesHelpers.js
```diff
@@ -104,11 +104,49 @@ function lookupCollectionInEvent(eventObject) {
...
+/**
+ * Build payload from rule for lambda invocation
+ *
+ * @param {RuleRecord} rule           - API rule
+ * @param {Object} [cumulusMeta]      - Optional cumulus_meta object
+ * @param {string} [asyncOperationId] - Optional ID for asynchronous operation
+ *
+ * @returns {Object} lambda invocation payload
+ */
+async function buildPayload(rule, cumulusMeta) {
```
npauzenga, line 116: This was confusing. At some point we made this `buildPayload` function in rulesHelpers but we ALSO had a `buildPayload` in the Rule model. It looks like we switched everything over to use the helper instead of the model but the helper didn't set the `asyncOperationId` (line 141) correctly. It's updated here and this is the only rule-related `buildPayload` now.

packages/api/lib/rulesHelpers.js
```diff
@@ -508,48 +546,10 @@ function recordIsValid(rule) {
...
-    template,
-    definition,
-    provider: rule.provider,
-    collection: rule.collection,
-    meta: get(rule, 'meta', {}),
-    cumulus_meta: cumulusMeta || get(rule, 'cumulus_meta', {}),
-    payload: get(rule, 'payload', {}),
-    queueUrl: rule.queueUrl,
-    executionNamePrefix: rule.executionNamePrefix,
-    asyncOperationId: asyncOperationId,
```
npauzenga, line 545: This was a problem. It should be rule.asyncOperationId as most calls don't pass in an explicit asyncOperationId.

packages/api/models/rules.js
npauzenga: Sure. What could go wrong?

https://github.com/nasa/cumulus/pull/3149/commits/e85d67e3d59650e810485c08d2bc9eff21d44fc1

packages/api/lib/rulesHelpers.js
npauzenga: It worked previously because we were mostly using Rule.buildPayload, not this one. These changes use exclusively this version.

packages/integration-tests/index.js
```diff
@@ -564,6 +562,42 @@ async function deleteRules(stackName, bucketName, rules, postfix) {
...
+/**
+ * Delete a rule's Kinesis Event Source Mappings
+ *
+ * @param {Object} rule - a Rule record as returned by the Rules API
+ * @returns {Promise<number>} - Number of rules deleted
+ */
+async function deleteRuleResources(rule) {
+  const kinesisSourceEvents = [
+    {
+      name: process.env.messageConsumer,
```
marchuffnagle, line 574: Let's use `getRequiredEnvVar` here as well.

packages/integration-tests/lambda.js
```diff
@@ -2,10 +2,26 @@
 
 const { lambda } = require('@cumulus/aws-client/services');
 
+/**
+ * Retrieve a rule's Kinesis Event Source Mappings
+ *
+ * @param {Object} uuid - unique identifier for a rule
```
marchuffnagle, line 8: Should this be a string?

packages/integration-tests/lambda.js
```diff
@@ -2,10 +2,26 @@
...
+ */
 async function getEventSourceMapping(uuid) {
   return await lambda().getEventSourceMapping({ UUID: uuid }).promise();
 }
 
+/**
+ * Delete a rule's Kinesis Event Source Mappings
+ *
+ * @param {Object} uuid - unique identifier for a rule
+ * @returns {Promise<void>}
```
marchuffnagle, line 19: Is this actually returning `void`?

marchuffnagle (changes requested):

packages/api/lib/rulesHelpers.js
```diff
@@ -673,9 +672,26 @@ async function createRuleTrigger(ruleItem) {
...
 }
 
+async function deleteOldEventSourceMappings(item) {
+  switch (item.rule.type) {
+  case 'kinesis':
+    await deleteKinesisEventSources(item);
+    break;
+  case 'sns': {
+    if (item.rule.arn) {
+      await deleteSnsTrigger(item);
```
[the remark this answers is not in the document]
npauzenga: So interestingly, this function isn't actually used anywhere. We mock in in a couple tests but no one is calling it so this problem never came up. It was (correctly) moved off of the Rule model but I don't see a reason for it to exist here.

https://github.com/nasa/cumulus/pull/3149/commits/96e3e5ba58f0f9fdf8e6074f5512fef11fd7875d

packages/api/lib/rulesHelpers.js
```diff
@@ -673,9 +672,26 @@ async function createRuleTrigger(ruleItem) {
   return newRuleItem;
 }
 
+async function deleteOldEventSourceMappings(item) {
+  switch (item.rule.type) {
+  case 'kinesis':
+    await deleteKinesisEventSources(item);
```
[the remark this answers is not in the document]
npauzenga: Same as above. I think we can just scrap this function.

https://github.com/nasa/cumulus/pull/3149/commits/96e3e5ba58f0f9fdf8e6074f5512fef11fd7875d

packages/api/lib/rulesHelpers.js
```diff
@@ -104,11 +104,48 @@ function lookupCollectionInEvent(eventObject) {
   });
 }
 
+/**
+ * Build payload from rule for lambda invocation
+ *
+ * @param {RuleRecord} rule           - API rule
```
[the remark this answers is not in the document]
npauzenga: Mission accomplished: https://github.com/nasa/cumulus/pull/3149/commits/f901007e0cf6d88ef39754f60b62ae4489d44a7e

We should discuss best practices here. It's probably worth adding an acceptance criteria on all these tickets to import/create the correct types if you're in a JS file. Otherwise it'll be a little scattered depending on who does the ticket or reviews if it actually gets done. I'll make a note for the next arch meeting.

packages/api/lib/rulesHelpers.js
```diff
@@ -104,11 +104,48 @@ function lookupCollectionInEvent(eventObject) {
...
+    template,
+    definition,
+    provider: rule.provider,
+    collection: rule.collection,
+    meta: get(rule, 'meta', {}),
+    cumulus_meta: cumulusMeta || get(rule, 'cumulus_meta', {}),
+    payload: get(rule, 'payload', {}),
+    queueUrl: rule.queueUrl,
+    executionNamePrefix: rule.executionNamePrefix,
+    asyncOperationId: rule.asyncOperationId,
```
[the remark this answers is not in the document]
npauzenga: https://github.com/nasa/cumulus/pull/3149/commits/66e85d95984c2f6c03d2abf5776f63468b6561ac

packages/api/lib/rulesHelpers.js
```diff
@@ -104,11 +104,48 @@ function lookupCollectionInEvent(eventObject) {
...
+async function buildPayload(rule, cumulusMeta) {
+  // makes sure the workflow exists
+  const bucket = process.env.system_bucket;
+  const stack = process.env.stackName;
+  const workflowFileKey = workflows.getWorkflowFileKey(stack, rule.workflow);
+
+  const exists = await s3Utils.fileExists(bucket, workflowFileKey);
+  if (!exists) throw new Error(`Workflow doesn\'t exist: s3://${bucket}/${workflowFileKey} for ${rule.name}`);
+
+  const definition = await s3Utils.getJsonS3Object(
```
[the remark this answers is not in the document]
npauzenga: Yeah. We need an effort for TypeScripting generally. I'll look and see if we have any tickets created and add this note if we do.

packages/api/lib/rulesHelpers.js
```diff
@@ -558,13 +557,13 @@ async function invokeRerun(rule) {
...
-/*
+/**
  * Updates rule trigger for rule
  *
- * @param {ApiRule}        original- Rule to update trigger for
- * @param {ApiRule}        updates - Updates for rule trigger
- * @param {Knex}           knex    - Knex DB Client
- * @returns {ApiRule}              - Returns new rule object
+ * @param {RuleRecord} original - Rule to update trigger for
+ * @param {Object} updates      - Updates for rule trigger
```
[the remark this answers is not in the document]
npauzenga: Oh interesting. So defining fields on an ad-hoc object if we're not going to create a type. This would be good to include in the instructions for devs when we make typing a more formal requirement on tickets.

https://github.com/nasa/cumulus/pull/3149/commits/d3e0a5ffc896d1e7f34a18e3fb75c033c9348e87

packages/integration-tests/index.js
```diff
@@ -564,6 +562,42 @@ async function deleteRules(stackName, bucketName, rules, postfix) {
...
+      eventType: 'arn',
+      type: {
+        arn: rule.rule.arn,
+      },
+    },
+    {
+      name: process.env.KinesisInboundEventLogger,
```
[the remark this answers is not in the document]
npauzenga: https://github.com/nasa/cumulus/pull/3149/commits/dc9917d037b7c80df99865746e940185417c9dc9

packages/integration-tests/index.js
```diff
@@ -564,6 +562,42 @@ async function deleteRules(stackName, bucketName, rules, postfix) {
   return rules.length;
 }
 
+/**
+ * Delete a rule's Kinesis Event Source Mappings
+ *
+ * @param {Object} rule - a Rule record as returned by the Rules API
+ * @returns {Promise<number>} - Number of rules deleted
```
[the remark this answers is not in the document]
npauzenga: https://github.com/nasa/cumulus/pull/3149/commits/437f01d3a9c18bf0a6f746d3791f8a3267703d8b

I think this is right. Part of the conversation here should be around how to handle external libraries and their return types.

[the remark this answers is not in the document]
npauzenga: Or should this be `Promise<unknown>`? 🤔

packages/integration-tests/lambda.js
npauzenga: Yup. This was re-written and not updated:
https://github.com/nasa/cumulus/pull/3149/commits/4df26b664bfa0933ca41ddc604bf6a12222af1de

packages/integration-tests/lambda.js
```diff
@@ -2,10 +2,26 @@
 
 const { lambda } = require('@cumulus/aws-client/services');
 
+/**
+ * Retrieve a rule's Kinesis Event Source Mappings
+ *
+ * @param {Object} uuid - unique identifier for a rule
+ * @returns {Promise<Object>} - details about an Event Source Mapping
```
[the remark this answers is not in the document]
npauzenga: So is `unknown` preferable when dealing with external library returns?

https://github.com/nasa/cumulus/pull/3149/commits/4df26b664bfa0933ca41ddc604bf6a12222af1de

packages/integration-tests/lambda.js
```diff
@@ -2,10 +2,26 @@
...
+ */
 async function getEventSourceMapping(uuid) {
   return await lambda().getEventSourceMapping({ UUID: uuid }).promise();
 }
 
+/**
+ * Delete a rule's Kinesis Event Source Mappings
+ *
+ * @param {Object} uuid - unique identifier for a rule
```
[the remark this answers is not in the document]
npauzenga: https://github.com/nasa/cumulus/pull/3149/commits/4df26b664bfa0933ca41ddc604bf6a12222af1de

packages/integration-tests/lambda.js
npauzenga: That's what the return of `deleteEventSourceMapping` claims but we're `promise()`ing it so I suppose it should either match the AWS return type:

`Promise<PromiseResult<AWS.Lambda.EventSourceMappingConfiguration, AWS.AWSError>>`

or `unknown` as above because we don't have all of those types imported? What do you think?

npauzenga: Updated here but open to discussion: https://github.com/nasa/cumulus/pull/3149/commits/4df26b664bfa0933ca41ddc604bf6a12222af1de

packages/integration-tests/index.js
npauzenga: https://github.com/nasa/cumulus/pull/3149/commits/dc9917d037b7c80df99865746e940185417c9dc9

packages/integration-tests/index.js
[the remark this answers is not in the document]
marchuffnagle: When in doubt, `Promise<unknown>`

packages/api/lib/rulesHelpers.js
```diff
@@ -14,15 +14,21 @@ const { sqsQueueExists } = require('@cumulus/aws-client/SQS');
 const { invoke } = require('@cumulus/aws-client/Lambda');
 const { RulePgModel } = require('@cumulus/db');
 const { ValidationError } = require('@cumulus/errors');
+const { getRequiredEnvVar } = require('@cumulus/common/env.js');
```
marchuffnagle, line 17:
```suggestion
const { getRequiredEnvVar } = require('@cumulus/common/env');
```

packages/api/lib/rulesHelpers.js
```diff
@@ -14,15 +14,21 @@ const { sqsQueueExists } = require('@cumulus/aws-client/SQS');
...
 const { listRules } = require('@cumulus/api-client/rules');
 const { removeNilProperties } = require('@cumulus/common/util');
 
 const { handleScheduleEvent } = require('../lambdas/sf-scheduler');
 const { isResourceNotFoundException, ResourceNotFoundError } = require('./errors');
-const Rule = require('../models/rules');
 
 const log = new Logger({ sender: '@cumulus/rulesHelpers' });
+
+/**
```
marchuffnagle, line 27: This should be moved up above the `const log ...` statement.

packages/integration-tests/index.js
```diff
@@ -564,6 +563,46 @@ async function deleteRules(stackName, bucketName, rules, postfix) {
...
 
+/**
+ * Delete a rule's Kinesis Event Source Mappings
+ *
+ * @param {Object} rule - a Rule record as returned by the Rules api
+ * @param {string} rule.name
+ * @param {Object} rule.rule
+ * @param {string} rule.rule.arn
+ * @param {string} rule.rule.logEventArn
+ * @returns {Promise<Object>[]} - Event Source Map deletion results
```
marchuffnagle, line 574: 1. Subtle difference, this is a Promise that resolves to an array of things, as opposed to an array of Promises that each resolve to a single value.
2. Let's use `unknown` here instead of `Object`.

```suggestion
 * @returns {Promise<unknown[]>} - Event Source Map deletion results
```

packages/api/lib/rulesHelpers.js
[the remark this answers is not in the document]
marchuffnagle: That's not a TS thing, that's a JSDoc thing: https://jsdoc.app/tags-param.html#parameters-with-properties

marchuffnagle (changes requested):

packages/api/lib/rulesHelpers.js
npauzenga: Ah yup.

https://github.com/nasa/cumulus/pull/3149/commits/ebee8226b6ce6305610ea5e568f4de77d56d69ed

packages/integration-tests/index.js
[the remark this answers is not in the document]
npauzenga: Ok, gotchya. Or when we don't control the type? If we're returning _directly_ from an external library?

marchuffnagle (approved):
Ship it.

packages/api/lib/rulesHelpers.js
```diff
@@ -558,13 +563,16 @@ async function invokeRerun(rule) {
...
- * @returns {ApiRule}              - Returns new rule object
+ * @param {RuleRecord} original - Rule to update trigger for
+ * @param {Object} updates      - Updates for rule trigger
+ * @param {string} updates.state
+*  @param {Object} updates.rule
+ * @param {unknown} updates.rule.value
+ * @param {Knex} knex           - Knex DB Client
+ * @returns {Promise<RuleRecord>}        - Returns new rule object
  */
 async function updateRuleTrigger(original, updates, knex) {
```
npauzenga, line 577: This was tricky. One of the main complications in the Rules work is that a lot of methods in the deprecated Rule model were _already_ duplicated here in rulesHelpers, prior to this PR. This PR switches everything over to use the helpers and fully removes the model but there were several problems there. The methods are duplicated but different in subtle ways. These helpers needed to be reconciled with the Rule methods while not breaking existing calls to the helper methods.

packages/api/tests/lib/test-sns-rules.js
```diff
@@ -64,82 +79,21 @@ test.afterEach.always(async (t) => {
...
-  t.is(updatedRule.name, rule.name);
-  t.is(updatedRule.state, 'DISABLED');
-  t.is(updatedRule.rule.type, rule.rule.type);
-  t.is(updatedRule.rule.value, rule.rule.value);
-  t.falsy(updatedRule.rule.arn);
-
-  t.teardown(() => rulesModel.delete(rule));
 });
 
-test.serial('enabling a disabled SNS rule and passing rule.arn throws specific error', async (t) => {
```
npauzenga, line 105: A lot of these tests were duplicates of existing tests in test-rulesHelpers.js

packages/api/lib/rulesHelpers.js
```diff
@@ -430,7 +473,7 @@ async function addSnsTrigger(item) {
       FunctionName: process.env.messageConsumer,
       Principal: 'sns.amazonaws.com',
       SourceArn: item.rule.value,
-      StatementId: `${item.name}Permission`,
+      StatementId: getSnsTriggerPermissionId(item),
```
npauzenga, line 476: This could use another set of eyes. Originally this helper used `${item.name}Permission` here. This function is an existing (not new to this PR) duplicate of a method on the Rule model, `Rule.addSnsTrigger`. That method used `getSnsTriggerPermissionId`. 

I'm switching it here and updating the tests for this helper + integration tests for Rules(!). https://github.com/nasa/cumulus/pull/3149/files#diff-86dca067c5034695cd77c20c8de88c752c0f2b906583e669f203ae3f940f0dcaR119

I'm thinking that, so long as it's consistent, either approach works. The issue is that it wasn't consistent. Definitely need that validated though.

packages/api/lib/rulesHelpers.js
```diff
@@ -594,10 +602,8 @@ async function updateRuleTrigger(original, updates, knex) {
       if (enabled && stateChanged && mergedRule.rule.arn) {
         throw new Error('Including rule.arn is not allowed when enabling a disabled rule');
       }
+
       let snsSubscriptionArn;
-      if (mergedRule.rule.arn) {
-        await deleteSnsTrigger(knex, mergedRule);
-      }
```
npauzenga, line 600: TODO: Not sure about this. This call to `deleteSnsTrigger` does NOT exist on the `Rule.updateRuleTrigger` method but does here.

packages/integration-tests/index.js
```diff
@@ -564,6 +563,46 @@ async function deleteRules(stackName, bucketName, rules, postfix) {
...
+ * Delete a rule's Kinesis Event Source Mappings
+ *
+ * @param {Object} rule - a Rule record as returned by the Rules api
+ * @param {string} rule.name
+ * @param {Object} rule.rule
+ * @param {string} rule.rule.arn
+ * @param {string} rule.rule.logEventArn
+ * @returns {Promise<unknown[]>} - Event Source Map deletion results
+ */
+async function deleteRuleResources(rule) {
```
marchuffnagle, line 576: Does this do about the same thing?

```js
async function deleteRuleResources(rule) {
  const deleteSourceEventMapping = (id) =>
    lambda.deleteEventSourceMapping(id)
      .catch((error) => {
        console.log(`Error deleting eventSourceMapping for ${rule.name}: ${error}`);
        if (error.code !== 'ResourceNotFoundException') throw error;
      });

  await Promise.all([
    deleteSourceEventMapping(rule.rule.arn),
    deleteSourceEventMapping(rule.rule.logEventArn),
  ]);
}
```

packages/integration-tests/index.js
```diff
@@ -564,6 +563,46 @@ async function deleteRules(stackName, bucketName, rules, postfix) {
...
+  const kinesisSourceEvents = [
+    {
+      name: getRequiredEnvVar('messageConsumer'),
+      eventType: 'arn',
+      type: {
+        arn: rule.rule.arn,
```
marchuffnagle, line 582: :can-of-worms: Is this an ARN or a UUID? The function that's consuming this value seems to think it's a UUID, and the AWS docs seem to agree.

packages/api/lib/rulesHelpers.js
npauzenga: Another, more compelling argument for using `getSnsTriggerPermissionId` is that's what is currently used in main:

https://github.com/nasa/cumulus/blob/master/packages/api/models/rules.js#L491

The rds-phase-3 branch is incorrect in having it set to `${item.name}Permission`. If we change the naming scheme users won't be able to update or remove existing permissions.

npauzenga: Discussed in a huddle. The consensus is that `getSnsTriggerPermissionId` is correct here and that the feature/rds-phase-3 branch is incorrect.

npauzenga (approved):
Leaving the approving review as the previous changes were approved. The updates here are from _another_ approved PR

https://github.com/nasa/cumulus/pull/3285

PR #3161, opened 2022-11-15
[CUMULUS-3116] Revert ElasticSearch Default Sort

npauzenga:
In working with GHRC, we found this to be the source of their CPU spikes and timeouts. Originally our default sort behavior was:
```
"sort": [
    { "timestamp": "desc" }
  ]
```

And it was updated [here](https://github.com/nasa/cumulus/pull/2993/files#diff-8b4a350e2afa9cf3d9dc8a49c0ed778a8d1eccf4d8cade8324817a24aab5cbd4) to:
```
"sort": [
    { "timestamp": "desc" }, { "_uid": "asc" }
  ]
```

Reverting the change (removing the _uid "tie-breaker") resolved the CPU usage issues for GHRC. The current best theory is that the _uid field isn't an efficient field to sort on. Its type isn't clear (it is a dynamically mapped field, not something we specify) but if it's not a numeric or keyword type, sorting could be less performant. https://www.elastic.co/guide/en/elasticsearch/reference/5.6/mapping-uid-field.html

It is also possible that _any_ sort more complex than a single field may impact performance at scale (100s of millions of records). In any case, it was discussed with the DAACs and there isn't a need to sort this thoroughly. If we revert the change and leave the results sorted by `{ "timestamp": "desc" }` alone, that should suffice for nearly all records. In the rare case that a `timestamp` conflicts, we cannot guarantee the order for those records.


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Fixed
+- **CUMULUS-3116**
+  - Reverted the default ElasticSearch sorting behavior to the pre-13.3.0 configuration
+  - Results from ElasticSearch are sorted by default by the `timestamp` field. This means that the order
+  is not guaranteed if two or more records have identical timestamps as there is no secondary sort/tie-breaker.
```
botanical, line 13: I wonder if this sort of note is helpful to add to the es-client package readme or something 🤔

botanical (approved):
This looks correct to me, had one comment but it's non-blocking

CHANGELOG.md
npauzenga: Sure, I think that makes sense to call out the default behavior: https://github.com/nasa/cumulus/pull/3161/commits/d8a662bbf6ed0a69f821c279b40852bd8f72fc95

Also, if users don't like it or it doesn't do what they expect, they can always update their API calls to sort differently.

PR #3176, opened 2022-11-21
CUMULUS-3097: Replacing Echo-Token with EDL-token for CMR

Nnaga1:

## Changes

* ...

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/audit-ci.json
+++ b/audit-ci.json
@@ -2,5 +2,5 @@
     "high": true,
     "pass-enoaudit": true,
     "retry-count": 20,
-    "allowlist": ["xmldom"]
+    "allowlist": ["xmldom", "knex", "json5"]
 }


packages/cmr-client/src/CMR.ts
```diff
@@ -37,45 +29,91 @@ const userIpAddress = (): Promise<string> =>
  * @private
  */
```
jennyhliu, line 30: CMR.updateToken needs to be refactored.  Suggest to add a new class file (e.g. EarthdataLog.ts) to have method getToken, createToken, revokeToken.

jennyhliu: suggest to move type definitiond outside of the method, and have a meaningful name.
see example, https://github.com/nasa/cumulus/blob/master/packages/oauth-client/src/OAuthClient.ts#L11

packages/cmr-client/src/CMR.ts
```diff
@@ -228,9 +266,8 @@ export class CMR {
 
     if (params.token) {
       if (this.oauthProvider === 'launchpad') headers.Authorization = params.token;
-      else headers['Echo-Token'] = params.token;
+      else headers['Edl-Token'] = params.token;
```
jennyhliu, line 269: Have you tested this?  Based on the [CMR doc](https://cmr.earthdata.nasa.gov/ingest/site/docs/ingest/api.html), bearer token is used for EDL token

jennyhliu (changes requested):
1. add integration test for creating and retrieving EDL token
2. set cmr_oauth_provider = 'earthdata' in your terraform.tfvars and https://github.com/nasa/cumulus/blob/master/example/deployments/cumulus/nnaga-ci-tf.tfvars, so EDL token is used instead of launchpad token for CMR requests, and verify that all integration tests pass

packages/cmr-client/src/CMR.ts
```diff
@@ -37,45 +29,91 @@ const userIpAddress = (): Promise<string> =>
...
+      const responseError = get(error, 'response.body.errors');
+      if (responseError) {
+        errorMessage = `${errorMessage}, CMR error message: ${JSON.stringify(responseError)}`;
+      }
+      log.error(errorMessage);
+      throw new Error(errorMessage);
+    }
+    return response2.access_token;
+  }
+  return response[0].access_token;
```
jennyhliu, line 116: only unexpired token should be retrieved

jennyhliu: It's easy to check if the token is still valid to avoid edge case, since it's in the return response.

packages/cmr-client/src/CMR.ts
jennyhliu: If you write integration test, you want to delete token you created during test.

packages/cmr-client/src/EarthdataToken.ts
```diff
@@ -0,0 +1,124 @@
...
+    cmrenv = '';
+  } else if (process.env.CMR_ENVIRONMENT === 'UAT' || process.env.CMR_ENVIRONMENT === 'SIT') {
+    cmrenv = process.env.CMR_ENVIRONMENT + '.';
+  } else {
+    throw new TypeError(`Invalid CMR environment: ${process.env.CMR_ENVIRONMENT}`);
+  }
+  const returnarray: Array<string> = [buff, cmrenv];
+  return returnarray;
+}
+
```
jennyhliu, line 30: 1. You can get encoded credential with this:
packages/integration-tests/api/EarthdataLogin.js
const auth = base64.encode(`${username}:${password}`);
2. Instead of checking env, we can pass in `edlEnv` variable (possible values should be PROD, OPS etc.), this way, the methods are more generic.
3. Create a method to return EDL url with env, default to SIT env.
4. It's better to create a EarthdataToken class with attribute username, password, edlEnv.  
packages/oauth-client/src can give you some ideas how to refactor the code.

packages/cmr-client/src/EarthdataToken.ts
```diff
@@ -0,0 +1,124 @@
...
+  let response: EarthdataPostTokenResponse;
+  try {
+    response = await got.post(`https://${data[1]}urs.earthdata.nasa.gov/api/users/token`,
+      {
+        headers: {
+          Authorization: `Basic ${data[0]}`,
+        },
+      }).json();
+  } catch (error) {
+    logDetails.credentials = username + ':' + password;
```
jennyhliu, line 79: NEVER log the password

jennyhliu (changes requested):

packages/cmr-client/src/EarthdataToken.ts
```diff
@@ -0,0 +1,124 @@
+// @ts-nocheck
+import get from 'lodash/get';
+import got from 'got';
+import Logger from '@cumulus/logger';
+
+import { EarthdataGetTokenResponse, EarthdataPostTokenResponse } from './types';
```
jennyhliu, line 6: If these types are only used here, we should move them to this file.

packages/cmr-client/src/EarthdataToken.ts
```diff
@@ -0,0 +1,124 @@
...
+  try {
+    response = await got.post(`https://${data[1]}urs.earthdata.nasa.gov/api/users/token`,
+      {
+        headers: {
+          Authorization: `Basic ${data[0]}`,
+        },
+      }).json();
+  } catch (error) {
+    logDetails.credentials = username + ':' + password;
+    log.error(error, logDetails);
```
jennyhliu, line 80: We don't want to log logDetails with username password

packages/cmr-client/src/EarthdataToken.ts
jennyhliu: 2. Instead of checking env, we can pass in `edlEnv` variable (possible values should be PROD, OPS etc.), this way, the methods are more generic.

jennyhliu: 3. Create a method to return EDL url given edl env, default to SIT env.

jennyhliu: 4. It's better to create a EarthdataLogin class with attribute username, password, edlEnv.  
packages/oauth-client/src can give you some ideas how to refactor the code.

packages/cmr-client/src/EarthdataToken.ts
```diff
@@ -0,0 +1,124 @@
...
+  // response: get a token from the Earthdata login endpoint using credentials if exists
+  let response: EarthdataGetTokenResponse;
+  try {
+    response = await got.get(`https://${data[1]}urs.earthdata.nasa.gov/api/users/tokens`,
+      {
+        headers: {
+          Authorization: `Basic ${data[0]}`,
+        },
+      }).json();
+  } catch (error) {
```
jennyhliu, line 45: The error handling code for all three methods can be refactored

packages/cmr-client/src/EarthdataToken.ts
```diff
@@ -0,0 +1,124 @@
...
+      errorMessage = `${errorMessage}, CMR error message: ${JSON.stringify(responseError)}`;
+    }
+
+    log.error(errorMessage);
+    throw new Error(errorMessage);
+  }
+  if (Object.keys(response).length === 0) {
+    return '';
+  }
+  return response[0].access_token;
```
jennyhliu, line 62: Suggest to get `unexpired` token or return undefined

packages/cmr-client/src/EarthdataToken.ts
```diff
@@ -0,0 +1,183 @@
...
+        const statusMessage = get(error, 'response.statusMessage', error.message);
+        let errorMessage = `Authentication error: Invalid Credentials, Authentication with Earthdata Login failed, statusCode: ${statusCode}, statusMessage: ${statusMessage}`;
+        const responseError = get(error, 'response.body.errors');
+        if (responseError) {
+          errorMessage = `${errorMessage}, CMR error message: ${JSON.stringify(responseError)}`;
+        }
+
+        throw new Error(errorMessage);
+      }
+      if (Object.keys(response).length === 0) {
```
jennyhliu, line 114: We should loop through the response and get the first unexpired token and return.  We can't assume how many tokens (2?) are there.

packages/cmr-client/src/EarthdataToken.ts
```diff
@@ -0,0 +1,183 @@
...
+        return this.createEDLToken();
+      }
+      if (Object.keys(response).length === 2) {
+        const date1 = new Date(response[0].expiration_date);
+        const date2 = new Date(response[1].expiration_date);
+        return date1 >= date2 ? 'Bearer' + response[1].access_token : 'Bearer' + response[0].access_token;
+      }
+      return 'Bearer ' + response[0].access_token;
+    }
+    return 'Bearer ' + this.token;
```
jennyhliu, line 124: `Bearer` part is specific for request header. The method should return token only.

packages/cmr-client/src/EarthdataToken.ts
```diff
@@ -0,0 +1,183 @@
...
+  getEDLurl(
+  ) {
+    switch (this.edlEnv) {
+      case 'PROD':
+      case 'OPS':
+        return 'https://urs.earthdata.nasa.gov';
+      case 'UAT':
+        return 'https://uat.urs.earthdata.nasa.gov';
+      case 'SIT':
+        return 'https://sit.urs.earthdata.nasa.gov';
```
jennyhliu, line 84:
```suggestion
     
```

packages/cmr-client/src/EarthdataToken.ts
```diff
@@ -0,0 +1,183 @@
...
+      let errorMessage = `Authentication error: Invalid Credentials, Authentication with Earthdata Login failed, statusCode: ${statusCode}, statusMessage: ${statusMessage}`;
+      const responseError = get(error, 'response.body.errors');
+      if (responseError) {
+        errorMessage = `${errorMessage}, CMR error message: ${JSON.stringify(responseError)}`;
+      }
+
+      throw new Error(errorMessage);
+    }
+    this.token = response.access_token;
+    return 'Bearer ' + response.access_token;
```
jennyhliu, line 150: same as above, return token only.

packages/cmr-client/src/EarthdataToken.ts
```diff
@@ -0,0 +1,183 @@
...
+ *  "password": "my-password",
+ *  "edlEnv": "my-cmr-environment",
+ *  "token" : "my-token"
+ * });
+ */
+export class EarthdataToken {
+  username: string;
+  password: string;
+  edlEnv: string;
+  token?: string;
```
jennyhliu, line 47: Is there use case you need to pass in the token? If not, remove from attribute list.

packages/cmr-client/src/CMR.ts
```diff
@@ -1,81 +1,42 @@
...
-
-  if (!response.body.token) throw new Error('Authentication with CMR failed');
-
-  return response.body.token.id;
+  const earthdataTokenObject = new EarthdataToken({
+    username: username,
+    password: password,
+    edlEnv: process.env.CMR_ENVIRONMENT,
+  });
+  return await earthdataTokenObject.getEDLToken();
```
jennyhliu, line 39: If earthdataTokenObject.getEDLToken returns undefined, create a new token

packages/cmr-client/src/EarthdataToken.ts
```diff
@@ -0,0 +1,183 @@
...
+      try {
+        response = await got.get(`${url}/api/users/tokens`,
+          {
+            headers: {
+              Authorization: `Basic ${buff}`,
+            },
+          }).json();
+      } catch (error) {
+        const statusCode = get(error, 'response.statusCode', error.code);
+        const statusMessage = get(error, 'response.statusMessage', error.message);
```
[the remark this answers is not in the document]
jennyhliu: EDL response could be different from CMR

packages/cmr-client/src/EarthdataToken.ts
```diff
@@ -0,0 +1,183 @@
...
+        let errorMessage = `Authentication error: Invalid Credentials, Authentication with Earthdata Login failed, statusCode: ${statusCode}, statusMessage: ${statusMessage}`;
+        const responseError = get(error, 'response.body.errors');
+        if (responseError) {
+          errorMessage = `${errorMessage}, CMR error message: ${JSON.stringify(responseError)}`;
+        }
+
+        throw new Error(errorMessage);
+      }
+      if (Object.keys(response).length === 0) {
+        return this.createEDLToken();
```
jennyhliu, line 115: To make code more clear, suggest to have a separate method to getToken.  We can have methods:
getToken (call retrieveToken and createToken)
retrieveToken (retrieve existing valid token if any)
createToken
revokeToken

Remove `EDL` from method name since they are methods in class EarthdataLogin

jennyhliu: Please add doc to class and methods

packages/cmr-client/src/EarthdataToken.ts
```diff
@@ -0,0 +1,183 @@
...
+  password: string;
+  edlEnv: string;
+  token?: string;
+
+  /**
+  * The constructor for the EarthdataToken class
+  *
+  * @param {string} params.username - Earthdata username
+  * @param {string} params.password - Earthdata password
+  * @param {string} params.edlEnv - the CMR environment of the user
```
jennyhliu, line 54: list valid values

packages/cmr-client/src/EarthdataToken.ts
jennyhliu: We want to make the code more clearer and maintainable.  The current getToken method retrieves the token from EDL, and also does this following and makes the logic hard to follow.
if (Object.keys(response).length === 0) {
        return this.createEDLToken();

We can do something for getToken like:
let token = await this.retrieveToken();
if (token === undefined) {
token = await this.createToken();
}

packages/cmr-client/src/EarthdataToken.ts
jennyhliu: ['PROD', 'OPS', 'SIT', 'UAT']

packages/cmr-client/src/EarthdataToken.ts
jennyhliu: We can't assume there are 1 or 2 EDL tokens, it might be 20 tokens.  We can just loop through the list and return the first valid one or undefined.

packages/cmr-client/src/EarthdataToken.ts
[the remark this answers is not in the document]
jennyhliu: yeah. make changes to work with EDL response

packages/cmr-client/src/EarthdataLogin.ts
```diff
@@ -0,0 +1,211 @@
...
+ *
+ * @typicalname earthdataLogin
+ *
+ * @example
+ * const { EarthdataLogin } = require('@cumulus/cmr-client');
+ *
+ * const earthdataLogin = new EarthdataLogin({
+ *  "username": "my-username",
+ *  "password": "my-password",
+ *  "edlEnv": "my-cmr-environment"
```
jennyhliu, line 38:
```suggestion
 *  "edlEnv": "my-edl-environment"
```

jennyhliu: Please update all 'CMR' to 'EDL', they are two different applications

jennyhliu: Add unit tests for EathdataLogin class.

jennyhliu: Add a test file test-EarthdataLogin.js and test each method in the class, tests should cover as much code as possible, e.g.
test('retrieveToken returns a valid token')
test('retrieveToken returns undefined when there is no valid token')
test('retrieveToken throws exception where invalid user credential')

packages/cmr-client/src/EarthdataLogin.ts
```diff
@@ -0,0 +1,202 @@
...
+    // response: get a token from the Earthdata login endpoint using credentials if exists
+    let response: EarthdataGetTokenResponse;
+    try {
+      response = await got.get(`${url}/api/users/tokens`,
+        {
+          headers: {
+            Authorization: `Basic ${buff}`,
+          },
+        }).json();
+    } catch (error) {
```
jennyhliu, line 123: To reduce the code redundancy of these methods, suggest to add a new method to handle the error (line 124-129)

CHANGELOG.md
```diff
@@ -108,6 +108,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- **CUMULUS-3097**
+  - Changed CMR package's token from Echo-Token to Earthdata Login (EDL) token in updateToken method
```
jennyhliu, line 112:
```suggestion
  - Changed `@cumulus/cmr-client` package's token from Echo-Token to Earthdata Login (EDL) token in updateToken method
```

jennyhliu (commented):
Nice work!  I have a couple of minor comments before approval.

packages/cmr-client/src/EarthdataLogin.ts
```diff
@@ -0,0 +1,201 @@
...
+      token = await this.createEDLToken();
+    }
+    return token;
+  }
+
+  /**
+   * This helper method is called in the functions that retrieve, create, and revoke EarthdataLogin
+   * tokens for error-handling. If API call made to the EarthdataLogin endpoint results in an error.
+   * The statuscode, statusmessage, error description, and error message are thrown and outputted.
+   *
```
jennyhliu, line 106: add comment:
@throws {Error} -...

packages/cmr-client/src/EarthdataLogin.ts
```diff
@@ -0,0 +1,201 @@
...
+   */
+  async handleHttpError(error: Error): Promise<void> {
```
jennyhliu, line 108: no need to declare as async

jennyhliu (approved):
Nice job!

PR #3182, opened 2022-12-01
Fix merge issues in main to feature/rds-phase-3

npauzenga:
In merging the latest from main into feature/rds-phase-3 we hit a few unit test failures. This PR fixes those failures and targets the feature branch with main. The intention is to merge this to main-to-feature-1129, then merge main-to-feature-1129 to feature/rds-phase-3 after a separate review.

### DIFF
--- a/packages/api/src/lib/granule-delete.ts
+++ b/packages/api/src/lib/granule-delete.ts
@@ -97,7 +97,12 @@ const deleteGranuleAndFiles = async (params: {
     logger.debug(`Successfully deleted granule ${apiGranule.granuleId} from ES datastore`);
     await deleteS3Files(apiGranule.files);
     logger.debug(`Successfully removed S3 files ${JSON.stringify(apiGranule.files)}`);
-    return;
+    return {
+      collection: apiGranule.collectionId,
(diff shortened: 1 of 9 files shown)


packages/api/src/lib/granule-delete.ts
```diff
@@ -97,7 +97,12 @@ const deleteGranuleAndFiles = async (params: {
     logger.debug(`Successfully deleted granule ${apiGranule.granuleId} from ES datastore`);
     await deleteS3Files(apiGranule.files);
     logger.debug(`Successfully removed S3 files ${JSON.stringify(apiGranule.files)}`);
-    return;
+    return {
+      collection: apiGranule.collectionId,
+      deletedGranuleId: apiGranule.granuleId,
+      deletionTime: Date.now(),
+      deletedFiles: apiGranule.files,
```
npauzenga, line 104: Ah I didn't realize this wasn't in feature at all. That _is_ curious why the test asserting `deletedFiles.length` passed. It should have been undefined.

Note to look into this further.

npauzenga (commented):
Looks good. OK to merge to the main-to-feature branch and re-review as a whole.

PR #3210, opened 2023-01-03
Update main README's shell snippets so that they can be easily copied and pasted

marchuffnagle:
A number of the code snippets had a leading `$`, which meant that you could not copy and paste them from GitHub into a terminal.


Jkovarik (approved):

marchuffnagle:
@Jkovarik Is there something special we have to do to get Bamboo reporting on a PR from a forked repo? Or some other thing we have to do for these forked PRs?

marchuffnagle:
Replaced by https://github.com/nasa/cumulus/pull/3243

PR #3211, opened 2023-01-03
Add python-version file

marchuffnagle:
The example/lambdas/python-processing package requires python 3.7, and bootstrapping will fail if it is not present. Adding a `.python-version` file that specifies Python 3.7, so developers using `pyenv` will have the right version set.


npauzenga (commented):
Nice. I think this makes sense if 3.7 is our baseline now (sounds like it is if you couldn't bootstrap at 3.6). We should probably update the pre-reqs in the README: https://github.com/nasa/cumulus#prerequisites

If we have it pinned to 3.7 like this will it blow up if someone is running a later version? I've understood the docs to say AT LEAST Python version X but maybe we should be explicit and say 3.7 if that's what we need. Not sure if other devs are running 3.8+ so might also be worth a shout to everyone that this might 💥 if they're running another version.

marchuffnagle:
> If we have it pinned to 3.7 like this will it blow up if someone is running a later version? I've understood the docs to say AT LEAST Python version X but maybe we should be explicit and say 3.7 if that's what we need. Not sure if other devs are running 3.8+ so might also be worth a shout to everyone that this might 💥 if they're running another version.

We actually require exactly v3.7: https://github.com/nasa/cumulus/blob/341ff54105633228cc895233758294c4f8031503/example/lambdas/python-processing/Pipfile#L16

I was using 3.12 and it wasn't accepting that.

npauzenga:
> > If we have it pinned to 3.7 like this will it blow up if someone is running a later version? I've understood the docs to say AT LEAST Python version X but maybe we should be explicit and say 3.7 if that's what we need. Not sure if other devs are running 3.8+ so might also be worth a shout to everyone that this might 💥 if they're running another version.
> 
> We actually require exactly v3.7:
> 
> https://github.com/nasa/cumulus/blob/341ff54105633228cc895233758294c4f8031503/example/lambdas/python-processing/Pipfile#L16
> 
> I was using 3.12 and it wasn't accepting that.

Ah ok. Maybe just update the README then to say 3.7. https://github.com/nasa/cumulus#prerequisites

marchuffnagle:
Replaced by https://github.com/nasa/cumulus/pull/3244

PR #3233, opened 2023-01-26
Master updates 012623

npauzenga:
This is _another_ set of changes from main to feature. It includes security vulnerabilities required for CI to pass and is targeting the other feature merge branch `feature-with-master-010623`.


packages/api/endpoints/granules.js
```diff
@@ -439,7 +439,25 @@ async function patchByGranuleId(req, res) {
...
-/* eslint-disable complexity */ //TODO
+/**
+ * Helper to check granule and collection IDs in queryparams
+ * against the payload body.
+ *
+ * @param {Object} body - update body payload
+ * @param {Object} req - express request object
+ * @returns {boolean} true if the body matches the query params
+ */
+function _granulePayloadMatchesQueryParams(body, req) {
```
npauzenga, line 450: This is new. We hit the complexity limit for `patch` so this is a stopgap to reduce it by a couple points. There's more that can be abstracted obviously but I'm just looking for an easy win the context of a merge PR.

PR #3237, opened 2023-01-30
CUMULUS-3181:Fixed sqsMessageRemover to correctly retrieve ENABLE sqs rules

jennyhliu:

Addresses [CUMULUS-3181: Rules in a disabled state triggers workflow](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3181)

## Changes

* Fixed sqsMessageRemover lambda to correctly retrieve ENABLE sqs rules
* The issue was introduced in v11.0.0
* fix test-granule-search-after test.
✖ endpoints › granules › default paginates correctly with search_after
The previous test adds data to es, and affects search_after test. Combines these two tests together.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/packages/api/tests/endpoints/test-rules.js
+++ b/packages/api/tests/endpoints/test-rules.js
@@ -298,7 +298,7 @@ test('CUMULUS-912 DELETE with pathParameters and with an invalid access token re
 
 test.todo('CUMULUS-912 DELETE with pathParameters and with an unauthorized user returns an unauthorized response');
 
-test.serial.only('default returns list of rules', async (t) => {
+test.serial('default returns list of rules', async (t) => {
   const response = await request(app)
     .get('/rules')
     .set('Accept', 'application/json')
--- a/packages/api/tests/endpoints/test-rules.js
+++ b/packages/api/tests/endpoints/test-rules.js
@@ -298,7 +298,7 @@ test('CUMULUS-912 DELETE with pathParameters and with an invalid access token re
 
 test.todo('CUMULUS-912 DELETE with pathParameters and with an unauthorized user returns an unauthorized response');
 
-test.serial('default returns list of rules', async (t) => {
+test.serial.only('default returns list of rules', async (t) => {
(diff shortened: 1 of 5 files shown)


packages/api/tests/endpoints/test-rules.js
```diff
@@ -309,6 +309,26 @@ test.serial('default returns list of rules', async (t) => {
...
+  t.is(results.length, 1);
+
+  const newResponse = await request(app)
+    .get('/rules?page=1&rule.type=sqs&state=ENABLED')
+    .set('Accept', 'application/json')
+    .set('Authorization', `Bearer ${jwtAuthToken}`)
+    .expect(200);
+
+  const { results: newResults } = newResponse.body;
+  t.is(newResults.length, 0);
```
botanical, line 329:
```suggestion
  t.is(newResponse.length, 0);
```

botanical (commented):
I think your updates are solid. I saw the AC of the ticket is `Only ENABLED (sqs) rules will trigger workflow` but I'm unsure how the changes to `packages/api/tests/lambdas/test-sqs-message-consumer.js` test that this is true. Should we update the integration test as well?

jennyhliu:
> I think your updates are solid. I saw the AC of the ticket is `Only ENABLED (sqs) rules will trigger workflow` but I'm unsure how the changes to `packages/api/tests/lambdas/test-sqs-message-consumer.js` test that this is true. Should we update the integration test as well?

sqs-message-consumer.js eventually calls rules private api to get the list of rules back, and I can only verify the correct parameters are passed to the fetchRules method. packages/api/tests/endpoints/test-rules.js verifies that `rule.type` and `state` can be used as search parameter for the listing.
I verified the sqsMessageConsumer cw log only retrieves the enabled sql rules by looking at the log message `processQueues found ${rules.length} active sqs rules`.

botanical (approved):
🙇

PR #3239, opened 2023-01-31
Improve linting of TS files

marchuffnagle:
I noticed a couple of changes that needed to be fixed in our eslint config for TypeScript files:

- Disabled the eslint [@typescript-eslint/dot-notation](https://typescript-eslint.io/rules/dot-notation/) rule, which is contrary to the [noPropertyAccessFromIndexSignature](https://www.typescriptlang.org/tsconfig#noPropertyAccessFromIndexSignature) rule in TypeScript.
- Disabled the [valid-jsdoc](https://eslint.org/docs/latest/rules/valid-jsdoc) rule. This rule is useful in JS files where we depend on JSDocs to give us type information, but is not necessary in TypeScript files where types are defined.


Jkovarik (approved):
The conflicting dot-notation change makes sense. 

 It would be good to highlight the second change to the team and why it matters before merging, but not having that redundant information in the function comments seems entirely acceptable to me.

marchuffnagle:
> not having that redundant information in the function comments seems entirely acceptable to me.

Worse than redundant is "conflicting", which is what I was finding in a lot of places. I'll bring it up in the next architecture meeting.

PR #3244, opened 2023-02-02
Add python-version file

marchuffnagle:
The example/lambdas/python-processing package requires python 3.7, and bootstrapping will fail if it is not present. Adding a .python-version file that specifies Python 3.7, so developers using pyenv will have the right version set.

This replaces https://github.com/nasa/cumulus/pull/3211


Jkovarik (approved):
👍🏻

PR #3263, opened 2023-02-13
Remove Granule DynamoDB Table

npauzenga:
Removes the Granule DynamoDB table. Tested deployment and no issues. 

See https://github.com/nasa/cumulus/pull/3269/files for another example of a Dynamo table removal.


packages/es-client/search.js
```diff
@@ -252,75 +252,6 @@ class BaseSearch {
     return response.detail !== recordNotFoundString;
   }
 
-  async granulesStats(key, value) {
```
npauzenga, line 255: I didn't find any references to this method. It looks like this may have either been deprecated or partially replaced by the `Stats` class.

Jkovarik (approved):
Reviewed all changes, and agree with the additional updates, reviewed TF changeset (looks good) and looked for additional references. 

approved -- :100: great work.

PR #3264, opened 2023-02-14
Updates to ES tunnel script and instructions

npauzenga:
Just a couple updates after using the script for the first time. I believe we require a a Key Pair on each of the Cumulus EC2 instances so this adds that behavior.


marchuffnagle:
I'm fine with updating the docs to indicate that you need an SSH keypair configured. I'd rather not update the script, though, because there are different ways to indicate which keypair to use. I'm using the default keypair, so you don't need to specify anything. If you're using something non-default, you can configure that in `~/.ssh/config`.

PR #3266, opened 2023-02-14
CUMULUS-3010 remove providers table

marchuffnagle:
This _should_ be done, but need to run the tests against it once the test fixes are merged into the release branch.


npauzenga (approved):
Looks good @nasamoduyebo! I pulled the branch and don't see other provider table references. If deployment succeeds and CI passes I think this is good to go.

PR #3323, opened 2023-04-05
CUMULUS-3106 - localstack upgrade + removing skip from tests

Nnaga1:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/packages/aws-client/tests/test-S3.js
+++ b/packages/aws-client/tests/test-S3.js
@@ -94,7 +94,6 @@ test('createBucket() creates a bucket', async (t) => {
   }
 });
 
-// TODO enable the test after CUMULUS-3106 fix
 test('deleteS3Objects() deletes s3 objects', async (t) => {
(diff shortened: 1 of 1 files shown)


packages/aws-client/tests/test-S3.js
```diff
@@ -95,7 +95,7 @@ test('createBucket() creates a bucket', async (t) => {
 });
 
 // TODO enable the test after CUMULUS-3106 fix
```
npauzenga, line 97: Can probably remove this comment

npauzenga (approved):
Looks good @Nnaga1! According to the ticket this _could_ impact the AWS SDK upgrade we have going in parallel. It seems like it will be alright because we're already using the newer version in places but it's worth flagging.

One comment that we can probably remove but otherwise ready to go.

PR #3332, opened 2023-04-14
CUMULUS-3215: Fix Reconciliation Reports

botanical:

Addresses [CUMULUS-3215: Fix Reconciliation Reports](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3215)

## Changes

* Updated Reconciliation Reports to propagate error to async operations so that the async operation can properly be set to `TASK_FAILED` when there is an issue generating the report
* Updated reconciliation reports to retry when there is a Knex/DB error `Connection terminated unexpectedly` and retry 3 times. 
* I didn't update integration tests since I know we are trying to stray away from adding more to those tests, but am open to adding some if there is a compelling reason! 

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


packages/api/lambdas/create-reconciliation-report.js
```diff
@@ -803,43 +838,56 @@ async function processRequest(params) {
...
+            Error: error.message,
+            Cause: errorify(error),
+          },
+        };
+        await reconciliationReportModel.update({ name: reportRecord.name }, updates);
+        throw new pRetry.AbortError(error);
+      }
+    },
+    {
+      retries: 3,
```
botanical, line 884: I hardcoded this value but am open to changing it

packages/api/lambdas/create-reconciliation-report.js
```diff
@@ -803,43 +838,56 @@ async function processRequest(params) {
...
+          await createGranuleInventoryReport(recReportParams);
+        } else if (reportType === 'ORCA Backup') {
+          await createOrcaBackupReconciliationReport(recReportParams);
+        } else {
+          // reportType is in ['Inventory', 'Granule Not Found']
+          await createReconciliationReport(recReportParams);
+        }
+        await reconciliationReportModel.updateStatus({ name: reportRecord.name }, 'Generated');
+      } catch (error) {
+        if (error.message.includes('Connection terminated unexpectedly')) {
```
botanical, line 867: If the error is a `Connection terminated unexpectedly` error, I will retry

jennyhliu: In which circumstances, there will be `Connection terminated unexpectedly` error?
Which db operation would throw the error?  Could we retry that operation instead of retry the report generation?
If the report generation is retried 2 times, will we get 3 reports or 1 report?

packages/api/lambdas/create-reconciliation-report.js
```diff
@@ -715,77 +734,93 @@ async function createReconciliationReport(recReportParams) {
...
+    // compare the CUMULUS holdings with the holdings in CMR
+    // -----------------------------------------------------
+    if (location !== 'S3') {
+      const cumulusCmrReport = await reconciliationReportForCumulusCMR({
+        bucketsConfig, distributionBucketMap, recReportParams,
+      });
+      report = Object.assign(report, cumulusCmrReport);
+    }
+  } catch (error) {
+    log.error(`Error caught in createReconciliationReport for reportKey ${reportKey}`);
```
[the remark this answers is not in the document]
botanical: > Can we move this logic to here for all reports?

If I moved the logic there, then there would be no report to reference and I would need to return the report object or retrieve it from S3. 

I updated to write the failed report to s3 because in the beginning of `createReconciliationReport` we write the report to S3 https://github.com/nasa/cumulus/pull/3332/files/676b39531abf20b2fc27d988e1d582cbcded46fe#diff-61b4ede5087a45c0d1191b106bc92b405c40e436ec41bf30bd9b86ef5bb2d634R738 but if we are okay with the report having no status if report generation fails, then I can remove this: 
https://github.com/nasa/cumulus/pull/3332/files/676b39531abf20b2fc27d988e1d582cbcded46fe#diff-61b4ede5087a45c0d1191b106bc92b405c40e436ec41bf30bd9b86ef5bb2d634R783-R790

What do you think @jennyhliu ?

packages/api/lambdas/create-reconciliation-report.js
```diff
@@ -182,97 +183,115 @@ async function createReconciliationReportForBucket(Bucket, recReportParams) {
...
+        searchParams: { bucket: Bucket },
+        sortColumns: ['key'],
+        granuleColumns: ['granule_id'],
+        collectionIds: recReportParams.collectionIds,
+        providers: recReportParams.providers,
+        granuleIds: recReportParams.granuleIds,
+      });
 
-  const pgFileSearchClient = new QuerySearchClient(query, 100);
+      const pgFileSearchClient = new QuerySearchClient(query, 100);
```
jennyhliu, line 198: Is it not possible to retry the operation under `QuerySearchClient`?

packages/api/lambdas/create-reconciliation-report.js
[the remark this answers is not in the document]
botanical: I think we need to write the failed report to S3 since that was the original behavior. I think it's also important to keep in mind that before, we didn't throw any errors so we would always write to S3 even when the report generation failed. Now, since we throw an error, we need to make sure to update the report in S3, otherwise the report status will be 'RUNNING'. @jennyhliu

packages/api/lambdas/create-reconciliation-report.js
botanical: https://github.com/nasa/cumulus/pull/3332/commits/c211dac7a088f6dff4d33297170135234210edf5

https://github.com/nasa/cumulus/pull/3332/commits/9308412966a1da9b21b4b68b62aef3789d806bf2

packages/db/src/lib/QuerySearchClient.ts
```diff
@@ -1,7 +1,10 @@
 import { Knex } from 'knex';
+import pRetry from 'p-retry';
+import Logger from '@cumulus/logger';
 
 import { BaseRecord } from '../types/base';
 
+const log = new Logger({ sender: '@db/QuerySearchClient' });
```
jennyhliu, line 7: I think only fetchRecords() needs to be retried

packages/api/lambdas/create-reconciliation-report.js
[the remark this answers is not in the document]
jennyhliu: ok make sense, thanks for explanation

packages/api/tests/lambdas/test-create-reconciliation-report.js
```diff
@@ -2151,3 +2152,177 @@ test.serial('Inventory reconciliation report JSON is formatted', async (t) => {
...
+    handler(event),
+    { message: 'ORCA error' }
+  );
+
+  const reportKey = `${t.context.stackName}/reconciliation-reports/${reportName}.json`;
+  const report = await getJsonS3Object(t.context.systemBucket, reportKey);
+  t.is(report.status, 'Failed');
+  t.is(report.reportType, 'ORCA Backup');
+});
+
```
jennyhliu, line 2237: Since we move the retry logic to QuerySearchClient.ts, move/add test to QuerySearchClient

packages/api/lambdas/internal-reconciliation-report.js
```diff
@@ -59,82 +61,104 @@ async function internalRecReportForCollections(recReportParams) {
...
-  let nextEsItem = await esCollectionsIterator.peek();
-  let nextDbItem = dbCollectionItems.length !== 0
-    ? translatePostgresCollectionToApiCollection(dbCollectionItems[0])
-    : undefined;
-
-  while (nextEsItem && nextDbItem) {
-    const esCollectionId = constructCollectionId(nextEsItem.name, nextEsItem.version);
-    const dbCollectionId = constructCollectionId(nextDbItem.name, nextDbItem.version);
+  const knex = recReportParams.knex || await getKnexClient();
+
```
jennyhliu, line 65: Same comment as for inventory report, we should avoid retry at this level.  Retry at the db query level if needed

packages/api/lambdas/internal-reconciliation-report.js
```diff
@@ -193,8 +217,28 @@ async function getCollectionsForGranules(recReportParams) {
...
     granuleIds,
   } = recReportParams;
-
-  const dbCollectionIds = await getAllCollectionIdsByGranuleIds(recReportParams);
+  let dbCollectionIds = [];
+
+  await pRetry(
+    async () => {
+      try {
+        dbCollectionIds = await getAllCollectionIdsByGranuleIds(recReportParams);
```
jennyhliu, line 225: retry at the lowest level function when possible.

packages/api/tests/lambdas/test-create-reconciliation-report.js
jennyhliu: delete these tests.

packages/api/lambdas/internal-reconciliation-report.js
jennyhliu: Do we know which reconciliation report function/operation throws `Connection terminated unexpectedly`, so we can avoid adding `await pRetry` everywhere?

packages/api/lambdas/internal-reconciliation-report.js
botanical: Yep this is still in progress, definitely plan to do that!

packages/api/tests/lambdas/test-create-reconciliation-report.js
botanical: Hm, why do you think I should delete these tests @jennyhliu ?

packages/api/lambdas/internal-reconciliation-report.js
botanical: For the user, it occurred in the `Inventory` report. I don't think this issue is limited to the report type, though. It's a general knex issue so it can happen when generating any report. @jennyhliu

packages/api/tests/lambdas/test-create-reconciliation-report.js
jennyhliu: Because the retry logic in now in `fetchRecords`, and we only need to test that instead.

botanical: @jennyhliu https://github.com/nasa/cumulus/pull/3332/commits/c3679b8e447af7aebf856bb4650fff597eeca7c2#diff-f6faf60179e0607db77cd33d9b72bda73294727ed1b2e12a00e96af5c929e748L2156

packages/api/lambdas/internal-reconciliation-report.js
```diff
@@ -59,82 +60,90 @@ async function internalRecReportForCollections(recReportParams) {
...
-  let nextDbItem = dbCollectionItems.length !== 0
-    ? translatePostgresCollectionToApiCollection(dbCollectionItems[0])
-    : undefined;
-
-  while (nextEsItem && nextDbItem) {
-    const esCollectionId = constructCollectionId(nextEsItem.name, nextEsItem.version);
-    const dbCollectionId = constructCollectionId(nextDbItem.name, nextDbItem.version);
+  const knex = recReportParams.knex || await getKnexClient();
+
+  try {
```
jennyhliu, line 65: Is it necessary to add try/catch here?  We can try/catch the error in upper caller function to simplify the code, and you have already added try/catch in createInternalReconciliationReport, and log error there.

packages/api/lambdas/reports/granule-inventory-report.js
```diff
@@ -38,59 +40,63 @@ async function createGranuleInventoryReport(recReportParams) {
...
-  const granulesSearchQuery = getGranulesByApiPropertiesQuery(
-    recReportParams.knex,
-    searchParams,
-    ['collectionName', 'collectionVersion', 'granule_id']
-  );
-  const pgGranulesSearchClient = new QuerySearchClient(
-    granulesSearchQuery,
-    100 // arbitrary limit on how items are fetched at once
-  );
+  try {
```
jennyhliu, line 43: same comment, why try/catch error if we don't need to handle the error?

packages/db/src/lib/QuerySearchClient.ts
```diff
@@ -29,12 +32,31 @@ class QuerySearchClient<RecordType extends BaseRecord> {
    * @throws
    */
   private async fetchRecords() {
-    this.records = await (
-      this.query
-        .offset(this.offset)
-        .limit(this.limit)
+    return await pRetry(
```
jennyhliu, line 35: I think we should have a common method, like
RetryOnDbConnectionTerminateError(methodName, options) {
  await pRetry(methodName, { retries: 3, onFailed , ...options})
}

packages/api/lambdas/internal-reconciliation-report.js
botanical: That's true, I can update

packages/api/lambdas/reports/granule-inventory-report.js
botanical: Yeah the main benefit would be logging where the error occurs but I can remove them too

packages/api/lambdas/internal-reconciliation-report.js
botanical: https://github.com/nasa/cumulus/pull/3332/commits/ff7b1d42578019db98a58d7ebc1032203f7e853b#diff-d06392e59ba6c4315b7509efe5ae98c2ad79eab742333baf145d95d4d9e266a8R65

packages/api/lambdas/create-reconciliation-report.js
```diff
@@ -194,17 +194,19 @@ async function createReconciliationReportForBucket(Bucket, recReportParams) {
...
   const pgFileSearchClient = new QuerySearchClient(query, 100);
 
-  log.info(`createReconciliationReportForBucket(S3 vs. PostgreSQL): ${Bucket}: ${JSON.stringify(recReportParams)}`);
   let okCount = 0;
   const onlyInS3 = [];
   const onlyInDb = [];
   const okCountByGranule = {};
+
+  log.info(`createReconciliationReportForBucket(S3 vs. PostgreSQL): ${Bucket}: ${JSON.stringify(recReportParams)}`);
   try {
```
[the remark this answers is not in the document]
botanical: https://github.com/nasa/cumulus/pull/3332/commits/ff7b1d42578019db98a58d7ebc1032203f7e853b#diff-61b4ede5087a45c0d1191b106bc92b405c40e436ec41bf30bd9b86ef5bb2d634R203

packages/api/lambdas/reports/granule-inventory-report.js
botanical: https://github.com/nasa/cumulus/pull/3332/commits/ff7b1d42578019db98a58d7ebc1032203f7e853b#diff-3cdf7146169c9c58cfbfd0e1652f49367ffcafafb2527f1979f5458fb77451e0R43

packages/db/src/lib/QuerySearchClient.ts
botanical: https://github.com/nasa/cumulus/pull/3332/commits/afb3b702976a0e4db8c5e63f6520c39e423ec898#diff-4407d17441d12af663b8a19415c93087978d8d8a6a80b3e7ec4ea9222890c388R8

packages/db/src/lib/retry.ts
```diff
@@ -0,0 +1,34 @@
...
+      } catch (error) {
+        if (error.message.includes('Connection terminated unexpectedly')) {
+          logger.error(`${error}. Retrying...`);
+          throw error;
+        }
+        throw new pRetry.AbortError(error);
+      }
+    },
+    {
+      ...retryConfig,
```
jennyhliu, line 27: put this after other parameters to override

packages/db/src/lib/retry.ts
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+import pRetry from 'p-retry';
+import Logger from '@cumulus/logger';
+
+type PromiseReturnType<T> = T extends (method: any) => infer R ? Promise<R> : never;
```
jennyhliu, line 6: not sure about the type definition here.
Should `method: any` be `method: function`?

packages/api/lambdas/create-reconciliation-report.js
```diff
@@ -839,6 +864,7 @@ async function processRequest(params) {
       },
     };
     await reconciliationReportModel.update({ name: reportRecord.name }, updates);
+    throw error;
```
jennyhliu, line 867: If the lambda fails, what happens to the AsyncOperation?  Does it retry the task or just fail the AsyncOperation?  Can we test it out?

CHANGELOG.md
```diff
@@ -7238,5 +7244,3 @@ Note: There was an issue publishing 1.12.0. Upgrade to 1.12.1.
 [v1.1.0]: https://github.com/nasa/cumulus/compare/v1.0.1...v1.1.0
 [v1.0.1]: https://github.com/nasa/cumulus/compare/v1.0.0...v1.0.1
 [v1.0.0]: https://github.com/nasa/cumulus/compare/pre-v1-release...v1.0.0
-
-[thin-egress-app]: <https://github.com/asfadmin/thin-egress-app> "Thin Egress App"
```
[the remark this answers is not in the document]
botanical: Hm yeah it seems like the automatic merge keeps deleting this

packages/db/src/lib/retry.ts
botanical: https://github.com/nasa/cumulus/pull/3332/commits/0962c4bd57ba8cc70c4df475c8fed21f1c06639e

packages/db/src/lib/retry.ts
botanical: https://github.com/nasa/cumulus/pull/3332/commits/0962c4bd57ba8cc70c4df475c8fed21f1c06639e

packages/api/lambdas/create-reconciliation-report.js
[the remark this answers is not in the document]
botanical: It sets the async operation to `TASK_FAILED` and retries for `Connection terminated unexpectedly`

packages/api/lambdas/internal-reconciliation-report.js
```diff
@@ -193,8 +197,13 @@ async function getCollectionsForGranules(recReportParams) {
   const {
     granuleIds,
   } = recReportParams;
-
-  const dbCollectionIds = await getAllCollectionIdsByGranuleIds(recReportParams);
```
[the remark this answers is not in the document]
botanical: https://github.com/nasa/cumulus/pull/3332/commits/d7a14883a4ad8281e240cf48fa30aae6ad78e0ed

packages/db/src/lib/retry.ts
```diff
@@ -0,0 +1,37 @@
...
+import { Knex } from 'knex';
+import pRetry from 'p-retry';
+import Logger from '@cumulus/logger';
+
+type PromiseReturnType<T> = T extends (
+  method: Promise<any> | Knex.QueryBuilder
+) => infer R ? Promise<R> : never;
+
+export const RetryOnDbConnectionTerminateError = async <T>(
+  method: Promise<any> | Knex.QueryBuilder,
```
jennyhliu, line 12: Does `Promise<any>` includes `Knex.QueryBuilder`?  Remove `Knex.QueryBuilder` if we can't define all specific types.

botanical: `Promise<any>` doesn't include `Knex.QueryBuilder` unfortunately 😕

jennyhliu (approved):
Nice work!

PR #3386, opened 2023-05-18
Cumulus 3307 pg version pin

etcart:
**Summary:** 
Addresses local bootstrap failures stemming from bug introduced in node pg version 8.11

## Changes

* pg version pinned to 8.10x in cumulus/db

## PR Checklist

- [X] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


packages/db/package.json
```diff
@@ -39,7 +39,7 @@
     "is-valid-hostname": "1.0.2",
     "knex": "2.4.1",
     "lodash": "^4.17.21",
-    "pg": "^8.3.0",
+    "pg": "~8.10.0",
```
charleshuang80, line 42: Not sure if this matters, the changelog says `v8.10.x`, but here it's `~8.10.0` (so I don't think it would go any higher than the .0 if I'm reading [this doc](https://docs.npmjs.com/about-semantic-versioning) correctly)

npauzenga:
Might we also need this in the `release-16.0.x` branch? I suspect bootstrapping will be blocked there as well. Probably need to cherry-pick the commit from main to the release branch once this PR is merged.

charleshuang80:
Ok -- updating the pg version in `db` and in `db-migration` did not work for me. I had to update the pg version in all the package.json files. Then it all worked.

charleshuang80 (approved):

PR #3391, opened 2023-05-24
CUMULUS-3252:Add orca copy_to_archive_adapter lambda to cumulus

jennyhliu:

Addresses [CUMULUS-3252 Add orca copy_to_archive_adapter lambda to cumulus](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3252)

## Changes

* see CHANGELOG.md

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


packages/oauth-client/src/EarthdataLoginClient.ts
```diff
@@ -53,7 +53,7 @@ export class EarthdataLoginClient extends OAuthClient {
       default:
         return new EarthdataLoginError(
           'UnexpectedResponse',
-          `Unexpected response: ${httpError.response.body}`
+          `Unexpected response: ${JSON.stringify(httpError.response.body)}`
```
sbquinlan, line 56: should we just use httpError.response.body.error?

sbquinlan (commented):
looks good to me, but I'll let more tenured people approve.

packages/oauth-client/src/EarthdataLoginClient.ts
jennyhliu: This error is not specific, so we want detailed error message.
I got "[Object, Object]" in one of the integration-test run before the fix, and still don't the detailed error is.

sbquinlan (approved):

PR #3393, opened 2023-05-24
CUMULUS-2625: Improve heap memory usage in queue-granules

sbquinlan:
**Summary:** Improve heap memory usage in queue-granules

Addresses [CUMULUS-2625](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2625). 

## Changes

* Change `groupAndBatchGranules` to be an Iterator, yielding chunks of granules at a time instead of the entire collection of chunks at once. 
* Optimize calls to the api by memoizing the responses and moving the calls out of the inner loops.

## Notes

* `groupAndBatchGranules` goes `groupby collection -> chunk -> groupby provider`. It would be more performant to do `groupby collection + provider -> chunk` but that does technically change the chunks. I'm assuming that's not intentional as I can't think of a reason to chunk before grouping by provider. 
* 800k granules in a single workflow call is pretty close to what you could expect Node to allocate on the heap to begin with (1.4 GB of heap space by default). I tested my changes up to 100k successfully in ECS. The current master can't do 10k in ECS, but this has alot more to do with api calls can memory.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests

### Ad-hoc Testing
The following is from manually testing memory using ```node --inspect``` and then connecting the Chrome Devtools to the process to record allocations while running `index.js` with fake granules and the api calls mocked out.

Memory before:
Screenshot 2023-05-23 at 4 54 45 PM

Memory after:
Screenshot 2023-06-05 at 9 30 08 PM

### Ad-hoc Load Testing

Please see https://github.com/nasa/cumulus/tree/test/CUMULUS-2625 for the task `GenerateGranules` and the associated workflow I used to load test these changes.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -113,6 +113,7 @@ module.exports = {
     ],
     strict: 'off',
     'guard-for-in': 'off',
+    'no-restricted-syntax': ['error', 'ForInStatement', 'LabeledStatement', 'WithStatement'],
     'object-shorthand': 'off',
     'space-before-function-paren': [
       'error',
--- a/packages/ingest/src/queue.js
+++ b/packages/ingest/src/queue.js
@@ -79,10 +79,10 @@ module.exports.enqueueParsePdrMessage = enqueueParsePdrMessage;
  * Enqueue a granule to be ingested
  *
  * @param {Object} params
- * @param {Object} params.granule - the granule to be enqueued for ingest
+ * @param {Object} params.granules - the granules to be enqueued for ingest
  * @param {string} params.queueUrl - the SQS queue to add the message to
- * @param {string} params.granuleIngestMessageTemplateUri - the S3 URI of template for
- * a granule ingest message
+ * @param {Object} params.messageTemplate - Message template for the workflow
+ * @param {Object} params.workflow - workflow name & arn object
  * @param {Object} params.provider - the provider config to be attached to the message
  * @param {Object} params.collection - the collection config to be attached to the
  *   message
@@ -96,52 +96,34 @@ module.exports.enqueueParsePdrMessage = enqueueParsePdrMessage;
 async function enqueueGranuleIngestMessage({
   collection,
   granules,
-  granuleIngestWorkflow,
   parentExecutionArn,
   pdr,
   provider,
-  stack,
-  systemBucket,
+  messageTemplate,
+  workflow,
   queueUrl,
   executionNamePrefix,
   additionalCustomMeta = {},
 }) {
-  const messageTemplate = await getJsonS3Object(systemBucket, templateKey(stack));
-  const { arn: ingestGranuleArn } = await getJsonS3Object(
-    systemBucket,
-    getWorkflowFileKey(stack, granuleIngestWorkflow)
-  );
-
-  const payload = { granules };
-
-  const workflow = {
-    name: granuleIngestWorkflow,
-    arn: ingestGranuleArn,
-  };
-
   const message = buildQueueMessageFromTemplate({
     messageTemplate,
     parentExecutionArn,
-    payload,
+    payload: { granules },
     workflow,
     customMeta: {
       ...additionalCustomMeta,
+      ...(pdr ? { pdr } : {}),
       collection,
       provider,
     },
     executionNamePrefix,
   });
 
-  if (pdr) message.meta.pdr = pdr;
-
-  const arn = buildExecutionArn(
+  await sendSQSMessage(queueUrl, message);
+  return buildExecutionArn(
     message.cumulus_meta.state_machine,
     message.cumulus_meta.execution_name
   );
-
-  await sendSQSMessage(queueUrl, message);
-
-  return arn;
 }
 exports.enqueueGranuleIngestMessage = enqueueGranuleIngestMessage;
 
--- a/tasks/queue-granules/index.js
+++ b/tasks/queue-granules/index.js
@@ -1,24 +1,34 @@
 'use strict';
 
 const get = require('lodash/get');
-const groupBy = require('lodash/groupBy');
-const chunk = require('lodash/chunk');
 const isNumber = require('lodash/isNumber');
+const memoize = require('lodash/memoize');
 const pMap = require('p-map');
 
 const cumulusMessageAdapter = require('@cumulus/cumulus-message-adapter-js');
-const { enqueueGranuleIngestMessage } = require('@cumulus/ingest/queue');
+const {
+  getWorkflowFileKey,
+  templateKey,
+} = require('@cumulus/common/workflows');
 const { constructCollectionId, deconstructCollectionId } = require('@cumulus/message/Collections');
 const { buildExecutionArn } = require('@cumulus/message/Executions');
+const { buildQueueMessageFromTemplate } = require('@cumulus/message/Build');
+const { getJsonS3Object } = require('@cumulus/aws-client/S3');
+const { sendSQSMessage } = require('@cumulus/aws-client/SQS');
+
 const {
+  collections: collectionsApi,
   providers: providersApi,
   granules: granulesApi,
 } = require('@cumulus/api-client');
-const CollectionConfigStore = require('@cumulus/collection-config-store');
 
-async function fetchGranuleProvider(prefix, providerId) {
+async function fetchGranuleProvider(event, providerId) {
+  if (!providerId || providerId === event.config.provider.id) {
+    return event.config.provider;
+  }
+
   const { body } = await providersApi.getProvider({
-    prefix,
+    prefix: event.config.stackName,
     providerId,
   });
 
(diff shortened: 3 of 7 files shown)


tasks/queue-granules/index.js
```diff
@@ -1,51 +1,109 @@
...
+ * generators that yield the chunks of granules one at a time.
  *
- * @param {Array<Object>} granules - list of input granules
- * @param {number} batchSize - size of batch of granules to queue
- * @returns {Array<Object>} list of lists of granules: each list contains granules which belong
- *                          to the same collection, and each list's max length is set by batchSize
+ * The *[Symbol.iterator] method loops over the collection looking for groups that have not
+ * already been yielded. Once the first granule of an unseen group is located, a new iterator is
+ * created that will search from that point in the collection of granules looking for other
+ * granules that to the same group.
```
botanical, line 46: This is a great description! However, is it missing a word? 
```suggestion
 * granules that belong to the same group.
```

tasks/queue-granules/index.js
```diff
@@ -62,6 +120,69 @@ function updateGranuleBatchCreatedAt(granuleBatch, createdAt) {
   }));
 }
 
+/**
+ * Constructs the partial params for enqueueGranuleIngestMessage. This is an optimization.
+ * This part of the message requires loading files out of S3. We don't want to do that everytime
+ * we create an SQS message considering that these parts at the same for every message, so this
```
botanical, line 126:
```suggestion
 * This part of the message requires loading files out of S3. We don't want to do that every time
 * we create an SQS message considering that these parts are the same for every message, so this
```
`everytime` should be `every time` and I think `at` is meant to be `are`?

.eslintrc.js
```diff
@@ -113,6 +113,7 @@ module.exports = {
     ],
     strict: 'off',
     'guard-for-in': 'off',
+    'no-restricted-syntax': ['error', 'ForInStatement', 'LabeledStatement', 'WithStatement'],
```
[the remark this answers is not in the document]
sbquinlan: @Nnaga1 without this change, there's a lint error that generators and iterators are not allowed because they require a poly-fill (specifically https://github.com/facebook/regenerator) but that's specific to running javascript in a browser and doesn't apply to node.js so I removed the "no generators" rule with this override.

tasks/queue-granules/index.js
```diff
@@ -62,6 +120,69 @@ function updateGranuleBatchCreatedAt(granuleBatch, createdAt) {
...
+    executionNamePrefix,
+  });
+
+  await sendSQSMessage(queueUrl, message);
+
+  return buildExecutionArn(
+    message.cumulus_meta.state_machine,
+    message.cumulus_meta.execution_name
+  );
+}
```
botanical, line 184: Is there a reason we shouldn't just update the function in the ingest package?

sbquinlan: yea def could do that, but I would do that refactor separately in all tasks rather than growing this PR to do that.

tasks/queue-granules/tests/index.js
```diff
@@ -830,3 +837,217 @@ test('updatedGranuleBatchCreatedAt updates batch granule object with correct cre
...
+  await queueGranules(event, testMocks);
+
+  const createdMap = Object.fromEntries(
+    testMocks.updateGranuleMock.getCalls()
+      .map(({ args: [params] }) => [params.granuleId, params.body.createdAt])
+  );
+  t.deepEqual(
+    testMocks.updateGranuleMock.getCalls()
+      .map(({ args: [params] }) => params)
+      .sort(({ granuleId: a }, { granuleId: b }) => a.localeCompare(b)),
```
botanical, line 901: What is the benefit of sorting? 🤔

botanical: What if we only compare the collectionId ?
Something like:
```
t.deepEqual(
    testMocks.updateGranuleMock.getCalls()
      .map(({ args: [{ collectionId }] }) => collectionId),
    event.input.granules.map(({ granuleId, dataType, version }) => { const collectionId = constructCollectionId(dataType, version); return collectionId })
  );
```

tasks/queue-granules/index.js
```diff
@@ -76,78 +197,92 @@ async function queueGranules(event, testMocks = {}) {
   const enqueueGranuleIngestMessageFn
     = testMocks.enqueueGranuleIngestMessageMock || enqueueGranuleIngestMessage;
 
-  const collectionConfigStore = new CollectionConfigStore(
-    event.config.internalBucket,
-    event.config.stackName
+  const memoizedFetchProvider = memoize(fetchGranuleProvider, (_, providerId) => providerId);
+  const memoizedFetchCollection = memoize(
+    collectionsApi.getCollection,
+    ({ collectionName, collectionVersion }) => constructCollectionId(
```
jennyhliu, line 203: `getCollection` takes `prefix, collectionName, collectionVersion`.  I'm wondering how the integration-test passes.

tasks/queue-granules/index.js
jennyhliu: We should update `enqueueGranuleIngestMessage` in the ingest package to take `granuleIngestWorkflowArn` and `cumulusMessageTemplate`, and that function is only used in `queue-granules` package for now.

jennyhliu (changes requested):

tasks/queue-granules/index.js
```diff
@@ -1,51 +1,109 @@
...
- * Group granules by collection and split into batches then split again on provider
+ * The purpose of this iterable is to avoid creating all the group and chunk arrays all
+ * at once to save heap space for large event inputs. This is done by using a sequence of
+ * generators that yield the chunks of granules one at a time.
  *
- * @param {Array<Object>} granules - list of input granules
- * @param {number} batchSize - size of batch of granules to queue
- * @returns {Array<Object>} list of lists of granules: each list contains granules which belong
- *                          to the same collection, and each list's max length is set by batchSize
+ * The *[Symbol.iterator] method loops over the collection looking for groups that have not
```
jennyhliu, line 43: Move the doc to the method

tasks/queue-granules/index.js
sbquinlan: getCollection is called with the prefix on line 231. I think you're confusing the second arg as the call to the api method but it's actually the key for the memoize.

tasks/queue-granules/index.js
sbquinlan: This is intentionally an overview of all the methods in the class so the doc block is on the class. I don't think it helps understand the code to move this particular sentence to that method.

jennyhliu (commented):
It's agreed that new task/package etc. should be written in typescript.  Since most methods are rewritten, I'm wondering if we should take the opportunity to convert it to typescript.  We can discuss with the team.

tasks/queue-granules/tests/iterator.js
```diff
@@ -0,0 +1,85 @@
...
+const test = require('ava');
+const { GroupedGranulesIterable } = require('..');
+
+function createGranule(granuleId, dataType, version, provider = undefined) {
+  return {
+    granuleId,
+    dataType,
+    version,
+    files: [],
+    ...(provider ? { provider } : {}),
```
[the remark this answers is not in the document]
sbquinlan: I prefer ternary because it's more explicit.

tasks/queue-granules/index.js
sbquinlan: Hm I'm trying to understand what documentation you'd like to see here. "so other developers can understand and use the method without looking into the underneath implementation." but the whole class is the thing you'd use, not just one method. The class has the documentation on it that explains how the whole things works so that you don't have to jump into the details. I feel like documenting each method requires the developer to read all the implementation details as you say. Would you like me to just additionally document the methods individually incase people would like some help understanding the implementation? I'm just not sure what the solution you'd like to see here is.

jennyhliu: Yeah, please add a few sentences to help people understand the logic of the individual method

tasks/queue-granules/tests/index.js
sbquinlan: Isn't the assertion about all the params more pedantic? I sort because the calls can be reordered by the chunking/grouping and deepEqual cares about the ordering I think.

packages/ingest/src/queue.js
```diff
@@ -96,52 +96,34 @@ module.exports.enqueueParsePdrMessage = enqueueParsePdrMessage;
...
       ...additionalCustomMeta,
+      ...(pdr ? { pdr } : {}),
       collection,
       provider,
     },
     executionNamePrefix,
   });
 
-  if (pdr) message.meta.pdr = pdr;
-
```
jennyhliu, line 136: good refactorings

jennyhliu (approved):
Great work!👍

PR #3404, opened 2023-06-05
CUMULUS-3253:Add orca_recovery_adapter lambda to cumulus

jennyhliu:

Addresses [CUMULUS-3253: Work with ORCA with other Lambda Adapters](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3253)

## Changes

* see CHANGELOG.md

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


tasks/orca-recovery-adapter/src/index.ts
```diff
@@ -0,0 +1,127 @@
...
+ *   that resolves to the output of step function execution
+ */
+export const getStateMachineExecutionResults = async (
+  params: {
+    executionArn: string,
+    retries?: number,
+    retryIntervalInSecond?: number,
+    maxRetryTimeInSecond?: number,
+  }
+) : Promise<AWS.StepFunctions.DescribeExecutionOutput> => {
```
sbquinlan, line 36:
```suggestion
): Promise<AWS.StepFunctions.DescribeExecutionOutput> => {
```

tasks/orca-recovery-adapter/src/index.ts
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Invoke ORCA Recovery workflow
+ *
+ * @param {HandlerEvent} event - input from the message adapter
+ * @returns {Promise<HandlerOutput>} - returns output from ORCA workflow
+ */
+export const invokeOrcaRecoveryWorkflow = async (
+  event: HandlerEvent
+) : Promise<HandlerOutput> => {
```
sbquinlan, line 72:
```suggestion
): Promise<HandlerOutput> => {
```

tasks/orca-recovery-adapter/src/index.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  const payload = {
+    payload: event.input,
+    config: event.config,
+  };
+
+  const currentWorkflowArn = buildExecutionArn(
+    event.cumulus_config?.state_machine || '',
+    event.cumulus_config?.execution_name || ''
```
sbquinlan, line 86:
```suggestion
    event.cumulus_config?.state_machine!,
    event.cumulus_config?.execution_name!
```

should we assert these aren't undefined? Feels like it'll produce a broken arn if they are empty string.

tasks/orca-recovery-adapter/src/index.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  try {
+    await sfn().startExecution(workflowParams).promise();
+  } catch (error) {
+    log.error(`Error starting ${childWorkflowArn}`, error);
+    throw error;
+  }
+
+  log.info(`About to get result from execution ${childWorkflowArn}`);
+  const executionResult = await getStateMachineExecutionResults({ executionArn: childWorkflowArn || '' });
```
sbquinlan, line 106: if childWorkflowArn is null then we shouldn't get here. Either we should use the typescript `!` syntax or we should assert that it is not null before this point. Alternatively `buildExecutionArn` could use overloads or conditional types to make it clear it returns string when both params are strings.

tasks/orca-recovery-adapter/src/index.ts
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Invoke ORCA Recovery workflow
+ *
+ * @param {HandlerEvent} event - input from the message adapter
+ * @returns {Promise<HandlerOutput>} - returns output from ORCA workflow
+ */
+export const invokeOrcaRecoveryWorkflow = async (
+  event: HandlerEvent
+) : Promise<HandlerOutput> => {
+  const workflowArn = process.env.orca_sfn_recovery_workflow_arn;
```
sbquinlan, line 73: For my own understanding, why can't we pull this from the event.config instead?

tasks/orca-recovery-adapter/src/index.ts
```diff
@@ -0,0 +1,127 @@
...
+    maxRetryTimeInSecond?: number,
+  }
+) : Promise<AWS.StepFunctions.DescribeExecutionOutput> => {
+  const {
+    executionArn,
+    retries = 50,
+    retryIntervalInSecond = 5,
+    maxRetryTimeInSecond = 1800,
+  } = params;
+  const result = await pRetry(
```
sbquinlan, line 43:
```suggestion
  return await pRetry(
```
nit: return directly

tasks/orca-recovery-adapter/src/index.ts
jennyhliu: I added empty string to avoid typescript error
`Argument of type 'string | undefined' is not assignable to parameter of type 'string'.`
It's a good idea to use non-null assertion operator.  Thanks!

sbquinlan (approved):
Nice!

PR #3406, opened 2023-06-08
Cumulus 3295 coverage ratcheting

etcart:
**Summary:**
implements coverage ratcheting to push thresholds up as code coverage improves, and facilitate lower thresholds as necessary

Addresses [CUMULUS-3295: coverage ratcheting]

## Changes

* addition of coverage.py in scripts
* call of coverage.py through npm scripts coverage:update and coverage:validate throughout the projects
* call of coverage:validate in cicd to ensure that coverage hasn't risen without updating thresholds
* introduction of a grace margin +- 3% to prevent undue hassle on developers.

## PR Checklist

- [ ] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests

for ad-hoc tests, see previous failed runs of CICD, done with bad thresholds to demonstrate functionality


packages/async-operations/package.json
```diff
@@ -15,7 +15,9 @@
     "tsc": "../../node_modules/.bin/tsc",
     "tsc:listEmittedFiles": "../../node_modules/.bin/tsc --listEmittedFiles",
     "tsc:watch": "../../node_modules/.bin/tsc -w",
-    "watch-test": "../../node_modules/.bin/tsc-watch --onsuccess 'npm test'"
+    "watch-test": "../../node_modules/.bin/tsc-watch --onsuccess 'npm test'",
+    "coverage:validate": "python ../../scripts/coverage_handler/coverage.py --noRerun",
```
sbquinlan, line 19: not big deal, but could you have just one 'coverage' script here and then call it from the root project scripts with the different flag passed in depending on the action?

lambdas/db-provision-user-database/.nycrc.json
```diff
@@ -1,6 +1,7 @@
 {
-  "statements": 100,
-  "functions": 100,
+  "extends": "../../nyc.config.js",
+  "lines": 97,
   "branches": 88,
-  "lines": 100
+  "statements": 97,
+  "functions": 97
```
sbquinlan, line 6: strange these are ints and the others are precision

sbquinlan (approved):
seems fine.

nyc.config.js
```diff
@@ -2,9 +2,9 @@
 
 const path = require('path');
 module.exports = {
-  all: true,
```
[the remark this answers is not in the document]
etcart: this is not strictly a part of scope, but a fix I encountered the need for in working.
and it says "based on all files" not simply the ones touched by tests, and opened up a possibility of inconsistency between coverage measured local vs in bamboo, most notably because nyc produces javascript files if you ask it to produce coverage report in html, which can get included under "all"

PR #3409, opened 2023-06-14
v16.0.1 alpha release for testing

npauzenga:
The purpose of this Alpha release is to test the dashboard against Cumulus Core v16. We have a few dashboard changes here that need to be tested: https://github.com/nasa/cumulus-dashboard/pull/1103 but we also need to generally test against v16.

This also includes this PR to fix serve.js for local API development: https://github.com/nasa/cumulus/pull/3408

The process will be:
1. Release Core `v16.0.1-alpha.0` as a pre-release and publish packages to NPM
2. Update the dashboard to use the new packages in this PR https://github.com/nasa/cumulus-dashboard/pull/1103
3. Run tests/CI
4. Manually evaluate SIT as necessary


packages/api/bin/serve.js
```diff
@@ -319,15 +320,23 @@ async function serveApi(user, stackName = localStackName, reseed = true) {
...
     checkEnvVariablesAreSet(requiredEnvVars);
 
-    // create tables if not already created
-    await pRetry(
-      async () => await checkOrCreateTables(stackName),
-      {
-        onFailedAttempt: (error) => console.log(
-          `Failed to Create Tables. Localstack may not be ready, will retry ${error.attemptsLeft} more times.`
-        ),
+    // Create reconciliation report table
```
npauzenga, line 323: On main we've removed the `checkOrCreateTables` call here which creates all of the Dynamo tables. This _also_ removes it from the v16 release and replaces it with the call to create the reconciliation report table only.

PR #3413, opened 2023-06-22
CUMULUS-2871: Implement bulk database writes for queue-granules

sbquinlan:
**Summary:** Implement bulk database writes for queue-granules

Addresses [CUMULUS-2871](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2871)

## Changes

* Added `getGranuleExists` method to db package to test for granule existence and test.
* Added `bulkUpdateGranules` api method to update multiple granules in the same collection at once and tests.
* Added api-client call to the new api method and test.
* Added the new call to the `queue-granules` task and updated tests.

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


packages/api/endpoints/granules.js
```diff
@@ -1075,6 +1077,74 @@ async function bulkReingest(req, res) {
   return res.status(202).send({ id: asyncOperationId });
 }
```
jennyhliu, line 1079: Why is this restricted to the same collection?
If this method is restricted to the same collection, the method and endpoint name should be updated to avoid confusion and misuse.

jennyhliu (commented):
The update significantly (if batch size is large) reduces API calls.  However, it still does individual granule update, not 'bulk' database update.

packages/api/endpoints/granules.js
sbquinlan: What would you suggest as a solution here?

sbquinlan:
@jennyhliu I still need this PR to do the bulk DB update you're suggesting.

PR #3423, opened 2023-07-06
CUMULUS-3287 + additional User PR

Nnaga1:

Addresses both User Contribution PRs: https://github.com/nasa/cumulus/pull/3383 and https://github.com/nasa/cumulus/pull/3353

## Changes

  - Added variable to allow the aws_ecs_task_definition health check to be configurable.
  - Added clarity to how the bucket field needs to be configured for the move-granules task definition
  
## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests

Addresses both: https://github.com/nasa/cumulus/pull/3383 and https://github.com/nasa/cumulus/pull/3353

### DIFF
--- a/tf-modules/cumulus_ecs_service/variables.tf
+++ b/tf-modules/cumulus_ecs_service/variables.tf
@@ -104,10 +104,10 @@ variable "health_check" {
   description = "Health check used by AWS ECS to determine containers health status. See https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_healthcheck"
   type = object({
     command     = list(string)
-    interval    = optional(number)
-    timeout     = optional(number)
-    retries     = optional(number)
-    startPeriod = optional(number)
+    interval    = number
+    timeout     = number
+    retries     = number
+    startPeriod = number
   })
   default = null
 }


tf-modules/cumulus_ecs_service/variables.tf
```diff
@@ -99,3 +99,15 @@ variable "log_destination_arn" {
...
+variable "health_check" {
+  description = "Health check used by AWS ECS to determine containers health status. See https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_healthcheck"
+  type = object({
+    command = list(string)
+    interval = number
+    timeout = number
+    retries = number
+    startPeriod = number
+  })
+  default = null
```
botanical, line 112: Something to note for the future is that, when we upgrade terraform, we will be able to set 
 ```
     interval = number
    timeout = number
    retries = number
    startPeriod = number
   ```
   
  to 
  ```
      interval = optional(number)
    timeout =  optional(number)
    retries =  optional(number)
    startPeriod =  optional(number)
```

since these are technically not required to set a health check. For now, users will need to set all those to null but that is fine.

botanical (approved):
Thanks for ushering these user contributions through @Nnaga1 🙇

PR #3442, opened 2023-07-25
Cumulus 3270

etcart:
**Summary:**

updates to python3.10 along with dependencies including cumulus-message-adapter-python and cumulus-process-py

## Changes

* update python version to 3.10
* update aws provider version
* update boto3 dependency to 1.26.90
* update cumulus-process dependency to 1.3.0
* update cumulus-message-adapter-python dependency to 2.2.0
* update cumulus base image to come loaded with python 3.10 installed
**requires run of cumulus image publisher while merging to master**
## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


bamboo/docker-compose.yml
etcart, line 1: this should always not have been hardcoded, but was a deal-breaker with testing bas eimage updates

bamboo/Dockerfile
etcart, line 1: 3.10 isn't readily accessible in the base image, but we can install it the hard way

bamboo/set-bamboo-env-variables.sh
etcart, line 1: needed as part of the above inclusion of $CUMULUS_BASE_IMAGE

charleshuang80 (approved):
suggestions for changelog updates. changes look good, though looks like CI choked on something.

bamboo/Dockerfile
npauzenga: Is that true for the later Debian versions e.g. https://github.com/nodejs/docker-node#nodebullseye ?

Might be, just curious if we're able to use a different version and skip the manual Python install.

example/lambdas/python-processing/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.7-buster
```
npauzenga, line 1: Just curious- why Debian v11 over v10 or v12?

bamboo/docker-compose.yml
npauzenga: Yeah, good change. `CUMULUS_BASE_IMAGE` was added after this Docker config so makes sense it wasn't updated here.

npauzenga (changes requested):
Looks good @etcart! Just a couple questions

bamboo/Dockerfile
etcart: it is, I haven't explored beyond the bounds of 16.9.0 versions, but I haven't found one that doesn't require me to do to the full install

example/lambdas/python-processing/Dockerfile
etcart: that's a fair question: I don't remember why I did that and I'm pretty sure there was no reason, I'm switching it back

charleshuang80: I think it was me - I just found bullseye as an image with 3.10 and used it. I didn't know what buster or bullseye meant, so TIL.

npauzenga (approved):
Nice work @etcart! Looks great to me!

PR #3462, opened 2023-08-18
Add logging when rules do not match collection

npauzenga:

Addresses [CUMULUS-3366](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3366)

## Changes

In the sqs-message-consumer Lambda we filter rules to trigger by the incoming collection. If no rule matches the incoming collection, nothing happens. This update adds a log if we don't have a rule that matches the collection.

I believe this addresses the concern but it's worth verifying. Additional context:

>Basically the issue is at the moment, if the dataVersion or collection in the CNM does not match the Rules, the CNM is lost with no indication it ever existed. The ticket I created is asking that the logic that checks the CNM to the rule logs a message if there was no match.

The CNM portion of this is _I believe_ not important to the fix. We're matching rules to collections in a Cumulus workflow using the CMA. Definitely worth verifying my assumptions here.

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


AssiduousPirate (approved):

jennyhliu (changes requested):
CNM message can also be delivered via kinesis, sns which is handled by message-consumer lambda

npauzenga:
> CNM message can also be delivered via kinesis, sns which is handled by message-consumer lambda

Thanks @jennyhliu! 

That's a good point. Thankfully the `message-consumer` lambda is using this same rulesHelper as `sqs-message-consumer` so I believe this change should support both Lambdas.

In `message-consumer` we call the helper here: https://github.com/nasa/cumulus/blob/master/packages/api/lambdas/message-consumer.js#L150

That then calls `collectionRuleMatcher`, which logs the mismatch.

How does that sound?

npauzenga:
> update changelog



> > so I believe this change should support both Lambdas.
> 
> Yeah, thanks. Please update the changelog to refer to the method updated.
> 
> Tested kinesis CNM record which doesn't match the rule: 2023-08-22T16:33:58.966Z 37db5cff-f12c-4a36-8c32-023bcbb64afe INFO { "level": "info", "message": "Rule collection name - L2_HR_PIXC_test-jl-rds-tf-KinesisTestTrigger-1692721924188 - or Rule collection version - 000 - does not match collection - {"type":"kinesis","name":"L2_HR_PIXC_test-jl-rds-tf-KinesisTestTrigger-1692721924188","version":"009","sourceArn":"arn:aws:kinesis:us-east-1:596205514787:stream/jl-rds-tf-KinesisTestTrigger-1692721924188-KinesisTestTriggerStream"}", "sender": "@cumulus/rulesHelpers", "timestamp": "2023-08-22T16:33:58.966Z" }

Ah of course. Thanks @jennyhliu! Here's the updated CL: https://github.com/nasa/cumulus/pull/3462/commits/aaa758d6f65a13c7766d173f760a00e3d3c0f5b0

jennyhliu (approved):
👍

PR #3481, opened 2023-09-13
CUMULUS-3218: Fix ingest lock count issue

jennyhliu:

Addresses [CUMULUS-3218: LPDAAC. Lock count issue with GlobalConnectionLimit](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3218)

## Changes

* ...

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


packages/ingest/src/lock.ts
```diff
@@ -81,54 +86,56 @@ async function addLock(
...
 ): Promise<boolean> {
-  if (provider.globalConnectionLimit === undefined) {
+  const { globalConnectionLimit, maxDownloadTime } = provider;
+  if (globalConnectionLimit === undefined) {
     return true;
   }
 
   // Fail if lock is not removed after 270 tries.
   if (counter > 270) {
+    log.debug(`The "${provider.id}" provider no lock available after ${counter} retries`);
```
botanical, line 125:
```suggestion
    log.debug(`The "${provider.id}" provider has no lock available after ${counter} retries`);
```

packages/ingest/src/lock.ts
```diff
@@ -14,63 +14,68 @@ export interface Lock {
...
 
 /**
-* Checks all locks and removes those older than five minutes. Returns a count
-* of locks that are not older than five minutes.
-*
-* @param {Object} bucket - The AWS S3 bucket with the locks to check
-* @param {Array} locks - The list of locks in the bucket
-* @returns {integer} - Number of locks remaining in bucket
-**/
+ * Checks all locks and removes those older than five minutes. Returns a count
```
botanical, line 17:
```suggestion
 * Checks all locks and removes those older than lock retention time. Returns a count
```

packages/ingest/src/lock.ts
```diff
@@ -14,63 +14,68 @@ export interface Lock {
...
+ * of locks that are not older than configured retention period or 5 minutes.
+ *
+ * @param {object} bucket - The AWS S3 bucket with the locks to check
+ * @param {Array} locks - The list of locks in the bucket
+ * @param {number} retentionTimeInSecond - lock retention time in second, default is 300
```
botanical, line 22:
```suggestion
 * @param {number} retentionTimeInSeconds - lock retention time in seconds, default is 300
```

packages/ingest/src/lock.ts
```diff
@@ -14,63 +14,68 @@ export interface Lock {
...
+ * @returns {integer} - Number of locks remaining in bucket
+ */
 export async function checkOldLocks(
   bucket: string,
-  locks: Lock[] = []
+  locks: Lock[] = [],
+  retentionTimeInSecond: number = 300
 ): Promise<number> {
-  const fiveMinutesAgo = Date.now() - (5 * 60 * 1000);
+  const expirationTimestamp = Date.now() - (retentionTimeInSecond * 1000);
```
botanical, line 30:
```suggestion
  const expirationTimestamp = Date.now() - (retentionTimeInSeconds * 1000);
```

packages/ingest/src/lock.ts
```diff
@@ -14,63 +14,68 @@ export interface Lock {
...
-*
-* @param {Object} bucket - The AWS S3 bucket to check
-* @param {string} providerName - The provider name
-* @returns {integer} - Number of current locks in the bucket
-**/
+ * Counts the number of locks in a bucket.
+ *
+ * @param {object} bucket - The AWS S3 bucket to check
+ * @param {string} providerName - The provider name
+ * @param {number} retentionTimeInSecond - lock retention time in second, default is 300
```
botanical, line 57:
```suggestion
 * @param {number} retentionTimeInSeconds - lock retention time in seconds, default is 300
```

packages/ingest/src/lock.ts
```diff
@@ -14,63 +14,68 @@ export interface Lock {
...
+  providerName: string,
+  retentionTimeInSecond?: number
 ): Promise<number> {
   const locks = await listS3ObjectsV2({
     Bucket: bucket,
     Prefix: `${lockPrefix}/${providerName}`,
   });
 
-  return checkOldLocks(bucket, locks);
+  return checkOldLocks(bucket, locks, retentionTimeInSecond);
```
botanical, line 70:
```suggestion
  return checkOldLocks(bucket, locks, retentionTimeInSeconds);
```

packages/ingest/src/lock.ts
```diff
@@ -14,63 +14,68 @@ export interface Lock {
...
 }
 
 async function addLock(
   bucket: string,
   providerName: string,
-  filename: string
+  granuleId: string
```
botanical, line 76: 👍

packages/ingest/src/lock.ts
```diff
@@ -81,54 +86,56 @@ async function addLock(
...
-    globalConnectionLimit?: number
+    globalConnectionLimit?: number,
+    maxDownloadTime?: number,
   },
-  filename: string,
+  granuleId: string,
   counter = 0
 ): Promise<boolean> {
-  if (provider.globalConnectionLimit === undefined) {
+  const { globalConnectionLimit, maxDownloadTime } = provider;
```
botanical, line 118: I think this line should go after this check
```
if (globalConnectionLimit === undefined) {
    return true;
  }
```

packages/api/lib/schemas.js
```diff
@@ -625,6 +625,11 @@ module.exports.provider = {
       title: 'Concurrent Connection Limit',
       type: 'integer',
     },
+    maxDownloadTime: {
+      title: 'Maximum download time in seconds for all granule files on a sync granule task',
+      type: 'integer',
+      description: 'The timeout is used together with globalConnectionLimit to limit concurrent downloads',
+    },
```
botanical, line 632: Nice! Should we update https://github.com/nasa/cumulus/blob/master/docs/configuration/data-management-types.md?plain=1#L63 as well?

packages/ingest/src/lock.ts
botanical: Actually, I am not sure anymore but am wondering if no globalConnectionLimit is provided, then below at 
```
if (count >= globalConnectionLimit) {
    log.debug(`The "${provider.id}" provider's globalConnectionLimit of "${provider.globalConnectionLimit}" has been reached.`);
    // wait for 5 second and try again
    await sleep(5000);
    return proceed(bucket, provider, granuleId, counter + 1);
  }
  ```

will it always be false since anything compared to `undefined` is false? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Greater_than#comparing_boolean_null_undefined_nan

packages/api/lib/schemas.js
jennyhliu: Added

packages/ingest/src/lock.ts
botanical: Ahh that makes sense

botanical (approved):
Changes look good. Thanks, @jennyhliu 🙇 ! I'm sure you're already doing it but let's also update https://github.com/nasa/cumulus-api/blob/master/content/providers.md as well

PR #3482, opened 2023-09-22
CUMULUS-3377: Add configurable metrics integration SNS access policy to report topics

botanical:

Addresses [CUMULUS-3377: Cloud Metrics would like a flag for metrics integrations that will update SNS topic access policies](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3377)

## Changes

* Added new variable to enable policy addition to SNS topics used for metrics reporting. 

## Testing
* Local and cross account testing has been done and is successful
* Testing has been done for the scenario that a user does not provide value for `report_sns_topic_subscriber_arns`

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


example/cumulus-tf/terraform.tfvars.example
```diff
@@ -126,3 +126,11 @@ additional_log_groups_to_elk = {
...
 }
 */
+
+# Optional, configure if needed. Adds policies to SNS report topics.
+#  If a Principal is specified as just an AWS account ID rather than an ARN,
+#    AWS silently converts it to the ARN for the root user,
+#    causing future terraform plans to differ. To avoid this problem,
+#    specify the full ARN.
+# add_report_topic_policy = true
+# metrics_account_id = ["arn:aws:iam::12345678:user/FederatedUser"]
```
jennyhliu, line 136: Suggest to have a more generic name like `report_sns_topic_subscriber_arns`

example/cumulus-tf/terraform.tfvars.example
```diff
@@ -126,3 +126,11 @@ additional_log_groups_to_elk = {
   "LogDescriptor" = "log_group_name"
 }
 */
+
+# Optional, configure if needed. Adds policies to SNS report topics.
+#  If a Principal is specified as just an AWS account ID rather than an ARN,
+#    AWS silently converts it to the ARN for the root user,
+#    causing future terraform plans to differ. To avoid this problem,
+#    specify the full ARN.
+# add_report_topic_policy = true
```
jennyhliu, line 135: Is it necessary to make it configurable?  sns topic has a policy (default or from terraform)

tf-modules/archive/ingest-reporting.tf
```diff
@@ -4,20 +4,266 @@ resource "aws_sns_topic" "report_executions_topic" {
...
+    principals {
+      type        = "AWS"
+      identifiers = ["*"]
+    }
+
+    resources = [
+      aws_sns_topic.report_executions_topic.arn,
+    ]
+
+    sid = "__default_statement_ID"
```
jennyhliu, line 64: Should we give  `sid` and `policy_id ` a real name?

example/cumulus-tf/terraform.tfvars.example
jennyhliu: Please check when the value is `[]`, the subscription from the owner account still works.
Also verify it works for multiple subscriber arns.

example/cumulus-tf/terraform.tfvars.example
botanical: @jennyhliu Yeah the topic will have a default policy, that's true. I added a configuration option because the ticket asked for it https://bugs.earthdata.nasa.gov/browse/CUMULUS-3377 
> We think it would be helpful for cumulus to add a flag for metrics integrations that when set to true, will update the access policy on the sns topics to match what we have in our documentation

But maybe since it is required for them to configure their metrics integration, maybe we make it required? Not sure 😅

jennyhliu: I meant we will create the policy anyway, w or w/o external subscribers

tf-modules/archive/ingest-reporting.tf
jennyhliu: Keep sid    = "__default_statement_ID" since it's default statement.  Add a 'sid' for custom statement?

tf-modules/archive/ingest-reporting.tf
```diff
@@ -4,20 +4,266 @@ resource "aws_sns_topic" "report_executions_topic" {
...
+    ]
+  }
+  policy_id = "__default_policy_ID"
+
+  statement {
+    actions = [
+      "SNS:Subscribe",
+      "SNS:SetTopicAttributes",
+      "SNS:RemovePermission",
+      "SNS:Receive",
```
jennyhliu, line 104: The default access policy doesn't have `SNS:Receive`

example/cumulus-tf/terraform.tfvars.example
botanical: Oh I see, yes the policy is created even without external subscribers. If you look at a topic's access policy that doesn't have these changes, you will notice it gets a default policy like:

```
{
  "Version": "2008-10-17",
  "Id": "__default_policy_ID",
  "Statement": [
    {
      "Sid": "__default_statement_ID",
      "Effect": "Allow",
      "Principal": {
        "AWS": "*"
      },
      "Action": [
        "SNS:GetTopicAttributes",
        "SNS:SetTopicAttributes",
        "SNS:AddPermission",
        "SNS:RemovePermission",
        "SNS:DeleteTopic",
        "SNS:Subscribe",
        "SNS:ListSubscriptionsByTopic",
        "SNS:Publish"
      ],
      "Resource": "arn:aws:sns:us-east-1:redacted:redacted-report-pdrs-topic",
      "Condition": {
        "StringEquals": {
          "AWS:SourceOwner": "redacted"
        }
      }
    }
  ]
}
```

tf-modules/archive/ingest-reporting.tf
botanical: Thank you! https://github.com/nasa/cumulus/pull/3482/commits/c51dc64f18f498676ca89d40b1b71ad5e84fd33d#diff-86e3ca943bb3976971bbb0a4e15dd36e19ed4ed9334be413f419d11523b29588R150-R156

tf-modules/archive/ingest-reporting.tf
botanical: What do you think of `sid = "__custom_sns_statement_ID"`? I updated the policy document order as well https://github.com/nasa/cumulus/pull/3482/commits/c51dc64f18f498676ca89d40b1b71ad5e84fd33d#diff-86e3ca943bb3976971bbb0a4e15dd36e19ed4ed9334be413f419d11523b29588R150-R156

example/cumulus-tf/terraform.tfvars.example
botanical: `[]` will now work, I updated the default value for the variable to be `[*]` https://github.com/nasa/cumulus/pull/3482/commits/c51dc64f18f498676ca89d40b1b71ad5e84fd33d#diff-923ccc457843eaf76dcda5594381ce2124f7cf184108342978c10da459c88aecR488 so if a user supplies `report_sns_topic_subscriber_arns = []` in their `terraform.tfvars`, the deployment will still succeed and use the default

tf-modules/archive/variables.tf
```diff
@@ -382,7 +382,7 @@ variable "add_report_topic_policy" {
   default = false
 }
 
-variable "metrics_account_id" {
+variable "report_sns_topic_subscriber_arns" {
   type = list
-  default = []
+  default = ["*"]
```
jennyhliu, line 387: This will allow anonymous user access

jennyhliu: I think you can use dynamic statement, like in
https://github.com/nasa/cumulus/blob/master/tf-modules/cumulus_distribution/iam.tf#L70
so this way, the policy is always created, and the statement for subscriber is dynamic

jennyhliu (approved):
Nice work👍

PR #3490, opened 2023-10-04
CUMULUS-3286:Handle get temporal info from non-existing CMR file

jennyhliu:

Addresses [CUMULUS-3286: Alter ORCA's granule record update behavior](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3286)

## Changes

* ...

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


packages/cmrjs/src/cmr-utils.js
```diff
@@ -1098,6 +1099,11 @@ async function getGranuleTemporalInfo(granule) {
 
   const cmrFilename = getS3UrlOfFile(cmrFile[0]);
 
+  if (!(await s3ObjectExists(parseS3Uri(cmrFilename)))) {
```
botanical, line 1102: Taking a closer look at `s3ObjectExists` do we need to await `headObject` in that function https://github.com/nasa/cumulus/blob/master/packages/aws-client/src/S3.ts#L191-L197 ? Can we add unit test to https://github.com/nasa/cumulus/blob/master/packages/aws-client/tests/test-S3.js?

botanical (approved):
Looks good

PR #3514, opened 2023-11-13
Update CloudWatchEvents to aws v3

npauzenga:

This updates CloudWatchEvents to the new AWS SDK. Note that this is the minimal set of changes to make it work, not the larger v3 syntax changes. Open to do those here as well.

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


packages/aws-client/src/CloudwatchEvents.ts
```diff
@@ -17,7 +18,7 @@ import { cloudwatchevents } from './services';
 export const putEvent = (
   name: string,
   schedule: string,
-  state: string,
+  state: RuleState,
```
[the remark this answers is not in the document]
npauzenga: Yeah, the change here is due to the AWS SDK enforcing a stricter typing on this argument. Previously it was a generic `string` but now they're looking for the `RuleState` type which, as you noted, is just
```
RuleState: {
    readonly DISABLED: "DISABLED"
    readonly ENABLED: "ENABLED";
}
```
This was caught by the TS compiler as an argument of type `string` is no longer valid.

example/spec/serial/ingestPdrWithNodeNameFtpSpec.js
```diff
@@ -81,14 +81,14 @@ describe('Ingesting from PDR', () => {
       pdrFilename = `${testSuffix.slice(1)}_${origPdrFilename}`;
       provider = { id: `s3_provider${testSuffix}` };
       testDataFolder = createTestDataPath(testId);
-
+      console.log('line 84');
```
botanical, line 84: Sorry to come into this PR unsolicited but can we remove the console logs? 😅 
```suggestion
```

example/spec/serial/ingestPdrWithNodeNameFtpSpec.js
```diff
@@ -119,13 +124,21 @@ describe('Ingesting from PDR', () => {
       } catch (error) {
         console.log(error);
       }
+      console.log('line 127');
+      console.log('testData::::', testData);
 
       const { newGranuleId, filePaths } = JSON.parse(new TextDecoder('utf-8').decode(testData.Payload));
+      console.log('payload:::', JSON.parse(new TextDecoder('utf-8').decode(testData.Payload)));
```
botanical, line 131:
```suggestion
```

example/spec/serial/ingestPdrWithNodeNameFtpSpec.js
```diff
@@ -119,13 +124,21 @@ describe('Ingesting from PDR', () => {
       } catch (error) {
         console.log(error);
       }
+      console.log('line 127');
+      console.log('testData::::', testData);
```
botanical, line 128:
```suggestion
```

example/spec/serial/ingestPdrWithNodeNameFtpSpec.js
```diff
@@ -138,7 +151,8 @@ describe('Ingesting from PDR', () => {
           { old: 'MOD09GQ.A2016358.h13v04.006.2016360104606', new: newGranuleId },
         ]
       );
-
+      console.log('bucketResp:::', bucketResp);
+      console.log('line 141::::'); // not getting here
```
botanical, line 155:
```suggestion
```

PR #3543, opened 2024-01-10
Cumulus 3540

etcart:
create a master parallel branch which removes reliance on CMR in integration tests, allowing us to merge work with reasonable confidence that it is functional before CMR is fixed
Addresses CUMULUS-3540
## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


example/spec/parallel/ingestGranule/IngestGranuleSuccessSpec.js
etcart, line 1: this turns off the tests directly checking if CMR has published (and replaces them with a dummy test to minimize changes)

PR #3552, opened 2024-01-19
[CUMULUS-3560] Address intermittent unit tests

npauzenga:
The goal here is to fix intermittent unit test failures. Previously we were seeing a very unhelpful `lerna ERR!` in @cumulus/api without any failed assertions. It also appeared that the stdout was swallowed by our version of Lerna. 

This PR updates Lerna to a later version in an attempt to get better errors. In doing so it _appears_ the failures have stopped. I have seen ~8 consecutive unit test runs without failures. This could be pure luck but it seems consistent. 

I think we can merge this PR because the Lerna upgrade is worthwhile in any case. If failures re-emerge we'll need to try something else (in addition to these changes).


package.json
```diff
@@ -24,7 +24,7 @@
     "test": "lerna run --ignore @cumulus/cumulus-integration-tests --concurrency 1 test",
     "test:ci": "./scripts/run_ci_unit_coverage.sh",
     "test:coverage": "lerna run --ignore @cumulus/cumulus-integration-tests --concurrency 1 test:coverage",
-    "coverage": "lerna run --ignore @cumulus/cumulus-integration-tests --concurrency 1 --since master coverage -- -- ",
```
npauzenga, line 27: This apparently changed in the latest version of Lerna. This was the syntax to pass an argument to child processes (python).

etcart (approved):
validated that the desired functionality is still there witht hese changes, looks great

PR #3575, opened 2024-02-06
Cumulus 3456 dla clarity

etcart:
**Summary:**
updates to the DeadLetterArchive to make debugging and improvements easier. including captured errors in logged DLA records
Addresses [CUMULUS-3456](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3456)

## Changes

* sf-event-sqs-to-db-records lambda captures errors and saves them in the "error" field in SQS message
* write-db-dlq-records-to-s3 lambda captures cumulusError field in SQS message and saves it to s3 record
* write-db-dlq-records-to-s3 lambda also parses and brings out to first layer other elements
    * collection
    * granules
    * execution
    * stateMachine
## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


packages/api/lambdas/sf-event-sqs-to-db-records/index.js
```diff
@@ -121,6 +121,72 @@ const writeRecords = async ({
...
+  let granule;
+  try {
+    const cumulusMessage = await getCumulusMessageFromExecutionEvent(executionEvent);
+    collection = extractCollection(cumulusMessage) || 'none';
+    granule = extractGranules(cumulusMessage) || 'none';
+  } catch {
+    collection = 'none';
+    granule = 'none';
+  }
+
```
etcart, line 179: the intention with this is to allow backwards compatibility by not changing fields that are already there, but adding new fields that a user *should* be able to just ignore if they're not interested.

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
```diff
@@ -121,6 +121,72 @@ const writeRecords = async ({
   });
 };
 
+function extractCollection(cumulusMessage) {
```
Jkovarik, line 124: Couple of comments: 

Docstrings with typing please https://wiki.earthdata.nasa.gov/pages/viewpage.action?pageId=292913776 are needed in plain JS. 

That said, just use ?. in the main method and remove all these.  

(also we should be always using  const = () => {} as the function syntax and avoid `function foo`, as const prevents accidental mutation (also for style consistency / format reasons, generally, although that's a change to be made wholesale here).

etcart:
I'm interested in the conversation about reformatting to be SQL friendly. what would that mean in this case? i've tried to follow a "don't break what people are already using" paradigm, but we can't be chained to the first pass at formatting forever.

do you think it's valid to capture the desired data without changing the format as a part of a version minor change and then do a breaking format change that rearranges this as apart of a version major change?

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
etcart: I forgot about the ? syntax, much cleaner

packages/api/lambdas/write-db-dlq-records-to-s3.js
```diff
@@ -31,7 +33,7 @@ async function handler(event) {
     await s3PutObject({
       Bucket: process.env.system_bucket,
       Key: `${process.env.stackName}/dead-letter-archive/sqs/${s3Identifier}.json`,
-      Body: sqsMessage.body,
+      Body: JSON.stringify(sqsMessage),
```
chuckwondo, line 36: Would you mind also adding the following?
```
ContentType: 'application/json'
```
Without this, AWS defaults to `application/octet-stream`, which prevents users in the AWS Console from being able to open JSON files in the browser. Instead, the browser forces the user to download the file, which becomes rather annoying.

This is true anywhere in the Cumulus codebase that writes JSON files (such as CMA messages written to S3).

Perhaps I should create a Cumulus issue requesting that `ContentType` be added to every call to `s3PutObject`?

Jkovarik (commented):
Adding a bit more review commentary, nits included, with apologies given it's work in progress if they're obvious.

etcart:
how do we feel about the structure of messages (SQL-ification aside) and data points captured?

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
```diff
@@ -121,6 +121,35 @@ const writeRecords = async ({
...
+
+  let cumulusMessage;
+  try {
+    cumulusMessage = await getCumulusMessageFromExecutionEvent(executionEvent);
+  } catch {
+    cumulusMessage = null;
+  }
+  const collection = cumulusMessage?.meta?.collection?.name || 'unknown';
+  const granules = cumulusMessage?.payload?.granules?.map((granule) => granule?.granuleId || 'unknown') || 'unknown';
+  return {
```
Jkovarik, line 143: So - `message` is currently of type `@cumulus/types/CumulusMessage`.   I *think* it's ok to change it to a different type (and noting this isn't yet typescripted) to add more fields, but there's a user contract here with respect to what *they* are expecting from a Cumulus message. 

Is there a good reason this couldn't/shouldn't be it's own object like 


{ 
   message: <CumulusMessage>
   <other things>:
}

Other than it requires changes to anything parsing the DLQ message?

Jkovarik (commented):
Comments on the message portion.   Thanks for your patience.

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
Jkovarik: As pointed out in slack, ^ isn't correct.    Based on the typings from the method here, it should be `EventBridgeEvent`. 

The concern here really is that this update is great, but because there are format assumptions inherent in the <thing> that's being put into the queue, we need to be sure that the downstream typings are updated (it's really unfortunate this isn't typed, yet) and that contracts are honored.     Apologies for the misdirection on the CumulusMessage type.

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
```diff
@@ -2,15 +2,91 @@
...
+
+    const parsed = JSON.parse(fileBody);
+    expect(parsed.stateMachine).toEqual('unknown');
+    expect(parsed.collection).toEqual('unknown');
+    expect(parsed.execution).toEqual('unknown');
+    expect(parsed.granules).toEqual('unknown');
+    expect(parsed.error).toEqual('CumulusMessageError: getMessageWorkflowStartTime on a message without a workflow_start_time');
+  });
+});
 describe('When a bad record is sent on the DLQ', () => {
```
etcart, line 90: should this (the old integration test) be removed? my intention was to make it redundant with the first, but it's not identical

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-index.js
```diff
@@ -386,7 +386,7 @@ test.serial('Lambda sends message to DLQ when writeRecords() throws an error', a
       WaitTimeSeconds: 10,
     });
   const dlqMessage = JSON.parse(Messages[0].Body);
```
etcart, line 388: new data has been added to the dlq message since this test was written, so to check for what it indents, change to asking for a superset of k/vs

CHANGELOG.md
```diff
@@ -7,14 +7,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 
 ### Breaking Changes
```
etcart, line 9: not sure if these should be set as "breaking"
it changes the structure of the message, but does so in as "non-breaking" a way as possible, adding new fields rather than changing anything about existing fields

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-index.js
```diff
@@ -484,3 +484,124 @@ test.serial('writeRecords() discards an out of order message that has an older s
...
+  t.is(numberOfMessagesAvailable, 1);
+  t.is(numberOfMessagesNotVisible, 0);
+  const { Messages } = await sqs()
+    .receiveMessage({
+      QueueUrl: t.context.queues.deadLetterQueueUrl,
+      WaitTimeSeconds: 10,
+      MaxNumberOfMessages: 3,
+    });
+
+  // console.log(Messages[1]);
```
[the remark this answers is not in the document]
etcart: removed

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-index.js
```diff
@@ -484,3 +484,124 @@ test.serial('writeRecords() discards an out of order message that has an older s
...
```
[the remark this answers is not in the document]
etcart: also removed

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
etcart: as of now, the DLQ message changes have been minimalized to capturing just the triggering error
the write-db-dlq-records-to-s3 lambda now expects (but defaults if not found) that error string and also parses out other data from the message body to put it at the top level.
this should be very similar, but it turned out once I investigated further that write-db-dlq-records-to-s3 was nesting the contents of the sqs message one layer deeper, such that I had to pull everything out a layer anyway.

packages/api/lambdas/write-db-dlq-records-to-s3.js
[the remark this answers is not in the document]
etcart: I'd like to revisit this with the current changes: I'm still using JSON.stringify, but no longer pulling out body here (I'm relying on parseSQSMessageBody to do it). but parseSQSMessageBody isn't stringifying something so I'm confused. let me know if the new version addresses your concerns

CHANGELOG.md
Jkovarik: I think this falls squarely into non-breaking.

Jkovarik (commented):
A few initial comments:

packages/api/tests/lambdas/test-write-db-dlq-records-to-s3.js
```diff
@@ -109,14 +109,233 @@ test.serial('write-db-dlq-records-to-s3 throws error if system bucket is not def
...
+          executionArn: 'execArn',
+        },
+        collection: 'aName',
+        granules: ['a', 'b'],
+        execution: 'execArn',
+        stateMachine: 'unknown',
+      },
+    },
+  ];
+  for (let i = 0; i < messages.length; i += 1) {
```
Jkovarik, line 334: Nit: consider removing iterator
```suggestion
  const results = await Promise.all(
    messages.map((message) => formatCumulusDLAObject(message.mangled))
  );
  results.foreach((result, index) => {
    t.deepEqual(result, messages[index].expected);
  });
});

```

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
Jkovarik: Discussed in huddle, and agreed should be removed as duplicative/meaningless

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
```diff
@@ -2,15 +2,91 @@
...
+      // fetch key for cleanup
+      const listResults = await listS3ObjectsV2({
+        Bucket: systemBucket,
+        Prefix: prefix,
+      });
+      failedMessageS3Key = listResults[0].Key;
+    } catch (error) {
+      fail(`Did not find expected S3 Object: ${error}`);
+    }
+    const s3Object = await getS3Object(systemBucket, failedMessageS3Key);
```
Jkovarik, line 79: This is deprecated, consider updating

CHANGELOG.md
```diff
@@ -8,15 +8,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
-  - Removed unused CloudWatch Logs AWS SDK client. This change removes the CloudWatch Logs
-    client from the `@cumulus/aws-client` package.
-- **CUMULUS-2890**
-  - Removed unused CloudWatch AWS SDK client. This change removes the CloudWatch client
-    from the `@cumulus/aws-client` package.
-    
-### Changed
 
+### Changed
+- **CUMULUS-3456**
```
Jkovarik, line 13: CL fomatting needs a fix

CHANGELOG.md
```diff
@@ -26,12 +22,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
    - Update incorrect docstring
 
 ### Changed
```
Jkovarik, line 24: Duplicate Changed header

Jkovarik (changes requested):
This work looks really good, just a couple of suggestions and typing concerns to resolve!

packages/message/src/DeadLetterMessage.ts
```diff
@@ -24,26 +57,25 @@ type UnwrapDeadLetterCumulusMessageReturnType = (
...
+      );
+    }
+    if (isCumulusMessageLike(messageBody)) {
+      return messageBody as CumulusMessage;
     }
-    // StepFunctionEventBridgeEvent case
-    const unwrappedMessageBody = await getCumulusMessageFromExecutionEvent(messageBody);
-    return await unwrapDeadLetterCumulusMessage(unwrappedMessageBody);
+    throw new TypeError('DeadLetter CumulusMessage in unrecognized format');
   } catch (error) {
```
charleshuang80, line 79: 1. so this error is an unwrap error, so doesn't need to be added as part of the DLA message?
2. I assume this is the case, but wanted to check - if we encounter this unwrap error are we going to see the full message in the logs so we can understand why the unwrap error was encountered?

etcart: this shouldn't appear in DLA, this is meant to remove duplicate code and clarify something. as previously written:
 - if message is in an unrecognized format (found by first not entering "if 'cumulus_meta' in obj" and "if 'Body' in obj", and then asking "is 'detail' not in obj?)
 - if an error is thrown somewhere in there
 
 this instead checks for truth in "if 'detail' in obj" and therefore treats it appropriately, and if it doens't fit into any of the buckets, treats that as an error. it should be functionally unchanged, but easier to read what's happening

packages/message/src/DeadLetterMessage.ts
```diff
@@ -24,26 +57,25 @@ type UnwrapDeadLetterCumulusMessageReturnType = (
...
+    if (isSQSRecordLike(messageBody)) {
       // AWS.SQS.Message/SQS.Record case
       const unwrappedMessageBody = parseSQSMessageBody(
-        messageBody
-      ) as CumulusMessage;
+        messageBody as SQSRecord | AWS.SQS.Message
+      ) as StepFunctionEventBridgeEvent;
       return await unwrapDeadLetterCumulusMessage(unwrappedMessageBody);
     }
-    if (!('detail' in messageBody)) {
```
etcart, line 40: originally *this* is checking for the "didn't fall into any of our prepared cases" edge case
while line 45 is the "if it's an event body" case

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-index.js
```diff
@@ -484,3 +484,124 @@ test.serial('writeRecords() discards an out of order message that has an older s
   )).status);
   t.is('completed', (await pdrPgModel.get(testKnex, { name: pdrName })).status);
 });
+
+test.serial('Lambda captures metadata on error', async (t) => {
```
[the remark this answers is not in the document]
etcart: this has been removed in the latest version

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-index.js
```diff
@@ -484,3 +484,124 @@ test.serial('writeRecords() discards an out of order message that has an older s
...
+      error: 'Error: Intentional failure: test case',
+      granule: [t.context.granule.granuleId],
+      execution: t.context.executionArn,
+      collection: toCamel(t.context.collection.name),
+      stateMachine: t.context.stateMachineArn,
+    }
+  );
+});
+
+test.serial('Lambda reports "none" on metadata collection failure', async (t) => {
```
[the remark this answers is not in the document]
etcart: removed in latest version

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
```diff
@@ -121,6 +121,64 @@ const writeRecords = async ({
...
+
+function extractStateMachine(executionEvent) {
+  try {
+    return executionEvent.detail.stateMachineArn;
+  } catch {
+    return null;
+  }
+}
+
+async function formatCumulusDLQMessage(message, error) {
```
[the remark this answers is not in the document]
etcart: removed (to different file)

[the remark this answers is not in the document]
etcart: and with docstring/typing included in the new location

CHANGELOG.md
etcart: moved to changed

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-index.js
etcart: they could be combined, but I kept them separate thinking it would make understanding easier:
'Lambda sends message to DLQ when writeRecords() throws an error' failed: you're not writing to the DLQ
'lambda captures error' failed: you're failing to pass through your error message (correctly)

packages/message/src/DeadLetterMessage.ts
etcart: this one is hard to say in text, I'd be happy to pair to clarify

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
```diff
@@ -145,7 +158,13 @@ const handler = async (event) => {
       return await writeRecords({ ...event, cumulusMessage, knex });
     } catch (error) {
       log.error(`Writing message failed: ${JSON.stringify(message)}`, error);
-      return sendSQSMessage(process.env.DeadLetterQueue, message);
+      return sendSQSMessage(
```
Jkovarik, line 161: This is throwing a typing error due to the parameters in sendSQSMessage being incorrect (the third param is optional)

packages/api/lambdas/write-db-dlq-records-to-s3.js
```diff
@@ -8,7 +8,54 @@ const { s3PutObject } = require('@cumulus/aws-client/S3');
...
+ */
+async function formatCumulusDLAObject(messageBody) {
+  const execution = messageBody?.detail?.executionArn || 'unknown';
+  const stateMachine = messageBody?.detail?.stateMachineArn || 'unknown';
+
+  let cumulusMessage;
+  try {
+    cumulusMessage = await getCumulusMessageFromExecutionEvent(messageBody);
+  } catch {
+    cumulusMessage = null;
```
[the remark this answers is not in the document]
Jkovarik: @etcart  I believe this still needs addressed

packages/api/lambdas/write-db-dlq-records-to-s3.js
```diff
@@ -8,7 +8,54 @@ const { s3PutObject } = require('@cumulus/aws-client/S3');
...
+    collection,
+    granules,
+    execution,
+    stateMachine,
+  };
+}
+/**
+ * Determine execution name from body
+ *
+ * @param {Object} cumulusMessageObject - cumulus message
```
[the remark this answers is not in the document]
Jkovarik: @etcart  I believe this still needs addressed

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-index.js
Jkovarik: Fair enough - there's leeway there in increased maintenance vs clarity.

packages/api/tests/lambdas/test-write-db-dlq-records-to-s3.js
```diff
@@ -109,14 +109,233 @@ test.serial('write-db-dlq-records-to-s3 throws error if system bucket is not def
...
 });
+
+test('formatCumulusDLAObject returns input message intact', async (t) => {
+  const message = {
+    a: 'b',
+  };
+  t.like(await formatCumulusDLAObject(message), message);
+});
+
+test('formatCumulusDLAObject returns details: collection, granules, execution, and stateMachine as found moved to top layer', async (t) => {
```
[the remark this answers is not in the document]
Jkovarik: This should have been noted as a nit.     Up to PR opener to resolve.

packages/api/lambdas/write-db-dlq-records-to-s3.js
```diff
@@ -8,7 +10,99 @@ const { s3PutObject } = require('@cumulus/aws-client/S3');
...
+ *
+ * @typedef {import('@cumulus/types').MessageGranule} MessageGranule
+ * @typedef {{granules: Array<MessageGranule>}} PayloadWithGranules
+*/
+
+/**
+ * @param {unknown} payload
+ * @returns {payload is PayloadWithGranules}
+ */
+function payloadHasGranules(payload) {
```
Jkovarik, line 25: Nice, appreciated this approach to typeguarding in annotated .js

packages/api/lambdas/write-db-dlq-records-to-s3.js
```diff
@@ -8,7 +10,99 @@ const { s3PutObject } = require('@cumulus/aws-client/S3');
...
+    return false;
+  }
+  return true;
+};
 
+/**
+ * Reformat object with key attributes at top level.
+ *
+ * @param {{ [key: string]: any }} messageBody - event bridge event as defined in aws-lambda
+ * @returns {Promise<Object>} - message packaged with metadata or 'unknown' where metadata not found
```
Jkovarik, line 57: With the correct typing, this is no longer correct.  Either fix, or remove/let automagic typing handle it.

Jkovarik (changes requested):
Really good work on the typing stuff, just a few more details to sort.     Thanks for taking the time to get the typing right!

packages/api/lambdas/write-db-dlq-records-to-s3.js
```diff
@@ -8,7 +10,99 @@ const { s3PutObject } = require('@cumulus/aws-client/S3');
...
+    && Array.isArray(payload.granules)
+  );
+}
+/**
+ *
+ * @typedef {import('aws-lambda').EventBridgeEvent} EventBridgeEvent
+*/
+/**
+ * @param {{ [key: string]: any }} event
+ * @returns {event is EventBridgeEvent}
```
charleshuang80, line 38: note: to be discussed as type in arch meeting

CHANGELOG.md
etcart: this should be properly fixed now

packages/api/lambdas/write-db-dlq-records-to-s3.js
```diff
@@ -1,14 +1,90 @@
...
+    stateMachine,
+  };
+}
+/**
+ * Determine execution name from body
+ *
+ * @param {Object} cumulusMessageObject - cumulus message
+ * @returns {string} - <executionName | 'unknown'>
+ */
 function determineExecutionName(cumulusMessageObject) {
```
charleshuang80, line 88: How different is the executionArn from the executionName? If they're the same, I'm wondering why we're grabbing them in different ways. If they're different, I'm wondering why we're grabbing both and if we should consolidate and pick the most useful one. Also thinking we should consolidate by using this method in the above function or eliminating this method and grabbing the info out of the massagedMessage/workedMessage in the handler below. All depends on the answer to the beginning question though.

packages/message/src/DeadLetterMessage.ts
charleshuang80: re-looking at this, I think we never got around to discussing question #2. Maybe we can talk about it.

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
```diff
@@ -1,93 +1,93 @@
...
+                  execution_name: executionName,
+                },
+                a: 'sldkj',
+              }),
+            },
+          }),
+        }],
+      }),
+    });
+    if ($metadata.httpStatusCode < 400) {
```
charleshuang80, line 46: is any code under 400 really ok, or are you actually looking for 20x?

charleshuang80:
finished looking through for now. In general I am ok with it, but have a few questions I just posted. Also, I know there are a few outstanding suggestions for data to pull.

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
etcart: I'm trying to be careful to get errors in the actual tests not here wherever possible so I've made as permissive as reasonable.
but if you think it's a better idea to look for 20x I'm open to it

packages/api/lambdas/write-db-dlq-records-to-s3.js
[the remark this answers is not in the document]
etcart: I believe this is fixed now

packages/api/lambdas/write-db-dlq-records-to-s3.js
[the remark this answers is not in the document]
etcart: I believe this is fixed now

packages/api/lambdas/write-db-dlq-records-to-s3.js
etcart: the difference is small:
in my function the executionArn is what I'm grabbing because it seems more useful to debugging,
int he existing code executionName is used to to name the s3 file being created.
do we risk breaking something if we change the s3 file to use arn vs name? (I assume we do)

packages/api/lambdas/write-db-dlq-records-to-s3.js
```diff
@@ -18,6 +65,13 @@ function determineExecutionName(cumulusMessageObject) {
   }
 }
 
+/**
+ * Lambda handler for saving DLQ reports to DLA in s3
+ *
+ * @param {Object} event - Input payload object
```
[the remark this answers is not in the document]
etcart: I'm not sure, the additional annotation is that it contains one field: Records, requires I define it as an "object"

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
charleshuang80: I'm not too concerned, as I don't know how likely it is, but seems like 10x and 30x would be accepted here. If you're trying to make sure it's not an error code then that makes sense.

packages/api/lambdas/write-db-dlq-records-to-s3.js
charleshuang80: I guess you made the change to check if it would break something or not? Just to be clear, I was not trying to say the change _had to_ be made, I just didn't know the difference and was wondering if it made sense to consolidate. I think the change looks good though as long as nothing breaks.

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
```diff
@@ -1,93 +1,93 @@
...
       }
-    });
+    );
+    const fileBody = await getObjectStreamContents(s3Object.Body);
+
+    const parsed = JSON.parse(fileBody);
+    expect(parsed.stateMachine).toEqual('unknown');
+    expect(parsed.collection).toEqual('unknown');
+    expect(parsed.execution).toEqual('unknown');
+    expect(parsed.granules).toEqual('unknown');
```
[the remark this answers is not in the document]
jennyhliu: If the attributes are not there, should we set them to `null` instead of arbitraray `unknown` string?

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
etcart: exactly, I'm trying to be lenient and just make sure it's not an error code

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
```diff
@@ -120,7 +122,17 @@ const writeRecords = async ({
     testOverrides,
   });
 };
+/**
+ * @typedef {import('aws-lambda').SQSRecord} SQSRecord
+ */
 
+/**
+ * Lambda handler for #TODO
```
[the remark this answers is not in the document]
etcart: finished out fixing this

packages/api/lambdas/write-db-dlq-records-to-s3.js
```diff
@@ -1,14 +1,90 @@
...
+ * @param {{[key: string]: any}} messageBody - event bridge event as defined in aws-lambda
+ * @returns {Promise<Object>} - message packaged with
+ * metadata or 'unknown' where metadata not found
+ * {
+ *   error: <errorString | 'unknown'>
+ *   collection: <collectionName | 'unknown'>
+ *   granules: <[granuleIds, ...] | 'unknown'>
+ *   execution: <executionArn | 'unknown'>
+ *   stateMachine: <stateMachineArn | 'unknown'>
+ *   ...originalAttributes
```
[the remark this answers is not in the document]
etcart: these are both included now

packages/message/src/DeadLetterMessage.ts
etcart: adding in failed messageBody

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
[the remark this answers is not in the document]
etcart: switchign to null

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
```diff
@@ -1,93 +1,93 @@
...
+      {
+        Bucket: systemBucket,
+        Key: failedMessageS3Key,
       }
-    });
+    );
+    const fileBody = await getObjectStreamContents(s3Object.Body);
+
+    const parsed = JSON.parse(fileBody);
+    expect(parsed.stateMachine).toEqual('unknown');
```
charleshuang80, line 87: should some of these be null now?

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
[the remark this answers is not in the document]
jennyhliu: @etcart granules can be null too.

charleshuang80 (dismissed):

charleshuang80 (commented):
Nothing big, a few questions and I think you need to also add the integration test that was modified.

packages/api/lambdas/write-db-dlq-records-to-s3.js
```diff
@@ -1,43 +1,142 @@
...
+    payload instanceof Object
+    && 'granules' in payload
+    && Array.isArray(payload.granules)
+  );
+}
+/**
+ * Reformat object with key attributes at top level.
+ *
+ * @param {DLQRecord | EventBridgeEvent} dlqRecord - event bridge event as defined in aws-lambda
+ * @returns {Promise<Object>} - message packaged with
```
Jkovarik, line 38: Declare type and return please!

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
```diff
@@ -145,7 +165,18 @@ const handler = async (event) => {
...
+        log.error('DeadLetterQueue not configured');
+        return undefined;
+      }
+      return sendSQSMessage(
+        process.env.DeadLetterQueue,
+        {
+          ...message,
+          error: error.toString(),
+        },
+        log
```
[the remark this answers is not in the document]
Jkovarik: That's a good question - functionally it's changing sender metadata from '@cumulus/aws-client/SQS' to '@cumulus/api/lambdas/sf-event-sqs-to-db-records'.   

There's probably a reasonable question re: why that change was made and/or if we want the logger to identify the lambda or the portion of code.     I have opinions, but it feels like something that should be a deliberate change/pattern.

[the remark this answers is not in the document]
Jkovarik: Actually, looking at this the issue is the typing in sendSQSMessage is incorrect, it doesn't have the log override as an optional param.  I'd vote for fixing that over this change, however <scope> is a thing.

packages/aws-client/src/Lambda.ts
```diff
@@ -2,12 +2,26 @@
...
+/**
+ * Bare check for EventBridge shape
+ *
+ * @param {{ [key: string]: any }} event
+ * @returns {event is EventBridgeEvent}
+ */
+export const isEventBridgeEvent = (event: Object): event is StepFunctionEventBridgeEvent => (
+  event instanceof Object
+  && 'detail' in event
+);
```
Jkovarik, line 23: Why is `event instanceof Object` needed here?    Is the intent to guard against someone ignoring typechecking in the raw .js? 

I... think I agree with the defensive intent here, but it might warrant a comment (or unit if that's not already demanded by coverage metrics).

packages/message/src/DeadLetterMessage.ts
```diff
@@ -1,21 +1,47 @@
...
+  && 'cumulus_meta' in message
+);
+
+/**
+ * Bare check for SQS message Shape
+ *
+ * @param {{ [key: string]: any }} message
+ * @returns {message is DLQRecord}
+ */
+export const isSQSRecordLike = (message: Object): message is DLQRecord | SQSRecord => (
```
Jkovarik, line 40: Thinking about isEventBridgeEvent/isSQSRecordLike, and typeguards in general: 

Should this be in the aws-client package?   SQSRecord doesn't seem to map to `@cumulus/message` as a package?

packages/message/src/DeadLetterMessage.ts
```diff
@@ -1,21 +1,47 @@
...
+const isCumulusMessageLike = (message: Object): message is CumulusMessage => (
+  message instanceof Object
+  && 'cumulus_meta' in message
+);
+
+/**
+ * Bare check for SQS message Shape
+ *
+ * @param {{ [key: string]: any }} message
+ * @returns {message is DLQRecord}
```
Jkovarik, line 38: This return type isn't correct.   Also I mentioned this to @npauzenga but I think we need to have a conversation as a team re: duplicative docstrings in .ts - annotating type shouldn't be required if we've done so in the code (and avoids duplicate updates)

packages/api/lambdas/write-db-dlq-records-to-s3.js
```diff
@@ -1,43 +1,142 @@
...
+ *   time: <timestamp(utc) | null>
+ *   status: <status | null>
+ *   collection: <collectionName | null>
+ *   granules: <[granuleIds, ...] | []>
+ *   execution: <executionArn | null>
+ *   stateMachine: <stateMachineArn | null>
+ *   ...originalAttributes
+ * }
+ */
+async function hoistCumulusMessageDetails(dlqRecord) {
```
Jkovarik, line 51: This method looks like it's private-duplicating some of the internal logic of `unwrapDeadLetterCumulusMessage` and/or making some sensible refactors - is there a reason we can't/shouldn't reuse that code and leave it common in the message package,  particularly given the other single use of that call? 

If not, does this have adaquate unit coverage?

packages/message/src/DeadLetterMessage.ts
```diff
@@ -24,30 +50,30 @@ type UnwrapDeadLetterCumulusMessageReturnType = (
...
-  messageBody: UnwrapDeadLetterCumulusMessageReturnType
-): Promise<UnwrapDeadLetterCumulusMessageReturnType> => {
+  messageBody: UnwrapDeadLetterCumulusMessageInputType
+): Promise<UnwrapDeadLetterCumulusMessageInputType> => {
   try {
-    if ('cumulus_meta' in messageBody) {
-      return messageBody;
-    }
-    if ('Body' in messageBody || 'body' in messageBody) {
+    if (isSQSRecordLike(messageBody)) {
```
Jkovarik, line 56: Excellent refactor :thumbsup:

packages/aws-client/src/SQS.ts
```diff
@@ -139,16 +143,25 @@ export const receiveSQSMessages = async (
   return <SQSMessage[]>(messages.Messages ?? []);
 };
 
+/**
+ * Extract SQS message body
+ *
+ * @param {SQSRecord | AWS.SQS.Message} message - SQS message
+ * @param {string} message.body - optional key for message body
+ * @param {string} message.Body - optional key for message body
+ * @returns {StepFunctionEventBridgeEvent | CumulusMessage} an AWS SQS message body
```
Jkovarik, line 152: This annotation appears to be incorrect (see other comment in the PR re: annotations/.ts as well)

packages/aws-client/src/SQS.ts
```diff
@@ -139,16 +143,25 @@ export const receiveSQSMessages = async (
...
+ * parsed into a CumulusMessage or EventBridgeEvent
+ */
 export const parseSQSMessageBody = (
   message: SQSRecord | AWS.SQS.Message
-): { [key: string]: any } =>
+): StepFunctionEventBridgeEvent | CumulusMessage | SQSRecord =>
   JSON.parse(get(message, 'Body', get(message, 'body')) ?? '{}');
 
 /**
  * Delete a given SQS message from a given queue.
```
Jkovarik, line 161: This jsdoc is throwing lint on being incomplete: 


```suggestion
/**
 * Deletes a message from an Amazon SQS queue.
 *
 * @param {string} QueueUrl - The URL of the Amazon SQS queue.
 * @param {string} ReceiptHandle - The unique identifier of the SQS message.
 * @returns {Promise} an AWS SQS response.
 * @throws Will throw an error if deletion fails.
 */```

File: packages/api/lambdas/write-db-dlq-records-to-s3.js (hunk 38 above)
Comment (Jkovarik) [reply packages/api/lambdas/write-db-dlq-records-to-s3.js:51]: Withdraw question re: units now that I've gotten to that file in the review :thumbsup:, the rest stands.

File: packages/api/lambdas/sf-event-sqs-to-db-records/index.js (hunk 34 above)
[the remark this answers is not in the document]
Comment (etcart) [reply packages/api/lambdas/sf-event-sqs-to-db-records/index.js:178]: it wasn't.
I thoguht originally that because the logger wasn't an optional it was required to pass in the log at the scope of whatever's calling it but... on looking further it ismeant to be optional

File: packages/aws-client/src/Lambda.ts (hunk 35 above)
Comment (etcart) [reply packages/aws-client/src/Lambda.ts:23]: precisely: not knowing where this might be used in a js file.

**Review by jennyhliu — COMMENTED** [review]
Should we have migration script/tool to populate the fields for existing DLA messages?

**Comment by etcart** [comment]
> Should we have migration script/tool to populate the fields for existing DLA messages?

some can't be populated (really just error) but for the rest...... maybe yea

File: packages/api/lambdas/write-db-dlq-records-to-s3.js (hunk 38 above)
Comment (etcart) [reply packages/api/lambdas/write-db-dlq-records-to-s3.js:52]: (I think) while this has some structural similarity to unwrap DeadLetterMessage (it does type checking from within a subset of the same types) it is not the same. i'm looking into how they could be joined, but fundamentally:
1 of these extracts a cumulusMessage object from within a questionably structured object
the other extracts some specific data from within a questionably structured object and packages it with the o riginal object

the obvious join would be to refactor this to a function that extracts all of these things and lets you do what you wish with it such that I can either capture that cumulusMessage and do something with it or capture those other pieces of data and do something with them, but prototyping it a little bit that adds a lot of logical complexity while reducing the amount of code by not very much.
at least by the route I've come up with so far

File: packages/aws-client/src/SQS.ts (hunk 41 above)
Comment (etcart) [reply packages/aws-client/src/SQS.ts:176]: I *think* this is corrected now, but no tthe way you said exactly. I'ev corrected the names to be capitalized correctly, but I haven't added the throws line? according to the aws docs this should report errors as http status codes (4xx codes). do you know otherwise?

File: packages/api/lambdas/write-db-dlq-records-to-s3.js (hunk 38 above)
Comment (etcart) [reply packages/api/lambdas/write-db-dlq-records-to-s3.js:52]: discussed with Jonathan and does appear to add too much complexity trying to make this refactor

File: packages/aws-client/src/SQS.ts (hunk 41 above)
Comment (etcart) [reply packages/aws-client/src/SQS.ts:174]: resolved in huddle

File: packages/message/src/DeadLetterMessage.ts (hunk 37 above)
Comment (etcart) [reply packages/message/src/DeadLetterMessage.ts:34]: corrected

File: packages/message/src/DeadLetterMessage.ts (hunk 36 above)
Comment (etcart) [reply packages/message/src/DeadLetterMessage.ts]: resolved

File: packages/api/tests/lambdas/test-write-db-dlq-records-to-s3.js [hunk 42]
```diff
@@ -107,16 +98,765 @@ test.serial('write-db-dlq-records-to-s3 throws error if system bucket is not def
...
+    messages.map((message) => hoistCumulusMessageDetails(message.mangled))
+  );
+  results.forEach((result, index) => {
+    t.deepEqual(result, messages[index].expected);
+  });
 });
 
-test('determineExecutionName returns "unknown" if getExecutionName throws error', (t) => {
-  t.is(determineExecutionName({}), 'unknown');
+test('hoistCumulusMessageDetails handles varying degrees of sqsMessage nestedness', async (t) => {
```
charleshuang80, line 701: Maybe I'm missing or misunderstanding something, but in the test validation from here down, I'm not seeing an `error` in the messages being compared against.

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
Jkovarik: Removing this comment as outdated/work has been done

packages/api/tests/lambdas/test-write-db-dlq-records-to-s3.js
Jkovarik: @Charles I'm pretty sure error is a field in hoistCumulusMessageDetails, and in the fixture as an `error` key , for example on `763`/`724`?

Jkovarik (changes requested):
Great work!    I just did a re-review based on the changes, thanks for addressing my initial concerns.  Just a couple of minor blocking comments and various nits - hit me up if you want to pair/discuss.

packages/message/tests/test-Providers.js
```diff
@@ -45,3 +47,25 @@ test('getMessageProviderId returns undefined if there is no provider ID', (t) =>
   });
   t.is(providerId, undefined);
 });
+test.only('isMessageProvider correctly filters for messageProviders', (t) => {
```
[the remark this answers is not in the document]
etcart: oof! thanks

packages/message/src/Providers.ts
```diff
@@ -1,18 +1,37 @@
...
   meta: {
     provider?: MessageProvider
   }
 };
 
+export const isMessageWithProvider = (
+  obj: any
+): obj is MessageWithProvider => (
+  isCumulusMessageLike(obj)
+  && 'meta' in obj
```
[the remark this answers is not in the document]
etcart: sort of:
that syntax generally doesn't satisfy typescript. in *this* case it will becuase 'isCumulusMessageLike' has the effect of stating that obj.meta.provider exists *buuuut* it doesn't actually check because it's doing that bare check. so while it'll *let* me get away with it, i've kept with the pattern as if it wont

packages/message/tests/test-DeadLetterMessage.js
```diff
@@ -112,3 +112,11 @@ test('unwrapDeadLetterCumulusMessage returns an non-unwrappable message', async
   };
   t.deepEqual(await unwrapDeadLetterCumulusMessage(testMessage), testMessage);
 });
+
+test('isDLQRecordLike correctly filters for DLQ record shaped objects', (t) => {
+  /* isDLQRecordLike first checks if object isSQSRecordLike,
+  * these test do not duplicate those tests
+  */
+  t.false(isDLQRecordLike({ body: '{a: "b"}' })); // object must contain an error attribute
```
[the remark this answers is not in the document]
etcart: done

packages/message/tests/test-Providers.js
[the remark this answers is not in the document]
etcart: done

packages/message/src/Providers.ts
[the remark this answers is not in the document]
Jkovarik: `in this case it will becuase 'isCumulusMessageLike' has the effect of stating that obj.meta.provider exists` -- I'm not sure I follow your meaning, given isCumulusMessageLike only validates `cumulus_meta` is in the object but doesn't have any impact on the typing at all other than the boolean result.    

isMessageProvider is a guard on if it's `MessageProvider`, which should have the correct boolean for `obj?.meta` when `meta` isn't defined (which is `isMessageWithProvider(undefined)` or `false`), and the same for `obj?.meta?.provider` when it's not defined, which is stating all three assertions and eliminating the extra 'is it an object check' on obj.meta (if obj.meta isn't, it passes in the literal and fails `obj instanceof Object`). 

Apologies if I'm being obtuse.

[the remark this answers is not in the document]
etcart: I don't think so, I'll change it

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
```diff
@@ -1,93 +1,171 @@
...
   afterAll(async () => {
     await deleteS3Object(
       systemBucket,
       failedMessageS3Key
     );
   });
+  it('is sent to the DLA and processed to have expected metadata fields', async () => {
+    executionArn = `execution-${randomString(16)}`;
+    const { $metadata } = await lambda().invoke({
+      FunctionName: `${stackName}-sfEventSqsToDbRecords`,
```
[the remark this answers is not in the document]
etcart: I assume you mean putting it in the beforeAll, it trips over the fact that I need to send *two* files. if I put it at the top and send two, I cna't easily distinguish them without parsing on the things I'm testing becuase it auto-appends a random string to ensure they're unique

packages/api/tests/lambdas/test-write-db-dlq-records-to-s3.js
etcart: error is present in those messages, but having up above verified that it *is* passed through, it should simply pass through

Jkovarik (approved):
Thanks for addressing my comments - I'm approving as I don't believe any of the outstanding comments should block merge.     Great work!

charleshuang80 (approved):

PR #3606, opened 2024-03-04
[CUMULUS-3519]: Update SQS and SNS to AWS SDK V3 Syntax

npauzenga:

Addresses [CUMULUS-3519: Update SQS and SNS to sdk v3 syntax](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3519)

## Changes

Updates SNS and SQS logic to use AWS SDK V3 syntax

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -4796,6 +4796,7 @@ test.serial('updateGranuleStatusToQueued() updates granule status in PostgreSQL/
     MaxNumberOfMessages: 2,
     WaitTimeSeconds: 10,
   });
+  console.log('Received Messages::', Messages);
```
npauzenga, line 4799: This test appears to be intermittently failing on main. Adding logging to help debug.

example/package.json
```diff
@@ -71,7 +73,6 @@
     "@cumulus/test-processing": "18.2.0"
   },
   "devDependencies": {
-    "@aws-sdk/client-ecs": "^3.447.0",
```
npauzenga, line 74: Not sure why this was a dev dependency but I think it needs to be a prod dependency.

packages/api/tests/endpoints/test-rules.js
```diff
@@ -1198,27 +1204,27 @@ test.serial('PATCH nullifies expected fields for existing rule in all datastores
...
-            Endpoint: process.env.messageConsumer,
-            SubscriptionArn: randomString(),
-          }],
-        })
-      ),
-      unsubscribe: () => (
-        Promise.resolve()
-      ),
-    });
+  const snsMock = mockClient(awsServices.sns());
```
paulpilone, line 1207: Oh this is neat. I don't know if I knew about `mockClient`.

packages/api/tests/endpoints/test-rules.js
```diff
@@ -1814,8 +1820,8 @@ test.serial('PATCH keeps initial trigger information if writing to PostgreSQL fa
     pgCollection,
   } = t.context;
 
-  const topic1 = await awsServices.sns().createTopic({ Name: randomId('topic1_') });
-  const topic2 = await awsServices.sns().createTopic({ Name: randomId('topic2_') });
+  const topic1 = await await awsServices.sns().send(new CreateTopicCommand({ Name: randomId('topic1_') }));
```
paulpilone, line 1823: Double `await`s here and a few places below this line.

packages/api/tests/lambdas/test-kinesis-consumer.js
```diff
@@ -146,9 +153,11 @@ test.serial('A kinesis message, should publish the invalid record to fallbackSNS
     Records: [validRecord, invalidRecord],
   };
   await handler(kinesisEvent, {}, testCallback);
-  const callArgs = publishStub.getCall(0).args;
-  t.deepEqual(invalidRecord, JSON.parse(callArgs[0].Message));
-  t.true(publishStub.calledOnce);
+
+  const publishCalls = snsMock.commandCalls(PublishCommand);
+
+  t.true(publishCalls.length > 0);
```
paulpilone, line 159: Not sure if it matters too much but should this say `publishCall.length === 1` to better match the previous assertion? There are a few assertions below similar to this one depending on what you decide to do with it.

packages/api/tests/lambdas/test-message-consumer.js
```diff
@@ -110,10 +120,13 @@ test('handler processes records as expected', async (t) => {
...
   t.true(fetchEnabledRulesStub.calledOnce);
 
-  t.true(publishMessageStub.withArgs({
+  const expectedArgs = {
     TopicArn: process.env.FallbackTopicArn,
     Message: JSON.stringify(erroringMessage),
-  }).calledOnce);
+  };
+  const publishCalls = snsMock.commandCalls(PublishCommand);
+  t.true(publishCalls.length > 0);
```
paulpilone, line 128: Another `calledOnce` vs `> 0`.

packages/api/tests/lib/writeRecords/test-write-granules.js
paulpilone: Any update to this comment (over a month old)?

paulpilone (changes requested):
Looks good! There are a few small syntax (specifically the double `await`s) that should probably be cleaned up but otherwise just some general questions. Should be quick to approve.

packages/api/tests/lib/writeRecords/test-write-granules.js
npauzenga: Not really. This was from the intermittent unit test failure investigation, before discovering the logging improvements. I saw this test fail a few times but it's been a while. Removing the log as it's not actively being investigated.
https://github.com/nasa/cumulus/pull/3606/commits/8727d897552c6a455d7d2c9301e28621da82c0c9

example/package.json
```diff
@@ -45,6 +45,8 @@
     ]
   },
   "dependencies": {
+    "@aws-sdk/client-ecs": "^3.447.0",
```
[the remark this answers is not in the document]
npauzenga: Woof. This looks like a bad merge. Fixed here: https://github.com/nasa/cumulus/pull/3606/commits/20af2f795db3ae155dda35348ff71fcc8af9be8f

packages/api/tests/endpoints/test-rules.js
npauzenga: Yeah! That's the new `aws-sdk-client-mock` library we're using. Works great. The only catch is that it stubs all of the service so it's important to explicitly reject calls you're not expecting. e.g. 
```
snsMock
    .onAnyCommand()
    .rejects()
     .on(serviceYouActuallyCareAbout)....
```
That way we're not sending unexpected commands to the mock and having it return `undefined`.

packages/api/tests/endpoints/test-rules.js
npauzenga: 🤦 Well. This is embarrassing. 

https://github.com/nasa/cumulus/pull/3606/commits/ff84c81f44548ffd29b0f653d67c858af46ae1e1

packages/api/tests/lambdas/test-kinesis-consumer.js
npauzenga: Yeah, makes sense. The half-baked idea was to future-proof the test because we really only care that the publish command is called, not how many times (if we were to add another call down the line). But... that doesn't make sense. Better to be explicit and guard against some accidental call that might be introduced. A change in behavior should require a test update.

packages/api/tests/lambdas/test-message-consumer.js
npauzenga: https://github.com/nasa/cumulus/pull/3606/commits/881e7bf190890a675b18f4d947f9d10ed1abc140

I don't think there's a good way to use the sinon `calledOnce` helper unfortunately. The `commandCalls` function returns a list of calls. Not seeing a great way to get the spy directly.

npauzenga:
Feedback addressed. Thanks @paulpilone! Good catches.

paulpilone (approved):
Look great!

PR #3608, opened 2024-03-07
CUMULUS-3610: Updates aws-client ES client to v3

paulpilone:

Addresses [CUMULUS-3610: B - Update ES code to AWS SDK v3](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3610)

## Changes

* Updates `aws-client` ES import to use v3 client. Fixes a few small tests that broke.
* This update does not impact `es-client` since that package does not use the `aws-client` elasticsearch module.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/packages/aws-client/package.json
+++ b/packages/aws-client/package.json
@@ -52,6 +52,7 @@
     "@aws-sdk/client-dynamodb-streams": "^3.447.0",
     "@aws-sdk/client-ec2": "^3.447.0",
     "@aws-sdk/client-ecs": "^3.447.0",
+    "@aws-sdk/client-elasticsearch-service": "^3.447.0",
     "@aws-sdk/client-kinesis": "^3.447.0",
     "@aws-sdk/client-kms": "^3.447.0",
     "@aws-sdk/client-lambda": "^3.447.0",
--- a/packages/aws-client/package.json
+++ b/packages/aws-client/package.json
@@ -52,7 +52,7 @@
     "@aws-sdk/client-dynamodb-streams": "^3.447.0",
     "@aws-sdk/client-ec2": "^3.447.0",
     "@aws-sdk/client-ecs": "^3.447.0",
-    "@aws-sdk/client-elasticsearch-service": "^3.447.0",
+    "@aws-sdk/client-elasticsearch-service": "^3.529.1",
     "@aws-sdk/client-kinesis": "^3.447.0",
     "@aws-sdk/client-kms": "^3.447.0",
     "@aws-sdk/client-lambda": "^3.447.0",
--- a/packages/aws-client/src/services.ts
+++ b/packages/aws-client/src/services.ts
@@ -15,7 +15,7 @@ import { SNS } from '@aws-sdk/client-sns';
 import { STS } from '@aws-sdk/client-sts';
 import { ECS } from '@aws-sdk/client-ecs';
 import { EC2 } from '@aws-sdk/client-ec2';
-import { ElasticsearchServiceClient } from '@aws-sdk/client-elasticsearch-service';
+import { ElasticsearchService } from '@aws-sdk/client-elasticsearch-service';
 
 import awsClient from './client';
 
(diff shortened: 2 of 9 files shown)


packages/aws-client/package.json
```diff
@@ -52,6 +52,7 @@
     "@aws-sdk/client-dynamodb-streams": "^3.447.0",
     "@aws-sdk/client-ec2": "^3.447.0",
     "@aws-sdk/client-ecs": "^3.447.0",
+    "@aws-sdk/client-elasticsearch-service": "^3.529.1",
```
npauzenga, line 55: Is there a reason to not use the same 3.447? There was some discussion about using the most recent vs. keeping them consistent.

packages/aws-client/src/services.ts
```diff
@@ -15,7 +15,7 @@ import { SNS } from '@aws-sdk/client-sns';
 import { STS } from '@aws-sdk/client-sts';
 import { ECS } from '@aws-sdk/client-ecs';
 import { EC2 } from '@aws-sdk/client-ec2';
-import * as AWS from 'aws-sdk';
```
npauzenga, line 18: Nice. That had to be satisfying 😄

packages/aws-client/src/services.ts
```diff
@@ -15,7 +15,7 @@ import { SNS } from '@aws-sdk/client-sns';
...
+import { ElasticsearchService } from '@aws-sdk/client-elasticsearch-service';
```
npauzenga, line 18: Discussed in huddles the difference between services and clients. From their docs https://github.com/aws/aws-sdk-js-v3, this appears correct. The service should allow v2 and v3 interfaces. The naming here is a bit different than other services, which is interesting but appears correct https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/elasticsearch-service/

packages/tf-inventory/tests/inventory.js
```diff
@@ -82,16 +82,17 @@ test.before(() => {
     });
 
   esStub = sinon.stub(es(), 'listDomainNames')
-    .returns({
-      promise: () =>
+    .returns(
+      (
```
npauzenga, line 86: Is this syntax correct? Shouldn't we be ok removing this paren? e.g. 
```
.returns(Promise.resolve({...}));
```

packages/tf-inventory/src/inventory.js
```diff
@@ -81,7 +81,7 @@ async function listAwsResources() {
   ec2Instances = [].concat(...ec2Instances.Reservations.map((e) => e.Instances));
   ec2Instances = ec2Instances.map((inst) => inst.InstanceId);
 
-  let esDomainNames = await es().listDomainNames().promise();
+  let esDomainNames = await es().listDomainNames();
```
npauzenga, line 84: Looks totally right to me. This seems like the most obvious spot for an issue but 🤷

npauzenga (changes requested):
Looks great @paulpilone! A couple quick comments but I don't see any issues. The CI failures are interesting. Curious to see if they're consistent.

packages/aws-client/package.json
paulpilone: I know there's been discussion about this but I'm not sure what benefit that offers us. If you do an `npm install @aws-sdk/client-elasticsearch-service@'^3.447.0' it installs 3.529.1. Given how the v3 SDK is distributed is there a good reason to _not_ install newer versions of the individual clients?

npauzenga: Yeah, I don't think there was a _super_ compelling reason to keep them consistent. Doesn't seem like it matters either way as we'll just pull the latest 3.x anyway 🤷 

Someone gave me this feedback so I'm just blindly passing it along 😆 

Makes sense to me to use the latest available at the time of release.

npauzenga (approved):
This looks great @paulpilone. I'm a little concerned about the CI issues so if there's anything we can do there to build confidence that we're not _introducing_ problems here, that would be great. I don't think we are but some kind of explanation would be great.

PR #3612, opened 2024-03-18
Cumulus 3609 dla format retro script

etcart:
**Summary:** adds a script that users can run to reformat their existing DLAmessages to the new format

Addresses [CUMULUS-3609: update script for new dla structure](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3609)

## Changes

* moves hoistDLAMessage to DeadLetterMessage.ts
* moves previous "scripts" folder to "ci_scripts" for clarity
* adds to "scripts" folder a new package whihc compiles to an executable js file to update DLA records

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


packages/aws-client/src/S3.ts
```diff
@@ -739,34 +740,32 @@ export const uploadS3FileStream = (
...
   const data = await s3().listObjects(params);
-  let contents = data.Contents || [];
+  if (!data.Contents) {
+    return [];
+  }
+  let contents = data.Contents.filter((obj) => obj.Key !== undefined) as Array<ListS3ObjectsOutput>;
   if (skipFolders) {
     // Filter out any references to folders
-    contents = contents.filter((obj) => obj.Key !== undefined && !obj.Key.endsWith('/'));
+    contents = contents.filter((obj) => !obj.Key.endsWith('/'));
```
etcart, line 768: I figure this should filter out undefined even if I haven't asked it to filter out directories

scripts/update-dla-format/main.ts
```diff
@@ -0,0 +1,249 @@
...
+ * @param skip skip files that are already present at the target directory, default false
+ */
+export const updateDLABatch = async (
+  bucket: string,
+  targetDirectory: string,
+  prefix: string,
+  skip: boolean = false
+) => {
+  const sourceDir = await parseS3Directory(prefix);
+  const subObjects = await listS3Objects(bucket, prefix);
```
jennyhliu, line 166: listS3Objects can skip Folders, but it doesn't list all the files.  listS3ObjectsV2 does recursive listing.

lambdas/dla-migration/src/index.ts
```diff
@@ -1,9 +1,162 @@
...
+ * @param targetDirectory
+ * @param prefix
+ * @param skip - skip files that are already present at the target directory, default false
+ */
+export const updateDLABatch = async (
+  bucket: string,
+  targetDirectory: string,
+  sourceDirectory: string
+): Promise<Array<boolean>> => {
+  const out = [];
```
jennyhliu, line 133: We can just add trailing /:
s3Path.replace(//?$/, '/')

lambdas/dla-migration/src/index.ts
```diff
@@ -1,9 +1,162 @@
...
+  const sourceDir = manipulateTrailingSlash(sourceDirectory, true);
+  const targetDir = manipulateTrailingSlash(targetDirectory, true);
+  for await (
+    const objectBatch of listS3ObjectsV2Batch({ Bucket: bucket, Prefix: sourceDir })
+  ) {
```
jennyhliu, line 138: I think we do need to check if the key is a json file in the dla directory. We don't want to process files in subdirectory.
if (s3ObjectKey && s3ObjectKey.endsWith('.json') && s3ObjectKey.lastIndexOf('/') === lastIndexOfDlaPathSeparator)
actually: 
lastIndexOfDlaPathSeparator = dlaPath.length -1;

jennyhliu: I'm confused with sourceDir and targetDir.  They are ${stackName}/dead-letter-archive/sqs/ and ${stackName}/dead-letter-archive/sqs/dateString/

lambdas/dla-migration/src/index.ts
```diff
@@ -1,9 +1,162 @@
...
+    const validKeys = objectBatch.map((obj) => obj.Key);
+    const targetPaths = validKeys.map(
+      (filePath) => filePath.replace(
+        sourceDir,
+        targetDir
+      )
+    );
+
+    const zipped: Array<[string, string]> = zip(validKeys, targetPaths) as Array<[string, string]>;
```
jennyhliu, line 147: I'm confused with the logic here.  Can we add logic for processing each key?

lambdas/dla-migration/src/index.ts
etcart: this is a little bit of a holdover from me wanting to put the outputs to an enitrely new directory.
the sourceDir and targetDir are, by default, the same (YYYY-MM-DD being within targetDir) but they could output to a new place

lambdas/dla-migration/src/index.ts
etcart: the function takes a sourceFile and a targetFile, and this joins them to each other so we can map over them

lambdas/dla-migration/src/index.ts
etcart: will we wan tto process files int eh dla subdirectory in the future?
this is set up to handle a dla directory already filled with date-string subdirectories.

lambdas/dla-migration/src/index.ts
etcart: you're right, this is a leftover from more complex logic that needed to do a lot more iwth this

packages/aws-client/src/S3.ts
```diff
@@ -779,39 +706,148 @@ export const uploadS3FileStream = (
...
+ *
+ * @param {Object} params - params for the s3.listObjectsV2 call
+ * @returns {Promise<Array>} resolves to an array of objects corresponding to
+ *   the Contents property of the listObjectsV2 response
+ *
+ * @static
+ */
+export const listS3ObjectsV2 = async (
+  params: ListObjectsV2Request
+): Promise<Array<ListS3ObjectsOutput>> => {
```
jennyhliu, line 761: The return type was Promise<ListObjectsV2Output['Contents']>.
Why do we need to filter on `obj.Key`? Every object has a key.

packages/aws-client/src/S3.ts
```diff
@@ -779,39 +706,148 @@ export const uploadS3FileStream = (
...
+      {
+
+        ...params,
+        ContinuationToken: listObjectsResponse.NextContinuationToken,
+      }
+    ));
+    discoveredObjects = listObjectsResponse.Contents ?? [];
+    yield discoveredObjects.filter((obj) => 'Key' in obj) as Array<ListS3ObjectsOutput>;
+  }
+}
```
jennyhliu, line 822: Good Method!👍
Same question about the filter.
Return type should be ListS3ObjectsOutput['Cotents']

lambdas/dla-migration/src/index.ts
jennyhliu: I prefer the simplest and cleanest solution.
If we want to process files in a different directory, it can be passed to the payload.  We want to avoid creating nested /date/date directories.

packages/aws-client/src/S3.ts
etcart: while every object *should* have a key, that object is define with Key as optional. so this layer guarantees it and returns an object type which likewise guarantees it

lambdas/dla-migration/src/index.ts
```diff
@@ -1,9 +1,106 @@
...
+    const objectBatch of listS3ObjectsV2Batch({ Bucket: bucket, Prefix: sourceDir })
+  ) {
+    const keys = objectBatch.map((obj) => obj.Key).filter(
+      (key) => key.lastIndexOf('/') === lastIndexOfDlaPathSeparator && key.endsWith('.json')
+    );
+    out.push(await pMap(
+      keys, (async (key) => updateDLAFile(bucket, key)),
+      {
+        concurrency: 5,
+        stopOnError: false,
```
jennyhliu, line 97: When `stopOnError: false` is used, it rejects with an aggregated error, we need to catch and log the error.
Can you add a test for the error case?

tsconfig.eslint.json
```diff
@@ -4,6 +4,7 @@
     "lambdas/**/*.ts",
     "packages/**/*.ts",
     "tasks/**/*.ts",
+    "scripts/**/*.ts",
```
jennyhliu, line 7: It's not used any more

packages/aws-client/src/S3.ts
```diff
@@ -779,39 +706,148 @@ export const uploadS3FileStream = (
       Body: fileStream,
     },
   });
+/**
+ * List of outputs from ListS3Objects functions (including V2)
+ * Guarantees that Key exists
+ */
```
jennyhliu, line 712: I'm not sure it's a good idea to reference `hidden` type _Object

lambdas/dla-migration/src/index.ts
```diff
@@ -1,9 +1,106 @@
...
+  listS3ObjectsV2Batch,
+  deleteS3Object,
+} from '@cumulus/aws-client/S3';
+import {
+  hoistCumulusMessageDetails,
+  extractDateString,
+  getDLARootKey,
+} from '@cumulus/message/DeadLetterMessage';
+import { DLARecord } from '@cumulus/types/api/dead_letters';
+import pMap from 'p-map';
```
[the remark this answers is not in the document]
etcart: to the first: done, to the second, how do we normally order these? alphabetical?

lambdas/dla-migration/src/index.ts
```diff
@@ -1,9 +1,106 @@
...
 
 const logger = new Logger({ sender: '@cumulus/dla-migration-lambda' });
+
+/**
+ * Manipulate target filepath to put file is YYYY-MM-DD sub-folder
+ *
+ * @param targetPath
```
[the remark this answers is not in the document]
etcart: done

lambdas/dla-migration/src/index.ts
```diff
@@ -1,9 +1,106 @@
...
+  sourceDirectory: string
+): Promise<Array<boolean>> => {
+  const out = [];
+  let sourceDir;
+  if (sourceDirectory.endsWith('/')) {
+    sourceDir = sourceDirectory;
+  } else {
+    sourceDir = `${sourceDirectory}/`;
+  }
+  const lastIndexOfDlaPathSeparator = sourceDir.lastIndexOf('/');
```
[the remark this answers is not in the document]
etcart: done

lambdas/dla-migration/src/index.ts
```diff
@@ -1,9 +1,106 @@
...
+  )
+);
+
+/**
+ * pull S3 Object from sourcePath, update it to new DLA structure and push it to targetPath
+ * noop if skip is true and an object already exists at targetPath
+ *
+ * @param bucket
+ * @param sourcePath
+ * @param targetPath
```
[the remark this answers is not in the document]
etcart: done

lambdas/dla-migration/package.json
```diff
@@ -24,7 +25,11 @@
   "dependencies": {
     "@cumulus/aws-client": "18.2.0",
     "@cumulus/logger": "18.2.0",
-    "lodash": "^4.17.21",
+    "@cumulus/message": "18.2.0",
+    "@cumulus/types": "18.2.0",
+    "lodash.clone": "^4.5.0",
```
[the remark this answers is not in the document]
etcart: uh.... because linter told me to. but looking into it linter advises it because lodash is a really big package, most of which we don't use

lambdas/dla-migration/src/index.ts
etcart: instead I've added an intenral catch and log to better use our boolean output from the updateDLAFile function. I do still need to add a test for *that* functionality though

tsconfig.eslint.json
etcart: right

packages/aws-client/src/S3.ts
etcart: how else would you suggest updating this, is an object which overrides it without inheriting it a better idea?

etcart: in terms of naming, my intention is to name it after its place in *our* s3 module. in this case I've overwritten the outputs of all of our ListS3Object* functions to return this type

jennyhliu: I prefer the old return type `Promise<ListObjectsV2Output['Contents']>`

package.json
jennyhliu, line 1: What's the change for?

packages/aws-client/src/S3.ts
etcart: but that isn't a correct type, as Key is guaranteed *not* undefined. switching back to that just means turning off type checking for lines that ask about Key, or de-optimizing our code to check for Key every time we use it

PR #3614, opened 2024-03-18
[CUMULUS-2600]: Update messaging related to CloudFront and API Gateways

npauzenga:
This is a doc-only update to ensure we're pointing users towards CloudFront distributions over API Gateways. 

https://bugs.earthdata.nasa.gov/browse/CUMULUS-3600


tclark-innovim (commented):
I believe CloudFront can provide both HTTP and HTTPS distribution URLs. Even though it is obvious, it might be worth stating that the HTTPS CloudFront distribution URLs should be used to ensure data in transit compliance.

PR #3616, opened 2024-03-20
CUMULUS-2902: Removes aws-sdk from es-client. Updates usage to AWS SDK v3.

paulpilone:

Partially addresses [CUMULUS-2902: D - (LAST) Remove all dependencies on AWS SDK v2](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2902).

## Changes

* Removes `es-client`'s direct dependencies on `aws-sdk` and `aws-elasticsearch-connector`.
* Updates `es-client` to use aws-sdk v3 credentials provider for fetching credentials.
* Adds custom AmazonConnection to author and sign requests to AWS for our Elasticsearch domain.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


packages/es-client/esAmazonConnection.js
```diff
@@ -0,0 +1,64 @@
+'use strict';
+
+const { Connection } = require('@elastic/elasticsearch');
+const aws4 = require('aws4');
```
npauzenga, line 4: This is probably alright. It's used by 11+ million people. There hasn't been a release in 4 years but it also doesn't pull in dependencies we should be concerned with.

packages/es-client/search.js
```diff
@@ -65,12 +56,19 @@ const esProdConfig = async (host) => {
     node = `https://${host}`;
   }
 
+  const credentialsProvider = fromNodeProviderChain({
+    clientConfig: {
+      region: process.env.AWS_REGION,
+    },
+  });
+
+  const credentials = await credentialsProvider();
```
npauzenga, line 65: Does the `credentials` object here look similar/the same as what `getCredentials` was previously returning? Makes sense if so.

npauzenga: Oh hmm I see in the mock below that it works a little different. We're using keys instead of user/pass. Are there any implications there? 🤔

npauzenga (changes requested):
Hey @paulpilone this looks good but I'm wondering about the switch from user/pass credentials to AWS keys. Is there anything we should be considering here? AWS keys eventually expire for example. If that happens you... should have other issues so I'm not sure that's a problem. Mostly curious because it's a functional change instead of a 1:1 replacement.

packages/es-client/esAmazonConnection.js
paulpilone: It seemed like a small enough library to continue using it _for now_.

packages/es-client/search.js
paulpilone: Good questions and I'm going to @Jkovarik in my responses because he's specifically working on a bug related to expired credentials.

The return type of credentials I believe is _close enough_ to what we were originally getting: 
**New:** https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-smithy-types/TypeAlias/AwsCredentialIdentityProvider/

**Old:** https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Credentials.html

The biggest difference (and something @Jkovarik might care about) is `expiration` vs `expireTime`. I don't think the previous code actually used `expireTime` and why we have the other bug that we do.

Do we _actually_ support user/pass and was that not just a relic of an easy mock? If so, we might need more than the Node provider chain. The Node provider chain looks at the following for creds (in order):

1. [Environment variables exposed via process.env](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-credential-providers/#fromenv)
2. [SSO credentials from token cache](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-credential-providers/#fromsso)
3. [Web identity token credentials](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-credential-providers/#fromtokenfile)
4. [Shared credentials and config ini files](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-credential-providers/#fromini)
5. [The EC2/ECS Instance Metadata Service](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-credential-providers/#fromcontainermetadata-and-frominstancemetadata)

npauzenga: Based on Slack conversations, I think this can be resolved. It appears that the mock was done using user/pass for simplicity rather than an attempt to emulate what actually happens.

npauzenga (approved):
Thanks for thinking this through @paulpilone. This can be merged.

PR #3618, opened 2024-03-26
CUMULUS-2902: Removes dependency of aws-sdk from all packages

paulpilone:

Addresses [CUMULUS-2902: D - (LAST) Remove all dependencies on AWS SDK v2](https://bugs.earthdata.nasa.gov/browse/CUMULUS-2902). This PR should be reviewed after https://github.com/nasa/cumulus/pull/3616 since it was originally branched from those changes.

## Changes

* Removes `aws-sdk` from all packages and replaces any usage of it with equivalent v3 syntax

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


example/lambdas/ftpPopulateTestLambda/webpack.config.js
```diff
@@ -27,7 +27,7 @@ module.exports = {
     }
   },
   externals: [
-    'aws-sdk',
+    '@aws-sdk',
```
npauzenga, line 30: I should probably know this but what are we saying here? What does this change do?

packages/api/webpack.config.js
```diff
@@ -80,7 +80,6 @@ module.exports = {
     }
   },
   externals: [
-    'aws-sdk',
     '@aws-sdk',
```
npauzenga, line 83: Same question as above (same for all of the webpack configs). ELI5 please 😆

packages/aws-client/src/client.ts
```diff
@@ -50,13 +45,7 @@ const awsClient = <T extends AWSClientTypes>(
     ...serviceOptions,
   };
   if (version) options.apiVersion = version;
-  if (inTestMode()) {
```
npauzenga, line 53: Presumably this was a V2 issue that V3 corrects?

packages/aws-client/src/client.ts
```diff
@@ -7,8 +6,6 @@ import { getServiceIdentifer } from './utils';
 
 const getRegion = () => process.env.AWS_DEFAULT_REGION || process.env.AWS_REGION || 'us-east-1';
 
-AWS.config.setPromisesDependency(Promise);
```
npauzenga, line 10: What did this do?

packages/aws-client/src/lib/S3MultipartUploads.ts
```diff
@@ -5,13 +5,14 @@ import range from 'lodash/range';
...
   CompleteMultipartUploadRequest,
+  CompleteMultipartUploadOutput as S3CompleteMultipartUploadOutput,
   CreateMultipartUploadRequest,
   UploadPartCopyRequest,
 } from '@aws-sdk/client-s3';
 
 import { s3 } from '../services';
 
-export interface CompleteMultipartUploadOutput extends AWS.S3.CompleteMultipartUploadOutput {
+export interface CompleteMultipartUploadOutput extends S3CompleteMultipartUploadOutput {
```
npauzenga, line 15: Hmm maybe we can clean this up? It looks like we're importing `CompleteMultipartUploadOutput` as an alias, then exporting `CompleteMultipartUploadOutput` as an extension of that alias?

I guess our interface already shared a name with an AWS interface but the namespace was different. Kinda gross but I see why you did it this way. The grossness was pre-existing.

Maybe we rename our interface? Also maybe this doesn't really matter except that it's a little confusing at first glance.

packages/aws-client/src/types.ts
```diff
@@ -36,6 +34,4 @@ export type AWSClientTypes =
     CloudWatchEvents |
     CloudFormation |
     Kinesis |
-    KMS |
-    AWS.Service |
-    AWS.DynamoDB.DocumentClient;
+    KMS;
```
npauzenga, line 37: Were we just not using these?

packages/tea-map-cache/webpack.config.js
```diff
@@ -9,7 +9,9 @@ module.exports = {
     path: path.resolve(__dirname, 'dist'),
     filename: 'index.js'
   },
-  externals: ['aws-sdk'],
+  externals: [
+    '@aws-sdk'
+  ],
```
npauzenga, line 14: Should this have the `@`? Also what's up with the `@`? 😄

npauzenga (changes requested):
This looks great @paulpilone! We were concerned about what might have been missed in this epic and it looks like you caught a few things. 

I had a couple clarifying questions but overall looks great.

packages/tea-map-cache/webpack.config.js
paulpilone: @npauzenga As discussed in Slack this was originally changed because we want (in theory) to exclude @aws-sdk (v3) from our packages since it's provided in the Lambda environment for us.

However your question made me think about this more and it's a good thing it did. We're still targeting Node 16 which does _not_ include the v3 SDK in the Lambda runtime. So we can't exclude it.

Also, I'm not convinced that was excluding it the way we thought it was because most of our packages depend on `aws-client` which does not have a webpack configuration. I've added a comment to our Node upgrade ticket (https://bugs.earthdata.nasa.gov/browse/CUMULUS-3433) to make sure this gets re-added and in a way that actually does what we expect.

packages/aws-client/src/types.ts
paulpilone: `AWS.Service` wasn't used and `AWS.DynamoDB.DocumentClient` became

```
export const dynamodbDocClient = (docClientOptions?: TranslateConfig, dynamoOptions?: object) =>
  DynamoDBDocument.from(
    awsClient(DynamoDB, '2012-08-10')(dynamoOptions),
    docClientOptions
  );
 ```

packages/aws-client/src/lib/S3MultipartUploads.ts
paulpilone: I added the alias because it was the path to least resistance but I think you're right and our interface should really be named something else. We already do something similar here: https://github.com/nasa/cumulus/pull/3618/files#diff-56edb8ddc2b0588b9a8b2f0bb5f33759f187a5bd46e30725d6e79cbe6ed46056L162 although I'm not sure us calling it `SQSMessage` is super clean either. 

Curious your opinion on if we should address both, this one, or neither in this PR.

packages/aws-client/src/client.ts
paulpilone: Correct. Looks like that `serviceIdentifier` issue was a v2 bug.

packages/aws-client/src/client.ts
paulpilone: AWS SDK v2 allowed you to configure the promise library it used and this is how you did it.

packages/aws-client/src/lib/S3MultipartUploads.ts
npauzenga: Yeah. What if we took a small step and instead of importing `CompleteMultipartUploadOutput` as an alias, we import it normally and rename _our_ interface `S3CompleteMultipartUploadOutput`?

At least that way we don't have something with exactly the same name as the AWS interface, which could(?) cause confusion.

packages/api/webpack.config.js
npauzenga: Punting this conversation to https://github.com/nasa/cumulus/pull/3618#discussion_r1561655895

example/lambdas/ftpPopulateTestLambda/webpack.config.js
npauzenga: Punting this conversation to https://github.com/nasa/cumulus/pull/3618#discussion_r1561655895

packages/tea-map-cache/webpack.config.js
npauzenga: Thanks for clarifying this. I think it makes sense to just remove the V2 external for now.

packages/aws-client/src/lib/S3MultipartUploads.ts
paulpilone: @npauzenga Pushed a small change that refactored that in this file. aws-client unit tests pass locally and waiting on CI now.

npauzenga (approved):
This looks great @paulpilone! Good to go pending CI.

PR #3620, opened 2024-04-02
Limited sqs size

etcart:
prevent possibility of sqs message overflow causing messages to fail. 

Addresses [CUMULUS-3678: limit sqs message size](https://bugs.earthdata.nasa.gov/browse/CUMULUS-XXX)

## Changes

* sqs messages over documented maximum size are truncated to 256kb

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/packages/aws-client/src/SQS.ts
+++ b/packages/aws-client/src/SQS.ts
@@ -9,7 +9,7 @@ import isObject from 'lodash/isObject';
 import isString from 'lodash/isString';
 import isNil from 'lodash/isNil';
 import { SQSRecord } from 'aws-lambda';
-import { Message, QueueAttributeName } from '@aws-sdk/client-sqs';
+import { Message, QueueAttributeName, SendMessageCommandOutput } from '@aws-sdk/client-sqs';
 
 import { StepFunctionEventBridgeEvent } from './Lambda';
 import { sqs } from './services';
@@ -78,13 +78,17 @@ export const sendSQSMessage = (
   queueUrl: string,
   message: string | object,
   logOverride: Logger | undefined = undefined
-) => {
+): Promise<SendMessageCommandOutput> => {
   const logger = logOverride || log;
   let messageBody;
   if (isString(message)) messageBody = message;
   else if (isObject(message)) messageBody = JSON.stringify(message);
   else throw new Error('body type is not accepted');
-
+  const maxLength = 262144;
+  if (messageBody.length > maxLength) {
+    const warningString = '...TruncatedForLength';
+    messageBody = `${messageBody.substring(0, maxLength - warningString.length)}${warningString}`;
+  }
   return sqs().sendMessage({
     MessageBody: messageBody,
     QueueUrl: queueUrl,
@@ -106,7 +110,7 @@ type ReceiveSQSMessagesOptions = {
  */
 export const receiveSQSMessages = async (
   queueUrl: string,
-  options: ReceiveSQSMessagesOptions
+  options: ReceiveSQSMessagesOptions = {}
 ): Promise<SQSMessage[]> => {
   const params = {
     QueueUrl: queueUrl,
--- a/packages/aws-client/tests/test-SQS.js
+++ b/packages/aws-client/tests/test-SQS.js
@@ -102,21 +102,21 @@ test('isSQSRecordLike filters correctly for sqs record shape', (t) => {
   but strictly checking is not compatible with the current use-case*/
 });
 
-test.only('sendSQSMessage truncates oversized messages safely', async (t) => {
+test('sendSQSMessage truncates oversized messages safely', async (t) => {
   const queueName = randomString();
   const queueUrl = await createQueue(queueName);
   const maxLength = 262144;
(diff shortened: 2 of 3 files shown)


vpnguye2:
Unit test passed locally

image

vpnguye2 (approved):
Looks great. 👍

packages/aws-client/src/SQS.ts
```diff
@@ -94,12 +96,17 @@ export const sendSQSMessage = (
...
   message: string | object,
   logOverride: Logger | undefined = undefined
-) => {
+): Promise<SendMessageCommandOutput> => {
   const logger = logOverride || log;
   let messageBody;
   if (isString(message)) messageBody = message;
   else if (isObject(message)) messageBody = JSON.stringify(message);
   else throw new Error('body type is not accepted');
+  const maxLength = 262144;
```
npauzenga, line 105: How'd we come to this size?

etcart: that is the max size of an sqs message as documented here https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/quotas-messages.html#:~:text=The%20maximum%20is%20262%2C144%20bytes,message%20payload%20in%20Amazon%20S3.

packages/aws-client/tests/test-SQS.js
```diff
@@ -100,3 +101,23 @@ test('isSQSRecordLike filters correctly for sqs record shape', (t) => {
...
+test('sendSQSMessage truncates oversized messages safely', async (t) => {
+  const queueName = randomString();
+  const queueUrl = await createQueue(queueName);
+  const maxLength = 262144;
+  const overflowMessage = '0'.repeat(maxLength + 2);
+  await sendSQSMessage(queueUrl, overflowMessage);
+
+  let recievedMessage = await receiveSQSMessages(queueUrl);
+  let messageBody = recievedMessage[0].Body;
+  t.true(messageBody.endsWith('...TruncatedForLength'));
```
BWex-NASA, line 114: Do you normally add negative test cases? Like if the message is under the max size, then verify the '...TruncatedForLength' message does not appear. In the event it adds it to all the messages regardless of size.

etcart: that's a good point, I'll add it

packages/aws-client/src/SQS.ts
npauzenga: Hmm so that's in bytes `262,144 bytes`. JS's `length` returns  UTF-16 "units".

>In UTF-16 encoding, every code unit is exact 16 bits long.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#utf-16_characters_unicode_code_points_and_grapheme_clusters:~:text=In%20UTF%2D16%20encoding%2C%20every%20code%20unit%20is%20exact%2016%20bits%20long

Is there a better way to determine the message size? If I'm thinking about this correctly (big *if*), the `length` JS would return is roughly twice as large as the max size SQS allows:

If `length` returns e.g. 262144 units, which are  each 16 bits long, and a byte is 8 bits...

It seems like 'length' may not be appropriate here?

packages/aws-client/tests/test-SQS.js
BWex-NASA: Maybe at the end add:

etcart: sure, we're checking implicitly just by checking that "it's unchanged", but verbosity==good in testing

BWex-NASA (approved):
Approved the latest commit for a negative test case.

BWex-NASA (approved):

packages/aws-client/src/SQS.ts
etcart: it seems like you're right. I've tested this with 262144 length vs 262144, but I *think* they're using the standard ascii/unicode split where *most* characters are 1 byte, but bytes >=128 are interpreted as the first of two in a unicode. I'll do some testing here to be sure and figure out how to correctly measure this

etcart: a fix for this is up now, thanks for that great catch!
there's an additional caveat that if a unicode character gets bifurcated, javascript encoding adds two *more* padding bytes that need to be dealt with. this latest change accounts for those things and tests them

BWex-NASA (approved):
Latest test changes appear good.

packages/aws-client/src/SQS.ts
npauzenga: Hmm is there another way to handle this besides inspecting the character length of the message? It's starting to feel like an anti-pattern if we have to continue to use string.length and accommodate for the difference between units.

I'm also a little concerned about not storing the original, untruncated message anywhere. 

We handle similar limitations with messages via the CMA by dumping the full message to S3 and referencing it in the message passed between functions in a step function. Is there something similar we could do here with SQS messages?

How was the issue initially encountered? Did a user report it?

PR #3621, opened 2024-04-03
Dla surety

etcart:
ensures that the DLA changes are simplified and non-breaking as far as possible

Addresses [CUMULUS-3629: ensure the updated dla structure doesn't cause problems](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3629)

## Changes

* HoistCumulusMessageDetails now de-nests nested body/Body contents and names the innermost version "body"
* dbRecordsDLQSpec (integration test) now ensures that the message output to DLA is handle-able by proces-s3-dead-letter-archive

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


packages/api/lambdas/process-s3-dead-letter-archive.js
```diff
@@ -19,12 +18,22 @@ const { writeRecords } = require('./sf-event-sqs-to-db-records');
...
+    * this *should* be called with format
+    * `${process.env.stackName}/dead-letter-archive/sqs/${messageId}`
+    * this path will also handle a call pointed to
+    * `${process.env.stackName}/dead-letter-archive/failed-sqs/${messageId}`
+    * */
+    return failedKey.replace('sqs/', 'failed-sqs/');
+  }
+  if (failedKey.includes('dead-letter-archive/')) {
+    // in case a weird path is passed, hopefully it contains dead-letter-archive/
+    return failedKey.replace('failed-dead-letter-archive/');
```
jennyhliu, line 32: What is `failed-dead-letter-archive`?  We have never used this path.

CHANGELOG.md
```diff
@@ -82,6 +82,8 @@ the CloudWatch logs for your async operations (e.g. `PREFIX-AsyncOperationEcsLog
   - Added user guide on querying dead-letter-archive messages using AWS Athena.
 
 ### Changed
+- **CUMULUS-3629**
+  - dla guarantees de-nested SQS message bodies, preferring outermost metadata as found.
```
jennyhliu, line 86: Can you also fix the s3 file name?

packages/api/lambdas/process-s3-dead-letter-archive.js
jennyhliu: I think no matter where the input messages come from, we always put failed messages to `${pathPrefix}/failed-sqs/${date}/${messageId}`;
date should be original eventDate, not the time the message is reprocessed

etcart: #1: it is using the original eventDate, not the time the message is reprocessed
#2: I had envisioned this as a failure case, that it will only go to these other paths if it *can't* parse the original path such that there is a clear ..../failed-sqs/.... to send it to.
if I'm going to guarantee that it will go into failed-sqs/ but *cannot* parse the /sqs/ folder from the path, can I guarantee that I'll be able to parse date and messageId?

CHANGELOG.md
etcart: done

packages/api/lambdas/process-s3-dead-letter-archive.js
jennyhliu: We can't put to `${process.env.stackName}/dead-letter-archive/failed-sqs/${date}/${messageId}` regardless the original path

jennyhliu: Operator can pass in different path, and there is no guarantee the path will have /sqs/.  So to make things simple, we can put the failed messages to predefined location, so the messages won't end up everywhere.

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
```diff
@@ -27,7 +36,7 @@ describe('when a bad record is ingested', () => {
   afterAll(async () => {
     await deleteS3Object(
       systemBucket,
-      failedMessageS3Key
+      leftoverS3Key
     );
   });
   it('is sent to the DLA and processed to have expected metadata fields', async () => {
```
jennyhliu, line 42: Can you break down this to multiple specs/it?

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
```diff
@@ -27,7 +36,7 @@ describe('when a bad record is ingested', () => {
   afterAll(async () => {
     await deleteS3Object(
       systemBucket,
-      failedMessageS3Key
+      leftoverS3Key
```
jennyhliu, line 39: s3 key for both `it` needs to be cleaned up

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
etcart: I can, it duplicates a bunch of the work, but would make the test results clearer: doing it

packages/api/lambdas/process-s3-dead-letter-archive.js
```diff
@@ -1,58 +1,58 @@
...
  */
-const generateNewArchiveKeyForFailedMessage = (failedKey) => {
-  // Split key with format `${process.env.stackName}/dead-letter-archive/sqs/${messageId}`,
-  const splitKey = failedKey.split('/sqs/');
-  const pathPrefix = splitKey[0];
-  const messageId = splitKey[1];
-  const date = moment.utc().format('YYYY-MM-DD');
-  return `${pathPrefix}/failed-sqs/${date}/${messageId}`;
-};
+const generateNewArchiveKeyForFailedMessage = (failedMessage) => (
```
jennyhliu, line 26: Is failedMessage actual message or s3 object?
There are places using this method, some pass in s3 object with key.

packages/api/lambdas/process-s3-dead-letter-archive.js
```diff
@@ -1,58 +1,58 @@
...
+  getDLAKey(process.env.stackName, failedMessage).replace('sqs/', 'failed-sqs/')
```
jennyhliu, line 27: Can your replace `/dead-letter-archive/sqs/` to be safe?

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
jennyhliu: you can change `it` to `describe`, and define some variables under that describe to avoid more setups.

example/spec/parallel/dbRecords/DeadLetterArchiveSpec.js
```diff
@@ -107,7 +116,98 @@ describe('when a bad record is ingested', () => {
...
+    expect(await s3ObjectExists({
+      Bucket: systemBucket,
+      Key: postRecoveryFailedKey,
+    })).toEqual(true);
+    leftoverS3Key = postRecoveryFailedKey;
+    await deleteAsyncOperation({
+      prefix: stackName,
+      asyncOperationid: deadLetterRecoveryAsyncOpId,
+    });
+  })
```
jennyhliu, line 210: missing ';'?

packages/api/lambdas/process-s3-dead-letter-archive.js
etcart: done

packages/api/lambdas/process-s3-dead-letter-archive.js
etcart: I've kept it functional with the old single input as first argument (and has added failover even in case of that usage pattern) with the new arguments using defaults if not supplied, to make sure we're compatible.
but within cumulus-core I've tracked down all calls to this and update their use

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
etcart: leftoverS3Key should be that, name changed because "leftover" more accurately describes it, where the one that calls process-s3-dead.... removes the old one

example/spec/parallel/dbRecords/DbRecordsDLQSpec.js
etcart: done, thanks for the pointer

jennyhliu (approved):

jennyhliu (approved):
👍

PR #3096, opened 2022-09-01
CUMULUS-3021: Support Collections with a collection.version containing slashes

jennyhliu:
**Summary:** Summary of changes

Addresses [CUMULUS-3021: Support Collections with a collection.version containing slashes](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3021)

## Changes

* Updated `@cumulus/api-client/collections` and `@cumulus/integration-tests/api` to encode
    collection version in the URI path
* Renamed example workflow `KinesisTriggerTest` to `CNMExampleWorkflow`.
* Added integration test to make sure we can ingest granules from collection  `ASCATB-L2-Coastal___Operational/Near-Real-Time`, and `sqsMessageConsumer` lambda and `CNMExampleWorkflow` handle such collection/granule

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Ad-hoc testing - Deploy changes and test manually
- [x] Integration tests

PR #3629, opened 2024-04-16
CUMULUS-3673: Fixes Granules API URI encoding by escaping path components.

paulpilone:

Addresses [CUMULUS-3673: Queue Granules fails for collections with "/" in version](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3673)

## Changes

* Updates Granules API so that paths containing granule and collection IDs are URI encoded.
* Fixes JSDoc linter warners in granules API handler

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


packages/api-client/src/granules.ts
```diff
@@ -15,21 +15,30 @@ type AssociateExecutionRequest = {
...
+): string =>
+  (collectionId
+    ? `/granules/${encodeURIComponent(collectionId)}/${encodeURIComponent(granuleId)}`
+    : `/granules/${encodeURIComponent(granuleId)}`); // Fetching a granule without a collectionId is supported but deprecated
+
 /**
  * GET raw response from /granules/{granuleId} or /granules/{collectionId}/{granuleId}
  *
- * @param {Object} params                                - params
+ * @param {object} params                                - params
```
npauzenga, line 29: There's conversation in Slack about docstrings generally. The consensus is that adding typings in docstrings in TypeScript files is probably redundant and prone to going stale (as seen here). I think we can generally remove type annotations from the docstrings and _maybe_ param descriptions in the case of (for example) `params.granuleId = a granule ID`.

I'd be inclined to at least remove the types here as we go but open to discussion. This idea hasn't really be codified.

packages/api-client/tests/test-granules.js
```diff
@@ -6,8 +6,8 @@ const granulesApi = require('../granules');
 
 test.before((t) => {
   t.context.testPrefix = 'unitTestStack';
-  t.context.granuleId = 'granule-1';
-  t.context.collectionId = 'fakeName___fakeVersion';
+  t.context.granuleId = randomId('gran/a-b-c-123');
```
npauzenga, line 9: Without using the encoder method, did this cause the tests to fail in the same way PODAAC saw?

npauzenga (changes requested):
Looks good @paulpilone! Just a couple comments. Do we need to extend this to other endpoints that deal with collections though? 

Also do we know what caused the break? It looks like it was working differently sometime in the past. I'm looking at this line in the ticket:

"This was not happening when we were on Cumulus 11, but after upgrading to Cumulus 16 this issue started occurring. This is a high priority issue for us because it blocks DMRPP generation for the above collection"

packages/api-client/src/granules.ts
paulpilone: Sounds good. My goal here was to eliminate all of the linter warnings and update things where it was obviously wrong or the change was obvious. I'll definitely remove the types for now.

packages/api-client/src/granules.ts
```diff
@@ -48,12 +57,7 @@ export const getGranuleResponse = async (params: {
     callback = invokeApi,
   } = params;
 
-  let path = `/granules/${collectionId}/${granuleId}`;
-
-  // Fetching a granule without a collectionId is supported but deprecated
-  if (!collectionId) {
-    path = `/granules/${granuleId}`;
-  }
+  const path = encodeGranulesURIComponent(granuleId, collectionId);
```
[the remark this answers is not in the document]
paulpilone: So I would hope not because the caller of this function isn't passing the path but rather the explicit collection and granule IDs. I'd say it's up to the caller to recognize that and not try to do any fancy encoding before hand. A quick search shows we're not encoding prior to calling this function so I think we're ok - assuming you're cool with my thought process.

packages/api-client/tests/test-granules.js
paulpilone: No, because it's all being mocked out with fake callbacks. I briefly talked to @Jkovarik about this and we discussed whether or not we should introduce an integration test. We leaned toward not adding one but curious what your thoughts are.

paulpilone:
> Looks good @paulpilone! Just a couple comments. Do we need to extend this to other endpoints that deal with collections though?

That was already done in an older PR https://github.com/nasa/cumulus/pull/3096 - just not carried over to granules.

> Also do we know what caused the break? It looks like it was working differently sometime in the past. I'm looking at this line in the ticket:
> 
> "This was not happening when we were on Cumulus 11, but after upgrading to Cumulus 16 this issue started occurring. This is a high priority issue for us because it blocks DMRPP generation for the above collection"

@npauzenga my guess (because I haven't done a deep dive into v11 vs v16) here is that v16 introduced looking up granules by collection ID and granule ID vs 11 only supporting granule ID: https://github.com/nasa/cumulus/blob/b5fc71399924b942406a7aaf8934c8feffc18989/packages/api-client/src/granules.ts#L443

The collection ID is what actually needs to be escaped here (even though I've escaped both in the change) and so was most likely introduced in 16 since the queue granules task now provides both collection and granule ID.

packages/api-client/src/granules.ts
paulpilone: Pushed a change that removes the types for now.

packages/api-client/src/granules.ts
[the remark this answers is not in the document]
npauzenga: Yup, absolutely. That seemed like the most likely scenario.

packages/api-client/src/granules.ts
npauzenga: Cool. We need to codify our intention with these docstrings. I think this is the right idea but we're sort of all over the place.

packages/api-client/tests/test-granules.js
npauzenga: Hmm interesting. Is there an existing integration test that could be updated with a granule/collectionId containing slashes? I tend to agree that adding one for this purpose isn't fun for anyone or useful.

At the end of the day our only real concern here is saying that we were able to replicate the problem and we believe this fix... fixes. If modifying an existing int test isn't feasible we could also do a manual one-off test with a deployed stack.

example/spec/parallel/cnmWorkflow/CnmWorkflowFromSqsSpec.js
```diff
@@ -71,6 +72,7 @@ const s3data = [
   '@cumulus/test-data/granules/ascat_20121029_010301_metopb_00588_eps_o_coa_2101_ovw.l2.nc',
 ];
 
+/** */
```
npauzenga, line 75: Probably remove

example/spec/parallel/cnmWorkflow/KinesisTestTriggerSpec.js
```diff
@@ -96,6 +96,7 @@ describe('The Cloud Notification Mechanism Kinesis workflow', () => {
   let scheduleQueueUrl;
   let failingWorkflowExecution;
 
+  /** */
```
npauzenga, line 99: Remove

npauzenga (commented):
This looks good @paulpilone! Just want to be sure: it looks to me like we should be able to remove `encodedConstructCollectionId` as that never should have existed in the first place. There are a couple specs (Lzards-related) that still use it. Is that intentional or should those also be updated?

Also the testing conversation. Are we all in agreement that what we have here is sufficient? @Jkovarik ?

paulpilone:
> This looks good @paulpilone! Just want to be sure: it looks to me like we should be able to remove `encodedConstructCollectionId` as that never should have existed in the first place. There are a couple specs (Lzards-related) that still use it. Is that intentional or should those also be updated?
> 
> Also the testing conversation. Are we all in agreement that what we have here is sufficient? @Jkovarik ?

I saw those tests and I wasn't 100% sure because my experience is non-existent with Lzards. _Looking_ at the test I'm not sure why it was necessary to encode it in the first place. Maybe we can chat about it or if you're aware of those tests and think it's ok to remove I'm happy to do it and push another commit. It was just a set of tests I wasn't comfortable updating and I didn't think it hurt to leave them be but I can definitely be convinced otherwise.

npauzenga (approved):
This is great @paulpilone. This can be merged pending CI. Thanks for thinking about it so carefully!

PR #3660, opened 2024-05-17
CUMULUS-3693: Update granules List endpoints to query postgres - range

jennyhliu:

Addresses [CUMULUS-3693: Update granules List endpoints to query postgres - range](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3693)

## Changes

* Added functionality to `@cumulus/db/src/search` to support range queries

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


Jkovarik (changes requested):
This change does not fix the problem - we should not be using promisify.

PR #3666, opened 2024-05-22
[CUMULUS-3671]: Update docs for Serverless V2

npauzenga:
These updates change a few links and shuffle wording to point to Serverless V2 info. For the most part the changes are relatively minor as we're not trying to duplicate AWS docs. There are, however, a couple areas that recommend configuration. I will comment in-line with my intention there.

### DIFF
--- a/docs/deployment/choosing_configuring_rds.md
+++ b/docs/deployment/choosing_configuring_rds.md
@@ -12,9 +12,9 @@ core deployment code expects this database to be provided by the [AWS RDS](https
 
 RDS databases are broadly divided into two types:
 
-- **Provisioned**: Databases with a fixed capacity in terms of CPU and memory capacity. You can find
-a list of the available database instance sizes in [this AWS documentation](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.DBInstanceClass.html).
-- **Serverless**: Databases that can scale their CPU and memory capacity up and down in response to database load. [Amazon Aurora](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/CHAP_AuroraOverview.html) is the service which provides serverless RDS databases.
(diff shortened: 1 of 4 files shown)


docs/upgrade-notes/serverless-v2-upgrade.md
```diff
@@ -0,0 +1,27 @@
+---
```
npauzenga, line 1: These instructions were hammered out in https://bugs.earthdata.nasa.gov/browse/CUMULUS-3670

docs/deployment/choosing_configuring_rds.md
```diff
@@ -77,39 +83,10 @@ Current security policy/best practices require use of a SSL enabled configuratio
 
 Cumulus can accommodate a self-signed/unrecognized cert by setting `rejectUnauthorized` as `false` in the connection secret.  This will result Core allowing use of certs without a valid CA.
 
-## Recommended Scaling Configuration for Aurora Serverless
```
npauzenga, line 80: This isn't applicable in V2. The most relevant configuration options here are the min and max capacity settings. We _could_ provide some guidance there but I don't believe we have that ready to go. Totally open to thoughts here as this was the most contentious part of these doc updates in my mind.

docs/deployment/choosing_configuring_rds.md
```diff
@@ -41,6 +41,12 @@ will be able to handle the spikes in your database load.
 
 ## General Configuration Guidelines
 
+### Aurora Serverless V2 Capacity Range
+
+Serverless V2 allows users to configure the minumum and maximum number of ACUs the custer should use.
```
npauzenga, line 46: See below comment. We could at some point make recommendations here but I don't think we have them now.

paulpilone (approved):

PR #3679, opened 2024-06-04
Cumulus 3742 record generation script

etcart:

Addresses [CUMULUS-3742 record generation script](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3742)

## Changes

* added script in scripts/generate_records to load bulk data into a cumulus database

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,586 @@
...
+const minimist = require('minimist');
+const cryptoRandomString = require('crypto-random-string');
+const fs = require('fs-extra');
+const Logger = require('@cumulus/logger');
+const {
+  GranulePgModel,
+  CollectionPgModel,
+  ProviderPgModel,
+  ExecutionPgModel,
+  GranulesExecutionsPgModel,
```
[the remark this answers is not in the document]
etcart: maybe? I've skipped the things that didn't occur to me as "need to uploading bulk for a test" I'd like to stick to scope, but if you feel like that belongs in scope fair enough

scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,586 @@
...
+process.env.DISABLE_PG_SSL = 'true';
+/**
+ * yield series of collection details
+ *
+ * @param {number} total - number of collections
+ * @param {boolean} repeatable - use consistent names versus pseudorandom
+ * @yields {CollectionDetails}
+ */
+function* yieldCollectionDetails(total, repeatable = true) {
+  for (let i = 0; i < total; i += 1) {
```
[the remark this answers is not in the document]
etcart: the fact that it's a generator: I'm trying not to hold onto data any more than necessary to trim ram usage as much as possible
the purpose is to generate a series of unique collection name/Ids.

scripts/generate_records/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+This script (generate_db_records.js) is meant to push up large quantities of realistic cumulus database entries for scaled testing purposes
+
+## Installation
+This can be installed with npm install in this directory (or will be installed as a part of cumulus when installing the whole of cumulus-core)
+
+generate_db_records.js is tested to run with both node v16.19.0 and v20.12.2
+
+## Configuration
+the script can be configured either through command line arguments of environment variables (or both), preferring command line arguments if both are supplied
```
charleshuang80, line 11: command line arguments _or_ environment variables

charleshuang80, line 11: command line arguments _or_ environment variables

scripts/generate_records/generate_db_records.js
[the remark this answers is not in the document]
charleshuang80: I would look into LP's data first. They definitely don't have any PDRs, so that can be added later. Rules, I'm not sure how much there are or if it matters for our testing for ES.

scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,593 @@
...
+/**
+ * add provider through providerPgModel call
+ *
+ * @param {Knex} knex
+ * @returns {Promise<string>}
+ */
+const addProvider = async (knex) => {
+  const providerJson = JSON.parse(fs.readFileSync(`${__dirname}/resources/s3_provider.json`, 'utf8'));
+  const providerModel = new ProviderPgModel();
+  const a = await providerModel.upsert(
```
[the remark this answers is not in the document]
charleshuang80: agree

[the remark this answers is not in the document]
etcart: whups, yes

scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,593 @@
...
+  granuleCumulusId,
+  granuleGranuleId,
+  fileCount,
+  models,
+  swallowErrors = false
+) => {
+  const fileModel = models.fileModel;
+  let uploaded = 0;
+  for (let i = 0; i < fileCount; i += 1) {
+    const file = /** @type {PostgresFile} */(fakeFileRecordFactory({
```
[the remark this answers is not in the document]
etcart: this is a typecast to satisfy ts-check. fakeFileRecordFactory returns type Omit<PostgresFile, "granule_cumulus_id">
but we explicitly set granule_cumulus_id by passing it in as a granule, and fileModel.upsert does demand a non-undefined granule_cumulus_id

scripts/generate_records/README.md
etcart: gotcha

etcart: https://github.com/nasa/cumulus/pull/3679/commits/c011c7aff24d5fa42151671a666dbb055d2d978d

scripts/generate_records/generate_db_records.js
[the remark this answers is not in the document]
etcart: https://github.com/nasa/cumulus/pull/3679/commits/365c780815a2b4bc3384a8a7e7a6059eff05fe56

scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,593 @@
...
+const uploadExecutions = async (
+  knex,
+  collectionCumulusId,
+  executionCount,
+  models,
+  swallowErrors = false
+) => {
+  const executionCumulusIds = [];
+  const executionModel = models.executionModel;
+  for (let i = 0; i < executionCount; i += 1) {
```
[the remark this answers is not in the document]
etcart: I'm doing some research on this, node doesn't stop me, but it leads to some weirdness
Array(5).map(async (i) => await something) appears to simply work, but I can't await that map function, so is the whole thing asynchronous? I need to investigate further

scripts/generate_records/generate_db_records.js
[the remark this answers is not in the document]
etcart: I'd like to look at scoping this out, LP has no pdfs in the snap database, and rules only a handful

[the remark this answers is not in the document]
etcart: after our conversation, I'm going to call this out of scope

scripts/generate_records/README.md
```diff
@@ -0,0 +1,23 @@
...
+| --- | :----: | :----: | ---: |
+| --collections <br>-c | COLLECTIONS | 1 | number of collections. number of granules will be <br> for *each* collection, not divided among them |
+| --granules_k <br> -g| GRANULES_K | 10 | number of granules, in thousands |
+| --executionsPerGranule <br> -e | EXECUTIONS_PER_GRANULE | 2:2 | number of executions *x* per <br> batch of granules *g* in format 'x:g' <br> \<executionsPerBatch>:\<granulesPerBatch> |
+| --files <br> -f | FILES | 1 | number of files per granule |
+| --concurrency <br> -C | CONCURRENCY | 1 | how many threads of parallelization |
+| --variance <br> -v| VARIANCE |  false | randomize executions and granules per batch, <br> adding up to 6 granules and or executions to a given batch |
+| --swallowErrors <br> -s|SWALLOW_ERRORS| true | swallow and move on from data data upload errors |
+
+concurrency should usually be >100,
```
charleshuang80, line 23: maybe move this comment to the explanation of concurrency?

scripts/generate_records/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Installation
+This can be installed with npm install in this directory (or will be installed as a part of cumulus when installing the whole of cumulus-core)
+
+generate_db_records.js is tested to run with both node v16.19.0 and v20.12.2
+
+## Configuration
+the script can be configured either through command line arguments or environment variables (or both), preferring command line arguments if both are supplied
+
+| Argument    | Environment | Default | Description |
```
charleshuang80, line 13: this set of options seems much cleaner, thanks!

scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,439 @@
...
+      models,
+      swallowErrors,
+    };
+
+    if (variance) {
+      //asking for variance adds some noise to batch executions vs granules
+      let _granulesPerBatch = 1;
+      for (let i = 0; i < numberOfGranules; i += _granulesPerBatch) {
+        _granulesPerBatch = granulesPerBatch + randomInt(6);
+        const _executionsPerBatch = executionsPerBatch + randomInt(6);
```
charleshuang80, line 194: I think this should also initially be declared outside of the `for` loop with a `let`, so it's not continuously declared with `const` for each iteration.

scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,439 @@
...
+ */
+function* yieldCollectionDetails(total, repeatable = true) {
+  for (let i = 0; i < total; i += 1) {
+    let suffix;
+    if (repeatable) {
+      suffix = `_test_${i.toString().padStart(3, '0')}`;
+    } else {
+      suffix = `_${cryptoRandomString({ length: 5 }).toUpperCase()}`;
+    }
+    yield {
```
charleshuang80, line 61: I think this yield block could use a bit of explanation similar to the other generator function below.

scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,439 @@
...
+    filesPerGranule,
+    granulesPerBatch,
+    executionsPerBatch,
+    models,
+    variance,
+    swallowErrors,
+  });
+  await pMap(
+    iterableParamGenerator,
+    // this lambda function swallows uploadDataBatch's return to prevent ballooning memory use
```
charleshuang80, line 280: that's interesting and cool

scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,439 @@
...
+ *   files: number,
+ *   granulesPerBatch: number
+ *   executionsPerBatch: number
+ *   collections: number
+ *   concurrency: number
+ *   variance: boolean
+ *   swallowErrors: boolean
+ * }}
+ */
+const parseArgs = () => {
```
charleshuang80, line 323: two things I thought about after our discussion on the options/args:
1. what are you doing to handle any options passed in that don't match the ones available?
2. to handle the close but not quite there options for things like executions_per_granules that you originally had, maybe you can grab those and give an error message like "did you mean executionsPerGranule?"

neither of these are completely necessary, maybe nice to have. Just something that came to mind after our discussion.

scripts/generate_records/README.md
etcart: https://github.com/nasa/cumulus/pull/3679/commits/b66b6f37ffe83d8936ac37bb4fc6a1cb3f2b6a69

scripts/generate_records/generate_db_records.js
etcart: https://github.com/nasa/cumulus/pull/3679/commits/f53114f0287cb4f1cd775df7fe492b2507a33e7b

scripts/generate_records/generate_db_records.js
etcart: https://github.com/nasa/cumulus/pull/3679/commits/3ba1b2e8c2befb1eb88aa179b05a79acfe6b9ebd
yea this is a leftover that's gone now in shifting to fake collections

scripts/generate_records/generate_db_records.js
etcart: I think because this is a developer tool I'm not gonna think too hard about it

scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,403 @@
...
+      models,
+      swallowErrors,
+    };
+
+    if (variance) {
+      //asking for variance adds some noise to batch executions vs granules
+      let _granulesPerBatch = 1;
+      let _executionsPerBatch;
+      for (let i = 0; i < numberOfGranules; i += _granulesPerBatch) {
+        _granulesPerBatch = granulesPerBatch + randomInt(6);
```
charleshuang80, line 169: I thought the utilization of the randomInt, was dependent on if variance is true. if that's the case, there should be some sort of if statement here, and I would think the _granulesPerBatch and _executionsPerBatch above should have the default value set to the non underscored value.

scripts/generate_records/generate_db_records.js
charleshuang80: totally fair, just wanted to share what I thought of

scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,403 @@
...
+      variance,
+      swallowErrors
+    );
+  }
+};
+
+if (require.main === module) {
+  main(
+  ).then(
+    (ret) => ret
```
charleshuang80, line 391: I've been debating if there should be a comment explaining why this line is necessary

scripts/generate_records/tests/test-db_record_loaders.js
```diff
@@ -0,0 +1,151 @@
...
+    t.context.knex,
+    granuleCumulusIds,
+    executionCumulusIds,
+    geModel
+  );
+  await Promise.all(geUploads.map(async (granuleExecution) => {
+    t.true(await geModel.exists(t.context.knex, granuleExecution));
+  }));
+});
+test('loadCollection() adds collections', async (t) => {
```
charleshuang80, line 128: I think these two 'redundant' tests are fine as is. An alternative for your consideration is to add a comment somewhere saying that loadCollection and loadProvider is not needed because it is required/used by __ and ___. But I suppose down the road if there is some change to the two functions that won't be fully tested by the other tests, it would be good to have these to modify.

scripts/generate_records/generate_db_records.js
etcart: I don't really know what to say except "js demands that I do it"

scripts/generate_records/tests/test-db_record_loaders.js
etcart: yea, it felt like a good thing to have even if only for that
unfortunately those pgmodels don't expose a way to check if it exists????

scripts/generate_records/db_record_loaders.js
```diff
@@ -0,0 +1,266 @@
...
+  collectionCumulusId,
+  executionCount,
+  model,
+  swallowErrors = false
+) => {
+  const executionCumulusIds = [];
+  for (let i = 0; i < executionCount; i += 1) {
+    const execution = fakeExecutionRecordFactory({ collection_cumulus_id: collectionCumulusId });
+    try {
+      const [executionOutput] = await model.upsert(knex, execution);
```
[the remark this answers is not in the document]
etcart: this is a great idea, I thought I couldn't do this because of potential unique-id conflicts, but having upped the length of random strings and wrapped inserts with a try catch this is dramatically faster, cleaner, errors are caught and also vanishingly rare

scripts/generate_records/db_record_loaders.js
```diff
@@ -0,0 +1,266 @@
...
+  model,
+  swallowErrors = false
+) => {
+  const granuleCumulusIds = [];
+  for (let i = 0; i < granuleCount; i += 1) {
+    const granule = /** @type {PostgresGranule} */(fakeGranuleRecordFactory({
+      granule_id: randomString(6),
+      collection_cumulus_id: collectionCumulusId,
+      provider_cumulus_id: providerCumulusId,
+      status: 'completed',
```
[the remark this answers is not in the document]
etcart: https://github.com/nasa/cumulus/pull/3679/commits/1afff01b557ecdcff2dc2b27a7878bf5b603bba6

scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,403 @@
...
+        bar.update(i);
+        // this passes out an object each time pMap (or another iteration) asks for the next index
+        // without holding onto it in memory
+        yield standardParams;
+      }
+    }
+
+    bar.stop();
+  }
+  const clujedIterable = {};
```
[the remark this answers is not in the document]
etcart: pMap demands an object which is iterable.
a generator function is not iterable, so we need to make an object and define it's iteration behavior to be a generator function.
the reason for the generator function is to not build an array and not hold a potentially massive set of objects, this way its only holding onto the params for the threads currently running

scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,396 @@
...
+      providerCumulusId,
+      filesPerGranule,
+      granulesPerBatch,
+      executionsPerBatch,
+      models,
+      swallowErrors,
+    };
+
+    //asking for variance adds some noise to batch executions vs granules
+    let _granulesPerBatch = 1;
```
charleshuang80, line 165: I understand wanting to start this at 1, but now if there's no variance, `_granulesPerBatch` will always be 1, is that what you want?

scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,396 @@
...
+        s: 'swallowErrors',
+      },
+      default: {
+        collections: process.env.COLLECTIONS || 1,
+        files: process.env.FILES || 1,
+        granulesK: process.env.GRANULES_K || 10,
+        executionsPerGranule: process.env.EXECUTIONS_PER_GRANULE || '2:2',
+        variance: process.env.VARIANCE || false,
+        concurrency: process.env.CONCURRENCY || 1,
+        swallowErrors: process.env.SWALLOW_ERRORS || true,
```
charleshuang80, line 323: I don't think you need to change anything, but just wanted to note that I feel a bit odd about the default for swallowErrors is true if nothing is passed in as an option, and it is set that way here, _yet_ everywhere else, swallowErrors is default set to false.

etcart: I've solved this by rethinking how this works, it's not set as a cmd line arg but just using pMap's stopOnError arg. (means whole batches can be lost if an error happens, but we ended up doing really small batches for high concurrency in practice so a potential loss like that is a drop in the bucket
https://github.com/nasa/cumulus/pull/3679/commits/285a0733acfb78aeb9f59bdf016298739686ef8e

scripts/generate_records/generate_db_records.js
etcart: good catch, this was a mistake
[error fix in batch params](https://github.com/nasa/cumulus/pull/3679/commits/01273d2e1224e806d547f1c82a8013415db60fd5)

scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,387 @@
+// @ts-check
+/* eslint-disable no-await-in-loop */
```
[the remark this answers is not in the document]
etcart: yea, it was removed from the loaders file, but there is still one place it's needed in here

scripts/generate_records/generate_db_records.js
```diff
@@ -0,0 +1,387 @@
...
+    granulesPerBatch,
+    executionsPerBatch,
+    collections,
+    variance,
+    concurrency,
+    swallowErrors,
+  } = parseArgs();
+  process.env.dbMaxPool = concurrency.toString();
+  const knex = await getKnexClient();
+  for (let collectionNumber = 0; collectionNumber < collections; collectionNumber += 1) {
```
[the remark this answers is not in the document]
etcart: this doesn't work because I don't have a way to batch the outermost loop of granules within collections, I nee to loop over the collections and don't have an insert function for sub-collections, or a feasible way to make one that I've been able to come up with

charleshuang80 (approved):

charleshuang80:
one last comment, adding here for posterity
I don’t know if it’s necessary immediately, but I think maybe with the next change/improvement. I see you have that last test that tests uploadDBGranules uploads a pile of entries, but you don’t check how many are actually added. Also, there’s no like “end to end” test that invokes the script with a set of arguments and ensures that a certain number of collections, granules, executions, files are all added to the DB.
Again, I think the tests we have now are fine since this isn’t a user facing thing. But as we add to it I think that sort of test would be helpful to have as it gets closer to something we’ll use quite a bit for setting up our testing env(s)

PR #3689, opened 2024-06-17
Excludes aws-sdk v3 from webpack packages

paulpilone:

Excludes @aws-sdk v3 as part of the webpack packaging since we're now requiring Node v20 and don't want to include all of the SDK packages with our code.

I did this by running through the opposite PR (removal of v2): https://github.com/nasa/cumulus/pull/3618

This change is necessary because after attempting to publish v18.3.0 some of the packages were too large to publish to npm.

## Changes

* Excludes @aws-sdk v3 from lambda packages.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/example/cumulus-tf/variables.tf
+++ b/example/cumulus-tf/variables.tf
@@ -362,7 +362,7 @@ variable "cumulus_process_activity_version" {
 variable "ecs_task_image_version" {
   description = "docker image version to use for Cumulus hello world task"
     type = string
-    default = "1.9.0"
+    default = "2.1.0-alpha"
 }
 
 variable "cumulus_test_ingest_image_version" {
--- a/example/cumulus-tf/variables.tf
+++ b/example/cumulus-tf/variables.tf
@@ -350,7 +350,7 @@ variable "rds_admin_access_secret_arn" {
 variable "async_operation_image_version" {
   description = "docker image version to use for Cumulus async operations tasks"
   type = string
-  default = "49"
+  default = "52"
 }
 
 variable "cumulus_process_activity_version" {
--- a/example/cumulus-tf/variables.tf
+++ b/example/cumulus-tf/variables.tf
@@ -362,7 +362,7 @@ variable "cumulus_process_activity_version" {
 variable "ecs_task_image_version" {
   description = "docker image version to use for Cumulus hello world task"
     type = string
-    default = "2.1.0-alpha"
+    default = "2.1.0"
 }
 
 variable "cumulus_test_ingest_image_version" {
--- a/packages/api/ecs/async-operation/Dockerfile
+++ b/packages/api/ecs/async-operation/Dockerfile
@@ -17,3 +17,5 @@ RUN cp -a /var/runtime/node_modules/. /home/task/node_modules/
 COPY index.js /home/task/
 
 CMD [ "node", "--harmony", "index.js" ]
+
+ENTRYPOINT [ ]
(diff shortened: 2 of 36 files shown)


paulpilone:
I tested that this regex works by running `npm run package` before and after the change and comparing the zip sizes. e.g. in the `tasks/sync-granule` lambda before excluding v3:

```
8230652 Jun 17 09:58 lambda.zip
```

and then after:

```
4704904 Jun 17 10:15 lambda.zip
```

Jkovarik (approved):
Looks good!

example/cumulus-tf/variables.tf
```diff
@@ -362,7 +362,7 @@ variable "cumulus_process_activity_version" {
 variable "ecs_task_image_version" {
   description = "docker image version to use for Cumulus hello world task"
     type = string
-    default = "1.9.0"
+    default = "2.1.0-alpha"
```
paulpilone, line 365: This will need to be updated once the final `cumulus-ecs-task` docker image is published.

paulpilone:
This is now passing CI with the updates to the `async_operations` and `cumulus-ecs-task` docker images: https://ci.earthdata.nasa.gov/browse/CUM-CBA3964-7

Related `cumulus-ecs-task` PR: https://github.com/nasa/cumulus-ecs-task/pull/66

packages/api/ecs/async-operation/package.json
```diff
@@ -22,10 +22,10 @@
     "coverage": "python ../../../../scripts/coverage_handler/coverage.py"
   },
   "dependencies": {
-    "@cumulus/aws-client": "18.2.0",
-    "@cumulus/db": "18.2.0",
-    "@cumulus/es-client": "18.2.0",
-    "@cumulus/logger": "18.2.0",
+    "@cumulus/aws-client": "18.3.0",
```
paulpilone, line 25: We got lucky here in that the dependency tree for this Docker image were all published as part of the incomplete 18.3.0 release. For the sake of getting things unblocked it made sense to just bump these dependencies to 18.3.0 instead of trying to publish some sort of alpha or version specific to this PR.

When 18.3.1 is released we should revisit this Dockerfile and bump the dependencies. 18.3.1 will not include aws-sdk v3 in the Lambda package so we want to verify everything still works - but it should.

npauzenga: So temporarily bump to 18.3, merge to main, do the 18.3.1 release and bump these? That makes sense to me. Just want to be sure to catch it during the release as I don't know if npm update will 🤔. Either way, sounds like a plan.

paulpilone: Yup that was my plan. It's the same thing we'd have to do if we pushed an "alpha" version it just so happened we got lucky and could use what was published of 18.3.0.

packages/api/ecs/async-operation/Dockerfile
```diff
@@ -1,18 +1,21 @@
...
 COPY package.json /home/task/
 RUN npm install
 
+RUN cp -a /var/runtime/node_modules/. /home/task/node_modules/
+
 COPY index.js /home/task/
 
 CMD [ "node", "--harmony", "index.js" ]
+
+ENTRYPOINT [ ]
```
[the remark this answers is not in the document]
paulpilone: Yes, the AWS Lambda ENTRYPOINT is set to a Lambda specific script. We obviously don't want to run that and a blank ENTRYPOINT in our Dockerfile overrides the other one.

PR #3691, opened 2024-06-20
moved to postgres functionality in cleanup executions

etcart:
**Summary:** update cleanExecutions Lambda to correctly interact with the database and clean up executions. write tests that more thoroughly test the internal workings

Addresses [CUMULUS-3385 Cumulus P3/V16+ is not correctly running cleanExecutions](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3385)

## Changes

* cleanExecutions lambda rewritten to interface with Postgres
* tests added that put executions in test DB and check direct expected functionality
* added UPDATE_LIMIT env configuration to allow batch processing expecting large backlog

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


Jkovarik:
@etcart  I haven't pulled this for a formal review - this looks to be handling the RDS portion of the fix, is the ES component of the solution intended as another PR?

etcart:
I had misunderstood and thought that this was meant to be independent of that

etcart:
ok, now with ES and Postgres both. I haven't solved the problem of "too many executions" since that's a separate ticket, but I can't if we'd prefer to do it here

packages/api/tests/lambdas/test-cleanExecutions.js
```diff
@@ -1,58 +1,470 @@
...
-
-const removeRecordsStub = sinon.spy();
+  t.context.knex = knex;
+  t.context.knexAdmin = knexAdmin;
+  const { esIndex, esClient, searchClient } = await createTestIndex();
+  t.context.esIndex = esIndex;
+  t.context.esClient = esClient;
+  t.context.searchClient = searchClient;
+  const records = [];
+  for (let i = 0; i < 10; i += 1) {
```
npauzenga, line 29: A little redundancy in test setup isn't terrible but there's overlap with another PR that's doing a very similar setup. I wonder if a test helper is in order.

https://github.com/nasa/cumulus/pull/3692/files#diff-723ecf957415d7198e55ff736cc0c60ca586962793d7b18d424bb8ab45dc3ae9R86

packages/api/tests/lambdas/test-cleanExecutions.js
```diff
@@ -1,58 +1,470 @@
...
-  process.env.completeExecutionPayloadTimeout = 100;
-  process.env.nonCompleteExecutionPayloadTimeout = 50;
-  await cleanExecutionPayloads(executionModel);
-  t.truthy(removeRecordsStub.calledWith(100, 50, false, false));
+const pgPayloadsEmpty = (entry) => !entry.final_payload && !entry.orginal_payload;
+
+const esPayloadsEmpty = (entry) => !entry.finalPayload && !entry.orginalPayload;
+
+test.serial('handler() handles complete expiration', async (t) => {
+  const env = clone(process.env);
```
npauzenga, line 72: Do we need to clone this? What happens if we just update in place?

packages/api/tests/lambdas/test-cleanExecutions.js
```diff
@@ -1,58 +1,470 @@
...
+    0,
+    true,
+    false
+  );
+  process.env.completeExecutionPayloadTimeoutDisable = 'false';
+  process.env.nonCompleteExecutionPayloadTimeoutDisable = 'true';
+  process.env.completeExecutionPayloadTimeout = expirationDays;
+
+  await handler();
+  await sleep(5000);
```
npauzenga, line 91: Hmm this seems like it could be fragile. Is there a better way to ensure the handler has completed?

packages/api/tests/lambdas/test-cleanExecutions.js
```diff
@@ -1,58 +1,470 @@
...
+  const model = new ExecutionPgModel();
+  let massagedPgExecutions = await Promise.all(
```
npauzenga, line 93: What do we mean by "massaged" here? Are they modified in some way below?

packages/api/tests/lambdas/test-cleanExecutions.js
```diff
@@ -1,58 +1,470 @@
...
+  );
+
+  massagedPgExecutions.forEach((massagedExecution) => {
+    if (massagedExecution.updated_at < expirationDate && massagedExecution.status === 'completed') {
+      t.true(pgPayloadsEmpty(massagedExecution));
+    } else {
+      t.false(pgPayloadsEmpty(massagedExecution));
+    }
+  });
+  let massagedEsExecutions = await t.context.searchClient.query({
```
npauzenga, line 106: I'm wondering if it would be better to split this test out into a couple assertions. Maybe one for PG and one for ES? Just thinking about removing ES in the near future. Might make it easier to just delete a whole test?

packages/api/tests/lambdas/test-cleanExecutions.js
```diff
@@ -1,58 +1,470 @@
...
+  process.env = env;
 });
 
-test.serial('Function throws error if passed invalid complete timeout', async (t) => {
-  process.env.nonCompleteExecutionPayloadTimeout = 100;
-  process.env.completeExecutionPayloadTimeout = 'notaninteger';
-  let actual;
-  await cleanExecutionPayloads(executionModel).catch((error) => {
-    actual = error.message;
+test.serial('handler() throws errors when misconfigured', async (t) => {
```
npauzenga, line 355: Nit: my preference would be to assert each known possible configuration issue separately. That's preference though so dealer's choice 😄

packages/api/tests/lambdas/test-cleanExecutions.js
```diff
@@ -1,58 +1,470 @@
...
+  process.env = env;
 });
 
-test.serial('Function throws error if passed invalid non-complete timeout', async (t) => {
-  process.env.nonCompleteExecutionPayloadTimeout = 'notaninteger';
-  process.env.completeExecutionPayloadTimeout = 100;
-  let actual;
-  await cleanExecutionPayloads(executionModel).catch((error) => {
-    actual = error.message;
+test.serial('handler() iterates through data in batches when updateLimit is set low', async (t) => {
```
npauzenga, line 379: Lower than what?

packages/api/tests/lambdas/test-cleanExecutions.js
npauzenga: Also it looks like there's a separate test file for ES executions. Do we need _any_ ES assertions here?

packages/api/tests/lambdas/test-clean_pg_executions.js
```diff
@@ -0,0 +1,196 @@
...
+test.afterEach.always(async (t) => {
+  await destroyLocalTestDb({
+    knex: t.context.knex,
+    knexAdmin: t.context.knexAdmin,
+    tesetDbName: t.context.testDbName,
+  });
+});
+
+test('getExpirablePayloadRecords()', async (t) => {
+  const executions = [
```
npauzenga, line 38: Could probably create a function to do this but not the end of the world.

packages/api/tests/lambdas/test-clean_pg_executions.js
```diff
@@ -0,0 +1,196 @@
+'use strict';
```
npauzenga, line 1: What if we rolled these tests into test-cleanExecutions? Would that make sense? Again thinking about ES going away so eventually we'll be left with:

test-cleanExecutions
test-clean_pg_executions

which isn't much of a distinction at that point.

packages/api/tests/lambdas/test-clean_pg_executions.js
```diff
@@ -0,0 +1,196 @@
...
+    inserted[0],
+    inserted[2],
+    inserted[7],
+    inserted[9],
+    inserted[10],
+    inserted[12],
+  ];
+
+  t.deepEqual(expected, expirable);
+  expirable = await getExpirablePayloadRecords(t.context.knex, moment().subtract(0, 'days').toDate(), 100);
```
npauzenga, line 106: Nit: preference to use constants and not reassign `expirable` / `expected`

packages/api/tests/lambdas/test-clean_pg_executions.js
```diff
@@ -0,0 +1,196 @@
...
+
+test('cleanupExpiredPGExecutionPayloads() cleans up expired payloads', async (t) => {
+  const env = clone(process.env);
+  process.env = localStackConnectionEnv;
+  process.env.PG_DATABASE = t.context.testDbName;
+  const completeExpirationDays = 3;
+  const nonCompleteExpirationDays = 5;
+
+  const records = [];
+  for (let i = 0; i < 10; i += 1) {
```
npauzenga, line 145: See previous comment re: creating test executions. Might be able to either share or create a common utility.

packages/db/src/models/execution.ts
```diff
@@ -81,6 +83,32 @@ class ExecutionPgModel extends BasePgModel<PostgresExecution, PostgresExecutionR
       });
     return executions;
   }
+
+  /**
+   * Fetches multiple items from Postgres that contain a payload before a date
+   * this is included unused as a part of PR to ask, does this belong here?
```
npauzenga, line 89: It's unused? In that case no 😄 

It's been a long time since we've had this debate about models BUT my opinion is that anything that's constrained to querying the datastore for a specific model is ok to have in the model vs. a lib. I'd say this belongs but you may find other opinions elsewhere 😄

scripts/generate_records/generate_db_executions.js
```diff
@@ -0,0 +1,217 @@
...
+  }
+) => {
+  await loadExecutions(
+    knex,
+    collectionCumulusId,
+    executionsPerBatch,
+    model,
+    {
+      updated_at: moment().subtract(randomInt(20), 'days').toDate(),
+      original_payload: Math.random() < 0.4 ? { a: 'b' } : null,
```
npauzenga, line 47: I wonder if we want to make this configurable. As in, maybe I just want to generate consistent executions and not have the random payloads. For this story it makes a ton of sense but as a more generic tool maybe have it optional.

CHANGELOG.md
```diff
@@ -105,6 +105,8 @@ operations (e.g. `PREFIX-AsyncOperationEcsLogs`).
     to granules table
 
 ### Added
+- **CUMULUS-3385**
+  - added upload_db_executions to dump large scale postgres executions
```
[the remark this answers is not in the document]
npauzenga: Thinking about it a bit more, maybe this doesn't make sense. They're separate use cases. Yours is for a larger-scale test in an environment other than local (though could be used locally) and the other is just really for unit tests. 

Keeping this comment here for kicks but I think it might be distinct enough to be moot.

npauzenga (changes requested):
Looks great @etcart! I added a few questions. Feel free to respond for clarity if anything isn't... clear.

CHANGELOG.md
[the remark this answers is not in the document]
etcart: yea, I have fixed the changeling, but the intent of this scrip tis different

packages/api/lambdas/cleanExecutions.js
```diff
@@ -1,48 +1,302 @@
...
+ * @returns {{
+ *  laterExpiration: Date,
+ *  completeExpiration: Date,
+ *  nonCompleteExpiration: Date
+ *  earlierExpiration: Date
+ * }}
+ */
+const getExpirationDates = (
+  completeTimeoutDays,
+  nonCompleteTimeoutDays,
```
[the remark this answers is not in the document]
etcart: I admit Ididn't super like it but, this comes from the original cleanExecutions. I've tried to replicate the cleanExecutions functionality as it previously existed, and that included being able to set the environment variables
process.env.completeExecutionPayloadTimeout
process.env.nonCompleteExecutionPayloadTimeout
process.env.completeExecutionPayloadTimeoutDisable
process.env.nonCompleteExecutionPayloadTimeoutDisable

to do.. this stuff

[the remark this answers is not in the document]
etcart: I wouldn't be opposed to simplifying it down if there's will to do it, but I don't want to break existing usages

packages/api/lambdas/cleanExecutions.js
```diff
@@ -1,48 +1,302 @@
...
+    runNonComplete
+  );
+  const knex = await getKnexClient();
+  const updateLimit = Number(process.env.UPDATE_LIMIT || 10000);
+  const executionModel = new ExecutionPgModel();
+  const executionRecords = await getExpirablePayloadRecords(
+    knex,
+    laterExpiration,
+    updateLimit
+  );
```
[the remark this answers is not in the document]
etcart: *if* they are not out of sync, they should remain in sync, although existing synchronicity issues will propagate. I wanted to keep them separate because successfully cleaning up everything seemed the priority. I dont' want to leave es records hanging around just because they aren't in pg for some reason.
on repeated runs the intention is that this gets *everything*

packages/api/lambdas/cleanExecutions.js
```diff
@@ -1,48 +1,302 @@
...
+  if (executionRecords.length === updateLimit) {
+    log.warn(`running cleanup for ${updateLimit} out of maximum ${updateLimit} executions. more processing likely needed`);
+  }
+  const concurrencyLimit = Number(process.env.CONCURRENCY || 100);
+  const limit = pLimit(concurrencyLimit);
+  const wipedPayloads = {
+    original_payload: null,
+    final_payload: null,
```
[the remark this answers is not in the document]
etcart: that's how the original cleanExecutions lambda worked, so I assume yes. if you have reason to think it's only original point me to it but keeping finals around will not eliminate the level of bloat we need to get rid of

packages/db/src/models/execution.ts
etcart: I've since removed this, I meant to before requesting PR, I decided the code was too specific to this script to belong in a standard model

scripts/generate_records/generate_db_executions.js
etcart: yea the exact proportion seemed like it needed to be arbitrary, this is about 2/3 records will have one or the other. I'm not real sure how to reasonably make this an option, but this is a dev tool so I'm assuming you can tweak it if you need. it's such a small amount of data that it won't really affect you if you don't need it

packages/api/lambdas/cleanExecutions.js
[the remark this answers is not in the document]
npauzenga: What if a delete for the ES record fails? I think it would then have successfully deleted from PG and not from ES, thus causing a sync issue that didn't exist prior?

packages/api/lambdas/cleanExecutions.js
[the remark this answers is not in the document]
npauzenga: Thanks. I think the mission of "get existing intentions actually working" is fine here.

packages/db/src/models/execution.ts
npauzenga: commit?

scripts/generate_records/generate_db_executions.js
npauzenga: I mean, you could add a flag that either adds this randomly-distributed payload or static payloads for each. 

For future use it might be nice to have realistic-looking payloads rather than { a: 'b' } as well. These types of scripts to generate large amounts of data are something we're needing more and more.

That said, I hear you that this doesn't need to do everything here and we can iterate as needed.

packages/api/tests/lambdas/test-clean_pg_executions.js
etcart: I believe this should be taken care of with some consolidation of tests

packages/api/tests/lambdas/test-clean_pg_executions.js
etcart: I believe this should be taken care of with some consolidation of tests

packages/api/lambdas/cleanExecutions.js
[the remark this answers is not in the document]
etcart: I've more tightly synced these in a way that I think solves this problem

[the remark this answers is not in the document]
npauzenga: I think this does make sense. `wipePayloads` will try to delete the ES payloads and (if successful) the PG payload. If the PG update fails, we still would have wiped the ES payload. I'm not sure it would be worth going to the trouble of reverting ES like we do in other updates. In the case that a PG update fails, we fail the promise. 

I think this makes sense. It may be worth bringing this up in a team setting to see if anyone disagrees with this approach. Being cautious because this will be an operation on potentially _many_ records.

PR #3728, opened 2024-07-12
CUMULUS-3608: Adds variable for sqs message consumer watcher message and time limits

paulpilone:

Addresses [CUMULUS-3608: Make sqs_message_consumer_watcher messageLimit configurable](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3608)

## Changes

* Exposes message and time limit Terraform vars for sqs_message_consumer_watcher Lambda

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,8 @@ degraded execution table operations.
   - Added `Bulk Execution Delete` migration type to async operations types
 - **CUMULUS-3742**
   - Script for dumping data into postgres database for testing and replicating issues
+- **CUMULUS-3608**
+  - Exposes variables for sqs_message_consumer_watcher messageLimit and timeLimit configurations.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


paulpilone:
I reviewed the sqs message consumer watcher Lambda code and didn't find anything obvious that should stop users from being able to configure these values. At most it would increase calls to S3; however, S3 actions are small potatoes when compared to the workflows this function ultimately kicks off.

I exposed both variables because it makes sense if you're going to increase the message limit you might also want to increase the timeout.

paulpilone:
Note of concern: exposing the timeout could result in the user setting that to a value greater than the actual Lambda timeout. I suppose this is something we should provide guidance on or else not expose the timeout at all.

npauzenga (changes requested):
This looks great @paulpilone. I think you're right and the trick here is guidance. We should probably have some rough idea of what numbers are safe or how to evaluate that. Also a note about the Lambda timeout to your earlier point would probably make sense.

npauzenga (approved):
Looks great @paulpilone! Thanks for adding details to the new params. I think we can leave it there and explicitly tell DAACs that are interested where to look.

PR #3746, opened 2024-07-25
CUMULUS-3238: Remove ElasticSearch dependency from Collection POST, PUT, and DEL endpoints

Nnaga1:

## Changes

* ...

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,13 +10,21 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - **CUMULUS-3238**
   - Removed elasticsearch dependency from collections endpoint
 - **CUMULUS-3239**
-  - Updated `execution` list api endpoint and added `ExecutionSearch` class to query postgres
+  - Updated `executions` list api endpoint and added `ExecutionSearch` class to query postgres
+- **CUMULUS-3240**
+  - Removed Elasticsearch dependency from `executions` endpoints
 - **CUMULUS-3639**
   - Updated `/collections/active` endpoint to query postgres
 - **CUMULUS-3640**
   - Removed elasticsearch dependency from granules endpoint
 - **CUMULUS-3641**
   - Updated `collections` api endpoint to query postgres instead of elasticsearch except if `includeStats` is in the query parameters
+- **CUMULUS-3642**
+  - Adjusted queries to improve performance:
+    - Used count(*) over count(id) to count rows
+    - Estimated row count for large tables (granules and executions) by default for basic query
+  - Updated stats summary to default to the last day
+  - Updated ExecutionSearch to not include asyncOperationId by default
 - **CUMULUS-3688**
   - Updated `stats` api endpoint to query postgres instead of elasticsearch
 - **CUMULUS-3689**
--- a/packages/api/endpoints/executions.js
+++ b/packages/api/endpoints/executions.js
@@ -17,12 +17,9 @@ const {
   CollectionPgModel,
   ExecutionPgModel,
   translatePostgresExecutionToApiExecution,
-  createRejectableTransaction,
   ExecutionSearch,
 } = require('@cumulus/db');
 const { deconstructCollectionId } = require('@cumulus/message/Collections');
-const { deleteExecution } = require('@cumulus/es-client/indexer');
-const { getEsClient, Search } = require('@cumulus/es-client/search');
 
 const { zodParser } = require('../src/zod-utils');
 const { asyncOperationEndpointErrorHandler } = require('../app/middleware');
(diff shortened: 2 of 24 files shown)


packages/api/endpoints/collections.js
```diff
@@ -238,44 +223,26 @@ async function del(req, res) {
...
     if (error instanceof RecordDoesNotExist) {
-      if (!(await esCollectionsClient.exists(collectionId))) {
-        log.info('Collection does not exist in Elasticsearch and PostgreSQL');
-        return res.boom.notFound('No record found');
-      }
-      log.info('Collection does not exist in PostgreSQL, it only exists in Elasticsearch. Proceeding with deletion');
-    } else {
-      throw error;
+      log.info(`Collection does not exist in PostgreSQL. Failed to delete collection with name ${name} and version ${version}`);
+      return res.boom.notFound('No record found');
```
npauzenga, line 239: I think we still need the `else throw error`, right? If the error is not RecordDoesNotExist, I believe this will swallow the error unless there's a catch further up I'm not looking at (which is possible).

packages/api/tests/endpoints/collections/delete-collection.js
```diff
@@ -386,112 +305,6 @@ test.serial('Attempting to delete a collection with an associated rule does not
   ));
 });
 
-test.serial('del() does not remove from Elasticsearch or publish SNS message if removing from PostgreSQL fails', async (t) => {
```
npauzenga, line 389: This test might still be valid without ES? e.g. `del() does not publish SNS message if removing from PostgreSQL fails`

packages/api/tests/endpoints/collections/update-collection.js
```diff
@@ -435,12 +352,11 @@ test.serial('PUT returns 400 for version mismatch between params and payload', a
   t.falsy(record);
 });
 
-test.serial('PUT does not write to Elasticsearch or publish SNS message if writing to PostgreSQL fails', async (t) => {
+test.serial('PUT does not write publish SNS message if writing to PostgreSQL fails', async (t) => {
```
npauzenga, line 355:
```suggestion
test.serial('PUT does not publish SNS message if writing to PostgreSQL fails', async (t) => {
```

npauzenga (changes requested):
This looks great @Nnaga1! Just getting into the ES removal stuff so this is helpful.

packages/api/endpoints/collections.js
npauzenga: All good. We can mark things as WIP and/or put them in Draft status in the future but not a big deal.

packages/api/endpoints/collections.js
```diff
@@ -21,11 +20,6 @@ const {
   CollectionSearch,
 } = require('@cumulus/db');
 const CollectionConfigStore = require('@cumulus/collection-config-store');
-const { getEsClient, Search } = require('@cumulus/es-client/search');
-const {
-  indexCollection,
```
etcart, line 26: I want to be clear that the intention is to leave this function (indexCollection) in the code base?
I see that it's used in the api to provide ES interfaces, is the plan to leave the api still able to interact with ES so people can manage their cluster even while we're not adding new data to it?

packages/api/endpoints/collections.js
```diff
@@ -176,7 +166,6 @@ async function put(req, res) {
   const {
     collectionPgModel = new CollectionPgModel(),
     knex = await getKnexClient(),
-    esClient = await getEsClient(),
     collectionConfigStore = new CollectionConfigStore(
       process.env.system_bucket,
```
etcart, line 170: this is slight scope creep so *nit* but
getRequiredEnvVar('system_bucket', process.env) here, and elsewhere in this file
(107, 108, 171, 172, 228, 229) would shift us towards better typing while you're in this file

packages/api/tests/endpoints/collections/delete-collection.js
```diff
@@ -161,6 +145,54 @@ test('Attempting to delete a collection without an Authorization header returns
   );
 });
 
+test.serial('del() does not publish a SNS message if removing from PostgreSQL fails', async (t) => {
```
etcart, line 148:
```suggestion
test.serial('del() does not publish an SNS message if removing from PostgreSQL fails', async (t) => {
```

packages/api/tests/endpoints/collections/update-collection.js
```diff
@@ -270,49 +236,7 @@ test.serial('PUT replaces an existing collection and correctly removes fields',
   });
 });
 
-test.serial('PUT replaces an existing collection in PG with correct timestamps', async (t) => {
```
etcart, line 273: why was this cut out?

etcart (changes requested):
mostly this is great, I'd like to clarify a few things to be sure they are what's intended

packages/api/endpoints/collections.js
etcart: yes exactly

etcart (approved):

etcart (approved):
this all looks good, ready to go.
my concerns and Nates have both been addressed and I don't see anything else that raises concerns

PR #3747, opened 2024-07-25
CUMULUS-3770: Investigate and test phased release approach for ElasticSearch Removal - Executions

Nnaga1:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


npauzenga:
@Nnaga1 is this reviewable? I don't see a WIP tag but it looks like it might be in progress still?

packages/db/src/search/field-mapping.ts
etcart, line 1: is it not possible here to re-use the Postgres to api translation tools here?
e.g. packages/db/src/translate/executions.ts
or if the purpose here is to be able to translate single fields, should that go in the translate/exeuctions.ts file

packages/db/src/search/ExecutionSearch.ts
etcart, line 1: does this Postgres search functionality not belong in the Postgres ExecutionModel?

packages/db/.nycrc.json
etcart, line 1: I'd like to see increased testing here. the new BaseSearch file has notably low coverage

packages/api/endpoints/executions.js
```diff
@@ -257,7 +234,7 @@ async function del(req, res) {
 async function searchByGranules(req, res) {
   const payload = req.body;
   const knex = await getKnexClient();
-  const granules = await getGranulesForPayload(payload, knex);
```
etcart, line 260: I can't find the getGranulesForPayload change,  is this just an update that got missed in a previous change?

etcart (changes requested):
Im have some questions, but the biggest one is this:
was there a conversation I missed where we decided to go for a split "Search" Postgres functionality that's distinct from main-line Postgres functionality and imitates ES

I'm concerned about ending up in a state a year from now where, for reasons that only make sense as a legacy, we have a "Postgres" bit and a "Postgres pretending to be elasticsearch" bit separated from each other

PR #3752, opened 2024-07-30
Temporarily allowlist fast-xml-parser

paulpilone:
Temporarily allowlist `fast-xml-parser` so as not to block work being merged to `master`. See https://bugs.earthdata.nasa.gov/browse/CUMULUS-3808 and https://bugs.earthdata.nasa.gov/browse/CUMULUS-3817 for follow up work.

### DIFF
--- a/audit-ci.json
+++ b/audit-ci.json
@@ -12,6 +12,7 @@
     "cacheable-request",
     "http-cache-semantics",
     "lodash.pick",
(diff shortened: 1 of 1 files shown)


npauzenga (approved):
👍 Good to go. We'll allow-list this temporarily, issue a fix before releasing, and revert this change.

PR #3759, opened 2024-08-01
CUMULUS-3817 update versions across the board

etcart:
**Summary:** increases aws-sdk package versions to 3.621.0, post high vulnerability inherited from fast-xml-parser

Addresses [CUMULUS-3817: remove fast-xml-vulnerability](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3817)

## Changes

* updates all aws-sdk dependencies to version 3.621.0
* except for aws-sdk/types which is set to "^3.609.0" (the latest version
## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests

Note that: this doesn't actually remove the allowlist entry for this, although it *does* remove the offending packages.
this is because our audit tool (and I couldn't find a good way around this) checks the declared dependencies of its dependencies (i.e. @cumulus/aws-client@18.3.2 declares aws-sdk/client-s3@<3.614.2, rather than what the source code says is the aws-sdk/client-s3 version in @cumulus/aws-client) so even though the offending packages has been updated to a non-vulnerable version, *this will still show as a vulnerability in the audit* until we release the next version

there will need to be a follow ticket once 18.3.3 is release to simply cut that line from the allowlist


jennyhliu (approved):
LGTM👍
Please update the PR title with ticket number

PR #3760, opened 2024-08-02
[CUMULUS-3229] - Remove ElasticSearch from Rules Endpoints and Update LIST Query

npauzenga:
This PR replaces the elasticsearch queries in /rules LIST with postgres queries and removes elasticsearch from the other Rule endpoints.


packages/db/src/translate/rules.ts
```diff
@@ -6,27 +6,47 @@ import { CollectionPgModel } from '../models/collection';
...
+// Rule queries can include fields from external tables like provider and collection
+interface RuleRecordWithExternals extends PostgresRuleRecord {
+  provider?: string,
+  collectionName?: string,
+  collectionVersion?: string,
+}
+
 export const translatePostgresRuleToApiRule = async (
-  pgRule: PostgresRuleRecord,
+  pgRule: RuleRecordWithExternals,
```
npauzenga, line 17: I wanted to use the existing translate method, not a new one that's mostly a copy but without DB queries. To do that though the types get a little weird. Open to feedback.

packages/db/src/search/RuleSearch.ts
```diff
@@ -0,0 +1,109 @@
...
+  */
+  protected buildInfixPrefixQuery(params: {
+    countQuery: Knex.QueryBuilder,
+    searchQuery: Knex.QueryBuilder,
+    dbQueryParameters?: DbQueryParameters,
+  }) {
+    const { countQuery, searchQuery, dbQueryParameters } = params;
+    const { infix, prefix } = dbQueryParameters ?? this.dbQueryParameters;
+    if (infix) {
+      [countQuery, searchQuery].forEach((query) => query.whereLike(`${this.tableName}.name`, `%${infix}%`));
```
npauzenga, line 81: The `name` field appears to be correct but this was confusing. The dash uses `name` and I don't think there's another field that makes sense so I _think_ this is correct.

packages/db/src/search/field-mapping.ts
```diff
@@ -249,6 +249,9 @@ const providerMapping : { [key: string]: Function } = {
 };
 
 const ruleMapping : { [key: string]: Function } = {
+  arn: (value?: string) => ({
```
npauzenga, line 252: I think these single-word mappings might be unnecessary 🤔. If the property can be directly used as a postgres column, I probably don't need these...

packages/db/src/translate/rules.ts
etcart: I like the approach and agree that we shouldn't make near-dupe functions if we can get away with it.
*but* I don't like the idea that this function can and is intended to be used without ever actually calling 3/4 of its parameters.
I foresee outside functions having to pass around knex and PGmodels for no reason to satisfy this

etcart: is there a sensible way to write this so that knex is optional?
catch if we need to call it and didn't instantiate it and throw an error so a dev knows they misunderstood?

npauzenga: Yeah, that's fair. Maybe we do a separate method for translating without the db calls but we call that method from the existing `translatePostgresRuleToApiRule`. So You have `translatePostgresRuleToApiRule` that requires a knex object and calls something like `translatePostgresRuleToApiRuleWithoutExternals`. Then we could use `translatePostgresRuleToApiRuleWithoutExternals` separately as necessary.

I'll refactor a bit and let ya know. Feel free to disagree with the approach. Happy to chat.

npauzenga: Oh ha you're right. `translatePostgresGranuleToApiGranule` calls `translatePostgresGranuleToApiGranuleWithoutDbQuery`. Didn't think about that when I replied to Ethan's original comment.

So yes, the original idea was to avoid that and just have one method. The downside is that you're passing a knex object and other params that you don't need in this case. And it's a bit more complex. The upside is that it's only one method.

Maybe I should just do what you all did for granules and executions because I accidentally arrived at the same conclusion and I don't think we want rules to be a unicorn.

npauzenga: I've finished reinventing the wheel you invented earlier: https://github.com/nasa/cumulus/pull/3760/commits/fd8f5bfdefec61549adc5ded6c396fcf5e3dbc8c

packages/db/src/search/RuleSearch.ts
npauzenga: Yeah, it's in the dash and logically I don't think the others work either.

packages/db/src/search/RuleSearch.ts
```diff
@@ -0,0 +1,109 @@
...
+  * @returns translated api records
+  */
+  protected async translatePostgresRecordsToApiRecords(
+    pgRecords: PostgresRuleRecord[],
+    knex: Knex
+  ): Promise<Partial<RuleRecord>[]> {
+    log.debug(`translatePostgresRecordsToApiRecords number of records ${pgRecords.length} `);
+
+    const apiRecords = pgRecords.map(async (record) => {
+      const apiRecord = await translatePostgresRuleToApiRule(record, knex);
```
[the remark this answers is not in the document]
npauzenga: That might have been true that if the initial query returned undefined, this might've tried again. Good catch.

In any case, we're not doing the branch logic now 😄  https://github.com/nasa/cumulus/pull/3760/commits/fd8f5bfdefec61549adc5ded6c396fcf5e3dbc8c

packages/db/tests/search/test-RuleSearch.js
```diff
@@ -0,0 +1,271 @@
...
+  const queryStringParameters = {
+    limit: 50,
+    prefix: 'fakeRule-1',
+  };
+  const dbSearch = new RuleSearch({ queryStringParameters });
+  const response = await dbSearch.query(knex);
+  t.is(response.meta.count, 11);
+  t.is(response.results?.length, 11);
+});
+
```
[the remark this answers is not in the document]
npauzenga: Fair question. I'll go through these again. My thinking here was along these lines:

If we're testing the RuleSearch class for e.g. does it support a term search for a boolean field? Is that really implemented in the RuleSearch class? Or is that actually implemented in the BaseClass and tested there? Same gist for range, number field, etc.

Then the others appear to be testing the field-mappings and ensuring that the correct fields are added. That's... maybe good? It's a test to ensure I added the fields there. I was originally thinking that wasn't an especially useful test but I do see your point.

I think the biggest argument for adding these is "do what the other records have done". I completely buy that and will re-evaluate 😄

packages/api/tests/endpoints/test-rules.js
```diff
@@ -406,7 +410,46 @@ test.serial('search returns correct list of rules', async (t) => {
   t.is(newResults.length, 0);
 });
 
-test('GET gets a rule', async (t) => {
+test.serial('search returns the expected fields', async (t) => {
```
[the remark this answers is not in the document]
npauzenga: 👍 https://github.com/nasa/cumulus/pull/3760/commits/aa596a74a795d6ab21a3ebc7c4bd6b87c481cbed

packages/db/tests/search/test-RuleSearch.js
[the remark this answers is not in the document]
npauzenga: Oh you're right. I was assuming BaseSearch had tests and it doesn't. It does feel like a lot of these assertions belong there instead of the child classes but I get it now. BaseSearch is abstract and can't be instantiated directly so testing it would mean exporting it. 

I think in order to not rock the boat I'll take your suggestion and add tests to my child class. It's possible this could come up again though and might warrant a refactor at some point. That or there's yet more I'm missing 😄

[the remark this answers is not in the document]
npauzenga: https://github.com/nasa/cumulus/pull/3760/commits/2c4654ba583933c2f93fafb35d1bd9a225bb4b76 Added the tests mentioned above with one exception. There aren't _really_ number fields on Rules. https://github.com/nasa/cumulus/pull/3760/commits/2c4654ba583933c2f93fafb35d1bd9a225bb4b76

packages/db/src/search/field-mapping.ts
npauzenga: Turns out they are actually required, even if we're mapping e.g. `arn` to `arn` as-is. That's maybe something that could be refactored but leaving it to match the other record types here.

packages/es-client/indexer.js
```diff
@@ -294,20 +294,6 @@ function indexReconciliationReport(esClient, payload, index = defaultIndexAlias,
   return genericRecordUpdate(esClient, payload.name, payload, index, type);
 }
```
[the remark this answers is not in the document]
npauzenga: I'm sort of thinking we should be removing this stuff as we go. I posted a thread in Slack so curious if there's pushback here or what the reasoning would be.

packages/api/tests/lambdas/test-index-from-database.js
```diff
@@ -307,7 +291,6 @@ test.serial('Lambda successfully indexes records of all types', async (t) => {
     searchEs('pdr', esAlias, '20'),
```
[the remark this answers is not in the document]
npauzenga: No to punt this but... https://bugs.earthdata.nasa.gov/browse/CUMULUS-3847 (punt)

Once we get there the work becomes deleting the test file wholesale. I think it's good to remove as we go (above conversation redux) but if we miss something 🤷. Probably best to handle the cleanup in a separate task in case there are issues.

packages/es-client/indexer.js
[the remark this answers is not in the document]
npauzenga: Yeah, sounds good. Thanks for flagging the test. I'll keep an eye on CI and let ya know when everything is green. The failure you caught is fixed here:

https://github.com/nasa/cumulus/pull/3760/commits/0e6c0c29624cce8380749c82cfe79036d2887cf2

I unknowingly changed the translate logic to always return a `collection` object (albeit with null properties) instead of `undefined`.

PR #3764, opened 2024-08-09
CUMULUS-3233/3234: Update Providers LIST endpoint to remove ElasticSearch dependency and query Postgres instead

Nnaga1:

## Changes

* ...

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Replace ElasticSearch Phase 2
 
+- **CUMULUS-3232**
+  - Update API PDR endpoints `DEL` and `GET` to not update Elasticsearch
 - **CUMULUS-3233**
   - Updated `providers` list api endpoint and added `ProviderSearch` class to query postgres
   - Removed Elasticsearch dependency from `providers` endpoints
--- a/packages/api/endpoints/pdrs.js
+++ b/packages/api/endpoints/pdrs.js
@@ -1,3 +1,5 @@
+//@ts-check
+
 'use strict';
 
 const router = require('express-promise-router')();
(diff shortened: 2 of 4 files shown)


packages/db/src/search/ProviderSearch.ts
```diff
@@ -0,0 +1,83 @@
...
+   */
+  protected buildInfixPrefixQuery(params: {
+    countQuery: Knex.QueryBuilder,
+    searchQuery: Knex.QueryBuilder,
+    dbQueryParameters?: DbQueryParameters,
+  }) {
+    const { countQuery, searchQuery, dbQueryParameters } = params;
+    const { infix, prefix } = dbQueryParameters ?? this.dbQueryParameters;
+    if (infix) {
+      [countQuery, searchQuery].forEach((query) => query.whereLike(`${this.tableName}.name`, `%${infix}%`));
```
paulpilone, line 58: @Nnaga1 I noticed that in this and a few other search classes (e.g. CollectionsSearch, ExecutionSearch) that the `whereLike` for both infix and prefix are the same. This is different than GranulesSearch where they're different in a way I'd expect.

Written as is wouldn't this code produce identical whereLike clauses regardless of whether an infix or prefix was provided?

paulpilone (approved):
Looks good to me pending CI. @Nnaga1 I just want to call out that this is my first real code review of the ES changes so if you want to get a second opinion you won't hurt my feelings!

packages/db/src/search/ProviderSearch.ts
```diff
@@ -0,0 +1,83 @@
...
+    super({ queryStringParameters }, 'provider');
+  }
+
+  /**
+   * Build basic query
+   *
+   * @param knex - DB client
+   * @returns queries for getting count and search result
+   */
+  protected buildBasicQuery(knex: Knex)
```
jennyhliu, line 28: We have implemented in BaseSearch, this can be removed.

packages/db/src/search/ExecutionSearch.ts
```diff
@@ -130,7 +130,7 @@ export class ExecutionSearch extends BaseSearch {
       [countQuery, searchQuery].forEach((query) => query.whereLike(`${this.tableName}.arn`, `%${infix}%`));
     }
     if (prefix) {
-      [countQuery, searchQuery].forEach((query) => query.whereLike(`${this.tableName}.arn`, `%${prefix}%`));
+      [countQuery, searchQuery].forEach((query) => query.whereLike(`${this.tableName}.arn`, `${prefix}%`));
```
jennyhliu, line 133: Can you update StatsSearch.buildInfixPrefixQuery as well?

Issue #3772, opened 2024-08-15
CNM tasks getting delayed


hailiangzhang:
At GES DISC, our CNM tasks are frequently delayed. The `sqsMessageConsumer` is supposed to run every minute, but CNM messages may not be dispatched for over 10 minutes.

hailiangzhang:
I think this is because `sqsMessageConsumer` need fetch all the enabled rules to dispatch the CNM messages, but at GES DISC, we have a large number of rules (>8,000 in UAT), which causes 2 issues:
1. `sqsMessageConsumer` is taking too long to fetch all the enabled rules (~ 30 seconds out of the 1 min cron interval)
2. Database connection frequently failed due to large number of queries.

paulpilone:
@hailiangzhang v18.3.4 has been released and includes this fix: https://github.com/nasa/cumulus/releases/tag/v18.3.4. Going to close this Issue.

PR #3773, opened 2024-08-15
Fix the issue of delays in CNM message dispatching

hailiangzhang:
**Summary:** This PR aims to address the delays in CNM message dispatching.
The change involves increasing the limit of records per page in the fetchRules function. This reduces the number of database connections when fetching the rules, which in turn decreases the total query time and lowers the risk of database connection failures.
After this fix, the issue of CNM message delay was addressed from our end; it also reduced the rule query time from ~ 30 seconds to 2-3 seconds.

Addresses [the issue of CNM tasks getting delayed](https://github.com/nasa/cumulus/issues/3772)

## Changes

* Increase limit of records per page to be 100 in fetchRules function

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


hailiangzhang:
NOTE: I haven't yet spent time adapting the tests, as I want the Cumulus core team to first confirm that this change is acceptable. If yes, I will proceed with updating the unit tests accordingly.

paulpilone:
@hailiangzhang In general this looks good but I had a question around whether or not we want to do this this way. I understand the want/need to specify a limit in the API request but I'm not sure if hardcoding a value is the best way to do it. `100` might work for _you_ but maybe not for someone else (in theory - I don't have any test cases to back this up).

Is this limit something we want to expose via a variable or maybe even support as an optional parameter to the function so that you can specify a limit as the caller of it?

hailiangzhang:
> @hailiangzhang In general this looks good but I had a question around whether or not we want to do this this way. I understand the want/need to specify a limit in the API request but I'm not sure if hardcoding a value is the best way to do it. `100` might work for _you_ but maybe not for someone else (in theory - I don't have any test cases to back this up).
> 
> Is this limit something we want to expose via a variable or maybe even support as an optional parameter to the function so that you can specify a limit as the caller of it?

Thanks for your comment @paulpilone!

Basically this is what I found is happening: `fetchRules` will eventually make database query via the `Search` class from `es-client`, which internally already has a [hardcoded value of 10](https://github.com/nasa/cumulus/blob/master/packages/es-client/search.js#L252) for the page size from its constructor. I don't know why 10 was chosen to be the default value, but this page size value [can be overwritten by its parameter](https://github.com/nasa/cumulus/blob/master/packages/es-client/search.js#L267).

I think `fetchRules` may be right place to overwrite this, because by looking through the whole codebase, it seems only being used by message consumer, which will get all the rules (or all the enabled rules). The real life is that: it's totally possible that some DAAC (like GES DISC and some others I believe) has thousands of collections, which could lead to ten thousands rules (>8000 in our case) depending on the number of streams they have. Now if we want to get all the ten thousands of rules with the page size of 10, it may be very inefficient and can easily hit the connection limit. I don't know what will be the optimal size, but 100 seems to work much much better (reduced the query time from 30 -> 2 seconds, and no more connection failure). Even for those who don't have so many rules, I don't see a negative impact by making this switch.

Sorry about the long description, but thanks again for your question Paul, and please let me know if we want to further optimize this value or move the configuration to a different level.

hailiangzhang:
> @hailiangzhang In general this looks good but I had a question around whether or not we want to do this this way. I understand the want/need to specify a limit in the API request but I'm not sure if hardcoding a value is the best way to do it. `100` might work for _you_ but maybe not for someone else (in theory - I don't have any test cases to back this up).
> 
> Is this limit something we want to expose via a variable or maybe even support as an optional parameter to the function so that you can specify a limit as the caller of it?

BTW,

hailiangzhang:
> @hailiangzhang Thanks for the detailed response! One concern @npauzenga and I had was where `fetchRules` was used and whether or not this could have impacts in places we weren't aware. Based on your response and some research it looks like it is only used by message consumer and this change seems reasonable.
> 
> My only suggestion would be to maybe instead of hardcoding the value we instead accept an optional argument to `fetchRules` (but if it's only being used by message consumer then you aren't actually writing the caller of this function, right?) or introduce it as a new terraform variable so that it's customizable per user.
> 
> It's possible I'm over thinking this and providing 100 here works _better_ as you've demonstrated and so it just makes sense to hard code that value so that the base search class value is overridden. I can definitely be talked into that if we agree this improves performance and there's no reason to set it to anything other than the max (100 - as defined by the ES Search class).

Thanks a lot for your reply @paulpilone !

That's exactly what's in my mind too:) Yes, `fetchRules` seems only being used by message consumers and I am glad that we agreed that this change is okay:)

And yes, I will pass it as a parameter (and adapt testing if needed). Make it configurable from the template may be even better, but so sorry that I am a little bit overloaded and may be we can do it later when needed.

Regarding 100, I picked it up because it's the max allowed by Search class, and I am open to suggestion about a different value, but to me 10 is definitely too small that will cause lots of issues:(

Sorry I don't have time today but will try to wrap it up before next Monday.

Thanks again @paulpilone and @npauzenga !

hailiangzhang:
> @hailiangzhang Thanks for the detailed response! One concern @npauzenga and I had was where `fetchRules` was used and whether or not this could have impacts in places we weren't aware. Based on your response and some research it looks like it is only used by message consumer and this change seems reasonable.
> 
> My only suggestion would be to maybe instead of hardcoding the value we instead accept an optional argument to `fetchRules` (but if it's only being used by message consumer then you aren't actually writing the caller of this function, right?) or introduce it as a new terraform variable so that it's customizable per user.
> 
> It's possible I'm over thinking this and providing 100 here works _better_ as you've demonstrated and so it just makes sense to hard code that value so that the base search class value is overridden. I can definitely be talked into that if we agree this improves performance and there's no reason to set it to anything other than the max (100 - as defined by the ES Search class).

@paulpilone and @npauzenga , I've made a few more commits to pass the page size as a parameter to the `fetchRules` function. I'm having some issues with my localstack setup to run the tests, but after looking through the tests related to `fetchRules`, it doesn't appear any adjustment/addition is needed, and my PR should not change the testing coverage so far.

Please let me know if you have any concerns. Thanks again for reviewing and providing comments promptly!

paulpilone:
@hailiangzhang The team discussed this in our architecture meeting and agreed this looks good. I've created a ticket for it and will prioritize in this sprint so that it can be released with v18.4.0. I'll be updating this PR with necessary change log comments and monitor our CI.

paulpilone:
Closing this PR for https://github.com/nasa/cumulus/pull/3779

hailiangzhang:
@paulpilone and @npauzenga , I noticed my commits were merged into [another PR](https://github.com/nasa/cumulus/pull/3779), and thank you so much for adding change log and more tests:)
If possible, could you patch this into Cumulus 18.3 (which is what we are currently upgrading to)? I really appreciate your time and help!

paulpilone:
@hailiangzhang Going to back port to a new version 18.3.4. Created a ticket here: https://bugs.earthdata.nasa.gov/browse/CUMULUS-3848

PR #3774, opened 2024-08-15
CUMULUS-3796: Prepare Release 18.4.0

Nnaga1:

## Changes

* ...

## PR Checklist

- [ x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


packages/api/ecs/async-operation/package.json
```diff
@@ -23,10 +23,10 @@
   },
   "dependencies": {
     "@aws-sdk/client-lambda": "^3.621.0",
-    "@cumulus/aws-client": "18.3.0",
```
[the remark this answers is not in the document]
paulpilone: @Nnaga1 I don't think that was ever intentional. I've found a few places where update misses. e.g. I had to bump something up from 18.3.2 -> 18.3.3 in my last PR back to master.

website/versions.json
```diff
@@ -1,4 +1,5 @@
 [
+  "18.4.0",
```
paulpilone, line 2: Not sure if this actually matters (but it might depending on how it's looked up) but I think this should be "v18.4.0"

packages/api/ecs/async-operation/package.json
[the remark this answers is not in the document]
paulpilone: Although that package.json is a little bit different. See https://bugs.earthdata.nasa.gov/browse/CUMULUS-3786

Bumping the version is the right thing to do but you'll also need to build and distribute that image. That's why it's captured in a separate ticket for later.

website/versions.json
etcart: a good catch

audit-ci.json
```diff
@@ -13,7 +13,6 @@
     "http-cache-semantics",
     "lodash.pick",
     "hoek",
-    "fast-xml-parser",
```
paulpilone, line 16: Need to make sure this gets back to master in eventual 18.4 merge into master.

website/versioned_sidebars/version-18.4.0-sidebars.json
```diff
@@ -0,0 +1,336 @@
+{
```
paulpilone, line 1: Based on the current pattern I think this file should be named `version-v18.4.0-sidebars.json`

website/versioned_docs/version-18.4.0/README.md
```diff
@@ -0,0 +1,36 @@
+---
+id: cumulus-docs-readme
```
paulpilone, line 2: Commenting on the README because I can't comment on a directory: this directory should be named version-v18.4.0

paulpilone (changes requested):
Small comments about version file names. It's possible these aren't necessary but I know we do some string matching on some of these files so probably better safe than sorry.

paulpilone (approved):
Looks good!

PR #3791, opened 2024-08-31
CUMULUS-3756 NDCUM-1264 Added excludeFileRegex configuration to update-granules-cmr-metadata-…

etcart:
**Summary:** Added excludeFileRegex configuration to update-granules-cmr-metadata-file-links and updated unit tests to exercise new file-exclusion feature.

Addresses [CUMULUS-3756: Add the ability to exclude URLs matching a specific pattern in the CMR metadata](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3756)

## Changes

* Updated the config schema for update-granules-cmr-metadata-file-links with excludeFileRegex
* Updated tasks/update-granules-cmr-metadata-file-links/index.js to read the excludeFileRegex configuration and, if it's specified, filter the files passed to updateCMRMetadata.
* Updated tasks/update-granules-cmr-metadata-file-links/tests/test-index.js to exercise both typical file exclusion functionality and exercise the case where the regex doesn't match any files in the files list.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


tasks/update-granules-cmr-metadata-file-links/index.js
```diff
@@ -89,6 +98,7 @@ async function updateGranulesCmrMetadataFileLinks(event) {
     .map(({ name, type }) => [name, type]));
 
   const cmrGranuleUrlType = get(config, 'cmrGranuleUrlType', 'both');
+  const excludeFileRegexPattern = get(config, 'excludeFileRegex', null);
```
jennyhliu, line 101: Why do we need to default to null? If we change it to:
`const excludeFileRegexPattern = get(config, 'excludeFileRegex');`
We can still do the check
`if (excludeFileRegexPattern)`

CHANGELOG.md
```diff
@@ -83,6 +83,10 @@ ElasticSearch, the `collections/granules/executions` API endpoints are updated t
 - **CUMULUS-3792**
   - Added database indexes to improve search performance
 
+### Added
+- **CUMULUS-3756**
+  - Added excludeFileRegex configuration to update-granules-cmr-metadata-file-links to allow files matching specified regex to be excluded when updating the Related URLs list. Defaults to the current behavior of excluding no files.
```
jennyhliu, line 88: Can we break this into multiple lines?

tasks/update-granules-cmr-metadata-file-links/index.js
```diff
@@ -41,15 +41,24 @@ async function updateEachCmrFileAccessURLs(
...
 ) {
   return await Promise.all(cmrFiles.map(async (cmrFile) => {
     const granuleId = cmrFile.granuleId;
     const granule = granulesObject[granuleId];
+    let files = granule.files;
+    if (excludeFileRegexPattern !== null) {
+      const excludeFileRegex = new RegExp(excludeFileRegexPattern);
+      files = granule.files.filter((file) => !file.key.match(excludeFileRegex));
+      if (files.length === granule.files.length) {
+        throw new Error(`No files matched the excludeFileRegex ${excludeFileRegexPattern}.  Found files: ${files.map((file) => file.key).join(', ')}`);
```
jennyhliu, line 55: Is this any error?  If no file matches the `excludeFileRegexPattern`, do all the granules have to have the DMRPP file?

tasks/update-granules-cmr-metadata-file-links/tests/test-index.js
```diff
@@ -240,6 +242,67 @@ test.serial('update-granules-cmr-metadata-file-links properly handles a case whe
...
+
+  await validateConfig(t, newPayload.config);
+  await validateInput(t, newPayload.input);
+
+  const filesToUpload = cloneDeep(t.context.filesToUpload);
+  await uploadFiles(filesToUpload, t.context.stagingBucket);
+  await updateGranulesCmrMetadataFileLinks(newPayload);
+
+  // TODO; instead of checking the resulting metadata, mock updateCMRMetadata
+  // and just verify that the function was called with the correct files
```
jennyhliu, line 290: Suggest to remove the TODO if we don't handle it here.

tasks/update-granules-cmr-metadata-file-links/index.js
```diff
@@ -102,7 +112,8 @@ async function updateGranulesCmrMetadataFileLinks(event) {
     cmrGranuleUrlType,
     config.distribution_endpoint,
     bucketTypes,
-    distributionBucketMap
+    distributionBucketMap,
+    excludeFileRegexPattern
```
jennyhliu, line 116: Actually `excludeFileRegex` is not a nested config field, so we can just pass `config.excludeFileRegexPattern` to this function like `config.distribution_endpoint`

tasks/update-granules-cmr-metadata-file-links/index.js
etcart: I don't think so? I'm not sure what you mean. files will be the remaining (non-filtered) files, and it throws an error if you have set a "bad" regex (no files were filtered) I think I want to change that to a warning, rather than an error

jennyhliu: Yeah, I also prefer to log a warning rather than an error, so the workflow can be used by granules which don't have the files matching excludeFileRegexPattern

tasks/update-granules-cmr-metadata-file-links/index.js
etcart: https://github.com/nasa/cumulus/pull/3791/commits/369d4bde6d68e5678f7f6af3726859eda9fc7b9f
changed

tasks/update-granules-cmr-metadata-file-links/index.js
etcart: https://github.com/nasa/cumulus/pull/3791/commits/1f7e9cc89fbcd6439e1f11726a83f719b3a4a266
yup, fixed it
and confirmed with mike dorfman that switching to logging wouldn't interfere with their use-case

jennyhliu (approved):

PR #3807, opened 2024-09-23
[CUMULUS-3902]: Update AWS-SDK errors to use the V3 error classes

npauzenga:
**Summary:** 

Addresses [CUMULUS-3902](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3902)

## Changes

The AWS SDK V3 update came with a new class structure for AWS errors. We have been using the old approach, which is now causing breakages. 

In the new V3 world, the SDK throws instances of AWS-defined Error classes. We can check for these errors in our handling by inspecting the error object with `instanceof`. This is how AWS recommends handling:

https://aws.amazon.com/blogs/developer/service-error-handling-modular-aws-sdk-js/

We can _also_ use the `name` property to identify an error but that isn't the first recommended approach. What we can definitely NOT do is use `code`. Instances that were only checking the `code` property are now likely to break.

Note: There are a LOT of errors we're creating in Cumulus that look like aws-sdk errors but are our own classes. This can make it tough to determine what needs to change. If you're seeing lots of errors that use the old style of handling, it's possible those are Cumulus errors (not aws-sdk) rather than missed instances that need correcting. Still worth double-checking though in case I missed any.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


packages/errors/src/index.ts
```diff
@@ -55,8 +62,9 @@ export const isWorkflowError = (error: Error) => error.name.includes('WorkflowEr
 /**
  * Returns true if the error is a DynamoDB conditional check exception.
  */
-export const isConditionalCheckException = (error: ErrorWithOptionalCode) =>
-  error.name === 'ConditionalCheckFailedException' || error.code === 'ConditionalCheckFailedException';
+export const isConditionalCheckException = (error: any) => {
+  return error.name === "ConditionalCheckFailedException";
```
npauzenga, line 66: This one appears to be a special case. The service throwing the error is DynamoDBDocument, which is from `@aws-sdk/lib-dynamodb`. That might work slightly differently. The error does appear to have the correct class inherited but `instanceof` does not work here where it does elsewhere.

packages/errors/src/index.ts
```diff
@@ -55,8 +55,8 @@ export const isWorkflowError = (error: Error) => error.name.includes('WorkflowEr
 /**
  * Returns true if the error is a DynamoDB conditional check exception.
  */
-export const isConditionalCheckException = (error: ErrorWithOptionalCode) =>
-  error.name === 'ConditionalCheckFailedException' || error.code === 'ConditionalCheckFailedException';
+export const isConditionalCheckException = (error: any) =>
+  error.name === 'ConditionalCheckFailedException';
```
npauzenga, line 59: This is a tough one. I'm fairly certain the issue here is that the library throwing this exception is `@aws-sdk/lib-dynamodb`, not one of the `client-x` packages. The clients appear to support `instanceof` via the expected chain of inheritance, the lib export... may not.

Switching this over to `instanceof` like the others does not work. The error LOOKS like it's a `ConditionalCheckFailedException` but the properties look different than other errors and, again, `instanceof` does not work despite the class presenting as we'd expect.

I've chosen to fall-back to the supported `name` property in this case. According to their docs:

"This change is backward compatible—all the existing error handling code will still work. You can still assert the error by the name value."

packages/integration-tests/index.js
```diff
@@ -142,7 +142,7 @@ async function getExecutionStatus(executionArn) {
     const { status } = await StepFunctions.describeExecution({ executionArn });
     return status;
   } catch (error) {
-    if (error.code === 'ExecutionDoesNotExist') return 'STARTING';
+    if (error instanceof StepFunctions.ExecutionDoesNotExist) return 'STARTING';
```
npauzenga, line 145: Trying to keep changes to a minimum and access these errors via existing imports. Happy to adjust if anyone feels strongly and e.g. import `ExecutionDoesNotExist` explicitly.

packages/errors/src/index.ts
```diff
@@ -42,7 +42,7 @@ export interface ErrorWithOptionalCode extends Error {
 }
 
 /**
- * Test to see if a given exception is an AWS Throttling Exception
+ * Test to see if a given exception is a Throttling Exception.
```
npauzenga, line 45: Things like this took some time to suss out. In this case ThrottlingException is a Cumulus-defined error, not an aws-sdk error:
```
export class ThrottlingException extends Error 
```

packages/errors/src/index.ts
```diff
@@ -42,7 +42,7 @@ export interface ErrorWithOptionalCode extends Error {
...
  */
 export const isThrottlingException = (err: ErrorWithOptionalCode) =>
   err.name === 'ThrottlingException' || err.code === 'ThrottlingException';
```
npauzenga, line 48: As above, we're defining this error, not aws.

packages/errors/src/index.ts
npauzenga: Also I _think_ we want the error type here to be `any` because of the above. We want to be able to inspect an object that looks like a `ConditionalCheckFailedException` class instance but doesn't respond to `instanceof`.

example/spec/helpers/kinesisHelpers.js
```diff
@@ -107,7 +107,7 @@ async function createKinesisStream(streamName) {
       try {
         return await kinesis().createStream({ StreamName: streamName, ShardCount: 1 });
       } catch (error) {
-        if (error.code === 'LimitExceededException') throw error;
+        if (error instanceof LimitExceededException) throw error;
```
npauzenga, line 110: An alternative for all of this error handling is helper functions like "isLimitExceed" that we could export from aws-client instead of importing error types into packages like this. I didn't do that because it felt like complexity for the sake of it but open to pushback.

packages/errors/src/index.ts
npauzenga: Yeah, so the worst case result here is still supported. It's not the most optimal/consistent but it won't cause problems... allegedly.

packages/integration-tests/index.js
npauzenga: Yeah, I do agree that it's better to only import what we need. I started to make the change but found why we're doing it this way, at least in some cases.

In some tests we're importing the entire `StepFunctions` module because we need to stub methods on that module. We can't do it exactly the same way if we import a member of the module. e.g.

```
sinon.stub(StepFunctions, 'getExecutionStatus').callsFake(...)
```

won't work if we import `getExecutionStatus` directly. 

That makes me more inclined to not mess with the imports as there are (again, in some cases) functional issues doing it one way vs. the other. I'm not sure the change is worth the trouble 🤔. Thoughts?

packages/errors/src/index.ts
npauzenga: It makes me a little worried about missing stuff but there are a few instances like this where it's ambiguous where the error is actually coming from.

example/spec/helpers/workflowUtils.js
```diff
@@ -286,7 +286,7 @@ async function stateMachineExists(stateMachineName) {
   try {
     await StepFunctions.describeStateMachine({ stateMachineArn });
   } catch (error) {
-    if (error.code === 'StateMachineDoesNotExist') return false;
+    if (error instanceof StepFunctions.StateMachineDoesNotExist) return false;
```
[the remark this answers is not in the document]
npauzenga: See above comment. I stumbled on the reason it's done this way, at least in some cases.

npauzenga:
>Great work Nate, quick question, was it only Kinesis and Step Functions that needed this? like those were the only spots? none of the others had it like this or need this change? Want to confirm that and why before approving

Good question. That's the trick here. We need to fix anything that inspects an AWS error for the `code` property to determine the error. That's the main concern and what I'm asserting I've done here. I searched for instances we were catching errors using `code` and, if they were AWS-created errors, updated them to this type check.

The secondary, "best effort" concern is updating _any_ aws error handling to use their classes and `instanceof` instead of `code` or `name` properties. I've done a lot of that here and we should do that going forward but it's not as critical because aws still explicitly supports `name`.

npauzenga:
Thank you so much for being thorough and double-checking me @Nnaga1! There are a couple instances you point out with `error.code` that I did see previously. Here's my reasoning:

https://github.com/nasa/cumulus/blob/7bac3407e6702c96969746392ac3f8d1ca47d7c0/packages/api/tests/lib/FileUtils/test-setS3FileSize.js#L15 This is an error type that we (Cumulus) are creating in a unit test and assigning a `code`. It appears that we don't actually inspect the `code` declared here. I updated it to `name` just for consistency and to match what all of the logic expects (though, again, nothing breaks or changes either way). https://github.com/nasa/cumulus/pull/3807/commits/67314314cf3187380e0021bed4e00c15d8852980

https://github.com/nasa/cumulus/blob/7bac3407e6702c96969746392ac3f8d1ca47d7c0/tasks/files-to-granules/index.js#L63 Yeah, this is logging but we should include the `name` for clarity. Including `code` is ok but not by itself.
https://github.com/nasa/cumulus/pull/3807/commits/adfef915f98abe5befb3185b4617c79e92aff94d

https://github.com/nasa/cumulus/blob/7bac3407e6702c96969746392ac3f8d1ca47d7c0/example/spec/parallel/startSF/startSfSpec.js#L266 This one logs the `code` and `name`. I think that's fine because `code` is still valid, just shouldn't be used for error handling.

The others are using `error.name` and, although not using type checking, it's still a valid, supported approach so I'm inclined to leave them and not mess with exporting/importing aws error types.

PR #3817, opened 2024-10-07
CUMULUS-3897 - Merges v18.5.0 CL updates

paulpilone:

Addresses [CUMULUS-3987: Prepare, Create, and Publish Release 18.5.x for Serverless v2](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3897)

## Changes

* Merges CL updates from v18.5.0 release

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,17 +8,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Migration Notes
 
-#### CUMULUS-3536 Upgrading from Aurora Serverless V1 to V2
-
-- The updates in CUMULUS-3536 require an upgrade of the postgres database.
-  Please follow [Upgrading from Aurora Serverless V1 to V2]
-  (https://nasa.github.io/cumulus/docs/next/upgrade-notes/serverless-v2-upgrade)
-
 ### Added
 
-- **CUMULUS-3536**
-  - Added `rejectUnauthorized` = false to db-provision-user-database as the Lambda
-    does not have the Serverless v2 SSL certifications installed.
 - **CUMULUS-3020**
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -8,17 +8,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Migration Notes
 
-#### CUMULUS-3536 Upgrading from Aurora Serverless V1 to V2
```
npauzenga, line 11: Hmmmmm 🤔 🤔 

18.5.0 is a backport. Do we want these entries to only appear in the 18.5.0 release? I don't think so because I think we want the next forward release (v19.1.0) to also have these entries because it is based on main, not this 18.x release series?

Does that make sense? I'm not 100% confident. I'm asserting we want the entries for these tickets in both releases.

npauzenga (changes requested):
Looks great @paulpilone. One comment we should iron out between our releases.

CHANGELOG.md
npauzenga: Wow this is confusing. After a Slack thread and too much time thinking about it, I suggest we not remove these entries on main. The 19.1.0 release will have them (e.g. 3536), even though they already exist on v19. 

Very open to argument though as this is weird 😆

paulpilone: @npauzenga discussion happening in Slack but one idea is to just update 18.5.0 entry to call out this is a backport and may or may not contain duplicate changes. We've done that before (e.g. 18.3.4):

>**Please note** changes in v18.3.4 may not yet be released in future versions, as this
is a backport/patch release on the v18.3.x series of releases.  Updates that are
included in the future will have a corresponding CHANGELOG entry in future releases.

PR #3826, opened 2024-10-15
Fix cumulus versions due to bad merge

npauzenga:
I'm not 100% sure which merge caused this but it appears that the Cumulus version v19.0.0 was overwritten on main with a previous version for the aws-client package. This breaks bootstrap.


npauzenga:
This is the merge we're trying to fix: https://github.com/nasa/cumulus/pull/3791/files#diff-f237ddd531cac477714c4b04fdef83ae812895abdf47d90c73c2c9ddf8897787

packages/aws-client/package.json
```diff
@@ -46,32 +46,32 @@
...
-    "@aws-sdk/client-cloudwatch-events": "^3.447.0",
-    "@aws-sdk/client-dynamodb": "^3.447.0",
-    "@aws-sdk/client-dynamodb-streams": "^3.447.0",
-    "@aws-sdk/client-ec2": "^3.447.0",
-    "@aws-sdk/client-ecs": "^3.447.0",
-    "@aws-sdk/client-elasticsearch-service": "^3.529.1",
-    "@aws-sdk/client-kinesis": "^3.447.0",
-    "@aws-sdk/client-kms": "^3.447.0",
-    "@aws-sdk/client-lambda": "^3.529.1",
-    "@aws-sdk/client-s3": "^3.447.0 <3.614.2",
```
npauzenga, line 60: If these are newer changes to avoid a vulnerability or something, this may be wrong. We'll see what CI says.

jennyhliu (approved):
LGTM

PR #3830, opened 2024-10-17
[CUMULUS-3921] Test new Orca version and update example deployment

npauzenga:
This updates the example deployment to the new Orca version. We had previously updated to v10 but it caused issues in our Cumulus Core development environment. This takes those changes to update to the Orca v10 series and includes the fix version 10.0.1, which addresses the issues we saw.

Basically, this is a revert of this revert with an updated package version: https://github.com/nasa/cumulus/pull/3821/files 😄


example/cumulus-tf/orca.tf
```diff
@@ -27,21 +27,21 @@ module "orca" {
   vpc_id                   = local.vpc_id
 
   ## OPTIONAL
-  tags                     = var.tags
+  tags                                = var.tags
+  deploy_rds_cluster_role_association = false
```
npauzenga, line 31: This is the new ENV that fixes our deployment issue. Initial deployment (for Cumulus Core) has it set to true, subsequent deployments `false`.

etcart: so to be clear, we, with our dev/ci stacks will need to set this to true by hand before deploying the first time with these Changes, and then back to false? or is this a global thing?

paulpilone: @etcart If _any_ stack has deployed this association with our shared RDS cluster then all subsequent stacks must be deployed with this set to `false`.

paulpilone: @npauzenga Should this value be set in the developers deployment tf vars instead of this file? Otherwise everyone (including the person that actually wants it set to true) will pick up changes to `true`/`false`.

npauzenga: Hmm @etcart you'll always have this set to `false`. The only time this is `true` for Core devs is on the very first deploy of Orca, when it creates the IAM role initially. Every subsequent run should be false.

@paulpilone hmm we could put it in our tfvars but the only time this would be set to true is if someone destroys the Orca stack or that IAM role. I'm a little hesitant because, in order to do that, we'd need it `true` in the example/variables (as that's the example deployment for everyone), and all dev's tfvars would need to be updated to set it `false`.

I'm sorta thinking it's less risk/easier to just leave it as part of the Orca module? If we end up having to do it again that person would temporarily flip the variable in orca.tf and deploy. If it happens a third time, maybe a variable? 😄

paulpilone: Looks good! Glad terraform is doing the sane thing here. This PR looks good to me.

paulpilone (approved):

PR #3873, opened 2024-12-02
WIP: [CUMULUS-3576]: Limit concurrent file downloads

npauzenga:
https://bugs.earthdata.nasa.gov/browse/CUMULUS-3576

The idea here is retaining `globalConnectionLimit` as a limit on the number of _granules_ that are synced concurrently. This adds a new config option called `fileDownloadPageSize`, which paginates the actual file downloads.


npauzenga:
Closing this as another approach will be required.

PR #3876, opened 2024-12-05
Ecarton/cumulus 3751 as separate task

etcart:

Addresses [CUMULUS-3751: task to move granules between collections (in s3 and cumulus datastores)](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3751)

## Changes

* new task move-granule-collections

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


package.json
```diff
@@ -33,7 +33,7 @@
     "bootstrap-no-build-no-scripts": "lerna bootstrap --no-ci --force-local --ignore-scripts",
     "bootstrap-no-build-no-concurrency": "lerna bootstrap --no-ci --force-local --concurrency 1",
     "bootstrap-no-build-quiet": "lerna bootstrap --no-ci --force-local --loglevel=error",
-    "ci:bootstrap": "lerna bootstrap --no-ci --force-local --ignore-scripts && lerna run prepublish",
```
etcart, line 36: this is while we're in here, but shouldn't the "not no scripts one" not ignore scripts?

packages/aws-client/src/S3.ts
etcart, line 1: extraneous but cleaning up t typing

packages/cmrjs/src/types.ts
etcart, line 1: some extra types to clarify in car-utils

packages/errors/src/index.ts
etcart, line 1: error to be used in move-granule-collections

tasks/move-granule-collections/README.md
etcart, line 1: TODO this needs to be not just a copy-pasta from move-granules

tasks/move-granule-collections/tests/test-index.js
BWex-NASA, line 1: Let me know if I am off base here, or missed the test case, but should you be concerned about moving the same granule that already was moved prior? Would that cause an overwrite to occur? Not sure if this in scope for the ticket.

etcart: it's not off base, it is fairly core to what we're trying to do here *but* I intended to solve this problem (and hope I have) see "handles partially moved files" and "handles granules that need no move.
the intention of this is to simulate a granule that has already been (at least partly) moved, in case it failed for some reason during processing previously (or a collection is passed in that doesn't chagne anything from the previous

PR #3881, opened 2024-12-11
[CUMULUS-3955] Update migration to remove vacuum statements

npauzenga:
This changes an existing db migration to remove VACUUM statements. If run in a Lambda env, these can cause timeouts on large datasets. See ticket for more details.


packages/db/src/migrations/20240728101230_add_table_indexes.ts
```diff
@@ -24,15 +24,6 @@ export const up = async (knex: Knex): Promise<void> => {
   await knex.raw('CREATE INDEX CONCURRENTLY IF NOT EXISTS providers_updated_at_index ON providers(updated_at)');
 
   await knex.raw('CREATE INDEX CONCURRENTLY IF NOT EXISTS rules_updated_at_index ON rules(updated_at)');
-
```
npauzenga, line 27: We didn't flag these initially but I think should also be removed.

npauzenga: I ran a search for `VACUUM` in all .ts files and didn't find any additional.

npauzenga: Yeah, the sql files in the migrations directory aren't meant to be run as part of a deployment. Those are for operators to execute manually.

PR #3885, opened 2024-12-16
Adds long PAN functionality to send-pan task

adtisdal-ASDC:
**Summary:** Adds Long PAN

## Changes

* Adds long Pan functionality
* Adds panType variable for configuring

## PR Checklist

- [ ] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests


tasks/send-pan/src/index.ts
```diff
@@ -41,16 +41,40 @@ async function sendPAN(event: HandlerEvent): Promise<HandlerOutput> {
   const { config, input } = event;
   const provider = config.provider;
   const remoteDir = config.remoteDir || 'pans';
+  const panType = config.panType || 'shortPan';
 
   const panName = input.pdr.name.replace(/\.pdr/gi, '.pan');
```
ckhowardASDC, line 46: Any chance you can make this file-extension uppercase? The spec/ICD says the extension is supposed to be .PAN.

adtisdal-ASDC: I changed it to uppercase

tasks/pdr-status-check/index.js
```diff
@@ -91,7 +91,7 @@ function buildOutput(event, groupedExecutions) {
 
   const parseFailedExecution = (execution) => {
     let reason = 'Workflow Failed';
-    if (execution.output) reason = JSON.parse(execution.output).exception;
+    if (execution.error) reason = execution.error;
```
adtisdal-ASDC, line 94: `execution.error` gets the error from the failed ingest granule workflow. The line `"ErrorPath": "$.exception.Cause"` has to be added into the `WorkflowFailed` step definition for the actual error to propagate. I added it to the README. 

https://docs.aws.amazon.com/step-functions/latest/dg/state-fail.html

adtisdal-ASDC: Also I used `ErrorPath` instead of `CausePath` since I wasn't 100 percent sure if Cause was being used for something else

tasks/send-pan/src/index.ts
```diff
@@ -41,16 +41,40 @@ async function sendPAN(event: HandlerEvent): Promise<HandlerOutput> {
...
+  const panName = input.pdr.name.replace(/\.pdr/gi, '.PAN');
   const uploadPath = path.join(remoteDir, panName);
 
   if (input.running.length !== 0) {
     throw new Error('Executions still running');
   }
-
-  const disposition = (input.failed.length > 0) ? 'FAILED' : 'SUCCESSFUL';
-  const pan = pdrHelpers.generateShortPAN(disposition);
+  let pan;
```
adtisdal-ASDC, line 52: We decided on 3 pan types:

- longPan: Only creates long PANs when there's more than one granule, else just a short
- shortPan: Always creates short PANs
- longPanAlways: Always creates long PANs

packages/api/lib/pdrHelpers.js
```diff
@@ -17,6 +18,38 @@ function generateShortPAN(disposition) {
   );
 }
 
+async function getGranuleFromExecution(executionArn) {
```
adtisdal-ASDC, line 21: I added this function for getting some of the granule information from the execution. There could be a better alternative I'm not familiar with

packages/api/lib/pdrHelpers.js
```diff
@@ -17,6 +18,38 @@ function generateShortPAN(disposition) {
...
+
+/**
+ * Generate Long PAN message
+ *
+ * @param {Object|string[]} executions - List of workflow executions
+ * @returns {string} the PAN message
+ */
+async function generateLongPAN(executions) {
+  const timeStamp = new Date();
+
```
jennyhliu, line 37: Do you have document about Long PAN?

packages/api/lib/pdrHelpers.js
```diff
@@ -17,6 +18,38 @@ function generateShortPAN(disposition) {
...
+  const longPan = new pvl.models.PVLRoot()
+    .add('MESSAGE_TYPE', new pvl.models.PVLTextString('LONGPAN'))
+    .add('NO_OF_FILES', new pvl.models.PVLNumeric(executions.length));
```
jennyhliu, line 40: Each execution can have multiple granules depending on the `QueueGranules` configuration `preferredQueueBatchSize`.
Each granule has multiple files.
Is `NO_OF_FILES` number of granules or number of files for this PDR?

packages/api/lib/pdrHelpers.js
```diff
@@ -17,6 +18,38 @@ function generateShortPAN(disposition) {
...
+  /* eslint-disable no-await-in-loop */
+  for (const exc of executions) {
+    const granule = await getGranuleFromExecution(exc.arn || exc);
+    longPan.add('FILE_DIRECTORY', new pvl.models.PVLTextString(granule.files[0].path));
+    longPan.add('FILE_NAME', new pvl.models.PVLTextString(granule.granuleId));
```
jennyhliu, line 45: For granules with multiple files, do we only report one file per granule?

tasks/send-pan/src/index.ts
jennyhliu: Is the implementation tested in a step function?

packages/api/lib/pdrHelpers.js
adtisdal-ASDC: That's how I thought it worked originally, but there should be a line for each file and not each granule. I'll be updating this PR

packages/api/lib/pdrHelpers.js
adtisdal-ASDC: Did not know that about the `preferredQueueBatchSize`. I'll update this also

packages/api/lib/pdrHelpers.js
adtisdal-ASDC: I can send one

jennyhliu:
@adtisdal-ASDC I have made some updates in the PR which has the commits from this PR https://github.com/nasa/cumulus/pull/3922
Please take a look to see if there are more updates needed. Thanks!

PR #3897, opened 2025-01-08
CUMULUS-3847: Remove any remaining ES indexing functions and tests

charleshuang80:
…loud Metrics usage

Addresses [CUMULUS-3847: Remove any remaining ES indexing functions and tests](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3847)

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


packages/api/lambdas/cleanExecutions.js
jennyhliu, line 1: Suggest to comment out the code, and put a dummy handler, so the lambda can still be created in tf-modules/archive/clean_executions.tf

packages/api/lambdas/bootstrap.js
```diff
@@ -4,6 +4,7 @@
  * It helps:
  *  - adding ElasticSearch index mapping when a new index is created
  */
+// is this needed for CloudMetrics?
```
jennyhliu, line 7: With es instance going away, this lambda shouldn't be needed.

packages/aws-client/tests/test-services.js
```diff
@@ -188,6 +188,7 @@ test('ec2() service defaults to localstack in test mode', async (t) => {
   );
 });
 
+// can this test be removed or is it needed for the granule CloudMetrics queries?
```
jennyhliu, line 191: With es instance going away, I think packages/aws-client/tests/services.js es and this test are not needed.

packages/api/lambdas/create-reconciliation-report.js
```diff
@@ -969,6 +954,7 @@ async function handler(event) {
   process.env.CMR_PAGE_SIZE = process.env.CMR_PAGE_SIZE || '200';
 
   //TODO: Remove irrelevant env vars from terraform after ES reports are removed
+  // can this be changed now?
```
jennyhliu, line 957: Yeah, we are not querying es anymore.

packages/api/tests/lambdas/test-bulk-operation.js
```diff
@@ -49,7 +49,7 @@ const esSearchStub = sandbox.stub();
 const esScrollStub = sandbox.stub();
 FakeEsClient.prototype.scroll = esScrollStub;
 FakeEsClient.prototype.search = esSearchStub;
-
+// is this es-client stuff still needed b/c granules and metrics?
```
jennyhliu, line 52: I think es-client is still needed. You can try to take it out.

CHANGELOG.md
```diff
@@ -92,6 +92,8 @@ External tooling making use of `searchContext` in the `GET` `/granules/` endpoin
   - Updated `@cumulus/api/bin/serveUtils` to no longer add records to ElasticSearch
   - Removed ElasticSearch from local API server code
   - Updated CollectionSearch to filter granule fields in addition to time frame for active collections
+- **CUMULUS-3847**
+  - list changes here
```
npauzenga, line 96: ?

packages/api/lambdas/cleanExecutions.js
```diff
@@ -2,138 +2,143 @@
...
+/**
+ * This lambda has been commented out because it needs to be rewritten for PG instead of running
+ * in ElasticSearch. This will be done in CUMULUS-XXXX
+ */
+
+// const { getEsClient, esConfig } = require('@cumulus/es-client/search');
+// const moment = require('moment');
 const Logger = require('@cumulus/logger');
-const { sleep } = require('@cumulus/common');
+// const { sleep } = require('@cumulus/common');
```
npauzenga, line 15: Remove comments

npauzenga: We should avoid having commented-out code generally. We can rely on the git diff if we need to reference what was here.

packages/api/lib/testUtils.js
```diff
@@ -620,20 +620,14 @@ const createAsyncOperationTestRecords = async (context) => {
   };
 };
 
+// this doesn't look like it's being used anywhere, should I remove it?
```
npauzenga, line 623: If it's not used, I'd say yes. It's an exported member so technically (possibly) a user-facing change for those using this package but... it's only test utils.

packages/async-operations/tests/test-async_operations.js
```diff
@@ -61,7 +61,7 @@ test.before(async (t) => {
   t.context.functionConfig = {
     Environment: {
       Variables: {
-        ES_HOST: 'es-host',
+        Timeout: 300,
```
npauzenga, line 64: What's this for?

packages/async-operations/tests/test-async_operations.js
```diff
@@ -357,7 +357,7 @@ test('getLambdaEnvironmentVariables returns expected environment variables', (t)
   const vars = getLambdaEnvironmentVariables(t.context.functionConfig);
 
   t.deepEqual(new Set(vars), new Set([
-    { name: 'ES_HOST', value: 'es-host' },
+    { name: 'Timeout', value: 300 },
```
npauzenga, line 360: Same as above, why this change?

tf-modules/archive/reconciliation_report.tf
```diff
@@ -15,8 +15,6 @@ resource "aws_lambda_function" "create_reconciliation_report" {
       CMR_HOST                         = var.cmr_custom_host
       DISTRIBUTION_ENDPOINT            = var.distribution_url
       ES_HOST                          = var.elasticsearch_hostname
```
npauzenga, line 17: What about ES_HOST?

CHANGELOG.md
charleshuang80: placeholder, I'm still making sure I've captured everything. will make sure this is filled out when I'm fairly certain

packages/api/lambdas/cleanExecutions.js
charleshuang80: hah. Well, I completely removed the lambda and test, and @jennyhliu suggested that I add them back and comment out except for a dummy handler. So I'm open to suggestions on why one method is better than the other...

packages/async-operations/tests/test-async_operations.js
charleshuang80: Since we're moving away from ES in general, ES indexing and all that, in the async operations tests I've changed all the places we're referencing ES operations to other options in those enums. Since we should probably be doing operations other than ES ones if ES is going away.

packages/api/lambdas/bulk-operation.js
```diff
@@ -94,7 +94,7 @@ async function applyWorkflowToGranules({
  * @param {number} [payload.concurrency]
  *   granule concurrency for the bulk deletion operation.  Defaults to 10
  * @param {Object} [payload.query] - Optional parameter of query to send to ES
- * @param {string} [payload.index] - Optional parameter of ES index to query.
+ * @param {string} [payload.index] - Optional parameter of ES index (metrics) to query.
```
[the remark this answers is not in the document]
charleshuang80: Yes, my understanding of why we need to keep the ES related functionality is that for some of the granules and bulk operations it will still query Cloud Metrics for data. I don't feel like totally understand how it all ties together, but that is what is mentioned in code and what I was informed of. So I thought it would be helpful to clarify as much as possible that the ES indexing and querying is going to Cloud Metrics and that's why it is staying around and not being removed as part of ES removal.

packages/api/lambdas/cleanExecutions.js
charleshuang80: (see comment at top of file)

packages/async-operations/tests/test-async_operations.js
charleshuang80: So just as I changed the `operationType`s away from ES Index, I'm trying to move away from the `ES_HOST`. I wanted to keep a variable in here and found that Timeout could be one that is used.

tf-modules/archive/reconciliation_report.tf
charleshuang80: So I removed these two variables in reference to the comment at the very end of packages/api/lambdas/create-reconciliation-report.js, where I also removed them from logging.
I will be writing a ticket to remove ES from all the other terraform files where they're not needed (minus Cloud Metrics) so I thought this should be part of that removal.

packages/api/lambdas/cleanExecutions.js
jennyhliu: If we remove lambda and test here, we have to remove the related tf code, that's why I suggested to have dummy handler since we want to keep the lambda with RDS implementation.
Now I think we can delete the code and leave only the dummy handler.

charleshuang80: Ok, I can remove the commented code, if everyone is in agreement there. 
What about the test file, should I remove the file altogether? Or if not, what should I leave in the file if not commented out tests?

tf-modules/archive/reconciliation_report.tf
charleshuang80: CUMULUS-3983

CHANGELOG.md
```diff
@@ -92,6 +92,12 @@ External tooling making use of `searchContext` in the `GET` `/granules/` endpoin
   - Updated `@cumulus/api/bin/serveUtils` to no longer add records to ElasticSearch
   - Removed ElasticSearch from local API server code
   - Updated CollectionSearch to filter granule fields in addition to time frame for active collections
+- **CUMULUS-3847**
+  - remove remaining ES indexing in code and tests
+  - for asyncOperations test data, change any ES related values to other options
+  - remove code from `@cumulus/api/lambdas/cleanExecutions` leaving a dummy handler, as the code worked with ES. lambda will be rewritten with CUMULUS-3982
+  - remove `@cumulus/api/endpoints/elasticsearch`, `@cumulus/api/lambdas/bootstrap`, and `@cumulus/api/lambdas/index-from-database`
+  - clarify in comments that ES functionality remaining in `@cumulus/api/lib/granules` is for working with CloudMetrics
```
npauzenga, line 100: Probably don't need this in the top-level release notes. The comments in the file (assuming they're there) are fine.

example/spec/serial/AsyncOperationRunnerFailingLambdaSpec.js
```diff
@@ -49,7 +49,7 @@ describe('The AsyncOperation task runner executing a failing lambda function', (
         id: asyncOperationId,
         taskArn: randomString(),
         description: 'Some description',
-        operationType: 'ES Index',
+        operationType: 'Bulk Granules',
```
npauzenga, line 52: Just for clarity? Seems great, just curious.

packages/api/endpoints/elasticsearch.js
```diff
@@ -1,233 +0,0 @@
...
-  const alias = req.params.alias || defaultIndexAlias;
-
-  const aliasIndices = await esClient.client.indices.getAlias({ name: alias })
-    .then((response) => response.body);
-
-  return res.send(Object.keys(aliasIndices));
-}
-
-// express routes
-router.put('/create-snapshot', createEsSnapshot);
```
npauzenga, line 221: Is there an argument for keeping any of this? Specifically the snapshot endpoint? Would this be useful for DAACs that still have an orphan ES instance?

packages/api/lambdas/bulk-operation.js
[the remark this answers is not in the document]
npauzenga: Ok, maybe just clarify "Cloud Metrics" or "Global Metrics" to make it clear what "metrics" means in this context?

packages/api/lambdas/create-reconciliation-report.js
```diff
@@ -900,9 +895,7 @@ async function processRequest(params) {
     location: buildS3Uri(systemBucket, reportKey),
   };
   let [reportPgRecord] = await reconciliationReportPgModel.create(knex, builtReportRecord);
-  let reportApiRecord = translatePostgresReconReportToApiReconReport(reportPgRecord);
-  await indexReconciliationReport(esClient, reportApiRecord, process.env.ES_INDEX);
-  log.info(`Report added to database as pending: ${JSON.stringify(reportApiRecord)}.`);
+  log.info(`Report added to database as pending: ${JSON.stringify(reportPgRecord)}.`);
```
npauzenga, line 898: Is there no value in logging the API-formatted record as we did previously? Generally users don't see the pg record, they get the translated version. But if we're just translating to log it... that may be overkill.

packages/api/lib/granules.js
```diff
@@ -234,10 +234,10 @@ function getTotalHits(bodyHits) {
 }
 
 /**
- * Returns an array of granules from ElasticSearch query
+ * Returns an array of granules from an ElasticSearch query
  *
  * @param {Object} payload
- * @param {string} [payload.index] - ES index to query
+ * @param {string} [payload.index] - ES (metrics) index to query
```
npauzenga, line 240: Same comment as above. Maybe clarify what 'metrics' means in this context.

packages/api/lib/testUtils.js
```diff
@@ -620,22 +617,6 @@ const createAsyncOperationTestRecords = async (context) => {
...
 
-const cleanupExecutionTestRecords = async (context, { arn }) => {
-  const {
-    knex,
-    executionPgModel,
-    esClient,
-    esIndex,
-  } = context;
-
-  await executionPgModel.delete(knex, { arn });
```
npauzenga, line 631: Don't we need this?

npauzenga (changes requested):
Had a couple comments. Mostly clarifying questions.

Are there any tests failing/blockers in CI?

packages/api/lambdas/create-reconciliation-report.js
charleshuang80: So yeah, the way this is written, we are translating it at the very end to return the api format record, but this seems to be the only place there was explicit logging. I'm not sure for logging purposes if api format is better than the pg format, so I'm open to doing it that way. But I did just log what we had instead of translating it for logging.
Also, thinking about it more, I'm open to adding some more logging like maybe at the end before returning the api format record. What do you think?

packages/api/lib/testUtils.js
charleshuang80: Hah, I mentioned in a previous comment that I did a search for the method `cleanupExecutionTestRecords` and did not find it being used anywhere. So I wondered if I should just remove it as I couldn't find it being used, and you confirmed...

example/spec/serial/AsyncOperationRunnerFailingLambdaSpec.js
charleshuang80: I'm sorry, I'm not sure I understand what you're asking here.

packages/api/endpoints/elasticsearch.js
charleshuang80: I don't know. If anything should be kept and we have a reason, that seems fine. But I don't know if any should be and I was just going with "remove all the ES things" except for what is needed for CloudMetrics and the other tickets I wrote. Please let me know if there's anything that should be put back.

charleshuang80:
> Had a couple comments. Mostly clarifying questions.
> 
> Are there any tests failing/blockers in CI?

Yes, I posted about an aws-client test I'm having issues with in slack.

CHANGELOG.md
npauzenga: Cool. I'll leave this comment unresolved for now.

packages/api/lambdas/cleanExecutions.js
npauzenga: Sounds good. Removing all commented code is certainly the preference as it's all in git if we need to get back to it.

packages/async-operations/tests/test-async_operations.js
npauzenga: I see. And being that this is a mock, we just want to prove that we get back some semi-realistic config object.

packages/api/lambdas/create-reconciliation-report.js
npauzenga: If it doesn't make sense to move this log statement or translated beforehand, maybe just update the message to make it clear that it's a PG-formatted record. It's a nit. I'm just concerned that someone sees this in the logs and gets confused as to why the naming conventions are different. More logging is good.

packages/api/lib/testUtils.js
npauzenga: Ah ok, sure. Didn't piece together that this was the function we weren't sure was used.

example/spec/serial/AsyncOperationRunnerFailingLambdaSpec.js
charleshuang80: yeah I got the types from /packages/types/api/async_operations. I switched it since we shouldn't be doing (most of) the ES things anymore. I considered removing the type, but Jonathan pointed out that past operations might still have the type. Not sure how long that will be for though.

packages/api/tests/lambdas/test-cleanExecutions.js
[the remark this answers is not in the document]
charleshuang80: removed as per convo during PI planning

packages/api/lambdas/cleanExecutions.js
charleshuang80: code was removed, dummy handler added, as per other thread in this file

packages/aws-client/tests/test-services.js
charleshuang80: when I tried removing ES, I found that some of the ES vars were for Metrics, so it seemed like ES needs to stay for these tests

packages/api/tests/lambdas/test-bulk-operation.js
charleshuang80: when I tried removing ES, I found that some of the ES vars were for Metrics, so it seemed like ES needs to stay for these tests

packages/aws-client/tests/test-services.js
charleshuang80: ES references removed

CHANGELOG.md
charleshuang80: this has been updated, and the line removed as suggested

packages/api/lambdas/cleanExecutions.js
charleshuang80: code removed but dummy handler left in place

packages/api/lib/testUtils.js
charleshuang80: removed as it was not used anywhere I could find

packages/async-operations/tests/test-async_operations.js
charleshuang80: I answered this in the other thread, is that sufficient?

packages/api/endpoints/elasticsearch.js
charleshuang80: the functionality for the endpoint was never actually implemented, so keeping this would not do anything.

packages/api/lambdas/create-reconciliation-report.js
charleshuang80: for consistency, it seems correct to keep the translation and logging of the api format. the translation is a quick function anyways. Added a comment for future reference because it might be confusing why the translation was done just for logging.
Also added a bit more logging (with translation)

packages/api/lib/granules.js
charleshuang80: this was done in a few places.

npauzenga (changes requested):
Thanks @charleshuang80! This is looking great. I think (pending anything CI uncovers) this should be good to go. I searched the repo and found a couple areas I'm curious about. Notably:
- https://github.com/nasa/cumulus/blob/CUMULUS-3847/packages/api/tests/performance/lib/test-write-granules.js#L75-L83

(And other tests that still import es-client) Should those be updated or is that scoped somewhere else?

charleshuang80:
> Thanks @charleshuang80! This is looking great. I think (pending anything CI uncovers) this should be good to go. I searched the repo and found a couple areas I'm curious about. Notably:
> 
> * https://github.com/nasa/cumulus/blob/CUMULUS-3847/packages/api/tests/performance/lib/test-write-granules.js#L75-L83
> 
> (And other tests that still import es-client) Should those be updated or is that scoped somewhere else?

I think it depends on the test. A few that utilize the cloud metrics functionality still need it I believe. But this test failed in CI and I've been working on removing the ES stuff and trying to figure out if we can still test the performance. I'll do another search, but if you have other specific examples you think should go or just aren't sure about please let me know.

charleshuang80:
> Thanks @charleshuang80! This is looking great. I think (pending anything CI uncovers) this should be good to go. I searched the repo and found a couple areas I'm curious about. Notably:
> 
> * https://github.com/nasa/cumulus/blob/CUMULUS-3847/packages/api/tests/performance/lib/test-write-granules.js#L75-L83
> 
> (And other tests that still import es-client) Should those be updated or is that scoped somewhere else?

So I fixed this file and tested it. Also found a few more references that I removed and tested in the latest commit. There are some references in like granules, test-executions, test-build-operation, and test-granules that I think need to stay for Cloud Metrics. If there are any other locations you think I missed please let me know.

npauzenga (approved):
Thanks @charleshuang80! I think you've got everything that jumps out. If there are areas that still have some ES cruft, we can remove as we find it. The important thing is that the tests pass and all of the core functionality still... functions 😄

PR #3909, opened 2025-01-27
CUMULUS-3983: remove all unneeded ES related values in tf modules

Nnaga1:

## Changes

* ...

## PR Checklist

- [X] Update CHANGELOG
- [X] Unit tests
- [ ] Integration tests


npauzenga:
> there are a lot of elasticsearch refs in `tf-modules/data-persistence/elasticsearch.tf:` , I think I should leave that alone right? based on the conversation in the arch meeting

What is the reasoning for leaving that module? Is it so that users don't end up in state where they have an ES instance that was deployed using TF but is no longer accessible via TF?

npauzenga (changes requested):
Thanks @Nnaga1! The changes look great. I had a question about a testing plan and about the data-persistence package but otherwise it's looking good. I did a couple cursory searches in our ES files and didn't find anything that jumped out.

npauzenga (approved):
This looks great @Nnaga1! I think we're good to merge to the feature branch as CI has deployed and tested successfully. If, when we test the removal of the ES cluster, we find additional removals, we can handle it as part of that ticket.

PR #3917, opened 2025-02-06
Release 20.0.0 merge to main

npauzenga:
This is the merge from the release branch back to main. All of the functional changes in the release are already in the destination so this is just taking the version bumps, changelog changes, and any docs that the release produced.


Jkovarik (approved):
💯  LGTM.

PR #3932, opened 2025-03-08
Cumulus 3960 merge to 18.5 release

etcart:
**Summary:** fold cumulus 3960 into feature branch

Addresses [CUMULUS-3960: update cumulus in cmr](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3960)

## Changes

* post-to-cmr task updated to handle updating collection in cmr for granule
* workflow reworked to pass oldGranules through meta

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/example/data/collections/s3_MOD09GQ-AZ_006_full_ingest_move/s3_MOD09GQ-AZ_006.json
+++ b/example/data/collections/s3_MOD09GQ-AZ_006_full_ingest_move/s3_MOD09GQ-AZ_006.json
@@ -1,6 +1,6 @@
 {
-  "name": "MOD09GQ",
-  "version": "007",
+  "name": "MOD09GQ-AZ",
+  "version": "006",
   "dataType": "MOD09GQ",
   "process": "modis",
   "duplicateHandling": "replace",
--- a/example/spec/helpers/Collections.js
+++ b/example/spec/helpers/Collections.js
@@ -37,7 +37,7 @@ const collectionExists = async (stackName, collection) => {
         }
         throw error;
       }
-      if (response.statusCode === 200) {
+      if (response.name && response.version) {
         return true;
       }
       return false;
@@ -126,4 +126,7 @@ const removeCollectionAndAllDependencies = async (params) => {
   }
 };
 
-module.exports = { removeCollectionAndAllDependencies };
+module.exports = {
+  collectionExists,
+  removeCollectionAndAllDependencies
+};
--- a/example/spec/helpers/Collections.js
+++ b/example/spec/helpers/Collections.js
@@ -37,7 +37,7 @@ const collectionExists = async (stackName, collection) => {
         }
         throw error;
       }
-      if (response.name === collection.name && response.version == collection.version) {
+      if (response.name === collection.name && response.version === collection.version) {
         return true;
       }
       return false;
--- a/example/spec/helpers/Collections.js
+++ b/example/spec/helpers/Collections.js
@@ -37,7 +37,7 @@ const collectionExists = async (stackName, collection) => {
         }
         throw error;
       }
-      if (response.name && response.version) {
+      if (response.name === collection.name && response.version == collection.version) {
         return true;
       }
       return false;
--- a/example/spec/helpers/Collections.js
+++ b/example/spec/helpers/Collections.js
@@ -128,5 +128,5 @@ const removeCollectionAndAllDependencies = async (params) => {
 
 module.exports = {
   collectionExists,
-  removeCollectionAndAllDependencies
+  removeCollectionAndAllDependencies,
 };
--- a/example/spec/parallel/changeGranuleCollections/ChangeGranuleCollectionWorkflowSpec.js
+++ b/example/spec/parallel/changeGranuleCollections/ChangeGranuleCollectionWorkflowSpec.js
@@ -1,5 +1,5 @@
 const { deleteExecution } = require('@cumulus/api-client/executions');
-const { collectionExists } = require('../../helpers/Collections');
+const { deleteCollection } = require('@cumulus/api-client/collections');
 const fs = require('fs');
 const { addCollections, addProviders } = require('@cumulus/integration-tests');
 const { deleteS3Object, s3ObjectExists } = require('@cumulus/aws-client/S3');
@@ -26,7 +26,7 @@ const {
 } = require('../../helpers/testUtils');
 const { waitForApiStatus } = require('../../helpers/apiUtils');
 const { setupTestGranuleForIngest } = require('../../helpers/granuleUtils');
-const { deleteCollection } = require('@cumulus/api-client/collections');
+const { collectionExists } = require('../../helpers/Collections');
 const workflowName = 'IngestAndPublishGranule';
 
 const granuleRegex = '^MOD09GQ\\.A[\\d]{7}\\.[\\w]{6}\\.006\\.[\\d]{13}$';
@@ -103,7 +103,7 @@ describe('The ChangeGranuleCollections workflow', () => {
       await addCollections(
         stackName,
         config.bucket,
-        collectionsDir
+        targetCollectionsDir
       );
     }
     // providers and test data
@@ -117,7 +117,6 @@ describe('The ChangeGranuleCollections workflow', () => {
         testSuffix
       ),
     ]);
-
 
     const inputPayloadJson = fs.readFileSync(inputPayloadFilename, 'utf8');
     // update test data filepaths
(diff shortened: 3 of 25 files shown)


tasks/post-to-cmr/.nycrc.json
```diff
@@ -5,6 +5,6 @@
   ],
   "statements": 94.0,
   "functions": 88.0,
-  "branches": 97.0,
+  "branches": 88.0,
```
Jkovarik, line 8: Wow, that's a big change.   I'm assuming that was upstream?

tasks/change-granule-collection-s3/src/index.ts
```diff
@@ -244,7 +243,8 @@ async function copyFileInS3({
   cmrObject: Object,
   s3MultipartChunksizeMb?: number,
 }): Promise<void> {
-  if (isCMRMetadataFile(targetFile)) {
+  log.warn('getting ready to post', JSON.stringify(sourceFile), JSON.stringify(targetFile));
+  if (isCMRFile(targetFile)) {
```
Jkovarik, line 247: Good change 👍🏻

example/data/collections/s3_MOD09GQ-AZ_006_full_ingest_move/s3_MOD09GQ-AZ_006.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "MOD09GQ",
-  "version": "007",
+  "name": "MOD09GQ-AZ",
```
Jkovarik, line 2: What was the reason for this change?

Jkovarik (commented):
Review still in progress, batch adding a set of comments/questions.

example/spec/parallel/changeGranuleCollections/ChangeGranuleCollectionWorkflowSpec.js
```diff
@@ -9,20 +9,22 @@ const {
...
   loadConfig,
-  createTestSuffix,
   createTimestampedTestId,
   uploadTestDataToBucket,
   createTestDataPath,
 } = require('../../helpers/testUtils');
 const { waitForApiStatus } = require('../../helpers/apiUtils');
 const { setupTestGranuleForIngest } = require('../../helpers/granuleUtils');
-const workflowName = 'IngestAndPublishGranuleWithOrca';
+const workflowName = 'IngestAndPublishGranule';
```
Jkovarik, line 27: This.... is probably a good change until we need Orca 👍🏻

example/spec/parallel/changeGranuleCollections/ChangeGranuleCollectionWorkflowSpec.js
```diff
@@ -35,11 +37,20 @@ const s3data = [
...
 const collectionsDir = './data/collections/s3_MOD09GQ_006_full_ingest';
 const targetCollectionsDir =
-  './data/collections/s3_MOD09GQ_007_full_ingest_move';
+  './data/collections/s3_MOD09GQ-AZ_006_full_ingest_move';
 
 const inputPayloadFilename =
   './spec/parallel/ingestGranule/IngestGranule.input.payload.json';
 
+async function getCMRClient(config) {
+  const lambdaFunction = `${config.stackName}-CreateReconciliationReport`;
```
Jkovarik, line 46: Nit: This .... is getting the CMR configuration directly from the deployed lambda environment variables.... which we are doing in other tests.       

We should probably, given how often we're doing this we consolidate it to a helper rather than boilerplate, but that doesn't have to be in this PR.

example/spec/parallel/changeGranuleCollections/ChangeGranuleCollectionWorkflowSpec.js
```diff
@@ -73,41 +86,40 @@ describe('The ChangeGranuleCollections workflow', () => {
...
+        ),
+        addProviders(
+          stackName,
+          config.bucket,
+          providersDir,
+          config.bucket
+        ),
+      ]);
+    } catch (error) {
+      console.log('error thrown in data setup', error);
```
Jkovarik, line 113: Shouldn't this throw as well?

tasks/post-to-cmr/.nycrc.json
etcart: it was, that's from jenny's change, I believe it was discussed

tasks/change-granule-collection-pg/schemas/config.json
```diff
@@ -101,6 +102,67 @@
     "s3MultipartChunksizeMb": {
       "type": ["number", "null"],
       "description": "S3 multipart upload chunk size in MB.  If none is specified, the default default_s3_multipart_chunksize_mb is used."
+    },
+    "oldGranules": {
```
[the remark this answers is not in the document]
etcart: no good reason, the value needs to be shunted into config, and then... either gotten there or shunted back into the main payload, and I just picked the simpler one. unless I'm missing an option 3?

example/data/collections/s3_MOD09GQ-AZ_006_full_ingest_move/s3_MOD09GQ-AZ_006.json
etcart: this needs to correspond to a collection that exists in cmr in order for post-to-cmr to not cry

example/spec/parallel/changeGranuleCollections/ChangeGranuleCollectionWorkflowSpec.js
```diff
@@ -73,41 +86,40 @@ describe('The ChangeGranuleCollections workflow', () => {
...
+    }
 
     const inputPayloadJson = fs.readFileSync(inputPayloadFilename, 'utf8');
     // update test data filepaths
     inputPayload = await setupTestGranuleForIngest(
       config.bucket,
       JSON.stringify({ ...JSON.parse(inputPayloadJson), pdr: undefined }),
       granuleRegex,
-      testSuffix,
+      '',
```
Jkovarik, line 122: Note: Reading through this method.... this should probably be fine....although based on the table structure of the PDRs table, we are probably updating the *same record* each time in the PDRs table if it's not cleaned up.   I can't see any reason that causes test instability,

Jkovarik (changes requested):
@etcart  this looks great 💯     No actionable feedback, just a couple of questions.  Should be quickly approve-able.

example/spec/parallel/changeGranuleCollections/ChangeGranuleCollectionWorkflowSpec.js
```diff
@@ -73,41 +86,40 @@ describe('The ChangeGranuleCollections workflow', () => {
       targetCollection.name,
       targetCollection.version
     );
-    provider = { id: `s3_provider${testSuffix}` };
+    provider = { id: 's3_provider' };
```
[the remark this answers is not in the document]
etcart: no strong reason, I purged collection unique-ness and assumed that provider needed it as well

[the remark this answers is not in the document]
Jkovarik: Fair enough - toss this in the nit/consider category then!

tasks/change-granule-collection-pg/schemas/config.json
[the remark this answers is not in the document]
Jkovarik: Well.... no, doing something like including an s3 object reference in the payload and having the lambda unpack it would be option 3, and that's sub-optimal. 

Config is generally for deployment static or ... well, config set values  whereas payload is for active things being done in the workflow.      The original object typing felt like it made a lot of sense at a gut level because of the *intent* of payload/config, but I don't know that it's worth a ton of time to consider until this task becomes something we're expecting users to intuit.   

I'll shove that off to nit category for now.

example/spec/parallel/changeGranuleCollections/ChangeGranuleCollectionWorkflowSpec.js
Jkovarik: Looking back at this, we should trap/ignore specific errors at a minimum *or* add logic that checks before creating and remove the try altogether.

etcart: I'm workign on logic checks (something else is going wrong that I don't understand) but yea, the original reason is that now it's not 100% that our collection doesn't exists because we're re-using the name

example/spec/helpers/Collections.js
```diff
@@ -37,7 +37,7 @@ const collectionExists = async (stackName, collection) => {
         }
         throw error;
       }
-      if (response.statusCode === 200) {
+      if (response.name && response.version) {
```
etcart, line 40: this doesn't return a status code, it return s collection. this was, in my testing, returning false no matter what, perhaps there's a better way to check this?
perhaps "if response.name === collection.name && response.version === collection.version"
but I wanted to push this up for consideration while I thought about it

Jkovarik (approved):
The updated commits look good, nice work.

PR #3943, opened 2025-03-21
CUMULUS-3961 Fixes for release of change-granule-collections

etcart:
**Summary:** QA fixes in release of this feature
Addresses [CUMULUS-3961: QA for change-granule-collections](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3961)

## Changes

* fixes config log in Postgres step that spams oldGranules and overwhelms logging
* adds maxRequestGranules parameter to api and api calls to limit overwhelming passes to bulkPatch and other api endpoints
* splits concurrency between s3 concurrency and api concurrency to play well with s3 limits
* limits granule requests to concurrency
* ensures s3 timeouts intelligently await s3 scale up before retrying
* removes an accidental no-await

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


tasks/change-granule-collection-pg/src/index.ts
```diff
@@ -77,9 +80,11 @@ function validateGranule(granule: ApiGranuleRecord): granule is ValidGranuleReco
 }
 
 function validateConfig(config: EventConfig): ValidEventConfig {
-  const newConfig = config;
+  const newConfig = clone(config) as ValidEventConfig;
   newConfig.concurrency = config.concurrency || 100;
-  return newConfig as ValidEventConfig;
+  newConfig.maxRequestGranules = config.maxRequestGranules || 1000;
```
etcart, line 85: warrants more code documentation what's happening here and what/why we're deleting

tasks/change-granule-collection-s3/src/index.ts
```diff
@@ -252,9 +251,13 @@ async function copyFileInS3({
         minTimeout: 2000,
         maxTimeout: 2000,
         onFailedAttempt: (error) => {
```
charleshuang80, line 253: So if I'm understanding this correctly, it will run `uploadCMRFile` with 5 retries, and this is stipulating what to do or log when it fails. Is that right?

tasks/change-granule-collection-s3/src/index.ts
```diff
@@ -332,7 +339,7 @@ async function copyGranulesInS3({
   await pMap(
     copyOperations,
     (operation) => operation(),
-    { concurrency: Number(concurrency || process?.env.concurrency || 100) }
+    { concurrency: concurrency }
```
charleshuang80, line 342: this is where you needed `50` before? so were you able to test and I assume your `pRetry` updates above are what helped?

tasks/change-granule-collection-pg/src/index.ts
```diff
@@ -77,9 +79,11 @@ function validateGranule(granule: ApiGranuleRecord): granule is ValidGranuleReco
 }
 
 function validateConfig(config: EventConfig): ValidEventConfig {
-  const newConfig = config;
+  const newConfig = config as ValidEventConfig;
   newConfig.concurrency = config.concurrency || 100;
-  return newConfig as ValidEventConfig;
+  newConfig.maxRequestGranules = config.maxRequestGranules || 1000;
+
```
charleshuang80, line 85: so `oldGranules` don't need to be deleted anymore?

tasks/change-granule-collection-pg/src/index.ts
```diff
@@ -177,8 +210,12 @@ async function changeGranuleCollectionsPG(
 
   const targetGranules = event.input.granules.filter(validateGranule);
   const oldGranulesByID: { [granuleId: string]: ValidGranuleRecord } = keyBy(oldGranules.filter(validateGranule), 'granuleId');
-  log.debug(`change-granule-collection-pg run with config ${JSON.stringify(config)}`);
-  for (const granuleChunk of chunkGranules(targetGranules, config.concurrency)) {
+
+  log.debug(`change-granule-collection-pg run with config ${JSON.stringify({
+    ...config,
+    oldGranules: undefined, // oldGranules needs to not be logged because it can be enormous
```
charleshuang80, line 216: I think specifying that log overflow and other hard to diagnose errors might be helpful.

charleshuang80 (commented):
I was curious what happened when long pan and all the release package.json changes were in, but it looks like that was fixed

tasks/change-granule-collection-s3/src/index.ts
etcart: yea, if it's a not a timeout, throw the error, otherwise give it a moment and try again

tasks/change-granule-collection-pg/src/index.ts
etcart: exactly, instead I'm skipping ghtem in the logging

tasks/change-granule-collection-s3/src/index.ts
etcart: exactly

etcart:
> I was curious what happened when long pan and all the release package.json changes were in, but it looks like that was fixed
yea this was originally branched from 18.5.3, I had to fix that to clean up the pr

tasks/change-granule-collection-pg/src/index.ts
```diff
@@ -124,8 +128,24 @@ async function cleanupS3File(
...
-    async () => await deleteS3Object(oldFile.bucket, oldFile.key),
-    { retries: 3, minTimeout: 2000, maxTimeout: 2000 }
+    () => deleteS3Object(oldFile.bucket, oldFile.key),
+    {
+      retries: 5,
+      minTimeout: 2000,
+      maxTimeout: 10000,
+      randomize: true,
+      onFailedAttempt:
+        /* istanbul skip next */
```
charleshuang80, line 138: what does this mean?

tasks/change-granule-collection-pg/src/index.ts
```diff
@@ -151,12 +171,28 @@ async function cleanupInS3(
...
+      return () => pRetry(
+        () => cleanupS3File(newFile, oldFile),
+        {
+          retries: 5,
+          minTimeout: 2000,
+          maxTimeout: 2000,
+          onFailedAttempt: (error) => {
+            if (error.toString().includes('RequestTimeout:')) {
+              log.warn(
+                `Error when deleting object ${newFile?.bucket}/${oldFile?.key} ::  ${error}, retrying`
```
charleshuang80, line 183: not sure if this is a typo or if I'm missing something. this is calling `cleanupS3File`, which is calling `deleteS3Object(oldFile.bucket, oldFile.key)`, so I'm not sure why the logging here is using `newFile?.bucket`.

tasks/change-granule-collection-pg/src/index.ts
charleshuang80: are you still adding any more comments here, or is it all added below?

tasks/change-granule-collection-pg/src/index.ts
etcart: this means to ignore it in code coverage, making it explicit that this won't, and really *can't* be unit tested

tasks/change-granule-collection-pg/src/index.ts
etcart: that's a great catch, actually that entire retry block Is a mistake, since immediately inside that block I run a more direct s3 call that I pRetry

tasks/change-granule-collection-s3/src/index.ts
```diff
@@ -59,6 +59,17 @@ function objectSourceAndTargetSame(
   return ((sourceFile.key === targetFile.key) && (sourceFile.bucket === targetFile.bucket));
 }
 
+export function logOrThrow(error: pRetry.FailedAttemptError, logString: string) {
```
charleshuang80, line 62: should this function be abstracted to use in the other task where it is retrying on this error else throwing?

etcart: I don't think so, only because the other task only uses this sort of thing once, and abstracting this means putting it in.... common? it's not a sticking point I'll do it if you really think we should, but it feels like a lot of effort for not much

tasks/change-granule-collection-s3/src/index.ts
```diff
@@ -67,7 +78,18 @@ async function metadataCollisionsMatch(
...
       `s3://${targetFile.bucket}/${targetFile.key}`
     ),
-    { retries: 5, minTimeout: 2000, maxTimeout: 2000 }
+    {
+      retries: 5,
+      minTimeout: 2000,
+      maxTimeout: 10000,
+      randomize: true,
+      onFailedAttempt:
+      /* istanbul ignore next */
```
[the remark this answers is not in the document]
etcart: Istanbul is nyc, our code coverage tool for node. nyc is what we actually call, but that's a wrapper for an app called Istanbul

tasks/change-granule-collection-pg/src/index.ts
etcart: this is it as I see it now

tasks/change-granule-collection-s3/src/index.ts
etcart: yea, it's possible we should, like, bake this into all S3 calls we make or something like that, but this is very specific to s3 timeout catchment

tasks/change-granule-collection-s3/src/index.ts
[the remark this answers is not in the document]
charleshuang80: ah, ok. so I can find it when searching, wondering if we might want to make that clear somewhere. probably no need to do anything right now, just mulling.

tasks/change-granule-collection-s3/tests/test-index.js
```diff
@@ -1208,3 +1208,27 @@ test.serial('changeGranuleCollectionS3 should parse fileName if not given in fil
     'MOD11A1.A2017200.h19v04.006.2017201090724.cmr.xml'
   ));
 });
+
+test('logOrThrow logs throws on a timeout, but passes otherwise', (t) => {
+  try {
+    throw new TypeError('RequestTimeout: this thing');
```
charleshuang80, line 1214: should there be an assertion for this to match?

tasks/change-granule-collection-s3/tests/test-index.js
```diff
@@ -1208,3 +1208,27 @@ test.serial('changeGranuleCollectionS3 should parse fileName if not given in fil
     'MOD11A1.A2017200.h19v04.006.2017201090724.cmr.xml'
   ));
 });
+
+test('logOrThrow logs throws on a timeout, but passes otherwise', (t) => {
```
charleshuang80, line 1212: I didn't think about testing this. But I think it would be good to test the retry or throw behavior, and if possible to match the expected logging. Can we test/simulate a timeout and a different error for at least one of the other functions that uses the logOrThrow? And can we test this in any of the integration tests?

charleshuang80: are the code comments to skip there because the retry or throw behavior cannot be tested properly in unit or integration tests?

etcart: the problem is I'm not sure how to reliably re-create a timeout case. I'm worried that's gonna make integration tests extremely unreliable in CI, and I'm not sure how to do it at all in units

the Istanbul ignores are because Istanbul is only for unit tests and I don't think unit testing that timeout failure case is a realistic thing to do

tasks/change-granule-collection-s3/tests/test-index.js
etcart: maybe? the assertion is "nothing goes wrong"
I can put. t.pass() to make it more explicit I think

etcart:
ok, I've added a test for one of these retry behaviors, along with what I think is the best fix for the s3 concurrency issue

tasks/change-granule-collection-s3/tests/test-index.js
```diff
@@ -1208,3 +1208,86 @@ test.serial('changeGranuleCollectionS3 should parse fileName if not given in fil
...
+        'this is a log output'
+      ),
+      {
+        message: 'some random error to throw',
+      }
+    );
+  }
+});
+
+test('s3 failures retry on timeout', async (t) => {
```
charleshuang80, line 1237: very nice

tasks/change-granule-collection-pg/src/index.ts
```diff
@@ -77,9 +81,12 @@ function validateGranule(granule: ApiGranuleRecord): granule is ValidGranuleReco
 }
 
 function validateConfig(config: EventConfig): ValidEventConfig {
```
charleshuang80, line 83: I know this function is super simple, but if we're unit testing all functions, we should have a simple test for it.

charleshuang80 (approved):
Looks good. Good work. Glad we got the retry/timeout tested in some form.
Added a comment about a function unit test that might be needed, but otherwise all looks good.

etcart:
> Looks good. Good work. Glad we got the retry/timeout tested in some form. Added a comment about a function unit test that might be needed, but otherwise all looks good.

woooooo!

PR #3953, opened 2025-04-02
CUMULUS-3944: update Terraform to use all lowercase columns names for aws_glue_catalog_table

Nnaga1:

Addresses https://bugs.earthdata.nasa.gov/browse/CUMULUS-3944

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- **CUMULUS-3788**
+  - Updated `@cumulus/launchpad-auth/getLaunchpadToken` to check if the token in s3 has been updated
+    before updating it with a new token
 - **CUMULUS-4077**
   - Update list/search endpoints and corresponding BaseSearch `@cumulus/db` logic to allow `countOnly` as a configuration-modifying query parameter that *only* returns a useful value for `meta.count` to allow users to get a count without returning results/incurring pagination/translation costs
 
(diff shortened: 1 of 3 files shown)


tf-modules/monitoring/dla.tf
```diff
@@ -44,27 +44,27 @@ resource "aws_glue_catalog_table" "dla_glue_table" {
     }
   
     columns {
-      name = "executionArn"
+      name = "executionarn"
```
npauzenga, line 47: When we update these names and apply, is there any chance we end up with duplicate columns? If I understand the ticket correctly, `executionArn` will always end up as `executionarn` in Glue. Changing it to all lowercase shouldn't result in both `executionarn` and `executionArn` columns. Is that correct?

Also were you able to replicate the output ASF saw? e.g. when you initially picked this up, you saw the changeset mentioned in the ticket?

npauzenga: From the ticket:
>but I've noticed that every time we deploy our terraform it seems to generate a plan to update the aws glue database column names

That's replicable without these changes?

npauzenga (changes requested):
Looks great @Nnaga1. Just had a couple questions about process.

tf-modules/monitoring/dla.tf
npauzenga: Ok, great. It seems like that validates the changes they were seeing and the fix removes the staged changes 👍

npauzenga (approved):
👍 Thanks @Nnaga1

PR #3957, opened 2025-04-04
CUMULUS-3788:Update launchpad token in s3

jennyhliu:

Addresses [CUMULUS-3788: Multiple DAACs: Frequent 401 errors in PostToCMR](https://bugs.earthdata.nasa.gov/browse/CUMULUS-3788)

## Changes

* Updated `@cumulus/launchpad-auth/getLaunchpadToken` to check if the token in s3 has been updated
    before updating it with a new token

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Changed
+
+- **CUMULUS-3788**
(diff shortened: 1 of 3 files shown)


jennyhliu:
@etcart I have added a comment to the ticket.  There will still be race conditions, but I think the fix will solve the issue DAACs are seeing.

etcart (approved):
talked through in person, will at least alleviate the problem, although I better solution might be called for in the future

PR #3971, opened 2025-04-23
Ecarton/fix webpack db

etcart:
hot fix exclude cloudflare:sockets in webpack so that we can publish

* because for some reason we could publish before without this
* but now we can't
* even though this bug was introduced in PG 8.11
* and we're on PG 8.13 and have been for a hot minute

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


Jkovarik (approved):
LGTM 👍🏻

PR #3974, opened 2025-04-25
Cumulus 4020

cqbanh:

Addresses [CUMULUS-4020: ElasticSearch (metrics) credential exposure ](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4020)

## Changes

* Added sanitizeSensitive function to mitigate credential exposure
* Update BaseSearch functions to sanitize the errors 

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


packages/es-client/search.js
```diff
@@ -29,6 +31,55 @@ const multipleRecordFoundString = 'More than one record was found!';
...
+ * Sanitizes sensitive data in error messages or logs
+ *
+ * @param {Error|string} input - The error object or message to sanitize
+ * @returns {Error|string} - Sanitized error or message
+ */
+const sanitizeSensitive = (input) => {
+  const sensitiveFields = [
+    process.env.METRICS_ES_PASS,
+    `${process.env.METRICS_ES_USER}:${process.env.METRICS_ES_PASS}`,
+  ].filter(Boolean).sort((a, b) => b.length - a.length);
```
etcart, line 44: why not remove [process.env.METRICS_ES_PASS, METRICS_ES_USER]?
this is a nit because this serves the intended purpose, but it seems like this introduces a bit of maintenance gotcha where it's important that these fields are sorted longest to shortest, and performed in order, such that it might break in the future, and simply leaving in ***:*** seems just as functional?

etcart (commented):
the changes as they stand look good, some small comments, but readily approvable.
however it looks like a section of the AC is missed here:
"Root cause of errors caused by special chars in passwords are understood, and tickets to remediate are created, as possible"
was this done and reported somewhere? I don't see it in the Jira ticket, apologies If I've just missed it

packages/es-client/search.js
```diff
@@ -358,23 +410,28 @@ class BaseSearch {
...
-    if (result.hits.total > 1) {
-      return { detail: multipleRecordFoundString };
-    }
-    if (result.hits.total === 0) {
-      return { detail: recordNotFoundString };
+      const resp = result.hits.hits[0]._source;
+      resp._id = result.hits.hits[0]._id;
+      return resp;
+    } catch (error) {
+      logger.error(sanitizeSensitive(error));
```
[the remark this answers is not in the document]
cqbanh: updated to use esCustomLogger.error(sanitizeSensitive(error));

packages/es-client/search.js
cqbanh: Removing METRICS_ES_PASS and METRICS_ES_USER might not be feasible with the current ES client version (5.6).
The esMetricsConfig uses these variables when metrics=true. Sanitizing them before logging seems like a good solution unless we update the ElasticSearch version.

I did remove process.env.METRICS_ES_PASS as a separate item and removed sorting as it adds complexity.

cqbanh:
> the changes as they stand look good, some small comments, but readily approvable. however it looks like a section of the AC is missed here: "Root cause of errors caused by special chars in passwords are understood, and tickets to remediate are created, as possible" was this done and reported somewhere? I don't see it in the Jira ticket, apologies If I've just missed it

Updated the code to handle URL-encoded username, password, and special characters. This should resolve the main issue here.

packages/es-client/search.js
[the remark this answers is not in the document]
etcart: wouldn't you then not need to call sanitizeSensitive here?
since the esCustomLogger.error is sanitizing internally?

packages/es-client/search.js
etcart: I'm not sure what you mean w/r/ to ES client version
if you remove password and user
```
const sensitiveFields = [
    process.env.METRICS_ES_USER,
    process.env.METRICS_ES_PASS,
  ].filter(Boolean);
```
this will remove the User name (if present) and the password (if present)
such that your logged output has a *****:***** value replacing the username and password
while removing the possibility that one is weird and borks the whole thing (you might see ecarton:****) if my username is configured weird, but my password is still protected

at least I think. I'm sorry if I'm off-base and misunderstanding something

etcart: we talked in more detail and determined that this was overcomplicating things, and might be a future work for more general logging

etcart (approved):
this looks great,
sorry for the wait on the re-review

Issue #3956, opened 2025-04-03
Granule Delete API Request Returns 400 Bad Request If Granule Doesn't Exist


jashkar:
### Current behavior: 
While using the Delete granule Cumulus API, we receive a 400 bad request if the granule doesn't exist in Cumulus.
### Expected behavior:
We expect a 404 not found or some other meaningful message if a granule doesn't exist.  We only expect a 400 bad request if the request was badly formed.
### Cumulus release: 
20.0.0

### Sample run:
```
$ curl -v -H "Authorization: Bearer $token" -X DELETE https://[cumulus url]/granules/[collection id]/[non-existing-granule]
{"statusCode":400,"error":"Bad Request","message":"Cannot read properties of undefined (reading 'cumulus_id')"}
```

### Related Cumulus logs:
```
2025-04-03T17:46:20.502Z	b92737ab-234b-457e-874b-85a41dbc92de	INFO	{
    "level": "debug",
    "message": "Initiating deletion of PG granule undefined",
    "sender": "@cumulus/api/granule-delete",
    "timestamp": "2025-04-03T17:46:20.502Z"
}
2025-04-03T17:46:24.084Z	b92737ab-234b-457e-874b-85a41dbc92de	ERROR	{
    "error": {
        "name": "TypeError",
        "message": "Cannot read properties of undefined (reading 'cumulus_id')",
        "stack": [
            "TypeError: Cannot read properties of undefined (reading 'cumulus_id')",
            "    at deleteGranuleAndFiles (/var/task/webpack:/granule-delete.ts:80:37)",
            "    at del (/var/task/webpack:/endpoints/granules.js:862:33)",
            "    at processTicksAndRejections (node:internal/process/task_queues:95:5)"
        ]
    },
    "level": "error",
    "message": "Cannot read properties of undefined (reading 'cumulus_id')",
    "sender": "@cumulus/api",
    "timestamp": "2025-04-03T17:46:24.084Z"
}

```

vhenckel:
PR: https://github.com/nasa/cumulus/pull/3989

PR #3989, opened 2025-05-19
Fixes Issue #3956

vhenckel:

Current behavior:
While using the Delete granule Cumulus API, we receive a 400 bad request if the granule doesn't exist in Cumulus.

Expected behavior:
We expect a 404 not found or some other meaningful message if a granule doesn't exist. We only expect a 400 bad request if the request was badly formed.

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


etcart:
I'm open to the idea of reviewing this, but it looks like there's a mangled merge in the beginning of this that duplicates a massive set of files with the suffix " 2.x" making it fairly hard to review.
can you clean those up?

etcart:
thanks, having a look at that one

PR #4000, opened 2025-06-16
CUMULUS-4077: Update reingestGranule to not set granule status when no granules payload

jennyhliu:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/example/cumulus-tf/cnm_response_task.tf
+++ b/example/cumulus-tf/cnm_response_task.tf
@@ -1,5 +1,5 @@
 locals {
-  cnm_response_version  = "3.1.0"
+  cnm_response_version  = "3.1.0-alpha.2-SNAPSHOT"
   cnm_response_filename = "cnmResponse-${local.cnm_response_version}.zip"
 }
 
@@ -8,9 +8,7 @@ resource "null_resource" "get_cnmResponse" {
     always_run = local.cnm_response_version
   }
   provisioner "local-exec" {
-    ## TODO uncomment when we have a cnm_response_version to support producerGranuleId
-    # command = "curl -s -L -o ${local.cnm_response_filename} https://github.com/podaac/cumulus-cnm-response-task/releases/download/v${local.cnm_response_version}/${local.cnm_response_filename}"
-    command = "curl -s -L -o ${local.cnm_response_filename} https://github.com/nasa/cumulus/releases/download/v20.1.3-alpha.2/cnmResponse.zip"
+    command = "curl -s -L -o ${local.cnm_response_filename} https://github.com/podaac/cumulus-cnm-response-task/releases/download/v${local.cnm_response_version}/${local.cnm_response_filename}"
   }
 }
 
--- a/packages/api/lambdas/bulk-operation.js
+++ b/packages/api/lambdas/bulk-operation.js
@@ -240,7 +240,6 @@ async function bulkGranuleReingest(
           ...apiGranule,
           ...(targetExecution && { execution: targetExecution }),
         };
-        await updateGranuleStatusToQueued({ apiGranule: apiGranuleToReingest, knex });
         await reingestHandler({
           apiGranule: apiGranuleToReingest,
           asyncOperationId: process.env.asyncOperationId,
--- a/packages/api/lib/ingest.js
+++ b/packages/api/lib/ingest.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const path = require('path');
+const get = require('lodash/get');
 
 const Lambda = require('@cumulus/aws-client/Lambda');
 const StepFunctions = require('@cumulus/aws-client/StepFunctions');
@@ -34,16 +35,22 @@ async function reingestGranule({
   updateGranuleStatusToQueuedMethod = updateGranuleStatusToQueued,
 }) {
   const knex = await getKnexClient();
-  await updateGranuleStatusToQueuedMethod({
-    apiGranule,
-    knex,
-    granulePgModel,
-  });
 
   const executionArn = path.basename(apiGranule.execution);
 
   const executionDescription = await StepFunctions.describeExecution({ executionArn });
   const originalMessage = JSON.parse(executionDescription.input);
+
+  // Only update the original granule to 'queued' if the original payload contains the granule.
+  // This avoids a situation where the original granule is updated to 'queued', but the
+  // reingest workflow creates a new granule, leaving the original granule stuck in 'queued'.
+  if (get(originalMessage, 'payload.granules.length', 0) > 0) {
+    await updateGranuleStatusToQueuedMethod({
+      apiGranule,
+      knex,
+      granulePgModel,
+    });
+  }
 
   const { name, version } = deconstructCollectionId(apiGranule.collectionId);
 
--- a/tasks/sync-granule/schemas/output.json.template
+++ b/tasks/sync-granule/schemas/output.json.template
@@ -37,6 +37,7 @@
           "version": {"type": "string"},
           "files": "{{files}}",
           "createdAt": {"type": "number"},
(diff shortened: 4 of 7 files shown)


packages/api/lambdas/bulk-operation.js
```diff
@@ -240,7 +240,6 @@ async function bulkGranuleReingest(
           ...apiGranule,
           ...(targetExecution && { execution: targetExecution }),
         };
-        await updateGranuleStatusToQueued({ apiGranule: apiGranuleToReingest, knex });
```
jennyhliu, line 243: This is redundant, reingestHandler sets the granule to queued

tasks/sync-granule/schemas/output.json.template
```diff
@@ -37,6 +37,7 @@
           "version": {"type": "string"},
           "files": "{{files}}",
           "createdAt": {"type": "number"},
+          "producerGranuleId": {"type": "string", "description": "Granule ID from the producer, if available"},
```
jennyhliu, line 40: Added to be consistent with output.json schema

packages/api/tests/lambdas/test-bulk-operation.js
```diff
@@ -722,23 +722,6 @@ test.serial('bulk operation BULK_GRANULE_REINGEST reingests granules returned by
   });
 });
 
-test.serial('bulk operation BULK_GRANULE_REINGEST sets the granules status to queued', async (t) => {
```
jennyhliu, line 725: granule status is set in reingestHandler(reingestGranule) and tested in packages/api/tests/lib/test-ingest.js

packages/api/tests/endpoints/granules/test-reingest-granules.js
```diff
@@ -96,7 +96,7 @@ test.after.always(async (t) => {
   });
 });
 
-test('PATCH request with reingest action queues granule and calls the reingestGranule function with expected parameters', async (t) => {
+test('PATCH request with reingest action calls the reingestGranule function with expected parameters', async (t) => {
```
jennyhliu, line 99: updateGranuleStatusToQueuedMethod is only called from within reingestGranule.

example/cumulus-tf/cnm_response_task.tf
```diff
@@ -8,9 +8,7 @@ resource "null_resource" "get_cnmResponse" {
     always_run = local.cnm_response_version
   }
   provisioner "local-exec" {
-    ## TODO uncomment when we have a cnm_response_version to support producerGranuleId
-    # command = "curl -s -L -o ${local.cnm_response_filename} https://github.com/podaac/cumulus-cnm-response-task/releases/download/v${local.cnm_response_version}/${local.cnm_response_filename}"
-    command = "curl -s -L -o ${local.cnm_response_filename} https://github.com/nasa/cumulus/releases/download/v20.1.3-alpha.2/cnmResponse.zip"
+    command = "curl -s -L -o ${local.cnm_response_filename} https://github.com/podaac/cumulus-cnm-response-task/releases/download/v${local.cnm_response_version}/${local.cnm_response_filename}"
```
Nnaga1, line 11: can you explain this change/why its needed? for context for me

packages/api/lib/ingest.js
```diff
@@ -34,17 +35,23 @@ async function reingestGranule({
...
   const executionArn = path.basename(apiGranule.execution);
 
   const executionDescription = await StepFunctions.describeExecution({ executionArn });
   const originalMessage = JSON.parse(executionDescription.input);
 
+  // Only update the original granule to 'queued' if the original payload contains the granule.
+  // This avoids a situation where the original granule is updated to 'queued', but the
+  // reingest workflow creates a new granule, leaving the original granule stuck in 'queued'.
+  if (get(originalMessage, 'payload.granules.length', 0) > 0) {
+    await updateGranuleStatusToQueuedMethod({
```
Nnaga1, line 48: why not change `updateGranuleStatusToQueuedMethod` to just not create a new granule then? is that whats creating a new granule upon reingest? 

why would it even get to reingest granule if there are no granules: `(get(originalMessage, 'payload.granules.length', 0) > 0) {`

packages/api/tests/endpoints/test-granules.js
```diff
@@ -554,6 +560,44 @@ test.serial('PATCH reingests a granule', async (t) => {
...
+  const body = response.body;
+  t.is(body.status, 'SUCCESS');
+  t.is(body.action, 'reingest');
+  t.true(body.warning.includes('overwritten'));
+
+  const updatedPgGranule = await getUniqueGranuleByGranuleId(
+    t.context.knex,
+    t.context.fakePGGranules[0].granule_id
+  );
+  t.is(updatedPgGranule.status, t.context.fakePGGranules[0].status);
```
Nnaga1, line 598: should there be something here explicitly to show that its NOT queued?

example/cumulus-tf/cnm_response_task.tf
jennyhliu: https://github.com/podaac/cumulus-cnm-response-task/releases/tag/v3.1.0-alpha.2-SNAPSHOT PODDAC has a release which supports producerGranuleId

packages/api/lib/ingest.js
jennyhliu: updateGranuleStatusToQueuedMethod updates current api granule to 'queued'.
Ingest workflow like CNMExampleWorkflow starts with payload which has no granules.

packages/api/tests/endpoints/test-granules.js
jennyhliu: It verifies updatedPgGranule.status is not changed and stays 'completed'

tasks/sync-granule/schemas/output.json.template
jennyhliu: It's missing in https://github.com/nasa/cumulus/pull/3997
Without the update, the build generates output.json without producerGranuleId, and it shouldn't

Nnaga1: ohhh ok just adding something that was missing from another PR, I guess thats fine 👍

example/cumulus-tf/cnm_response_task.tf
Nnaga1: ohhhhh ok, thats cool they made that and let us use it like this, I guess maybe they have some stake in this work so 🤷 ok makes sense 👍

packages/api/lib/ingest.js
Nnaga1: so where in the workflow does this happen: `but the// reingest workflow creates a new granule,` when the message is being written to sfEventDbRecords?

packages/api/tests/endpoints/test-granules.js
Nnaga1: that seems explicit enough honestly 👍

packages/api/lib/ingest.js
jennyhliu: reingest action starts a workflow based on the previous execution on the granule, when the workflow completes, the granules will be written to db by sfEventDbRecords.
https://github.com/nasa/cumulus/blob/jl/CUMULUS-4077/packages/api/lib/ingest.js#L58

If the workflow creates a new unique granuleId, then a new granule is added.
https://github.com/nasa/cumulus/blob/jl/CUMULUS-4077/example/cumulus-tf/cnm_workflow.asl.json#L68

Nnaga1: ohhhh ok, that makes sense, it should only be queued if the payload contains the original granules so if it does create a new granule from no payload that wouldnt also be queued, ok

PR #4001, opened 2025-06-17
CUMULUS-4039: allow slashes in provider name and collection name (in api-client)

charleshuang80:

Addresses [CUMULUS-4039: allow slashes in provider name and collection name](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4039)

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


packages/api-client/src/collections.ts
```diff
@@ -63,7 +63,7 @@ export const updateCollection = async (params: {
       httpMethod: 'PUT',
       resource: '/{proxy+}',
       headers: { 'Content-Type': 'application/json' },
-      path: `/collections/${collection.name}/${encodeURIComponent(collection.version)}`,
+      path: `/collections/${encodeURIComponent(collection.name)}/${encodeURIComponent(collection.version)}`,
```
npauzenga, line 66: Nit: Just for the sake of consistency could we do a helper function here like we have in `encodeGranulesURIComponent` in api-client/src/granules.ts?

npauzenga: Also could you remind me where the decoding happens:? The api-client receives the request, encodes the params, and then passes it off to the appropriate API endpoint. Does the... router... handle decoding the URI encoded params? It looks like it does happen, just having a total lapse as to... where 😆

packages/api-client/src/providers.ts
```diff
@@ -56,7 +56,7 @@ export const deleteProvider = async (params: {
     payload: {
       httpMethod: 'DELETE',
       resource: '/{proxy+}',
-      path: `/providers/${providerId}`,
+      path: `/providers/${encodeURIComponent(providerId)}`,
```
npauzenga, line 59: Same comment as above, we could probably use a path helper here for consistency. Annoying comment, I'm sorry.

packages/api-client/tests/test-collections.js
```diff
@@ -9,7 +9,7 @@ const { randomString } = require('../../common/test-utils');
 test.before((t) => {
   process.env.stackName = randomString();
   t.context.testPrefix = 'unitTestStack';
-  t.context.collectionName = 'testCollection';
+  t.context.collectionName = 'testCollection/name';
```
npauzenga, line 12: Is there any value in testing other special characters in addition to `/`?

packages/api-client/tests/test-collections.js
```diff
@@ -19,7 +19,7 @@ test('deleteCollection calls the callback with the expected object', async (t) =
     payload: {
       httpMethod: 'DELETE',
       resource: '/{proxy+}',
-      path: `/collections/${t.context.collectionName}/${encodeURIComponent(t.context.collectionVersion)}`,
+      path: `/collections/${encodeURIComponent(t.context.collectionName)}/${encodeURIComponent(t.context.collectionVersion)}`,
```
npauzenga, line 22: Do we need tests (maybe not here) that assert the actual encoding works and is able to decode the params? Perhaps not if that logic is buried in one of our packages. Just want to be sure it's tested somewhere by someone.

npauzenga (changes requested):
Looks great @charleshuang80. Just questions for my information, nothing crazy.

packages/api-client/src/collections.ts
charleshuang80: So my reading of the `encodeGranulesURIComponent` is that it is checking to see if `collectionId` is provided, and based on that, is creating a path (that does some encoding) either with or without the `collectionId`. So I think that is different use case, and the reason for the helper is that there is actually some logic being run through and the helper is allowing that to not be repeated everywhere.
So the `collection.version` was already being encoded, without a helper. Following that, it seems that the provider shouldn't need a helper if it's just one part being encoded. And then for using a helper here, I'm not sure how much we're actually saving or making more convenient by creating a helper, since we are just encoding one more part of the path and there is no logic involved.
Happy to discuss this further if you disagree, but it seems to me that adding the helpers aren't necessary and wouldn't really help all that much.

charleshuang80: I looked into it briefly, but I'm actually not sure where the decoding happens. I can't see in the `api-client` lib where or how the router is being used. But if the router is involved, then I would suspect that is what handles it, because it is the path that needs to be interpreted properly.
If this isn't a satisfactory answer, I can keep digging or we can ask in the team slack.

packages/api-client/src/providers.ts
charleshuang80: See my above longer answer

packages/api-client/tests/test-collections.js
charleshuang80: I'm only testing `/` as that is what was specifically asked for and what I'm trying to address. I think unless there are specific characters we think will be used or will be problematic, I'm not sure there is value in testing anything else. And I am not aware of any other special characters that will or may be used, but if there are any please let me know.

packages/api-client/tests/test-collections.js
charleshuang80: So, I'm just following the setup of the pre-existing test(s). I would have preferred to actually be creating and testing against entries in the DB, but I didn't see that in any of the tests I looked at for this package. The granules tests almost look like they might, but they're actually doing a lot of mocking and not really creating (or cruding) anything.
Is there another place I should add tests, or is there another test I should emulate?
I _think_ that the integration test working should mean all this stuff works, what do you think?

packages/api-client/src/collections.ts
npauzenga: Yeah `encodeGranulesURIComponent` does do more than just encode but it's still a nice wrapper to simplify a bit. I see we didn't go that way for e.g. providers though so I guess it's fine to not use a helper here.

So long as the decoding works as expected and you're able to either assert in an api-client test that it decodes properly or see in CloudWatch logs that the encoded params are decoded before being passed, that's fine.

packages/api-client/tests/test-collections.js
npauzenga: Do other api-client tests test different characters? Is this the first api-client test that tests encoding at all?

packages/api-client/tests/test-collections.js
npauzenga: >Is there another place I should add tests, or is there another test I should emulate?

Are there other api-client tests that deal with encoding/decoding issues? Test data that includes special characters? If you can assert that this is well tested because it follows the pattern of other endpoints, great.

packages/api-client/tests/test-collections.js
charleshuang80: Other api-client tests do _not_ test different characters. These are not the first tests for encoding. The collection tests were already testing encoding of collectionVersion, and the granule tests test the encoding of the helper method you mentioned, so tests encoding of the collectionId and the granuleId. All those tests just test `/` encoding (they also include `-`, but that's not a char that requires special encoding to my knowledge).

packages/api-client/tests/test-collections.js
charleshuang80: So... I would assert that this follows the pattern of the other endpoints. But not much more than that. I found nothing dealing with testing that decoding worked properly in these unit tests. But in the integration test I modified, the collectionVersion has a `/`, so I believe that was testing the encoding and decoding worked, and I _think_ that is where the verification that the encoding and decoding of the new values with `/`s is also happening. (maybe that's where the 'well tested' can be asserted?)
I am definitely not opposed to adding more or better testing. But I don't know where to do that, and I'm not sure what type of tests to add.

npauzenga: But you are confident this meets the requirements? I'm ok with the tests here matching the others if YOU have enough confidence that this is working as expected. In cases like this I tend to either add additional tests OR do some ad-hoc testing to ensure the user gets what they expect. If additional automated tests don't make sense because there isn't a good spot for them and this feature is mostly handled by library functions, etc. that's fine, so long as we can confidently assert that this does what the ticket calls for.

npauzenga: Discussed in a huddle. The existing tests have been updated with the use case in the ticket's AC. Modifying those tests according to the ticket's AC led to successful reproduction of the failure, adding the encoding fixed the errors. I'd call this successful testing of the AC.

npauzenga (approved):
Nice work @charleshuang80

PR #4002, opened 2025-06-18
Ecarton/cumulus 4123 archive database

etcart:
**Summary:** adding "archived" column to granules and executions databases, along with associated code and tests

Addresses [CUMULUS-4123: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4123)

## Changes

* adds migration to incorporate archived column to granules and archives
* alters tests to account for the default false archived value
* adds GranuleSearch and ExecutionSearch tests to check archived query

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Notable Changes
 
+- **CUMULUS-4194**
+  - update cumulus-process to 1.5.0
 - **CUMULUS-4131**
   - Users upgrading to this release will be required to update their terraform version to at least 1.12.2. Reference migration instructions are included at [https://nasa.github.io/cumulus/docs/next/upgrade-notes/upgrade-terraform-1.12](https://nasa.github.io/cumulus/docs/next/upgrade-notes/upgrade-terraform-1.12)
 - **CUMULUS-4176**
@@ -21,11 +23,51 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
   - Updated Terraform version requirement to `>=1.12.2`
   - Updated AWS provider requirement to `5.100.x`
 
+### Changed
+
+- **CUMULUS-4142, CUMULUS-4144**
+  - Updated the S3 credentials endpoint attached to TEA to delete the access token after successful authentication.
+  - Configured both Cumulus distribution and S3 credentials to set the SameSite attribute on cookies in the response.
+
 ### Fixed
 
 - **CUMULUS-4177**
   - Update form-data sub-dependency to safe version ^4.0.4
 - **CUMULUS-4174**
+  - Fix broken CreateReconciliationReportSpec test cleanup
+- **CUMULUS-4170**
+  - Upgrade Node Docker image from buster to bullseye for a compatible debian version
+
+## [v20.2.1] 2025-08-14
+
+### Changed
+
+- **CUMULUS-4142, CUMULUS-4144**
+  - Updated the S3 credentials endpoint attached to TEA to delete the access token after successful authentication.
+  - Configured both Cumulus distribution and S3 credentials to set the SameSite attribute on cookies in the response.
+
+## [v20.2.0] 2025-08-06 Updated
+
+### Migration Notes
+
+From this release forward, Cumulus Core will be tested against PostgreSQL v17. Users
+should migrate their datastores to Aurora PostgreSQL 17.4+ compatible data
+stores as soon as possible after upgrading to this release.
+
+#### Database Upgrade
+
+Users utilizing the `cumulus-rds-tf` module should reference [cumulus-rds-tf upgrade
+instructions](https://nasa.github.io/cumulus/docs/upgrade-notes/upgrade-rds-cluster-tf-postgres-17).
+
+### Fixed
+
+- **CUMULUS-4176**
+  - Updated `engine_version` default value to `17.4` in `example/rds-cluster-tf/variables.tf`.
+  - Updated `tf-modules/cumulus-rds-tf` module to take additional parameter `enable_upgrade` in support of
+    migration from Aurora PostgreSQl v13 to v17.
+- **CUMULUS-4177**
+  - Update form-data sub-dependency to safe version ^4.0.4
+- **CUMULUS_4174**
   - Fix broken CreateReconciliationReportSpec test cleanup
 - **CUMULUS-4170**
   - Upgrade Node Docker image from buster to bullseye for a compatible debian version
@@ -8646,7 +8688,9 @@ Note: There was an issue publishing 1.12.0. Upgrade to 1.12.1.
 ## [v1.0.0] - 2018-02-23
 
 
-[Unreleased]: https://github.com/nasa/cumulus/compare/v20.1.2...HEAD
+[Unreleased]: https://github.com/nasa/cumulus/compare/v20.2.1...HEAD
+[v20.2.1]: https://github.com/nasa/cumulus/compare/v20.2.0...v20.2.1
+[v20.2.0]: https://github.com/nasa/cumulus/compare/v20.1.2...v20.2.0
 [v20.1.2]: https://github.com/nasa/cumulus/compare/v20.1.1...v20.1.2
 [v20.1.1]: https://github.com/nasa/cumulus/compare/v20.0.1...v20.1.1
 [v20.0.1]: https://github.com/nasa/cumulus/compare/v20.0.0...v20.0.1
--- a/bamboo/bootstrap-tf-deployment.sh
+++ b/bamboo/bootstrap-tf-deployment.sh
@@ -98,3 +98,6 @@ echo "Deploying Cumulus example to $DEPLOYMENT"
   -var "pdr_node_name_provider_bucket=$PDR_NODE_NAME_PROVIDER_BUCKET" \
   -var "rds_admin_access_secret_arn=$RDS_ADMIN_ACCESS_SECRET_ARN" \
   -var "orca_db_user_password=$ORCA_DATABASE_USER_PASSWORD" \
+  -var "metrics_es_host=$METRICS_ES_HOST" \
+  -var "metrics_es_username=$METRICS_ES_USER" \
+  -var "metrics_es_password=$METRICS_ES_PASS" \
--- a/docs/development/release.md
+++ b/docs/development/release.md
@@ -1,7 +1,8 @@
 ---
 id: release
+title: Versioning and Releases
+hide_title: false
 ---
-# Versioning and Releases
 
 ## Versioning
 
@@ -13,7 +14,7 @@ Read more about the semantic versioning [here](https://docs.npmjs.com/getting-st
 
 :::note
 
-This is only necessary when preparing a release for a new major version of Cumulus (e.g. preparing to go from `6.x.x` to `7.0.0`).
+This is only necessary when preparing a release for a new major version of Cumulus (e.g., going from `6.x.x` to `7.0.0`) or for a minor version release that includes significant updates.
 
 :::
 
@@ -26,17 +27,20 @@ You should create an entirely new deployment for this testing to replicate the e
 Pre-release testing steps:
 
 1. Checkout the [cumulus-template-deploy](https://github.com/nasa/cumulus-template-deploy) repo
-2. Update the deployment code to use the latest release artifacts if it wasn't done already. For example, assuming that the latest release was `5.0.1`, update the deployment files as follows:
+2. Update the deployment code to use the latest release artifacts if it wasn't done already. For example, assuming that the latest release was `20.0.1`, update the deployment files as follows:
 
     ```text
+     # in rds-cluster-tf/main.tf
+    source = "https://github.com/nasa/cumulus/releases/download/v20.0.1/terraform-aws-cumulus-rds.zip"
+
     # in data-persistence-tf/main.tf
-    source = "https://github.com/nasa/cumulus/releases/download/v5.0.1/terraform-aws-cumulus.zip//tf-modules/data-persistence"
+    source = "https://github.com/nasa/cumulus/releases/download/v20.0.1/terraform-aws-cumulus.zip//tf-modules/data-persistence"
 
     # in cumulus-tf/main.tf
-    source = "https://github.com/nasa/cumulus/releases/download/v5.0.1/terraform-aws-cumulus.zip//tf-modules/cumulus"
+    source = "https://github.com/nasa/cumulus/releases/download/v20.0.1/terraform-aws-cumulus.zip//tf-modules/cumulus"
     ```
 
-3. For both the `data-persistence-tf` and `cumulus-tf` modules:
+3. For the `rds-cluster-tf`, `data-persistence-tf` and `cumulus-tf` modules:
    1. Add the necessary backend configuration (`terraform.tf`) and variables (`terraform.tfvars`)
       - You should use an entirely new deployment for this testing, so make sure to use values for `key` in `terraform.tf` and `prefix` in `terraform.tfvars` that don't collide with existing deployments
    2. Run `terraform init`
@@ -44,10 +48,13 @@ Pre-release testing steps:
 4. Checkout the `master` branch of the `cumulus` repo
 5. Run a full bootstrap of the code: `npm run bootstrap`
 6. Build the pre-release artifacts: `./bamboo/create-release-artifacts.sh`
-7. For both the `data-persistence-tf` and `cumulus-tf` modules:
+7. For all three modules:
    1. Update the deployment to use the built release artifacts:
 
       ```text
+      # in rds-cluster-tf/main.tf
+      source = "[path]/cumulus/terraform-aws-cumulus-rds.zip"
+
       # in data-persistence-tf/main.tf
       source = "[path]/cumulus/terraform-aws-cumulus.zip//tf-modules/data-persistence"
 
@@ -59,7 +66,7 @@ Pre-release testing steps:
    3. Run `terraform init`
    4. Run `terraform apply`
 8. Review the `CHANGELOG.md` for any post-deployment migration steps and confirm that they are successful
-9. Delete your test deployment by running `terraform destroy` in `cumulus-tf` and `data-persistence-tf`
+9. Delete your test deployment by running `terraform destroy` in `cumulus-tf`, `data-persistence-tf` and `rds-cluster-tf`
 
 ## Updating Cumulus version and publishing to NPM
 
@@ -173,7 +180,7 @@ If there is exists a PR in the cumulus-dashboard repo with a name containing: "V
 
 ### 4. Publish async_operations image to Docker Hub
 
-For consistency and security reasons, a [new docker image for async_operations should be created and published](../../packages/api/ecs/async-operation/README.md) for each release and pushed to Docker Hub and possibly AWS ECR. Make sure to follow the instructions for **Updating the Cumulus deployment configuration**
+For consistency and security reasons, a [new docker image for async_operations should be created and published](https://github.com/nasa/cumulus/blob/master/packages/api/ecs/async-operation/README.md) for each release and pushed to Docker Hub and possibly AWS ECR. Make sure to follow the instructions for **Updating the Cumulus deployment configuration**
 
 ### 5. Update CHANGELOG.md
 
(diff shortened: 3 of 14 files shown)


CHANGELOG.md
```diff
@@ -32,6 +32,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 - **CUMULUS-4108**
   - Added standalone lambda function code to scan and terminate old instances when they pass their 90 day expiration
+- **CUMULUS-4123**
+  - add "archived" column to granules and executions tables
+  - add "archived" field to associated data types and schemas
```
npauzenga, line 83: (From Slack) Do we have an ADR?

packages/db/src/migrations/20250617190412_add_archived_and_index.ts
```diff
@@ -0,0 +1,41 @@
+import { Knex } from 'knex';
+
+export const up = async (knex: Knex): Promise<void> => {
```
npauzenga, line 3: Do we have a sense for how long this takes on very large tables?

Is the intention here with the conditional to allow users to populate this column separately?

packages/db/src/migrations/20250617190412_add_archived_and_index.ts
```diff
@@ -0,0 +1,41 @@
...
+  if (!(await knex.schema.hasColumn('granules', 'archived'))) {
+    await knex.schema.table('granules', (table) => {
+      table.boolean('archived')
```
npauzenga, line 6: Is it worth having this as NotNull?

packages/db/src/translate/executions.ts
```diff
@@ -45,6 +45,7 @@ export const translatePostgresExecutionToApiExecutionWithoutDbQuery = ({
     asyncOperationId,
     collectionId,
     parentArn,
+    archived: executionRecord.archived,
```
npauzenga, line 48: Depending on what we're doing with null values, we could explicitly make this a boolean. Are we going to support `null` `archived` values?

npauzenga: It looks like yes if it's an optional field. In that case we could translate this to something like `archived: executionRecord.archived? true: false,`

packages/db/tests/search/test-ExecutionSearch.js
```diff
@@ -622,3 +628,27 @@ test('ExecutionSearch only returns count if countOnly is set to true', async (t)
...
+    archived: true,
+  };
+  const dbSearch = new ExecutionSearch({ queryStringParameters });
+  const response = await dbSearch.query(knex);
+  response.results.forEach((ExecutionRecord) => {
+    t.is(ExecutionRecord.archived, true);
+  });
+});
+
+test('ExecutionSearch with archived: false pulls only non-archive granules', async (t) => {
```
npauzenga, line 644: Depending on what we decide re: null values, it may be worth a test. I could also be skipping an important piece here where null isn't valid.

CHANGELOG.md
npauzenga: Specifically I'm curious what the intention is when something is `archived: true`. Is it just that it doesn't appear in search results? Does it come back from a LIST query? I'm assuming it'll return via GET as that's explicit...

packages/db/src/migrations/20250617190412_add_archived_and_index.ts
etcart: I think it *is* worth not-nulling it. because the migration sets it as 'default: false' it shouldn't really be *able* to be null, but that can at least be more explicit here.

CHANGELOG.md
etcart: yes, I've posted an ADR, although I'm still editing it.
https://wiki.earthdata.nasa.gov/pages/viewpage.action?pageId=454330127

etcart: done editing it.
the idea is that it doesn't appear *by default* in *dashboard* search results. a LIST query would, by default, return whatever, unless the parameter "archived: true", or "archived: false" is added. (as written right now, perhaps that could be different if you think it should, but I don't think it should

packages/db/src/migrations/20250617190412_add_archived_and_index.ts
etcart: I'll get some exact numbers with the lp snapshot but it was seconds

etcart: by "the conditional" do you mean this 
` if (!(await knex.schema.hasColumn('granules', 'archived'))) {`
statement?
this is simply "don't blow yourself up"
if it tries to alter the table schema to add a column that already exists, it'll fail.
or do you mean a different conditional

packages/db/src/migrations/20250617190412_add_archived_and_index.ts
etcart: set it to notNullable()

packages/db/src/translate/executions.ts
etcart: or Boolean(executionRecord.archived) but I think making it notNullable should solve this

CHANGELOG.md
npauzenga: Nice. What is the plan for user docs on this? Are we gonna add a page to https://nasa.github.io/cumulus/docs/category/features or similar? Not suggesting it needs to be done for this PR, just that it's done for a release. It would probably be good to spell out how users can actually use the flag in their queries (as well as the cron job you're working).

packages/db/src/migrations/20250617190412_add_archived_and_index.ts
npauzenga: Yup, that's the one. It's good to have regardless but I was wondering if we were expecting some users to populate this column _outside_ of the migrations for some reason. That was why I was thinking we had this check.

It is a query and queries take <an amount> of time during the limited window the db migration lambda is running but I would expect it to be fast. If this whole migration executes against the LP snapshot in a few seconds, well under the timeout, that's great.

packages/db/src/types/execution.ts
```diff
@@ -17,10 +17,12 @@ export interface PostgresExecution {
   updated_at?: Date | null,
   url?: string | null,
   workflow_name?: string | null,
+  archived?: boolean
```
npauzenga: Shouldn't be optional if we make it notNull

packages/db/src/types/granule.ts
```diff
@@ -4,6 +4,7 @@ export interface PostgresGranuleUniqueColumns {
   collection_cumulus_id: number,
 }
 export interface PostgresGranule extends PostgresGranuleUniqueColumns {
+  archived?: boolean,
```
npauzenga: Same as above, shouldn't be optional

packages/db/src/migrations/20250617190412_add_archived_and_index.ts
npauzenga: Agreed that hopefully it shouldn't be possible to be null. This would be just for the sake of being explicit.

Dumb question: Does this `defaultTo` impact new record writes? If I PUT or POST a new granule without the `archived` field, will it populate automagically as a `false`?

etcart:
> Looks great @etcart. Added a couple minor questions/comments. Also just curious -- is there a use case for marking the Collection `archived`? Right now if a user has several Collections that they know are old/archivable, they'd have to update each granule to `archived: true`. Is there value/would it be easier to do it at the Collection level? This is completely making up a use case that I'm not sure exists so totally fine if that doesn't match what they're expecting.

I don't think so? If a collection stops being added to, then eventually all of its granules would fall off to an archived state. Past that I think this implies that archival is more in some way than just a "should we look for it in the dashboard" and I think we can consider expanding the functionality to apply to all tables of we think there's reason to, but at least in this iteration, there's no significant optimization to be made archiving a collection.
If I marked a collection as archived, would granule search follow up with every grabule's collection to see if it should be thrown out? I think adding a join to check collection archived state would be *worse* not better from a performance perspective

packages/db/src/migrations/20250617190412_add_archived_and_index.ts
etcart: yea, it'll magically be false

packages/db/src/migrations/20250617190412_add_archived_and_index.ts
etcart: for the lp snapshot:
add archived column to granules table (163328300 granules) took  53.283 ms
add archived column to execution table (389771061 executions) took 37.760 ms
however as I've gone and revisited adding the index this takes a hot minute. I think this is because previously I was juicing the CPU of the database, and maybe it's worth testing with the CPU turned way up, but realistically I don't think that's the play, this makes me think the indexing should be done manually

packages/db/src/types/execution.ts
etcart: fixed

packages/db/src/types/granule.ts
etcart: fixed

packages/db/tests/search/test-ExecutionSearch.js
etcart: I think this is fixed with null being declared invalid

packages/db/src/migrations/20250617190412_add_archived_and_index.ts
npauzenga: Is this covered in the docs? I saw a lot added and will need to be reviewed. But are you asserting that this is covered there? In the past have we done something similar where we add the column and recommend adding the index separately? I believe so, just want to be sure we're matching the instructions/format.

CHANGELOG.md
npauzenga: I see the docs added here. Will review.

docs/features/record_archival.md
```diff
@@ -0,0 +1,56 @@
...
+id: db_record_archive
+title: Database Record Archival
+hide_title: false
+---
+
+This documentation explains the database record archival and associated functionality.
+
+## Database Record Archive Column
+
+The cumulus database tables "granules" and "executions" contain a field "archived" which is outside of normal data as it contains nothing useful to the end user, but instead is part of query structure to optimize database search.
```
npauzenga: What do you mean "normal" data? Do we need to make this distinction?

docs/features/record_archival.md
```diff
@@ -0,0 +1,56 @@
...
+
+A granule or execution will by default be false, but once old enough (age set by the DAAC) will be flagged with archived=true.
```
npauzenga:
```suggestion
A granule or execution will by default be `archived=false`, but once old enough (age set by the DAAC) will be flagged with `archived=true`.
```
Just trying to be as explicit as possible.

docs/features/record_archival.md
```diff
@@ -0,0 +1,56 @@
...
+
+This makes no material difference to the content or state of the record, but does allow a granule or execution search query to access un-archived records more rapidly and at lower cost by eliminating from consideration a majority of records.
+
+## Cumulus Dashboard Search
+
+The cumulus dashboard, by default, performs searches against granules and executions excluding records that are archived. if necessary, a user can toggle archived searching on in order to look at older records, but accepts a performance hit in doing so.
```
npauzenga: Is this logic in Core or in the Dashboard? If the default behavior is set on the dash we probably don't want to talk about that here as, if that needs to change in the future, we likely won't remember to change it here. 

I think we could call out that the dash supports it and link to the docs there (or something similar that removes the dash specific implementation from these docs).

docs/features/record_archival.md
```diff
@@ -0,0 +1,56 @@
...
+        archived: false,
+        limit: 20,
+        sort_key: ['-updated_at']
+    }
+})
+```
+
+This query would ask for the most recent 20 granules from the collection 'COLLECTION1' *which are not archived*.
+
+The key reason to do this is performance, a search with archived: false will be more performant in most cases than a general search as it allows the query to eliminate old, out of consideration records, more simply and rapidly.
```
npauzenga:
```suggestion
The key reason to do this is performance, a search with `archived: false` will be more performant in cases where records have been archived and therefor removed from these query results.
```

docs/features/record_archival.md
```diff
@@ -0,0 +1,56 @@
...
+    query: {
+        collectionId: 'COLLECTION1',
+        archived: false,
+        limit: 20,
+        sort_key: ['updated_at']
+    }
+})
+```
+
+ will be similar performance or even marginally worse than searching without setting archived: false.
```
npauzenga: Do we have a suggestion we can share? Is it "be mindful of your sorting"?

docs/upgrade-notes/archived_column_indexing.md
```diff
@@ -0,0 +1,93 @@
...
+title: Add and Index Archived Column
+---
+
+## Background
+
+To optimize record searching we are adding a boolean column to archive granules and executions records. these should be btree indexed, adding very little additional size to the database while allowing us to better make the most common queries from the database: getting the most recent records according to a search query.
+
+## Apply Changes in Production Environment
+
+With a large database (e.g. any daac production database), these columns and indices should be applied manually to ensure lambda timeouts are not exceeded
```
npauzenga:
```suggestion
With a large database (e.g. any DAAC production database), these columns and indices should be applied manually to ensure lambda timeouts are not exceeded
```

npauzenga: I think we should include some indication of how much time. This is a good example but doesn't necessarily need to be as detailed. I imagine though that the first question is going to be "How long does this take?"

[docs/features/record_archival.md](https://github.com/nasa/cumulus/blob/master/docs/upgrade-notes/update-cumulus_id-type-indexes-CUMULUS-3449.md#apply-the-changes-in-production-environment)

docs/upgrade-notes/archived_column_indexing.md
```diff
@@ -0,0 +1,93 @@
...
+    From AWS CLI: aws ssm start-session --target `EC2 Instance ID`.
+
+    :::note Remember to take a note on which instance you run the commands.
+
+2. Install tmux and postgres client
+
+    ```sh
+    sudo yum install -y tmux
+    # Amazon Linux 2
+    sudo amazon-linux-extras install postgresql13
```
npauzenga: 13? Not 17?

docs/upgrade-notes/archived_column_indexing.md
```diff
@@ -0,0 +1,93 @@
...
+
+    psql -h <Endpoint for writer instance> -p <Port for database or 5432> -d <cumulus database name> -U <database admin user> -W
+    #e.g. psql -h cumulus-dev-rds-cluster.cluster-xxx.us-east-1.rds.amazonaws.com -p 5432 -d cumulus_test_db -U cumulus_test -W
+
+    # Use -f option to run the SQL commands from a file, -o option to write output to file
+    psql -h <Endpoint for writer instance> -p <Port for database or 5432> -d <cumulus database name> -U <database admin user> -f 20250617190412_add_archived_and_index.sql -W
+    ```
+
+    The following are SQL commands, and  20250617190412_add_archived_and_index.sql is available
+    [here](https://raw.githubusercontent.com/nasa/cumulus/master/packages/db/src/migrations/ 20250617190412_add_archived_and_index.sql):
```
npauzenga: Link is broken with an extra space

docs/upgrade-notes/archived_column_indexing.md
```diff
@@ -0,0 +1,93 @@
...
+
+    ```sql
+ALTER TABLE granules ADD COLUMN IF NOT EXISTS archived BOOLEAN NOT NULL DEFAULT FALSE;
+ALTER TABLE granules ADD COLUMN IF NOT EXISTS archived BOOLEAN NOT NULL DEFAULT FALSE;
```
npauzenga: executions

docs/upgrade-notes/archived_column_indexing.md
```diff
@@ -0,0 +1,93 @@
...
+5. Verify the updates
+
+     We can verify that the tables are updated successfully by checking the `\d tablename` results from psql, the indexes created should be listed.
+
+     If the concurrent index query fails for any reason, you may have an `invalid` index - if this occurs,
+     make sure to drop and create the index again to avoid resources being used for the invalid index.
+
+6. Close the session
+
+    Close the tmux session after the task is complete by `exit` or `Ctrl-b x`.
```
npauzenga, line 94: Were you able to walk through these? It might be good to do and copy/paste these commands to ensure there aren't errors carried over from the docs these were copied from.

packages/db/src/migrations/20250617190412_add_archived_and_index.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE granules ADD COLUMN IF NOT EXISTS archived BOOLEAN NOT NULL DEFAULT FALSE;
+ALTER TABLE granules ADD COLUMN IF NOT EXISTS archived BOOLEAN NOT NULL DEFAULT FALSE;
```
npauzenga: eep

packages/db/src/migrations/20250617190412_add_archived_and_index.ts
npauzenga: Yup, answered this question.

docs/features/record_archival.md
npauzenga: Yeah, I think you can just skip making this distinction. Users could update that field one way or the other as they like.

docs/features/record_archival.md
npauzenga: Yup, I'd remove this. You could add a line somewhere like "The dashboard supports this junk" and leave it there.

docs/features/record_archival.md
etcart: sure, I'm changing it

docs/features/record_archival.md
etcart: ok, pulling it

docs/features/record_archival.md
etcart: the proper "key takeaway" is to set your archive timeout to be a few months further back than you expect to be querying for debug purposes, but that doesn't really fit into this PR since that's a functionality to comein a later ticket

docs/upgrade-notes/archived_column_indexing.md
etcart: yup, good catch

docs/upgrade-notes/archived_column_indexing.md
etcart: thanks

docs/upgrade-notes/archived_column_indexing.md
etcart: I'll go trhough them today and make sure, it's a good point. they all look right, but a full pass is needed

CHANGELOG.md
```diff
@@ -36,6 +36,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
```
npauzenga, line 83: Should probably add a note with the title of the new docs page

docs/upgrade-notes/archived_column_indexing.md
```diff
@@ -0,0 +1,94 @@
+---
+id: add_and_index_archived_column
```
npauzenga, line 2: I think we need an entry for both of these new docs in sidebars.js

docs/upgrade-notes/archived_column_indexing.md
```diff
@@ -0,0 +1,94 @@
...
+    psql -h <Endpoint for writer instance> -p <Port for database or 5432> -d <cumulus database name> -U <database admin user> -f 20250617190412_add_archived_and_index.sql -W
+    ```
+
+    The following are SQL commands, and  20250617190412_add_archived_and_index.sql is available
+    [here](https://raw.githubusercontent.com/nasa/cumulus/master/packages/db/src/migrations/20250617190412_add_archived_and_index.sql):
+
+    ```sql
+    ALTER TABLE granules ADD COLUMN IF NOT EXISTS archived BOOLEAN NOT NULL DEFAULT FALSE;
+    ALTER TABLE executions ADD COLUMN IF NOT EXISTS archived BOOLEAN NOT NULL DEFAULT FALSE;
+    CREATE INDEX IF NOT EXISTS executions_archived_index ON executions (archived);
```
npauzenga: What is the difference between this and using "CONCURRENTLY" as we have in other migrations? e.g.
```
  await knex.raw('CREATE INDEX CONCURRENTLY IF NOT EXISTS executions_parent_cumulus_id_index ON executions(parent_cumulus_id)');

```

etcart: it does need ot be done concurrently, sorry good catch. it is done that way in the typescript migration and the sql file
Concurrently means don't block other operations while doing it

docs/upgrade-notes/archived_column_indexing.md
etcart: done

CHANGELOG.md
etcart: done

docs/upgrade-notes/archived_column_indexing.md
etcart: I have, at this point, hand tested all of these (and fixed some things as necessary)

PR #4010, opened 2025-07-08
Ecarton/exp ec2 cleanup

etcart:
**Summary:** adds an ec2 cleanup lambda

Addresses [CUMULUS-4108: rotate ec2 instances automatically](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4108)

## Changes

* Implements a lambda function to scan and terminate expired ec2 instances
* implements a terraform spec to automatically stand-up or update the associated lambda function

this is currently "disabled" (dry run flag set to true) pending approval

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


scripts/aws_cleanup/index.py
```diff
@@ -0,0 +1,131 @@
...
+        f'''never found timeout key for {
+            instance_object['InstanceId']
+        }'''
+    )
+    return False
+
+
+def get_instances_to_clean(
+    describe_func,
+    today_func: Callable[[], datetime] = datetime.today,
```
charleshuang80, line 74: I'm not sure if I'm missing something, but why does this method need the `today_func`, if it is also being set/defined in the other method that actually uses it?

scripts/aws_cleanup/main.tf
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  tags = {
+    Environment = "production"
+    Application = "sandbox"
+  }
+}
+
+resource "aws_cloudwatch_event_rule" "daily_ec2_cleanup" {
+  name                = "daily_ec2_cleanup"
```
charleshuang80, line 65: So this lambda is just for us, right? It won't impact our users because it's not in tf-modules?
Wondering if like a weekly cleanup is a possibility (with the rules we need to follow), and if so what that cost difference looks like. Do you know? Probably minor, but just curious.

scripts/aws_cleanup/index.py
```diff
@@ -0,0 +1,131 @@
...
+            )
+            if rotate_date < today_func():
+                return True
+            return False
+    logger.warning(
+        f'''never found timeout key for {
+            instance_object['InstanceId']
+        }'''
+    )
+    return False
```
charleshuang80, line 69: Is this the desired behavior? That if there's no Rotate By or timeout_key, it will just log and return False? So if that info is missing an instance could linger around forever?

scripts/aws_cleanup/test_index.py
```diff
@@ -0,0 +1,178 @@
...
+                ],
+            },
+            lambda: datetime(2022, 2, 1),
+        )
+        is False
+    )
+    os.environ = env
+
+
+def test_should_be_cleaned_up_false_on_tag_not_found():
```
charleshuang80: Is there a way to check for the log message for no tag?

scripts/aws_cleanup/index.py
etcart: yea, I think this this desired behavior.
I don't want to be wiping out ec2 instances randomly, and in particular they won't have a Rotate By key for the first few hours after they're created. I could be convinced to delete by default as well, but this would make ec2 instances that are hand-made and meant to stay around a non-starter

scripts/aws_cleanup/main.tf
etcart: yea, I imagine it will cost 1/7 the amount, but not much, this runs for well under a minute, and once a day (or once a week) with minimal resources. by my estimate it's about 25 cents to run the function once

scripts/aws_cleanup/index.py
etcart: this is purely for being able to test it,  need to be able to pass it an anonymous function that returns a static date so it doesn't break when the current date goes too high or whatever

scripts/aws_cleanup/test_index.py
etcart: yup, added

scripts/aws_cleanup/index.py
charleshuang80: ok, makes sense in that case.

scripts/aws_cleanup/index.py
charleshuang80: Yeah, I was wondering if there might be cases where we want one to stick around.

scripts/aws_cleanup/main.tf
charleshuang80: Did you have an answer for the first two questions?

etcart: sorry, yes, this is just for us, and won't be deployed as a part of any deployment since it's not in tf-modules

PR #4013, opened 2025-07-15
CUMULUS-4062: Add producerGranuleId to LZARDS metadata

Nnaga1:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/example/spec/parallel/lzardsBackup/LzardsBackupWorkflowSpec.js
+++ b/example/spec/parallel/lzardsBackup/LzardsBackupWorkflowSpec.js
@@ -157,7 +157,6 @@ describe('The Lzards Backup workflow ', () => {
         }
       }
       expect(lambdaOutput.payload.granules[0].granuleId).toEqual('FakeGranule1');
-      expect(lambdaOutput.payload.granules[0].producerGranuleId).toEqual('FakeGranule1');
     });
   });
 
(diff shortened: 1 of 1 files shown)


tasks/lzards-backup/src/index.ts
```diff
@@ -184,6 +186,7 @@ export const postRequestToLzards = async (params: {
             granuleId,
             provider,
             createdAt,
+            producerGranuleId,
```
jennyhliu, line 189: It's possible producerGranuleId is not set in granule object, in that case, it should be set to granuleId.
Please add unit test for the case producerGranuleId isn't set.

jennyhliu:
Add test case for LzardsBackup with uniquified granuleId.  Add LzardsBackupWithUniqueGranuleIdsSpec.js or adding test to existing -UniqueGranuleIds spec.

tasks/lzards-backup/src/index.ts
Nnaga1: I can try to, the ticket never mentioned adding tests so I didnt do that

Nnaga1: added unit test 👍

PR #4021, opened 2025-07-21
WIP: Feature/cumulus 4080

npauzenga:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/example/data/collections/L2_HR_PIXC-000-unique/L2_HR_PIXC-000.json
+++ b/example/data/collections/L2_HR_PIXC-000-unique/L2_HR_PIXC-000.json
@@ -6,7 +6,7 @@
   "granuleId": "^.*$",
   "granuleIdExtraction": "^(.*)(\\.h5|\\.cmr)",
   "reportToEms": false,
-  "url_path": "{cmrMetadata.Granule.Collection.ShortName}___{cmrMetadata.Granule.Collection.VersionId}/{substring(file.fileName, 0, 3)}",
+  "url_path": "{cmrMetadata.Granule.Collection.ShortName}___{cmrMetadata.Granule.Collection.VersionId}/{cmrMetadata.Granule.granuleId}",
   "sampleFileName": "L2_HR_PIXC_product_0001-of-4154.h5",
   "files": [
     {
--- a/example/data/collections/L2_HR_PIXC-000-unique/L2_HR_PIXC-000.json
+++ b/example/data/collections/L2_HR_PIXC-000-unique/L2_HR_PIXC-000.json
@@ -6,7 +6,7 @@
   "granuleId": "^.*$",
   "granuleIdExtraction": "^(.*)(\\.h5|\\.cmr)",
   "reportToEms": false,
-  "url_path": "{cmrMetadata.Granule.Collection.ShortName}___{cmrMetadata.Granule.Collection.VersionId}/{cmrMetadata.Granule.granuleId}",
+  "url_path": "{cmrMetadata.Granule.Collection.ShortName}___{cmrMetadata.Granule.Collection.VersionId}/{cmrMetadata.Granule.GranuleUR}",
   "sampleFileName": "L2_HR_PIXC_product_0001-of-4154.h5",
   "files": [
     {


docs/features/duplicate_granule_handling.md
```diff
@@ -0,0 +1,60 @@
+---
+id: duplicate_granule_handling
+title: Cumulus Duplicate Granule Handling
+hide_title: false
+---
+
+Collections and Workflows can be configured to assign a unique `granuleId` to each Granule. This is useful in cases where two Granules share the same `granuleId` but must both be ingested into the Cumulus system. Previously, a Granule with a conflict on the `granuleId` key would cause an ingest failure for that Granule.
```
jennyhliu: `a Granule with a conflict on the `granuleId` key would cause an ingest failure for that Granule.` This is true for granuleId across collections, but granules with identical granuleId within the same collection will fail only when duplicateHandling is error.

docs/features/duplicate_granule_handling.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Granule Uniquification
+
+Cumulus has implemented a set of changes that allow multiple Granules with identical `granuleIds` to be ingested and managed independently. This is done by replacing the `granuleId` with a unique value and storing the original `granuleId` value as a new field, `producerGranuleId`. This new `producerGranuleId` field can then be used to track and correlate the Granule to the original data holdings.
+
+### Granule Uniquification Algorithm and Task Component
+
+The process Cumulus Core uses to generate a unique `granuleId` can be [found here](https://github.com/nasa/cumulus/blob/feature/duplicate-granule-924/packages/ingest/src/granule.ts#L479-L487).
```
jennyhliu: The link will change, maybe just link to future master branch and use the package and method name.

docs/features/duplicate_granule_handling.md
```diff
@@ -0,0 +1,67 @@
...
+
+To make the process easier, Cumulus Core also provides a Task Component that will uniquely identify each Granule in a payload and return a modified Granule object containing a unique `granuleId` and a `producerGranuleId` containing the original `granuleId`.
+
+## Configuration
+
+### Collection Configuration
+
+Cumulus Collections can be configured to replace the Granule's existing `granuleId` with a unique value. The originnal `granuleId` will then be stored in a new field, `producerGranuleId`.
+
+To configure a Collection to support uniquely identifying Granules, the following params are used in the Collection's JSON:
```
jennyhliu, line 31: But the uniquification can be configured in the other places like rule.meta, or hardcode the value in workflow task configuraton.

docs/features/duplicate_granule_handling.md
```diff
@@ -0,0 +1,67 @@
...
+
+```json
+"meta": {
+    "uniquifyGranuleId": true,
+    "hashLength": 6
+  }
+```
+
+If `uniquifyGranuleId` is `true` and the Collection is ingested using a workflow that includes the `AddUniqueGranuleId` Task, the `granuleId` will be uniquely generated using the process [here](https://github.com/nasa/cumulus/blob/feature/duplicate-granule-924/packages/ingest/src/granule.ts#L479-L487). The `hashLength` specifies how many characters the randomized hash contains. More characters offer a greater chance of uniqueness.
```
jennyhliu: Same comment as above, the link will be invalid.

docs/features/duplicate_granule_handling.md
```diff
@@ -0,0 +1,67 @@
...
+or
+
+```json
+"url_path": "{cmrMetadata.Granule.Collection.ShortName}___{cmrMetadata.Granule.Collection.VersionId}/{cmrMetadata.Granule.GranuleUR}",
+```
+
+If that is NOT desirable and using the original, non-unique value is preferred, that is still possible. The Collection would need to be configured to use the `producerGranuleId`, which represents the original `granuleId` value without any uniquification, or a completely different value.
+
+```json
+"url_path": "{cmrMetadata.Granule.Collection.ShortName}___{cmrMetadata.Granule.Collection.VersionId}/{granule.producerGranuleId}/",
```
jennyhliu, line 62: Do we want to update moveGranule task to use granule.granuleId value if granule.producerGranuleId is undefined?

packages/ingest/src/url-path-template.js
```diff
@@ -57,8 +60,21 @@ function templateReplacer(context, submatch) {
     // args[0] is either an object path or a constant
     //   e.g. extractPath(file.path) or extractPath('/a/b/c')
     // assume args[0] is an object path if it starts with a key of context
-    const isObjectPath = Object.keys(context).includes(args[0].split('.')[0]);
+    let isObjectPath = Object.keys(context).includes(args[0].split('.')[0]);
     const jsonPathValue = get(context, args[0], isObjectPath ? undefined : args[0]);
+
+    // fall back to second argument if using the firstDefined function and
+    // the first argument doesn't produce a jsonPathValue
+    if (name === 'firstDefined' && !jsonPathValue && args[1]) {
```
npauzenga: There are several ways to do this but I'm going in this direction because I'd like the user to have to specify that they want this fallback behavior. They do that by using `firstDefined` in the `url_path`. I believe this is acceptable because, in order to arrive at this situation, they would have had to modify the Collection to use `producerGranuleId`. This gives them an option to do that but with a fallback.

An alternative could be doing it directly in move-granules but, if we do that, we're obfuscating what's happening a bit. I chose this route because it seems a little clearer what they're intending to happen.

 In our use case here, we're concerned about a Collection that's configured with a `producerGranuleId` in the url_path that may not exist for some reason (although it really should if they've configured correctly). In this case it will fall back to the second value provided in this new function.

npauzenga: This does make `templateReplacer` aware of one of the possible functions that can be passed in the template. We need this logic _at the point the string values are interpolated_ so it seemed like the cleanest but open to suggestions.

docs/features/duplicate_granule_handling.md
npauzenga: Yup, see comment below. New logic exists in move-granule in the ingest utility we're using to parse the url_path.

docs/features/duplicate_granule_handling.md
npauzenga: 👍 Updated that section: https://github.com/nasa/cumulus/pull/4021/commits/5d24744de181c7ebbf52becf19919ba948cd5533

docs/features/duplicate_granule_handling.md
npauzenga: Ah yes. I forgot to call this out. I updated the links to use master and I'll find a ticket to add an AC that double-checks these links at the end once the feature is merged.

npauzenga: https://github.com/nasa/cumulus/pull/4021/commits/5d24744de181c7ebbf52becf19919ba948cd5533

docs/features/duplicate_granule_handling.md
npauzenga: Added separate sections to cover the other configuration options: https://github.com/nasa/cumulus/pull/4021/commits/5d24744de181c7ebbf52becf19919ba948cd5533

docs/features/duplicate_granule_handling.md
npauzenga: Same as above

packages/ingest/src/url-path-template.js
```diff
@@ -33,6 +33,9 @@ function evaluateOperation(name, args) {
   case 'extractPath': {
     return path.dirname(valueStr);
   }
+  case 'firstDefined': {
```
jennyhliu: lodash has a method defaultTo(value, defaultValue), and defaultTo is probably a better name.

packages/ingest/src/url-path-template.js
```diff
@@ -57,8 +60,21 @@ function templateReplacer(context, submatch) {
     // args[0] is either an object path or a constant
```
jennyhliu: I think we can move the logic from line 60-79 to evaluateOperation method, and add  firstDefined specific code there. This seems to work.
```
function evaluateOperation(name, args, context) {
  // args[0] is either an object path or a constant
  //   e.g. extractPath(file.path) or extractPath('/a/b/c')
  // assume args[0] is an object path if it starts with a key of context
  console.log(evaluateOperation, name, args, context);
  const isObjectPath = Object.keys(context).includes(args[0].split('.')[0]);
  const jsonPathValue = get(context, args[0], isObjectPath ? undefined : args[0]);

  if (name !== 'firstDefined' && !jsonPathValue) throw new Error(`Could not resolve path ${args[0]}`);
  const valueStr = jsonPathValue;
  switch (name) {
  case 'extractYear': {
    return new Date(valueStr).getUTCFullYear();
  }
  case 'extractMonth': {
    return (new Date(valueStr).getUTCMonth() + 1).toString();
  }
  case 'extractDate': {
    return new Date(valueStr).getUTCDate().toString();
  }
  case 'extractHour': {
    return new Date(valueStr).getUTCHours().toString();
  }
  case 'dateFormat': {
    return moment.utc(valueStr).format(args[1]);
  }
  case 'substring': {
    return String.prototype.substring.apply(String(valueStr), args.slice(1));
  }
  case 'extractPath': {
    return path.dirname(valueStr);
  }
  case 'firstDefined': {
    const isObjectPath1 = Object.keys(context).includes(args[1].split('.')[0]);
    const jsonPathValue1 = get(context, args[1], isObjectPath1 ? undefined : args[1]);
    return jsonPathValue || jsonPathValue1;
  }
  default:
    throw new Error(`Could not support operation ${name}`);
  }
}
```

packages/ingest/src/url-path-template.js
jennyhliu: Need to add the function to https://github.com/nasa/cumulus/blob/master/docs/workflows/workflow-configuration-how-to.md?plain=1#L238

docs/features/duplicate_granule_handling.md
```diff
@@ -0,0 +1,76 @@
...
+
+Collections and Workflows can be configured to assign a unique `granuleId` to each Granule. This is useful in cases where two Granules share the same `granuleId` but must both be ingested into the Cumulus system. Previously, a Granule with a conflict on the `granuleId` key within the same Collection would cause either an ingest failure or overwrite the previous granule, depending on `duplicateHandling` configuration.
+
+## Granule Uniquification
+
+Cumulus has implemented a set of changes that allow multiple Granules with identical `granuleIds` to be ingested and managed independently. This is done by replacing the `granuleId` with a unique value and storing the original `granuleId` value as a new field, `producerGranuleId`. This new `producerGranuleId` field can then be used to track and correlate the Granule to the original data holdings.
+
+### Granule Uniquification Algorithm and Task Component
+
+The process Cumulus Core uses to generate a unique `granuleId` can be [found here](https://github.com/nasa/cumulus/blob/master/packages/ingest/src/granule.ts#L479-L487).
```
jennyhliu: Can we add the method name instead of line number?

docs/features/duplicate_granule_handling.md
```diff
@@ -0,0 +1,76 @@
...
+To configure a Collection to support uniquely identifying Granules, the following params are used in the Collection's JSON:
+
+```json
+"meta": {
+    "uniquifyGranuleId": true,
+    "hashLength": 6
+  }
+```
+
+If `uniquifyGranuleId` is `true` and the Collection is ingested using a workflow that includes the `AddUniqueGranuleId` Task, the `granuleId` will be uniquely generated using the process [here](https://github.com/nasa/cumulus/blob/master/packages/ingest/src/granule.ts#L479-L487). The `hashLength` specifies how many characters the randomized hash contains. More characters offer a greater chance of uniqueness.
```
jennyhliu: The AddUniqueGranuleId task configuration is necessary to use the configuration from collection

jennyhliu: Even if collection configuration doesn't have uniquifyGranuleId set to true, they may still want to update url_path if the granule is uniquified through other workflow configuration.

packages/ingest/src/url-path-template.js
npauzenga: Yup, makes sense. https://github.com/nasa/cumulus/pull/4021/commits/48fe4c4777eee5c438b3270a49de982b6d8f28dc

packages/ingest/src/url-path-template.js
npauzenga: Sure, thanks for thinking about this. That's definitely nicer than having `templateReplacer` handle this specific case.

https://github.com/nasa/cumulus/pull/4021/commits/76606cd63eec57f25e8bebdf8c6542b073405546

docs/features/duplicate_granule_handling.md
npauzenga: Is that possible in GH? I can create a "permalink" but that still references the line number, presumably attached to a specific version of the code:

https://github.com/nasa/cumulus/blob/87a653cb9a848aeb4eebcd463c4fd5ba2a61ae21/packages/ingest/src/granule.ts#L479

But that doesn't get us much. Do you know of a better way? Or do you mean adding the method name in text but including a link to line numbers?

docs/features/duplicate_granule_handling.md
npauzenga: Yup, fair enough: https://github.com/nasa/cumulus/pull/4021/commits/717ed80d749e7511bc26a99c4fc5955ecca667d3

docs/features/duplicate_granule_handling.md
jennyhliu: Add the method name in the text, remove the line number from the link.
We can link to the generateUniqueGranuleId document @Nnaga1 when we have one.

packages/ingest/src/url-path-template.js
```diff
@@ -3,14 +3,24 @@ const moment = require('moment');
...
+function evaluateOperation(name, args, context) {
+  // args[0] is either an object path or a constant
+  //   e.g. extractPath(file.path) or extractPath('/a/b/c')
+  // assume args[0] is an object path if it starts with a key of context
+  const isObjectPath = Object.keys(context).includes(args[0].split('.')[0]);
+  const jsonPathValue = get(context, args[0], isObjectPath ? undefined : args[0]);
+
+  if (name !== 'defaultTo' && !jsonPathValue) throw new Error(`Could not resolve path ${args[0]}`);
+
+  const valueStr = jsonPathValue;
```
jennyhliu: This can be removed by replacing jsonPathValue variable with valueStr

docs/features/duplicate_granule_handling.md
jennyhliu: Should we include example of AddUniqueGranuleId configuration?
https://github.com/nasa/cumulus/blob/feature/duplicate-granule-924/example/cumulus-tf/cnm_workflow.asl.json#L55-L88

docs/workflows/workflow-configuration-how-to.md
```diff
@@ -244,6 +244,7 @@ The following functions are implemented:
 * `dateFormat` - takes a second argument describing how to format the date, and
   passes the metadata date string and the format argument to
   [moment().format()](https://momentjs.com/docs/#/displaying/format/)
+* `defaultTo` - takes two arguments and returns the first defined (not `null` or `undefined`) value
```
jennyhliu, line 249: Can you help add `extractPath` and `substring`?
Also add "supports multiple and nested operations" and example
'{extractPath({substring(file.source, 6)})}'

docs/features/duplicate_granule_handling.md
npauzenga: Fair enough https://github.com/nasa/cumulus/pull/4021/commits/717ed80d749e7511bc26a99c4fc5955ecca667d3

docs/features/duplicate_granule_handling.md
npauzenga: Sure. I'm linking to that workflow rather than copy/paste: https://github.com/nasa/cumulus/pull/4021/commits/5505d9f32b1aff285e24f5a134cfad0e135f0cd5

packages/ingest/src/url-path-template.js
npauzenga: https://github.com/nasa/cumulus/pull/4021/commits/8dd3475ed672e623aa4d1231d282b8caacabf2c4

docs/workflows/workflow-configuration-how-to.md
npauzenga: Yup, no problem:

npauzenga: Oh, sorry: https://github.com/nasa/cumulus/pull/4021/commits/2d02653ebc4c4629d4545469fc8dc50adef5e967

CHANGELOG.md
```diff
@@ -27,6 +27,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     `producerGranuleId` identifier is set in updated CMR metadata
 
 ### Added
+- **CUMULUS-4080**
```
jennyhliu: Can you move the changelog down, order by ticket number?
Also add operation name 'defaultTo' to the entry.

npauzenga: Sure but we should let folks know if we want to make that a standard. https://github.com/nasa/cumulus/pull/4021/commits/39fd3cf6f5fffc2642e4c4b7f74d89ff34f82a39

docs/features/duplicate_granule_handling.md
```diff
@@ -0,0 +1,79 @@
...
+If that is NOT desirable and using the original, non-unique value is preferred, that is still possible. The Collection would need to be configured to use the `producerGranuleId`, which represents the original `granuleId` value without any uniquification, or a completely different value.
+
+```json
+"url_path": "{cmrMetadata.Granule.Collection.ShortName}___{cmrMetadata.Granule.Collection.VersionId}/{granule.producerGranuleId}/",
+```
+
+Taking it a step further, depending on the workflow configuration, users can specify a fallback if the `producerGranuleId` isn't available for a given Granule.
+
+```json
+"url_path": "{cmrMetadata.Granule.Collection.ShortName}___{cmrMetadata.Granule.Collection.VersionId}/{granule.producerGranuleId || granule.granuleId}",
```
jennyhliu: missing 'defaultTo' operation

docs/workflows/workflow-configuration-how-to.md
```diff
@@ -241,9 +241,18 @@ The following functions are implemented:
 * `extractMonth` - returns the month, formatted as MM
 * `extractDate` - returns the day of the month, formatted as DD
 * `extractHour` - returns the hour in 24-hour format, with no leading zero
+* `extractPath` - returns the path only, not including the file name e.g.`/data/test/abc.xml` returns `/data/test/`
```
jennyhliu: Should be `/data/test` without trailing /

docs/features/duplicate_granule_handling.md
npauzenga: https://github.com/nasa/cumulus/pull/4021/commits/e19cb0da3a82209ba4fa7f354bc3544f19b48907

docs/workflows/workflow-configuration-how-to.md
npauzenga: https://github.com/nasa/cumulus/pull/4021/commits/1bde9e4172e6ed93625e7ee9c6389aa303c2b933

jennyhliu (approved):
Great work as always 👍

PR #4025, opened 2025-07-28
CUMULUS-4069: Migration script and document for producer_granule_id

jennyhliu:

Addresses [CUMULUS-4069: Migration script and document for producer_granule_id](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4069)

## Changes

* Added migration script and instructions to add the producer_granule_id column
    to the granules table and populate it in the production environment.

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


docs/upgrade-notes/update-granules-to-include-producer_granule_id.md
```diff
@@ -0,0 +1,200 @@
...
+
+```sql
+SELECT granule_id, COUNT(*) AS count
+FROM granules
+GROUP BY granule_id
+HAVING COUNT(*) > 1
+ORDER BY count DESC;
+```
+
+The migration script will abort if there are duplicate grarnule_id values in the granules table.
```
Nnaga1: should we include any reconciliation steps for this part or no, would it be just to run this again pretty much?

docs/upgrade-notes/update-granules-to-include-producer_granule_id.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Apply the Changes in Production Environment
+
+For large databases (e.g., when the `granules` table contains more than 100,000 rows), updates must
+be applied manually, as the commands can take a significant amount of time. Since `ALTER TABLE`
+commands require an **exclusive lock** on the table, and populating the new column is time-consuming,
+it is recommended to **quiesce all database activity** during this process. This means shutting down
```
Nnaga1: i would change:
`shutting down` -> `pausing`

packages/db/src/migrations/20250425134823_granules_add_producer_granule_id.py
```diff
@@ -0,0 +1,194 @@
+import os
+import getpass
+from datetime import datetime
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
```
Nnaga1, line 5: whats the difference btwn this file and the sql file? for my own context, they do the same thing right? is this file supposed to be an alternative to the SQL one? (in case they want to use tmux or something)

jennyhliu: The python script updates producer_granule_id in batch with multiple workers, and significantly reduces the migration time, and there are other features like recovery.
20250425134823_granules_add_producer_granule_id.sql updates producer_granule_id in batch, but all the updates are still in the same transaction as defined by DO$$ PL/pgSQL, take very long time to complete with large database. This file is for reference and can be run on small database.

docs/upgrade-notes/update-granules-to-include-producer_granule_id.md
jennyhliu: It's up the DAAC to determine which granules to delete/keep.

Nnaga1: so kind of like implicitly saying `make sure there are no duplicate granule_ids before running` , ok 👍

packages/db/src/migrations/20250425134823_granules_add_producer_granule_id.py
Nnaga1: so
python one -> smaller dbs, more manual
sql one -> larger, one transaction, longer

kind of like providing options, thats pretty cool!

jennyhliu: Actually, SQL commands are for reference only. python script can be run manually with multiple workers. If it's a dev db, no need to run the migration, bootstrap lambda will automatically update the database as part of the data-persistence module deployment.

docs/upgrade-notes/update-granules-to-include-producer_granule_id.md
```diff
@@ -0,0 +1,201 @@
...
+
+    Once installed, a `tmux` session is started with two windows. Alternatively, you can open two
+    concurrent SSM sessions to the same EC2 instance and start a separate tmux session from each.
+
+    The primary window is used to run the migration script, while the secondary window is used
+    to monitor the database. When the operator's shift ends or monitoring is no longer needed,
+    the tmux session can be detached and reattached later as needed.
+
+4. Run Migration Script
+    The database login credentials can be retrieved from the prefix_db_login secret.
```
Nnaga1: `prefix_db_login secret` -> `<prefix>_db_login secret`, just for consistency

docs/upgrade-notes/update-granules-to-include-producer_granule_id.md
```diff
@@ -0,0 +1,201 @@
...
+    When the migration script is running, perform step 5 to monitor the commands.
+
+    ```sh
+    curl -o /home/ssm-user/20250425134823_granules_add_producer_granule_id.py https://raw.githubusercontent.com/nasa/cumulus/master/packages/db/src/migrations/20250425134823_granules_add_producer_granule_id.py
+
+    tmux new-session -s CumulusUpgrade -n add-producer_granule_id
+    python3 /home/ssm-user/20250425134823_granules_add_producer_granule_id.py
+    ```
+
+    The actual number of rows updated in each batch may be less than BATCH_SIZE because cumulus_id values may not increase by exactly 1.
```
Nnaga1: maybe for this line, you can add ```:::note``` ? bc it seems kind of important but just nit

packages/db/src/migrations/20250425134823_granules_add_producer_granule_id.py
Nnaga1: ok, but basically its another option in case the sql one isn't preferred or isnt in line with their use case (smaller db, need to do ops asap, etc.)

packages/db/src/migrations/20250425134823_granules_add_producer_granule_id.py
```diff
@@ -0,0 +1,194 @@
...
+                  IF NOT EXISTS (
+                    SELECT 1
+                    FROM information_schema.columns
+                    WHERE table_name = '{TABLE_NAME}' AND column_name = '{COLUMN_NAME}'
+                  ) THEN
+                    ALTER TABLE {TABLE_NAME} ADD COLUMN {COLUMN_NAME} TEXT;
+                    COMMENT ON COLUMN {TABLE_NAME}.{COLUMN_NAME} IS 'Producer Granule Id';
+                  END IF;
+                END$$;
+            """)
```
etcart, line 74: this code which allows user input, even command line input, should really avoid  direct character substitution for SQL injection reasons.
as per https://www.psycopg.org/docs/sql.html
this call should probably be something like:
```
from psycopg2 import sql
cur.execute(sql.SQL("""
   DO $$
   BEGIN
     IF NOT EXISTS (
       SELECT 1
       FROM information_schema.columns
       WHERE table_name = {table_name} AND column_name = {column_name}
     ) THEN
       ALTER TABLE {table_name} ADD COLUMN {column_name} TEXT;
       COMMENT ON COLUMN {table_name}.{column_name} IS 'Producer Granule Id';
     END IF;
   END$$;
   """.format(
       column_name=sql.Identifier(COLUMN_NAME),
       table_name=sql.Identifier(TABLE_NAME),
     )
   )
```

this should do some filtering on those values during execution to make Sure they are properly formatted as identifiers and not full of pernicious characters

packages/db/src/migrations/20250425134823_granules_add_producer_granule_id.py
```diff
@@ -0,0 +1,194 @@
...
+                cur.execute(f"""
+                    SELECT MIN(cumulus_id), MAX(cumulus_id)
+                    FROM {TABLE_NAME}
+                    WHERE {COLUMN_NAME} IS NULL;
+                """)
+            else:
+                cur.execute(f"""
+                    SELECT MIN(cumulus_id), MAX(cumulus_id)
+                    FROM {TABLE_NAME};
+                """)
```
etcart, line 92: same formatting best practice concerns here and just above

packages/db/src/migrations/20250425134823_granules_add_producer_granule_id.py
jennyhliu: Good advise.
COLUMN_NAME and TABLE_NAME are hardcoded constant, and are not from user input.
Should I replace them with actual values, maybe?

jennyhliu: Updated to use actual COLUMN_NAME and TABLE_NAME values.
[74354a8](https://github.com/nasa/cumulus/pull/4025/commits/74354a81bde103680cddee3fbf52d88a0fce4039)

etcart: yea, if those don't need to be flexible (and I don't think they do) then that also solves this concern

PR #4037, opened 2025-08-07
[CUMULUS-4119]: Integration Test Updates

npauzenga:
Update 8.19: 

I've added a new workflow execution to our integration tests. The purpose here is to cover the primary use case for IART-924 explicitly. We should be able to ingest a granule with the same `granuleId` as a previously ingested granule in a separate collection. The new granule should be ingested as normal, given a unique ID, and should share the same `producerGranuleId` as the previous "original" granule.

---

The original purpose of this ticket was to evaluate and add any missing integration tests for IART-924. After evaluating our test plan here:
-  https://wiki.earthdata.nasa.gov/display/CUMULUS/Duplicate+Granules%3A+Testable+Requirements

It looks to me like we don't need more Integration Tests for IART-924. The work is tested here:
-  https://github.com/nasa/cumulus/blob/feature/duplicate-granule-924/example/spec/parallel/cnmWorkflow/KinesisTestTriggerWithUniqueGranuleIdsSpec.js

That spec runs a workflow that contains the new task and runs an ingest. There is also an integration test here that checks the CMR metadata for producerGranuleId:  
- https://github.com/nasa/cumulus/pull/3997/files#diff-5decbba358493e90d4dd7ad765c77250e8288653bb23e36a4b5e8b7fe5c1f82fR657

I am claiming that we don't need tests for Lzards or Orca because of the way uniquification works. Once the granule has been given a unique granule_id and ingests properly, it acts like any other granule in the system. There is nothing in that granule that would cause it to act differently than any other. This may get more complicated when we discuss duplicate handling but in this epic, we're making the granules unique with the express purpose of handling them just like all the others without special logic.

I've added a new column to the testing plan wiki with notes on each assertion listed. https://wiki.earthdata.nasa.gov/display/CUMULUS/Duplicate+Granules%3A+Testable+Requirements

I'm open to feedback here as the claim is "do nothing". Our integration tests for this feature (so far) are testing from the Kinesis Trigger, which isn't perhaps an immediately obvious place for all of this but it also doesn't totally matter as the trigger isn't super relevant to the feature.


jennyhliu (commented):
What is the status of the tests here?
https://github.com/nasa/cumulus/pull/3981/files#diff-7cf08dcd03b23ecb59c1e30d23ddd41ad65ad3458b2eccfe44699c1e994f1dc9R254

What are the tests covering 1). same producerGranuleId same collection 2) same producerGranuleId cross collections?
Most of our tests(if not all) already uniquify granuleIds during test setup.

npauzenga:
> What is the status of the tests here? https://github.com/nasa/cumulus/pull/3981/files#diff-7cf08dcd03b23ecb59c1e30d23ddd41ad65ad3458b2eccfe44699c1e994f1dc9R254
> 
> What are the tests covering 1). same producerGranuleId same collection 2) same producerGranuleId cross collections? Most of our tests(if not all) already uniquify granuleIds (original producerGranuleId) during test setup.

The tests in the 924 feature branch _should_ be complete. I'm not sure I understand the question though. The tests we have thus far should cover uniquification of granuleId but no duplicate detection logic.

jennyhliu:
> > What is the status of the tests here? https://github.com/nasa/cumulus/pull/3981/files#diff-7cf08dcd03b23ecb59c1e30d23ddd41ad65ad3458b2eccfe44699c1e994f1dc9R254
I think I will enable the assertions in example/spec/parallel/ingestGranule/IngestGranuleWithDuplicateHandlingUniqueSpec.js in my PR

example/spec/parallel/cnmWorkflow/data/rules/kinesis/unique/L2_HR_PIXC_kinesisRule.json
```diff
@@ -0,0 +1,18 @@
+{
```
npauzenga: This is just moved to a new directory

example/spec/parallel/cnmWorkflow/KinesisTestTriggerWithUniqueGranuleIdsSpec.js
```diff
@@ -147,7 +154,7 @@ describe('The Cloud Notification Mechanism Kinesis workflow with Unique GranuleI
     // granuleId will be uniquified
     granuleId = record.product.name;
     producerGranuleId = record.product.name;
-    recordIdentifier = randomString();
+    recordIdentifier = `granule1_${randomString()}`;
```
npauzenga, line 187: Just for clarity

example/spec/parallel/cnmWorkflow/KinesisTestTriggerWithUniqueGranuleIdsSpec.js
```diff
@@ -456,6 +455,96 @@ describe('The Cloud Notification Mechanism Kinesis workflow with Unique GranuleI
...
+      });
+
+      // Get the uniquified granuleId
+      const uniqueTaskOutput = await lambdaStep.getStepOutput(workflowExecution2.executionArn, 'AddUniqueGranuleId');
+      const duplicateGranule = uniqueTaskOutput.payload.granules[0];
+      granuleId2 = duplicateGranule.granuleId;
+    });
+
+    afterAll(async () => {
+      await removePublishedGranule({
```
jennyhliu: Why do we delete granule1 in this block? Shouldn't it be deleted where it was created?
Also the next test case `describe('Workflow fails because SyncGranule fails'` ingests a failed granule with the same granuleId as granule 1.
We probably can clean the two granules as part of cleanUp() function.

example/spec/parallel/cnmWorkflow/KinesisTestTriggerWithUniqueGranuleIdsSpec.js
```diff
@@ -456,6 +455,96 @@ describe('The Cloud Notification Mechanism Kinesis workflow with Unique GranuleI
...
+        collectionId: constructCollectionId(ruleOverride.collection.name, ruleOverride.collection.version),
+      });
+      await removePublishedGranule({
+        prefix: testConfig.stackName,
+        granuleId: granuleId2,
+        collectionId: constructCollectionId(ruleOverride2.collection.name, ruleOverride2.collection.version),
+      });
+      console.log(`\nDeleting rule ${ruleOverride2.name}`);
+      const rules = await readJsonFilesFromDir(ruleDirectory2);
+      await deleteRules(testConfig.stackName, testConfig.bucket, rules, ruleSuffix);
```
jennyhliu: Where are the collection and execution cleaned up?

example/spec/parallel/cnmWorkflow/data/rules/kinesis/unique/L2_HR_PIXC_kinesisRule.json
jennyhliu: Why? This is used for KinesisTestErrorSpec.js and KinesisTestTriggerSpec.js as well.

example/spec/parallel/cnmWorkflow/KinesisTestTriggerWithUniqueGranuleIdsSpec.js
```diff
@@ -456,6 +455,96 @@ describe('The Cloud Notification Mechanism Kinesis workflow with Unique GranuleI
...
+    });
+
+    it('Executes successfully', () => {
+      expect(executionStatus2).toEqual('SUCCEEDED');
+    });
+
+    it('Makes the new granule with matching producerGranuleId available', async () => {
+      const originalGranule = await getGranule({
```
jennyhliu: I'm not sure if we want to call them originalGranule and duplicateGranule, they are two different granules from two collections.

example/spec/parallel/cnmWorkflow/KinesisTestTriggerWithUniqueGranuleIdsSpec.js
```diff
@@ -456,6 +455,96 @@ describe('The Cloud Notification Mechanism Kinesis workflow with Unique GranuleI
...
+
+      const duplicateGranule = await getGranule({
+        prefix: testConfig.stackName,
+        granuleId: granuleId2,
+        status: 'completed',
+        collectionId: constructCollectionId(ruleOverride2.collection.name, ruleOverride2.collection.version),
+      });
+
+      expect(originalGranule.producerGranuleId).toEqual(duplicateGranule.producerGranuleId);
+      expect(originalGranule.granuleId !== duplicateGranule.granuleId).toBeTruthy();
```
jennyhliu: Should this be `toBeTrue()`?

npauzenga: Yeah, makes more sense. https://github.com/nasa/cumulus/pull/4037/commits/aae96d04c3fe0aba0adc0fb7aa322f30c3f0b7dd

example/spec/parallel/cnmWorkflow/KinesisTestTriggerWithUniqueGranuleIdsSpec.js
npauzenga: Sure, we can do `granule1` and `granule2` and not mess with what makes this granule "original" vs. "duplicate". https://github.com/nasa/cumulus/pull/4037/commits/a1538eb21074264a5eca60943e120b14f3d9c9d2

example/spec/parallel/cnmWorkflow/KinesisTestTriggerWithUniqueGranuleIdsSpec.js
npauzenga: They are immortal.

Update: added collection and execution deletion to the cleanUp function: https://github.com/nasa/cumulus/pull/4037/commits/894d9f240a471c07653697205e340c29797b9a4b

I'll also move granule cleanup in another commit.

example/spec/parallel/cnmWorkflow/data/rules/kinesis/unique/L2_HR_PIXC_kinesisRule.json
npauzenga: Yeah, I did it this way because we're setting rule overrides. We can do that though so long as the json files are in separate directories, even if they're nested. I re-arranged to leave the original in place and create the new one in a new subdirectory. It works because `addRules` doesn't scan for files in subdirectories.

https://github.com/nasa/cumulus/pull/4037/commits/a5db113cc0c66f991acd33ec04e6470bc8507c69

There's an alternative here where we can update `addRules` to support multiple overrides for multiple files but that seems like overkill.

example/spec/parallel/cnmWorkflow/KinesisTestTriggerWithUniqueGranuleIdsSpec.js
npauzenga: Yeah, cleaned this up a bit. I moved the collection, granule, and execution cleanup to the `cleanUp` helper and added the third granule to be cleaned up https://github.com/nasa/cumulus/pull/4037/commits/a5db113cc0c66f991acd33ec04e6470bc8507c69

example/spec/parallel/cnmWorkflow/KinesisTestTriggerWithUniqueGranuleIdsSpec.js
```diff
@@ -147,7 +188,7 @@ describe('The Cloud Notification Mechanism Kinesis workflow with Unique GranuleI
     // granuleId will be uniquified
     granuleId = record.product.name;
```
jennyhliu: It's hard to tell which granuleId is which.
Sorry, I took back my another comment about originalGranuleId and duplicateGranuleId. granuleId1 and granuleId2 are confusing.
Can we rename some of these Ids and also add comment/context next to them?  e.g,
save original granuleId to originalGranuleId

Original Name | Suggested Name | Reason / Context
-- | -- | --
granuleId | uniqueGranuleId | First generated granuleId (from record)
granuleId2 | uniqueGranuleIdDuplicateDifferentCollection | Second generated granuleId (from record2: duplicate record different collection)
granuleId3 | uniqueGranuleIdErrorCase | Third granuleId, generated during failing execution (from record)
record2 | duplicateRecordDifferentCollection | record2 uses same producerGranuleId but from a different collection
recordIdentifier2 | duplicateRecordIdentifier | Identifier for the duplicate/second record
ruleOverride | primaryRuleOverride | For the original record (primary case)
ruleOverride2 | duplicateRuleOverride | For the second record (duplicate collection case)
workflowExecution | primaryWorkflowExecution | Matches first record
workflowExecution2 | duplicateWorkflowExecution | Matches second record
executionStatus2 | duplicateExecutionStatus | Matches duplicateWorkflowExecution, refers to status of second/duplicate execution
executionStatus | primaryExecutionStatus | for clarity and consistency with naming

executionStatus (in error case) → errorExecutionStatus

npauzenga: Thanks @jennyhliu. See what you think of this: https://github.com/nasa/cumulus/pull/4037/commits/334974631a137444584aa3be9ac7ac2915b73042

Apologies that the declarations at the top shifted quite a lot after alpha-sorting. Makes it tougher to read in the diff.

jennyhliu (approved):
Good work! Thanks for addressing my comments.

PR #4046, opened 2025-08-18
CUMULUS-4110: Add workflow_configurations to tf-modules to control which record types are written

jennyhliu:

Addresses [CUMULUS-4110: Allow users to configure workflows to NOT report granules on start](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4110)

## Changes

- Added the workflow_configurations variable to the tf-modules/ingest and
    tf-modules/cumulus modules.
    Its property sf_event_sqs_to_db_records_types has been added to 
    workflow_template.json under the cumulus_meta field to control which record
    types should be written to the database during different workflow execution statuses.
- Updated the SfSqsReport task to set meta.reportMessageSource in the Cumulus message.
- Updated the @cumulus/api/sfEventSqsToDbRecords Lambda to determine which
    record types should be written to the database based on the
    cumulus_meta.sf_event_sqs_to_db_records_types and meta.reportMessageSource fields.
    By default, all record types will be written to the database.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/tasks/sf-sqs-report/README.md
+++ b/tasks/sf-sqs-report/README.md
@@ -6,6 +6,8 @@ Note that the initial and final reporting of an execution/granule/PDR status is 
 
 If the task's input includes a `payload` key, the value of the key is returned as the output of the task, otherwise the output will be an empty object.
 
+The task sets `meta.reportMessageSource` in the Cumulus message, so lambdas such as `sfEventSqsToDbRecords` know where the Cumulus message is coming from.
+
 ## About Cumulus
 
 Cumulus is a cloud-based data ingest, archive, distribution and management prototype for NASA's future Earth science data streams.


example/spec/parallel/ingestGranule/IngestGranuleWithDuplicateHandlingUniqueSpec.js
```diff
@@ -54,7 +54,7 @@ function failOnSetupError(setupErrors) {
   }
 }
 
-describe('The Ingest Granules workflow with duplicates in the payload and uniquification enabled', () => {
+xdescribe('The Ingest Granules workflow with duplicates in the payload and uniquification enabled', () => {
```
jennyhliu, line 57: Disable due to error:
> An error was thrown in an afterAll
error properties: Object({ statusCode: 409, apiMessage: '{"statusCode":409,"error":"Conflict","message":"Cannot delete provider s3_provider_test-jl-rds-tf-IngestGranuleSuccess-1755953078611 with associated granules."}' })

In order to delete granule, we need uniquified granuleId, but the workflow execution fails and the test for checking execution status is xit.

CHANGELOG.md
```diff
@@ -145,6 +145,21 @@ Please follow the instructions before upgrading Cumulus
   - Add integration testing for duplicate granule workflows. This includes new
     specs and workflows in the `ingestGranule`, `discoverGranules`,
     `lzardsBackup`, `cnmWorkflow`, and `orca` specs.
+- **CUMULUS-4110**
+  - Added the `workflow_configurations` variable to the `tf-modules/ingest` and
+    `tf-modules/cumulus` modules.
+    Its property `sf_event_sqs_to_db_records_types` has been added to
```
Jkovarik:
```suggestion
    The property `sf_event_sqs_to_db_records_types` has been added to
```

CHANGELOG.md
```diff
@@ -145,6 +145,21 @@ Please follow the instructions before upgrading Cumulus
...
     specs and workflows in the `ingestGranule`, `discoverGranules`,
     `lzardsBackup`, `cnmWorkflow`, and `orca` specs.
+- **CUMULUS-4110**
+  - Added the `workflow_configurations` variable to the `tf-modules/ingest` and
+    `tf-modules/cumulus` modules.
+    Its property `sf_event_sqs_to_db_records_types` has been added to 
+    `workflow_template.json` under the `cumulus_meta` field to control which record
+    types should be written to the database during different workflow execution statuses.
+  - Updated the `SfSqsReport` task to set `meta.reportMessageSource` in the Cumulus message.
+  - Updated the `@cumulus/api/sfEventSqsToDbRecords` lambda to determine which
```
Jkovarik, line 155: Nit: we should probably be more specific here - there's assumed context that a reader knows this means PDRs, Granules and Executions.

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
```diff
@@ -41,11 +41,84 @@ const {
...
+
+/**
+@typedef {Object} RecordWriteFlags
+@property {boolean} shouldWriteExecutionRecords
+@property {boolean} shouldWriteGranuleRecords
+@property {boolean} shouldWritePdrRecords
+*/
+
+/**
+ * @typedef {Object} CumulusMessage
```
Jkovarik: This is already, more or less, defined in the `@cumulus/message` package - when possible we should be using common typings/interfaces for consistency between lambdas/packages.    I think we should consider updating this to import the @typedef from the types package rather than relying on a custom Message type.

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
```diff
@@ -41,11 +41,84 @@ const {
...
+/**
+ * Determines whether a record of the given type should be written to the database.
+ *
+ * @param {RecordType[] | undefined} configuredRecordTypes - An optional list of record types
+ *   that are allowed to be written. If undefined, all record types are allowed.
+ * @param {RecordType} recordType - The type of record to check.
+ * @returns {boolean} True if the record type should be written; otherwise, false.
+ */
+const isRecordTypeWritable = (configuredRecordTypes, recordType) =>
+  (configuredRecordTypes === undefined || configuredRecordTypes.includes(recordType));
```
Jkovarik: Super tiny nit: do we want to have an assumed behavior if the user specifies `null` in the message JSON?     Or, more accurately, should we be defensive here and assume everything should be written if we don't get RecordType[]?

The case here presumes someone is modifying `.meta`, but it's important to note that  `.meta` *is user modifiable space*

Jkovarik: Also - from a defensive standpoint, given this is a textual field, should we consider changing the configuration such that it's opt-out? 

Meaning we have this set to `configuredRecordTypes.includes(recordType)` - but that means if the array is`['branules', 'executions', 'pdrs'] `it won't write granules, right? 

(this comment is a draft, I need to check to see how this is getting to this code block, may remove if comment doesn't make sense)

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
```diff
@@ -95,31 +175,49 @@ const writeRecords = async ({
...
+    });
+  }
 
-  return writeGranulesFromMessage({
-    cumulusMessage,
-    executionCumulusId,
-    knex,
-    testOverrides,
-  });
+  if (shouldWriteGranuleRecords) {
```
Jkovarik, line 191: Does it make sense to *ever* not write an execution?    In what circumstances would this be a thing a user should do in configuration?    

If it does, we should consider if we're violating provanance requirements for a granule in allowing this, and /or the metrics implications if we have not done so already, and ensure this relationship is fully feature-documented (e.g. we need work explicitly scoped to document this change)

Jkovarik: Also `writeGranulesFromMessage` has this typed as a required string value - which makes the next call where it's defined as 

` * @param {number}            params.executionCumulusId - Execution ID the granule was written from`

even more awkward - the typing in writeGranules is not in good shape, clearly. 

Regardless if we follow that down the chain, the DB package (which also has it defined as a number, but an unrequired one) eventually has the following logic: 

```
  if (executionCumulusId) {
    await granulesExecutionsPgModel.upsert(knexTransaction, {
      granule_cumulus_id: pgGranule.cumulus_id,
      execution_cumulus_id: executionCumulusId,
    });
  }

```

which leads me to suggest:  "update typing in the write granules chain and make sure we're not creating a new typing/logic concern (we shouldn't be based on my read through)"

Jkovarik: Also - the above commentary was added without 2582dd738b54947ac07ee4419292fd738bfdd0f1 pulled in 😓, still reviewing.

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
```diff
@@ -95,31 +175,49 @@ const writeRecords = async ({
...
+    return writeGranulesFromMessage({
+      cumulusMessage,
+      executionCumulusId,
+      knex,
+      testOverrides,
+    });
+  }
+
+  if (shouldWriteExecutionRecords && !shouldWriteGranuleRecords && executionCumulusId) {
```
Jkovarik: If I'm reading this right, this upserts association with granules *if* the granules are in the database, meaning it's attempting to handle the case where there is a granule, but the workflow is configured to not write (and therefore it won't be updated), but we do want the execution association? 

Should this logic be  something like`if (executionCumulusId && !shouldWriteGranuleRecords)`?    Explicit > implicit for sure, but I'm not sure the config flag should be driving this so much as if we have an execution from writing to associate here.

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -843,6 +844,62 @@ const updateGranuleFromApi = async (granule, knex) => {
...
+ * @throws {Error} - Throws on unexpected database errors.
+ */
+const writeGranuleExecutionAssociationsFromMessage = async ({
+  cumulusMessage,
+  executionCumulusId,
+  knex,
+  granulePgModel = new GranulePgModel(),
+  granulesExecutionsPgModel = new GranulesExecutionsPgModel(),
+}) => {
+  if (!messageHasGranules(cumulusMessage)) {
```
Jkovarik, line 919: We should go with DRY here rather than duplicating code from the writeGranulesFromMessage code, generally.

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -843,6 +844,62 @@ const updateGranuleFromApi = async (granule, knex) => {
...
+    log.info('No granules found in message. Skipping granules-executions cross-reference write.');
+    return undefined;
+  }
+
+  const granules = getMessageGranules(cumulusMessage);
+  const granuleIds = granules.map((granule) => granule.granuleId);
```
Jkovarik, line 925: Typing: granules[] is unknown from this method, which makes granule unknown, and cascades into the rest of this method typing. 

See above comment re: DRY - that may help, or may make it worse.    Open to discussion on bounds here, but we should try to properly type annotate new code in api/lib.

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -843,6 +844,62 @@ const updateGranuleFromApi = async (granule, knex) => {
...
+
+  const results = await Promise.allSettled(
+    granuleCumulusIds.map((granuleCumulusId) =>
+      granulesExecutionsPgModel.upsert(knex, {
+        granule_cumulus_id: granuleCumulusId,
+        execution_cumulus_id: executionCumulusId,
+      }))
+  );
+
+  const failures = results.filter((result) => result.status === 'rejected');
```
Jkovarik: Same here re: DRY

packages/api/lib/writeRecords/write-granules.js
```diff
@@ -843,6 +844,62 @@ const updateGranuleFromApi = async (granule, knex) => {
...
+ *   in the workflow message.
+ * @param {Knex} params.knex - A Knex client instance for interacting with PostgreSQL.
+ * @param {GranulePgModel} [params.granulePgModel] - Optional override for the GranulePgModel.
+ * @param {GranulesExecutionsPgModel} [params.granulesExecutionsPgModel] - Optional
+ *   override for the GranulesExecutionsPgModel.
+ * @returns {Promise<object[]|undefined>} - Results from Promise.allSettled()` for
+ *   each granule, or `undefined` if no granules exist.
+ * @throws {Error} - Throws on unexpected database errors.
+ */
+const writeGranuleExecutionAssociationsFromMessage = async ({
```
Jkovarik, line 912: Nit: consider under/private annotating this if it's a helper and not a lib export (can see this either way)

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -467,6 +468,94 @@ test.serial('_writeGranule will not allow a running status to replace a complete
...
+      { granule_id: granuleId, collection_cumulus_id: collectionCumulusId }
+    )).status);
+    t.true(await t.context.granulesExecutionsPgModel.exists(
+      knex,
+      { execution_cumulus_id: executionCumulusId }
+    ));
+  });
+
+test.serial(
+  'writeGranuleExecutionAssociationsFromMessage() does not save granule-execution associations '
```
Jkovarik, line 529: Great test title, and answered a prior question I had.   :thumbsup:

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -467,6 +468,94 @@ test.serial('_writeGranule will not allow a running status to replace a complete
   );
 });
 
+test.serial('writeGranuleExecutionAssociationsFromMessage() saves granule-execution associations to PostgreSQL',
```
Jkovarik, line 472: Saying this entirely with understanding that it might not already have coverage in the other block, I think we should have coverage around 

```

  const failures = results.filter((result) => result.status === 'rejected');
  if (failures.length > 0) {
    const allFailures = failures.map((failure) => failure.reason);
    const aggregateError = new AggregateError(allFailures);
    log.error('Failed writing some granule-execution associations.', aggregateError);
    throw aggregateError;
  }
  ```

*particularly* if this is an exported lib method intended for import elsewhere.

File: packages/api/tests/lib/writeRecords/test-write-granules.js [hunk 14]
```diff
@@ -467,6 +468,94 @@ test.serial('_writeGranule will not allow a running status to replace a complete
...
+    )).status);
+
+    cumulusMessage.meta.status = 'completed';
+    await writeGranuleExecutionAssociationsFromMessage({
+      cumulusMessage,
+      executionCumulusId,
+      knex,
+    });
+
+    t.is('running', (await t.context.granulePgModel.get(
```
Jkovarik: Is the intent of this assertion to test that the granule hasn't been modified?    If so this *probably* should be a deep comparison of the OG granule rather than just the status field.

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -467,6 +468,94 @@ test.serial('_writeGranule will not allow a running status to replace a complete
...
+      knex,
+      { granule_id: granuleId, collection_cumulus_id: collectionCumulusId }
+    )).status);
+    t.true(await t.context.granulesExecutionsPgModel.exists(
+      knex,
+      { execution_cumulus_id: executionCumulusId }
```
Jkovarik: We should include the granuleId here as well, given the DB isn't reset per test.

packages/integration-tests/cmr.js
```diff
@@ -196,8 +198,10 @@ function generateCmrXml(granule, collection, additionalUrls) {
   };
 
   xmlObject.Granule.OnlineAccessURLs.forEach((url) => {
+    const replacementId = (matchFilesWithProducerGranuleId && granule.producerGranuleId)
```
Jkovarik, line 201: Nit: just a helper, but could this logic cause inadvertant badness we care about (dev accidentally sets this to true, but passes in granule that for bad test construction reasons doesn't have producerGranuleId), or is this intentional? 

If the latter, nit: docstring, if the former, consider if we should help ourselves out and throw if the flag is set and there's no `producerGranuleId`

tasks/sf-sqs-report/index.js
```diff
@@ -19,6 +19,7 @@ async function reportSQSMessage(event) {
     meta: {
       ...meta,
       status: 'running',
+      reportMessageSource: 'SfSqsReport',
```
Jkovarik, line 22: Nit: typing here might document why this message is structured like that.   Definitely not blocking given the scope/context of the ticket, however if the logic on this is 'lambda' or 'else' then this string winds up being meaningless and we might want to consider a boolean, or better yet documenting an annotated type for this key. 

Nit: More generally we should consider namespacing this, there's a minute chance that users might collide with the key.

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-index.js
```diff
@@ -493,3 +499,214 @@ test.serial('Lambda captures error type on error', async (t) => {
...
+      await t.context.granulePgModel.exists(
+        t.context.testKnex,
+        { granule_id: granuleId, collection_cumulus_id: collectionCumulusId }
+      )
+    );
+  }
+);
+
+test.serial(
+  'writeRecords() writes all records to PostgreSQL when meta.reportMessageSource is "lambda", '
```
Jkovarik: Just to be certain - what are the expected user states/configurations where this is set to lambda?

tf-modules/cumulus/variables.tf
```diff
@@ -612,4 +612,43 @@ variable "deploy_cumulus_workflows" {
...
+          sf_event_sqs_to_db_records_types = {
+            IngestAndPublishGranule = {
+              running = ["execution", "pdr"]
+            }
+          }
+        }
+  EOF
+
+  type = object({
+    sf_event_sqs_to_db_records_types = optional(map(map(list(string))), {})
```
Jkovarik, line 648: Nit/non-blocking this might be a good candidate for evaluating the inputs re: discussion in standup this morning.

example/spec/parallel/ingestGranule/IngestGranuleWithDuplicateHandlingUniqueSpec.js
Jkovarik: I feel like this is non-blocking for this ticket, but shouldn't we resolve the state of this test prior to merging the feature?  

@npauzenga?

example/spec/parallel/ingestGranule/IngestGranuleSuccessSpec.js
```diff
@@ -1419,7 +1419,7 @@ describe('The S3 Ingest Granules workflow', () => {
         expect(definition.Comment).toEqual('Ingest Granule');
 
         // definition has all the states' information
-        expect(Object.keys(definition.States).length).toBe(14);
+        expect(Object.keys(definition.States).length).toBe(15);
```
Jkovarik, line 1422: Just for posterity - this update was due to the modification to the template that included the message source?

example/data/collections/s3_MOD09GQ_006_full_ingest/s3_MOD09GQ_006.json
```diff
@@ -14,8 +14,7 @@
     "orca": {
       "excludedFileExtensions": [".met"]
     },
-    "s3MultipartChunksizeMb": 16,
-    "uniquifyGranuleId": false
```
Jkovarik, line 18: This change will result in roughly 8 tests (the ones that aren't overriding this value)  moving from non-unqified testing to uniqified granule testing - are we certain that's not creating testing gaps around full ingest behavior where we aren't uniqifying granules?

example/cumulus-tf/variables.tf
```diff
@@ -475,3 +475,42 @@ variable "dead_letter_recovery_memory" {
   default = 1024
   description = "The amount of memory in MB to reserve for the dead letter recovery Async Operation Fargate Task"
 }
+
+variable "workflow_configurations" {
+  description = <<EOF
```
Jkovarik, line 480: Nice docstring - we should go through and do this sort of thing for most of our configurations.

Jkovarik (changes requested):
This is really great, impactful work, really well done 🎉  

I have questions/comments we should discuss/work through, but this is looking really good!

example/data/collections/s3_MOD09GQ_006_full_ingest/s3_MOD09GQ_006.json
jennyhliu: No, `"uniquifyGranuleId": false` is the default behavior.

example/spec/parallel/ingestGranule/IngestGranuleSuccessSpec.js
jennyhliu: `GranuleStatusReport` step is added to IngestAndPublishGranule definition, so there are 15 states now

packages/api/tests/lambdas/sf-event-sqs-to-db-records/test-index.js
jennyhliu: Good catch. It has been changed to no longer look for a specific value.

example/spec/parallel/ingestGranule/IngestGranuleSuccessSpec.js
npauzenga: This is probably because the unique granule task was added to the workflow (and represents an additional state)? I had something similar in another test.

example/spec/parallel/ingestGranule/IngestGranuleWithDuplicateHandlingUniqueSpec.js
npauzenga: Yeah, I think skipping it here is fine. The intention with this test file was originally to cover IART-924 but Jenny's previous work made it mostly redundant for now. 

I made the claim earlier that it's worth hanging onto this file for the next work in IART-923. https://wiki.earthdata.nasa.gov/display/CUMULUS/Duplicate+Granules%3A+Testable+Requirements

tf-modules/cumulus/variables.tf
jennyhliu: [6f20d64](https://github.com/nasa/cumulus/pull/4046/commits/6f20d64271b7c92f1579f97bae839143b1c1841a)

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
jennyhliu: Yes, we still want the granule-execution association even if granule is not updated.
Updated if logic: [570c6e9](https://github.com/nasa/cumulus/pull/4046/commits/570c6e9410ee27e8c662a71082df17cab6686c24)

packages/api/tests/lib/writeRecords/test-write-granules.js
jennyhliu: Yes, [8c50851](https://github.com/nasa/cumulus/pull/4046/commits/8c50851c93060ff776b6aa886eabfcd151b8929a)

packages/api/tests/lib/writeRecords/test-write-granules.js
jennyhliu: [8c50851](https://github.com/nasa/cumulus/pull/4046/commits/8c50851c93060ff776b6aa886eabfcd151b8929a)

packages/api/tests/lib/writeRecords/test-write-granules.js
jennyhliu: [7935617](https://github.com/nasa/cumulus/pull/4046/commits/7935617d585c585ee6e9a119abc2b4172f7a30af)

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
jennyhliu: [d979710](https://github.com/nasa/cumulus/pull/4046/commits/d979710ab6ed7894b7fcf2d1ab535620d1b22337)

packages/api/lib/writeRecords/write-granules.js
Jkovarik: Just DRY, not typing :)

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
Jkovarik: Typing updates look good - had a couple of minor edit suggestions related to typing for this file I added to 253f5db4d88 /branch pushed for your perusal.

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
Jkovarik: @jennyhliu  I saw this was still open - do we still need to discuss?

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
Jkovarik: @jennyhliu  I still think this is a legit question, assuming it's not been already discussed at length.

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
Jkovarik: Great - and thanks for the comment, I realize I asked that prior to all of the commits yesterday.   Resolving.

packages/api/lib/writeRecords/write-granules.js
Jkovarik: Looked this up, agree implementation at L926 isn't really a dupe and has a custom log line. 

Downgrading to a nit, as that sort of abstraction isn't a terribly helpful mod

packages/api/lib/writeRecords/write-granules.js
Jkovarik: I think this could be DRY'd up to a helper, e.g. 47722b64879.     Willing to drop that line of thought though, suggestion only.

packages/api/lib/writeRecords/write-granules.js
Jkovarik: So I *think* overloading the method in the Granules src *should* do the trick, combined with a type guard/transform in the .js code: 

0ba5951c0b9

This is a suggestion only, and probably needs some harsh evaluation given it's adding process cycles to our granule write logic to ensure types.

packages/api/tests/lib/writeRecords/test-write-granules.js
```diff
@@ -556,6 +560,63 @@ test.serial(
   }
 );
 
+test.serial(
```
Jkovarik, line 563: Nice test!

tf-modules/cumulus/variables.tf
Jkovarik: That looks really good, I'm leaving this open only to remind me to manually test

example/data/collections/s3_MOD09GQ_006_full_ingest/s3_MOD09GQ_006.json
Jkovarik: Ah, I see, you're just removing an explicit default.    Got it.

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
Jkovarik: Commit brought in, resolving

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
Jkovarik: Discussed in huddle, this is being handled in ingest terraform module validation.       This lands squarely as a defensive nit that would could add further validation here.

packages/api/lambdas/sf-event-sqs-to-db-records/index.js
Jkovarik: - Concern about granule linkup addressed in future work
- Going to discuss concern about execution reporting in 8/27 arch meeting.   Can resolve as that discussion will hold up 4058.

packages/api/lib/writeRecords/write-granules.js
Jkovarik: from huddle/PR discussion: This was merged in from the suggestion branch ✅

packages/api/lib/writeRecords/write-granules.js
Jkovarik: Per huddle, included.

packages/api/lib/writeRecords/write-granules.js
Jkovarik: From huddle: this is being exported, just a bad comment 😅

packages/integration-tests/cmr.js
Jkovarik: Answer: Jonathan should learn to read ternary ops properly 😅

tasks/sf-sqs-report/index.js
Jkovarik: Huddle discussion: maybe add something to the task README?

PR #4052, opened 2025-08-28
CUMULUS-4206: sf_event_sqs_to_db_records_types record type list must include "execution" and "pdr"

jennyhliu:

Addresses [CUMULUS-4206: Update to 4110: sf_event_sqs_to_db_records_types record type list must include "execution" and "pdr"](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4206)

## Changes

* Updated terraform workflow_configurations variable validation

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


example/cumulus-tf/variables.tf
```diff
@@ -483,9 +483,11 @@ variable "workflow_configurations" {
 
     - `sf_event_sqs_to_db_records_types`: An optional nested map that controls which record types
       ("execution", "granule", "pdr") should be written to the database for each workflow and
-      workflow status("running", "completed", "failed").
+      workflow status ("running", "completed", "failed").
       This configuration is used by the `@cumulus/api/sfEventSqsToDbRecords` Lambda.
 
+      Currently, both "execution" and "pdr" must be written to the database, so the record type list must include both.
+
```
npauzenga, line 490: Below, on line 514 we're saying the default is `running = ["execution", "pdr"]` but shouldn't that be an empty object? Or `running = ["execution", "pdr", "granule"]`?

tf-modules/ingest/variables.tf
```diff
@@ -267,12 +269,20 @@ variable "workflow_configurations" {
           keys(var.workflow_configurations.sf_event_sqs_to_db_records_types[workflow])
         ) :
         contains(["running", "completed", "failed"], status) &&
+        alltrue([
```
npauzenga, line 272: What do you think about a test for this? e.g. `writeRecords() throws error and writes no records to PostgreSQL if 'execution' is not included`

CHANGELOG.md
```diff
@@ -154,6 +154,8 @@ Please follow the instructions before upgrading Cumulus
     The property `sf_event_sqs_to_db_records_types` has been added to
     `workflow_template.json` under the `cumulus_meta` field to control which record
     types should be written to the database during different workflow execution statuses.
+    Currently, both "execution" and "pdr" must be written to the database, so the
+    record type list must include both.
```
npauzenga, line 158: I know we talked about this and I think these changes are good but I do wonder if it makes more sense, now that we have required input, to make it a list of records/status that are skipped rather than written. e.g. `sf_event_sqs_to_db_skip_record_writes`.

I don't think it's a deal-breaker either way as we're throwing an appropriate error but it could potentially be more confusing for users this way if they have options they really can't modify.

npauzenga: If we think there's a case in the future to modify this logic to make it possible to remove other records, that makes more sense.

If not though, I do wonder about either making it a "skip" list or making the configuration option only applicable to our current use case,, e.g. `skip_running_granule_writes`. I could see value in having something more generic but I'm not sure if we're gonna use it? 

None of this should be a deal-breaker on the PR. I don't think it matters all that much.

npauzenga (changes requested):
This looks great @jennyhliu. I added a couple clarifying questions, nothing major.

example/cumulus-tf/variables.tf
jennyhliu: This the default value for the example deployment, so we don't specify the value for each deployment.  In the tf-modules the default value is
default = {
    sf_event_sqs_to_db_records_types = {}
  }

tf-modules/ingest/variables.tf
jennyhliu: The validation happens during deployment, `terraform apply` will throw error if recordType list doesn't include `execution`.
writeRecords function still supports the recordType list without execution or even empty list.

CHANGELOG.md
jennyhliu: > If we think there's a case in the future to modify this logic to make it possible to remove other records, that makes more sense.

I believe this is possible, and it’s the main reason we opted for an inclusive list from the start.

example/cumulus-tf/variables.tf
npauzenga: Don't we want the example deployment to be `sf_event_sqs_to_db_records_types = {}`? Sorry, I may just be misunderstanding.

tf-modules/ingest/variables.tf
npauzenga: Ah, of course. That makes sense. Not really testable logic.

CHANGELOG.md
npauzenga: Ok. I don't think we want to change the direction here anyway.

example/cumulus-tf/variables.tf
jennyhliu: We want example workflow IngestAndPublishGranule only writes ["execution", "pdr"] on workflow starts, so we have it as default for example, otherwise we need to set the variable with this value for each deployment.  The variable description indicates the default value of this variable.
https://github.com/nasa/cumulus/pull/4052/files#diff-923ccc457843eaf76dcda5594381ce2124f7cf184108342978c10da459c88aecL499-L502

npauzenga: Yup, fair enough. Thanks for explanation!

PR #4054, opened 2025-08-29
Doc updates for CUMULUS-4011

npauzenga:
Doc updates for CUMULUS-4011

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -174,6 +174,10 @@ Please follow the instructions before upgrading Cumulus
   - Added an optional `includeTimestampHashKey` parameter to the `generateUniqueGranuleId` function in the `@cumulus/ingest/granule`, with a default value of `false`.
   - Added an optional `includeTimestampHashKey` configuration to the `add-unique-granuleId` and `parse-pdr tasks`, also with a default value of `false`.
   - Added a documentation page titled `"Generate Unique GranuleId"` to explain the algorithm for generating unique `granuleIds`.
+- **CUMULUS-4028**
+  - Update AddUniqueGranuleId task to output the input payload in addition to the modified granules.
+  - Added 'unique' version of ingest_and_publish granule workflow for 'uniquiy' feature ingest tests
+
 - **CUMULUS-4209**
   - Updated the `producer_granule_id` migration script to disable autovacuum before the
     migration and re-enable it afterward to improve performance.
--- a/example/cumulus-tf/ingest_and_publish_granule_unique_workflow.asl.json
+++ b/example/cumulus-tf/ingest_and_publish_granule_unique_workflow.asl.json
@@ -1,7 +1,85 @@
 {
   "Comment": "Ingest Granule",
-  "StartAt": "SyncGranule",
+  "StartAt": "AddUniqueGranuleId",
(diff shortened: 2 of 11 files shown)


docs/features/record_write_options.md
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+}
+```
+
+:::caution
+There is a drawback to skipping the initial `granule` record write in PDR workflows. If the "running" granule is not written, there will be no link between the incoming `granule` and the incoming `pdr` until that `granule` is written. That linkage works via `granule.pdrName`, which requires the `granule` be written to the primary datastore. In this case it may be necessary to introduce an additional write step to store the `granule` _after_ the unique `granuleId` has been generated via the `add-unique-granuleID` task.
+
+[The sfs-sqs-report Task](https://github.com/nasa/cumulus/blob/master/tasks/sf-sqs-report/README.md) can be used for this purpose. If inserted after the `add-unique-granuleID` Task, the `sfs-sqs-report` Task will store the `granule` correctly with the uniquely-generated `granuleId` and `pdrName`.
+:::
```
npauzenga, line 43: @Jkovarik and @jennyhliu this block is meant to cover the PDR issue. I might need more discussion here though as I'm not sure if it's true that the `sfs-sqs-report` task will magically work and store the `pdrName` in this way.

docs/features/granule_uniquification.md
```diff
@@ -218,7 +219,46 @@ Taking it a step further, depending on the workflow configuration, users can spe
...
+        IngestAndPublishGranule = {
+            running = ["execution", "pdr"]
+        }
+    }
+}
+```
+
+`execution` and `pdr` records will be written in the `running` state. `granule` will not because it is not specified. As of this writing, `execution` and `pdr` must always be written. `granule` is currently the only record type that can be skipped in this way. This behavior is subject to change in the future, see below documentation for the most current Record Write Options.
+
+**Important**: If skipping the initial Granule record write is desired in a **PDR workflow**, additional modification may be required. See [the Record Write Options documentation](record_write_options.md) for details on this use case as well as general documentation on skipping record writes.
```
npauzenga, line 261: Added to cover PDR workflow concern.

docs/features/duplicate_granule_granule-id-hashing-approach.md
jennyhliu, line 1: Do we want to link this page to the sidebar?

docs/features/record_write_options.md
jennyhliu: When the payload has both granules and pdr, the granules written will have pdr_cumulus_id.

npauzenga: So does that mean additional logic is required to get the PDR from that ID and assign pdrName?

npauzenga: - Update task/lambda names for consistency.
- The PDR and granule is required to be in the payload when the report task is run (for this situation).
- Better describe the linkage issue

npauzenga: Addressed here: https://github.com/nasa/cumulus/pull/4054/commits/fd19317368a5ef681aa479a2b997d561842df1f3

PR #4061, opened 2025-09-08
CUMULUS-4200: Update Terraform files to clarify `metrics_es_host`

charleshuang80:

Addresses [CUMULUS-4200: Update Terraform files to clarify `metrics_es_host`](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4200)

## Changes

* Added description to `metrics_es_host` in variables files
* added validation rule for `metrics_es_host`

## PR Checklist

- [x] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/tf-modules/archive/variables.tf
+++ b/tf-modules/archive/variables.tf
@@ -2,12 +2,12 @@
 
 variable "async_operation_image" {
   description = "docker image to use for Cumulus async operations tasks"
-  type        = string
+  type = string
 }
 
 variable "background_queue_url" {
   description = "Queue URL to use for throttled background operations (e.g. granule reingest)"
-  type        = string
+  type = string
 }
 
 variable "cmr_client_id" {
@@ -63,12 +63,12 @@ variable "ecs_cluster_name" {
 
 variable "ecs_execution_role" {
   description = "Object containing name and ARN of IAM role for initializing ECS tasks"
-  type        = object({ name = string, arn = string})
+  type = object({ name = string, arn = string})
 }
 
 variable "ecs_task_role" {
   description = "Object containing name and ARN of IAM role for running ECS tasks"
-  type        = object({ name = string, arn = string})
+  type = object({ name = string, arn = string})
 }
 
 variable "kinesis_inbound_event_logger_lambda_function_arn" {
@@ -112,21 +112,21 @@ variable "token_secret" {
 }
 
 variable "urs_client_id" {
-  description = "The URS app ID"
   type        = string
+  description = "The URS app ID"
 }
 
 variable "urs_client_password" {
-  description = "The URS app password"
   type        = string
+  description = "The URS app password"
 }
 
 # Optional
 
 variable "api_gateway_stage" {
-  description = "The API Gateway stage to create"
   type        = string
   default     = "dev"
+  description = "The API Gateway stage to create"
 }
 
 variable "api_port" {
@@ -135,14 +135,14 @@ variable "api_port" {
 }
 
 variable "api_reserved_concurrency" {
-  type    = number
+  type = number
   default = 8
 }
 
 variable "api_url" {
-  description = "If not specified, the value of the API Gateway endpoint is used"
   type        = string
   default     = null
+  description = "If not specified, the value of the API Gateway endpoint is used"
 }
 
 variable "buckets" {
@@ -179,14 +179,14 @@ variable "cmr_search_client_config" {
 
 variable "lambda_memory_sizes" {
   description = "Configurable map of memory sizes for lambdas"
-  type        = map(number)
-  default     = {}
+  type = map(number)
+  default = {}
 }
 
 variable "lambda_timeouts" {
   description = "Configurable map of timeouts for lambdas"
-  type        = map(number)
-  default     = {}
+  type = map(number)
+  default = {}
 }
 
 variable "lambda_subnet_ids" {
@@ -227,12 +227,12 @@ variable "metrics_es_host" {
 }
 
 variable "metrics_es_password" {
-  type    = string
+  type = string
   default = null
 }
 
 variable "metrics_es_username" {
-  type    = string
+  type = string
   default = null
 }
 
@@ -259,8 +259,8 @@ variable "private_archive_api_gateway" {
 
 variable "rds_connection_timing_configuration" {
   description = "Cumulus rds connection timeout retry timing object -- these values map to knex.js's internal use of  https://github.com/vincit/tarn.js/ for connection acquisition"
-  type        = map(number)
-  default     = {
+  type = map(number)
+  default = {
       acquireTimeoutMillis: 90000
       createRetryIntervalMillis: 30000,
       createTimeoutMillis: 20000,
@@ -296,9 +296,9 @@ variable "tags" {
 }
 
 variable "urs_url" {
-  description = "The URL of the Earthdata Login site"
   type        = string
   default     = "https://uat.urs.earthdata.nasa.gov"
+  description = "The URL of the Earthdata Login site"
 }
 
 variable "users" {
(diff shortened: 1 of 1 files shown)


tf-modules/archive/variables.tf
```diff
@@ -210,17 +210,23 @@ variable "launchpad_passphrase" {
 }
 
 variable "metrics_es_host" {
-  type = string
-  default = null
+  description = "Domain name (not URL) of the Cloud Metrics API."
+  type        = string
+  default     = null
+
```
jennyhliu, line 216: I think this is the only place we need to add validation.
tf-modules/archive is where metrics_es_host is defined and used.  Any other places use archive module, the validation applies.

tf-modules/archive/variables.tf
```diff
@@ -210,17 +210,23 @@ variable "launchpad_passphrase" {
...
+  validation {
+    condition     = !startswith(var.metrics_es_host, "https://")
+    error_message = "metrics_es_host domain name should not include `https://`."
+  }
 }
```
jennyhliu, line 227: metrics_es_host can be null, and can't starts with 'http://'.  This is an improved version:
  
 ```
validation {
    condition = (
      var.metrics_es_host == null ||
      (
        !startswith(trimspace(var.metrics_es_host), "https://") &&
        !startswith(trimspace(var.metrics_es_host), "http://")
      )
    )
    error_message = "metrics_es_host should be a domain name only (e.g., 'metrics.example.com') and must not start with 'http://' or 'https://'."
  }
```

File: example/cumulus-tf/variables.tf [hunk 3]
```diff
@@ -325,81 +331,81 @@ variable "rds_admin_access_secret_arn" {
...
-  default = {
-    cleanExecutions = 400           # archive
-    DistributionApiEndpoints = 400  # cumulus_distribution
-    s3-credentials-endpoint  = 400  # distribution
-    HelloWorld = 400                # ingest
-    s3-replicator = 400             # s3-replicator
-    TeaCache = 400                  # tea-map-cache
+  type        = map(number)
+  default     = {
+    cleanExecutions          = 400 # archive
```
jennyhliu: Are these format updates done by IDE auto-formatting?  Updating these in a whole file reduces the traceability of the code changes, and are not necessary.

example/cumulus-tf/terraform.tfvars.example
```diff
@@ -113,8 +113,9 @@ saml_assertion_consumer_service = "<Cumulus API endpoint>/saml/auth, e.g. https:
 # key_name      = "MY-KEY"
 
 # Optional, uncomment if needed.
+# NOTE: metrics_es_host is the domain name, not the URL, so should not include `https://`
```
jennyhliu: After we add the validation code, this is not necessary.

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Notable Changes
+
+- **CUMULUS-4200**
```
jennyhliu, line 188: This isn't a significant change. Current users don't need to be aware of it or make any updates.

example/cumulus-tf/variables.tf
jennyhliu: Additionally, the formats of the variables are still inconsistent in this file.
I think we should revert these formatting changes in the PR.

CHANGELOG.md
charleshuang80: This is what was blocking all of ASF's work on bignbit for the entire last PI. In the conversations had with other DAACs, including NSIDC, it seems that they have not been able to do successful CloudMetrics queries with the bulk granules endpoint, so it seems that this maybe be an issue for other DAACs as well. For that reason, I think it's worth making a bit more noticeable so that they can check their terraform files.

example/cumulus-tf/variables.tf
charleshuang80: I made the changes to make things more consistent. The majority of the variables had the `description` at the top, above `type` and `default` but not all of them did. Many but not all of the variables had the `=` indented to align but not all. I think those things combined made the file quite messy and this change makes things more consistent. The traceability just means that for whatever variable is being investigated you just need to look back one more commit, I don't think that makes that big of a difference. IMHO, I think having the files be more consistent outweighs the additional effort of needing to look back one more commit.

example/cumulus-tf/variables.tf
```diff
@@ -275,8 +275,14 @@ variable "thin_egress_jwt_secret_name" {
 }
 
 variable "metrics_es_host" {
-  type    = string
-  default = null
+  description = "Domain name (not URL) of the Cloud Metrics API."
+  type        = string
+  default     = null
+
+  validation {
```
charleshuang80: from the comment below, you don't think the validation is needed in this file? this will read from tf-modules/archive module?

tf-modules/cumulus/variables.tf
```diff
@@ -369,8 +369,14 @@ variable "log_destination_arn" {
 }
 
 variable "metrics_es_host" {
-  type    = string
-  default = null
+  description = "Domain name (not URL) of the Cloud Metrics API."
+  type        = string
+  default     = null
+
+  validation {
```
charleshuang80: So based on the above comment, we don't need the validation here? I think it makes sense since `metrics_es_host` is specified in the `archive` module in `cumulus/archive.tf`. But we still need to declare the variable here? Sorry I don't quite get it, just trying to fully understand.

example/cumulus-tf/terraform.tfvars.example
charleshuang80: That's may be true, but I don't think it hurts to keep it in. Especially if folks were already confused by this.

jennyhliu: When in confusion, users should look at variable description instead

example/cumulus-tf/variables.tf
jennyhliu: But the formats are still not consistent in the file.  There is no document in terraform saying description should be at the top even in their examples.

example/cumulus-tf/variables.tf
jennyhliu: It's not needed here.

tf-modules/cumulus/variables.tf
jennyhliu: just like functions, variable has to be declare here, but not validation.

example/cumulus-tf/variables.tf
jennyhliu: I actually prefer the format `cleanExecutions = 400`. When you add another item to the list, if the name is longer than the existing ones, you don't need to align the format again and lose the traceability.

example/cumulus-tf/terraform.tfvars.example
charleshuang80: Well, we didn't have anything in the description before, so users had nothing else to go by. I can remove this since we are adding info in the description. But in general, I think that providing more information/explanation is usually better because otherwise you might end up with not enough (as we did in this case). If we're going by the same reasoning of taking out all unnecessary or duplicative comments, we should probably strip out many of the comments above, as the users should just refer to documentation elsewhere.

example/cumulus-tf/variables.tf
charleshuang80: Ok, so if we look at https://developer.hashicorp.com/terraform/language/block/variable#complete-configuration and https://developer.hashicorp.com/terraform/language/block/variable#complete-configuration, then it does look like they have `description` below `default`. But, they don't specify anything about order. As for the equal signs, they have everything aligned within the specific block/section. And then in the Examples https://developer.hashicorp.com/terraform/language/block/variable#complete-configuration, they follow the general order of arguments and align the equal signs. Except for the very last example where the equal signs are not aligned.
I know this is pretty low stakes and not that important, but IMO, I think it would be a good idea to make things consistent, and the general order they use and aligning the equal signs seems a good example to follow, unless you have another suggestion to follow. I also think if we did that with many of our other areas of code some parts would be easier to read. But, is your stance that we should just leave things be if nothing is wrong? And that we shouldn't try and improve the file and make things more consistent? So if people add things in an inconsistent manner, and it gets through peer review and into the codebase, we should not ever try and make things cleaner or consistent? I think this may be a question I'd like to ask the team about, so might bring it up during Arch meeting.

tf-modules/archive/variables.tf
charleshuang80: this was helpful, thank you.

charleshuang80:
@jennyhliu, my terraform/stack is acting funny, so I'm not able to verify that the validation works when I put in a value with https and then try and deploy. Would you be able to verify that the validation works, or have you done that already?

jennyhliu:
> @jennyhliu, my terraform/stack is acting funny, so I'm not able to verify that the validation works when I put in a value with https and then try and deploy. Would you be able to verify that the validation works, or have you done that already?

Yes, the validation added works well.

PR #4062, opened 2025-09-08
CUMULUS-4191: Update lambdas to filter rules based on provider in message

jennyhliu:

Addresses [CUMULUS-4191: Allow for the matching of Cumulus rules provider to the provider included in CNM message](https://bugs.earthdata.nasa.gov/browse/CUMULUS-4191)

## Changes

* Updated `messageConsumer` and `sqsMessageConsumer` Lambdas to apply rule filtering
   based on the provider from the record message.

## PR Checklist

- [x] Update CHANGELOG
- [x] Unit tests
- [x] Integration tests


example/spec/parallel/cnmWorkflow/LambdaEventSourceSpec.js
```diff
@@ -133,15 +133,23 @@ describe('When adding multiple rules that share a kinesis event stream', () => {
...
-        console.log(`Dropping record onto ${streamName}, recordIdentifier: ${recordIdentifier}.`);
-        await putRecordOnStream(streamName, record);
+        const swotRecord = {
+          provider: `PODAAC_SWOT${testSuffix}`,
+          collection: `L2_HR_PIXC${testSuffix}`,
+          bucket: 'random-bucket',
+          identifier: swotRecordIdentifier,
+        };
+
+        console.log(`Dropping record onto ${streamName}, recordIdentifier: ${swotRecordIdentifier}.`);
```
etcart: we probably need to log both here, either two separate logs for swot and then nisar, or `recordIdentifiers: ${swot}, ${hisar}`

etcart: otherwise I predict people being confused debugging something when it at first appears that nisar is never added to stream

example/spec/parallel/cnmWorkflow/LambdaEventSourceSpec.js
```diff
@@ -150,21 +158,26 @@ describe('When adding multiple rules that share a kinesis event stream', () => {
...
+          swotRecordIdentifier,
           workflowArn,
           maxWaitForSFExistSecs,
           1
         );
       });
     });
 
-    it('runs the HelloWorldWorkflow for L2_HR_PIXC and not MOD09GQ', async () => {
+    it('runs the HelloWorldWorkflow for rule and record with collection L2_HR_PIXC and provider PODAAC_SWOT', async () => {
```
etcart: I'd like this to be clearer although I admit I'm not sure how.
it('the workflow filters rules based on provider in message') <- clunky I know
it('runs a workflow, filtering the provider to PODAAC_SWOT based on ....')

etcart: I think part of my confusion comes from the fact that as I'm parsing the ticket we're supposed to be filtering 
the *rule* based on the provider. but as I'm parsing these tests we're filtering the *provider* based on the rule (we're disabling the rule, and then testing which provider ended up in the step Input). if I'm totally off base, then I apologize but also I think something needs to change here to make it clearer what's happening for future debugging

packages/api/lambdas/sqs-message-consumer.js
```diff
@@ -81,9 +81,13 @@ async function dispatch(queueUrl, message) {
   await archiveSqsMessageToS3(queueUrl, message);
 
   const eventObject = JSON.parse(message.Body);
-  const eventCollection = rulesHelpers.lookupCollectionInEvent(eventObject);
 
-  const rulesToSchedule = rulesHelpers.filterRulesbyCollection(rulesForQueue, eventCollection);
```
etcart, line 86: just kidding, this was a bad read

etcart: wait, no I'm sorry I'm totally off-base there, I see it now

packages/api/tests/lambdas/test-message-consumer.js
```diff
@@ -9,67 +9,72 @@ const { PublishCommand } = require('@aws-sdk/client-sns');
...
-test.afterEach.always(() => {
-  fetchEnabledRulesStub.reset();
-  snsMock.reset();
-  queueMessageStub.reset();
+test.afterEach.always((t) => {
+  t.context.sandbox.restore();
+  t.context.snsMock.reset();
 });
 
 test('handler processes records as expected', async (t) => {
```
etcart: can I request more detail than just "as expected"?

etcart: feel free to ignore, I realize this part isn't this pr, sorry

packages/api/tests/lambdas/test-message-consumer.js
etcart, line 1: I appreciate the general cleanup and improvement in this file

example/spec/parallel/cnmWorkflow/LambdaEventSourceSpec.js
jennyhliu: [205715c](https://github.com/nasa/cumulus/pull/4062/commits/205715cde7d774ad5360b29000abdb9d55d2fbde) Updated, along with a couple of other updates.

example/spec/parallel/cnmWorkflow/LambdaEventSourceSpec.js
jennyhliu: Updated to:
it('runs the workflow for the record-matching-enabled rule, which has the same collection L2_HR_PIXC and provider PODAAC_SWOT'

packages/api/tests/lambdas/test-message-consumer.js
jennyhliu: updated to:
test('handler correctly processes mixed record types and handles errors via fallback'

packages/api/tests/lambdas/test-message-consumer.js
jennyhliu: The current setup doesn't support multiple tests, so we need to isolate the context setup for each test.

example/spec/parallel/cnmWorkflow/LambdaEventSourceSpec.js
etcart: much clearer, thank you

PR #4066, opened 2025-09-12
CUMULUS-4146: Determine how we mark granule as 'hidden’ vs ‘active'

Nnaga1:

## Changes

* ...

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,20 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Notable Changes
 
+### Migration Notes
+
+- This release updates all core integration deployments to target [cumulus-message-adapter v1.5.0](https://github.com/nasa/cumulus-message-adapter/releases/tag/v1.5.0).  It is suggested that users update their deployment to utilize the updated CMA.  Updates are *not* required for compatibility in custom lambdas.
+
 ### Changed
+
+- **CUMULUS-4155**
+  - Update Cumulus integration tests to utilize:
+    - Cumulus Message Adapter: v2.0.5
+    - Cumulus Message Adapter-py: v2.4.0
+    - Cumulus Process: 1.6.0
+  - Update all Python dependencies to use boto >=1.40.29
+  - Update all Core integration lambdas to use Python 3.12
+  - Update external CNM lambdas to run on Java 21 in integration
 
 - **CUMULUS-4191**
   - Updated `messageConsumer` and `sqsMessageConsumer` Lambdas to apply rule filtering
--- a/bamboo/Dockerfile
+++ b/bamboo/Dockerfile
@@ -7,15 +7,15 @@ libreadline-dev libsqlite3-dev wget curl llvm libncurses5-dev libncursesw5-dev \
 xz-utils tk-dev libffi-dev liblzma-dev python3-pip
 # install python
 WORKDIR /opt
-ARG PYTHON_VERSION=3.10
-ARG PYTHON_VERSION_PATCH=9
+ARG PYTHON_VERSION=3.12
+ARG PYTHON_VERSION_PATCH=11
 RUN wget -c https://www.python.org/ftp/python/$PYTHON_VERSION.$PYTHON_VERSION_PATCH/Python-$PYTHON_VERSION.$PYTHON_VERSION_PATCH.tgz
 RUN tar -zxf Python-$PYTHON_VERSION.$PYTHON_VERSION_PATCH.tgz
 WORKDIR "/opt/Python-$PYTHON_VERSION.$PYTHON_VERSION_PATCH"
 RUN ./configure --enable-optimizations && make -j4 && make altinstall
 
 WORKDIR /
-RUN update-alternatives --install /usr/bin/python python /usr/local/bin/python3.10 1 && \
+RUN update-alternatives --install /usr/bin/python python /usr/local/bin/python3.12 1 && \
     update-alternatives --install /usr/bin/pip pip /usr/bin/pip3 1
 RUN python -m pip install -U pip
 RUN npm config --global set script-shell /bin/bash
--- a/bamboo/bootstrap-tf-deployment.sh
+++ b/bamboo/bootstrap-tf-deployment.sh
@@ -34,11 +34,11 @@ echo "terraform {
 
 if [[ $NGAP_ENV = "SIT" ]]; then
   BASE_VAR_FILE="sit.tfvars"
-  CMA_LAYER_VERSION=18
+  CMA_LAYER_VERSION=19
   ROLE_BOUNDARY=NGAPShRoleBoundary
 else
   BASE_VAR_FILE="sandbox.tfvars"
-  CMA_LAYER_VERSION=21
+  CMA_LAYER_VERSION=22
   ROLE_BOUNDARY=NGAPShNonProdRoleBoundary
 fi
 
--- a/example/cumulus-tf/cnm_response_task.tf
+++ b/example/cumulus-tf/cnm_response_task.tf
@@ -26,7 +26,7 @@ resource "aws_lambda_function" "cnm_response_task" {
   s3_key           = aws_s3_bucket_object.cnm_response_lambda_zip.id
   handler          = "gov.nasa.cumulus.CNMResponse::handleRequestStreams"
   role             = module.cumulus.lambda_processing_role_arn
-  runtime          = "java17"
+  runtime          = "java21"
   timeout          = 300
   memory_size      = 256
   source_code_hash = aws_s3_bucket_object.cnm_response_lambda_zip.etag
--- a/example/cumulus-tf/cnm_to_cma_task.tf
+++ b/example/cumulus-tf/cnm_to_cma_task.tf
@@ -27,7 +27,7 @@ resource "aws_lambda_function" "cnm_to_cma_task" {
   s3_key           = aws_s3_bucket_object.cnm_to_cma_lambda_zip.id
   handler          = "gov.nasa.cumulus.CnmToGranuleHandler::handleRequestStreams"
   role             = module.cumulus.lambda_processing_role_arn
-  runtime          = "java17"
+  runtime          = "java21"
   timeout          = 300
   memory_size      = 128
   source_code_hash = aws_s3_bucket_object.cnm_to_cma_lambda_zip.etag
--- a/example/cumulus-tf/python-reference-task.tf
+++ b/example/cumulus-tf/python-reference-task.tf
@@ -4,7 +4,7 @@ resource "aws_lambda_function" "python_reference_task" {
   source_code_hash = filebase64sha256("${path.module}/../lambdas/python-reference-task/dist/lambda.zip")
   handler          = "initial_task.handler"
   role             = module.cumulus.lambda_processing_role_arn
-  runtime          = "python3.10"
+  runtime          = "python3.12"
   timeout          = 300
   memory_size      = 1556
 
--- a/example/cumulus-tf/variables.tf
+++ b/example/cumulus-tf/variables.tf
@@ -333,7 +333,7 @@ variable "async_operation_image_version" {
 variable "cumulus_process_activity_version" {
     description = "docker image version to use for python processing service"
     type = string
-    default = "4"
+    default = "5"
 }
 
 variable "ecs_task_image_version" {
@@ -345,7 +345,7 @@ variable "ecs_task_image_version" {
 variable "cumulus_test_ingest_image_version" {
     description = "docker image version to use for python test ingest processing service"
     type = string
-    default = "17"
+    default = "18"
 }
 variable "ecs_custom_sg_ids" {
   description = "User defined security groups to add to the Core ECS cluster"
--- a/example/lambdas/python-processing/Dockerfile
+++ b/example/lambdas/python-processing/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.10-buster
+FROM python:3.12.11-slim-bookworm
 ADD . /source
 WORKDIR /source
 RUN pip install pipenv && pipenv install --system --deploy --ignore-pipfile
--- a/example/lambdas/python-processing/Pipfile
+++ b/example/lambdas/python-processing/Pipfile
@@ -4,13 +4,13 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-boto3 = "~=1.26.90"
-cumulus-process = "==1.5.0"
-cumulus-message-adapter-python = "==2.3.0"
-six = "==1.12.0"
+boto3 = "~=1.40.29"
+cumulus-process = "==1.6.0"
+cumulus-message-adapter-python = "==2.4.0"
+six = ">=1.17.0"
 
 [dev-packages]
 pylint = ">=1.8.2"
 
 [requires]
-python_version = "3.10"
+python_version = "3.12"
--- a/example/lambdas/python-processing/Pipfile.lock
+++ b/example/lambdas/python-processing/Pipfile.lock
@@ -1,11 +1,11 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "3f1d8b234eb3b287ea41b8d1ff3d8bc62df94db6210a5dfb5c87fe188eb90819"
+            "sha256": "10c9b9fa80ff58338a5562bdc192878df78d9f422b316cf42e9dc21d7934f32f"
         },
         "pipfile-spec": 6,
         "requires": {
-            "python_version": "3.10"
+            "python_version": "3.12"
         },
         "sources": [
             {
@@ -26,41 +26,41 @@
         },
         "boto3": {
             "hashes": [
-                "sha256:9e7242b9059d937f34264125fecd844cb5e01acce6be093f6c44869fdf7c6e30",
-                "sha256:fa85b67147c8dc99b6e7c699fc086103f958f9677db934f70659e6e6a72a818c"
+                "sha256:61d5f9975c54ff919a24ff9d472c6c09c8a443a083fe56d30c04fc22d22ac42b"
             ],
             "index": "pypi",
-            "markers": "python_version >= '3.7'",
-            "version": "==1.26.165"
+            "markers": "python_version >= '3.9'",
+            "version": "==1.40.31"
         },
         "botocore": {
             "hashes": [
-                "sha256:6f35d59e230095aed7cd747604fe248fa384bebb7d09549077892f936a8ca3df",
-                "sha256:988b948be685006b43c4bbd8f5c0cb93e77c66deb70561994e0c5b31b5a67210"
+                "sha256:4033a00f8c6a4b5c3acb30af9283963123917227a437e5fc165189d07bd3cf8a",
+                "sha256:9496b91bbe40ed01d341dc8f6ff0492d7f546e80f1a862b00ec5bc9045fa3324"
             ],
-            "markers": "python_version >= '3.7'",
-            "version": "==1.29.165"
+            "markers": "python_version >= '3.9'",
+            "version": "==1.40.31"
         },
         "cumulus-message-adapter": {
             "hashes": [
-                "sha256:fe98bae50b87c0682c46e855e746680bb69b9439e1ee86b3129e6ea2d0af6ff7"
+                "sha256:815b957fde9a1cd35c6b0a8f1c71ed7181d73d608b7222c462e3d6f2bbdb0232"
             ],
-            "markers": "python_version ~= '3.10'",
-            "version": "==2.0.4"
+            "markers": "python_version ~= '3.12'",
+            "version": "==2.0.5"
         },
         "cumulus-message-adapter-python": {
             "hashes": [
-                "sha256:ff55768fae327db5aad1239b2186685f4085a2d6a6adb5ed29cda32f79b43abc"
+                "sha256:d53f8d5d01582ef5960a3c24846de5ea354abf1273f01002825c99f6b050e707"
             ],
             "index": "pypi",
-            "version": "==2.3.0"
+            "markers": "python_version ~= '3.12' and python_version < '4.0'",
+            "version": "==2.4.0"
         },
         "cumulus-process": {
             "hashes": [
-                "sha256:d1889b1befccd574116b0399bdc7afccbfb776597e3b9820b90606ed748cc9be"
+                "sha256:13de9282bbeaa7797c7a0cd97537cd2564098d7284efefcc5ee73cc1a5a122de"
             ],
             "index": "pypi",
-            "version": "==1.5.0"
+            "version": "==1.6.0"
         },
         "dicttoxml": {
             "hashes": [
@@ -144,7 +144,7 @@
                 "sha256:37dd54208da7e1cd875388217d5e00ebd4179249f90fb72437e91a35459a0ad3",
                 "sha256:a8b2bc7bffae282281c8140a97d3aa9c14da0b136dfe83f850eea9a5f7470427"
             ],
-            "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'",
+            "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2'",
             "version": "==2.9.0.post0"
         },
         "python-json-logger": {
@@ -156,28 +156,28 @@
         },
         "s3transfer": {
             "hashes": [
-                "sha256:b014be3a8a2aab98cfe1abc7229cc5a9a0cf05eb9c1f2b86b230fd8df3f78084",
-                "sha256:cab66d3380cca3e70939ef2255d01cd8aece6a4907a9528740f668c4b0611861"
+                "sha256:ea3b790c7077558ed1f02a3072fb3cb992bbbd253392f4b6e9e8976941c7d456",
+                "sha256:eff12264e7c8b4985074ccce27a3b38a485bb7f7422cc8046fee9be4983e4125"
             ],
-            "markers": "python_version >= '3.7'",
-            "version": "==0.6.2"
+            "markers": "python_version >= '3.9'",
+            "version": "==0.14.0"
         },
         "six": {
             "hashes": [
-                "sha256:3350809f0555b11f552448330d0b52d5f24c91a322ea4a15ef22629740f3761c",
-                "sha256:d16a0141ec1a18405cd4ce8b4613101da75da0e9a7aec5bdd4fa804d0e0eba73"
+                "sha256:4721f391ed90541fddacab5acf947aa0d3dc7d27b2e1e8eda2be8970586c3274",
+                "sha256:ff70335d468e7eb6ec65b95b99d3a2836546063f63acc5171de367e834932a81"
             ],
             "index": "pypi",
-            "markers": "python_version >= '2.6' and python_version not in '3.0, 3.1, 3.2, 3.3'",
-            "version": "==1.12.0"
+            "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2'",
+            "version": "==1.17.0"
         },
         "urllib3": {
             "hashes": [
-                "sha256:0ed14ccfbf1c30a9072c7ca157e4319b70d65f623e91e7b32fadb2853431016e",
-                "sha256:40c2dc0c681e47eb8f90e7e27bf6ff7df2e677421fd46756da1161c39ca70d32"
+                "sha256:3fc47733c7e419d4bc3f6b3dc2b4f890bb743906a30d56ba4a5bfa4bbff92760",
+                "sha256:e6b01673c0fa6a13e374b50871808eb3bf7046c4b125b216f6bf1cc604cff0dc"
             ],
-            "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3, 3.4, 3.5'",
-            "version": "==1.26.20"
+            "markers": "python_version >= '3.9'",
+            "version": "==2.5.0"
         }
     },
     "develop": {
@@ -194,7 +194,7 @@
                 "sha256:0633f1d2df477324f53a895b02c901fb961bdbf65a17122586ea7019292cbcf0",
                 "sha256:44f54bf6412c2c8464c14e8243eb163690a9800dbe2c367330883b19c7561049"
             ],
-            "markers": "python_version < '3.11'",
+            "markers": "python_version >= '3.8'",
             "version": "==0.4.0"
         },
         "isort": {
@@ -215,11 +215,11 @@
         },
         "platformdirs": {
             "hashes": [
-                "sha256:3d512d96e16bcb959a814c9f348431070822a6496326a4be0911c40b5a74c2bc",
-                "sha256:ff7059bb7eb1179e2685604f4aaf157cfd9535242bd23742eadc3c13542139b4"
+                "sha256:abd01743f24e5287cd7a5db3752faf1a2d65353f38ec26d98e25a6db65958c85",
+                "sha256:ca753cf4d81dc309bc67b0ea38fd15dc97bc30ce419a7f58d13eb3bf14c4febf"
             ],
             "markers": "python_version >= '3.9'",
-            "version": "==4.3.8"
+            "version": "==4.4.0"
         },
         "pylint": {
             "hashes": [
@@ -230,44 +230,6 @@
             "markers": "python_full_version >= '3.9.0'",
             "version": "==3.3.8"
         },
-        "tomli": {
-            "hashes": [
-                "sha256:023aa114dd824ade0100497eb2318602af309e5a55595f76b626d6d9f3b7b0a6",
-                "sha256:02abe224de6ae62c19f090f68da4e27b10af2b93213d36cf44e6e1c5abd19fdd",
-                "sha256:286f0ca2ffeeb5b9bd4fcc8d6c330534323ec51b2f52da063b11c502da16f30c",
-                "sha256:2d0f2fdd22b02c6d81637a3c95f8cd77f995846af7414c5c4b8d0545afa1bc4b",
-                "sha256:33580bccab0338d00994d7f16f4c4ec25b776af3ffaac1ed74e0b3fc95e885a8",
-                "sha256:400e720fe168c0f8521520190686ef8ef033fb19fc493da09779e592861b78c6",
-                "sha256:40741994320b232529c802f8bc86da4e1aa9f413db394617b9a256ae0f9a7f77",
-                "sha256:465af0e0875402f1d226519c9904f37254b3045fc5084697cefb9bdde1ff99ff",
-                "sha256:4a8f6e44de52d5e6c657c9fe83b562f5f4256d8ebbfe4ff922c495620a7f6cea",
-                "sha256:4e340144ad7ae1533cb897d406382b4b6fede8890a03738ff1683af800d54192",
-                "sha256:678e4fa69e4575eb77d103de3df8a895e1591b48e740211bd1067378c69e8249",
-                "sha256:6972ca9c9cc9f0acaa56a8ca1ff51e7af152a9f87fb64623e31d5c83700080ee",
-                "sha256:7fc04e92e1d624a4a63c76474610238576942d6b8950a2d7f908a340494e67e4",
-                "sha256:889f80ef92701b9dbb224e49ec87c645ce5df3fa2cc548664eb8a25e03127a98",
-                "sha256:8d57ca8095a641b8237d5b079147646153d22552f1c637fd3ba7f4b0b29167a8",
-                "sha256:8dd28b3e155b80f4d54beb40a441d366adcfe740969820caf156c019fb5c7ec4",
-                "sha256:9316dc65bed1684c9a98ee68759ceaed29d229e985297003e494aa825ebb0281",
-                "sha256:a198f10c4d1b1375d7687bc25294306e551bf1abfa4eace6650070a5c1ae2744",
-                "sha256:a38aa0308e754b0e3c67e344754dff64999ff9b513e691d0e786265c93583c69",
-                "sha256:a92ef1a44547e894e2a17d24e7557a5e85a9e1d0048b0b5e7541f76c5032cb13",
-                "sha256:ac065718db92ca818f8d6141b5f66369833d4a80a9d74435a268c52bdfa73140",
-                "sha256:b82ebccc8c8a36f2094e969560a1b836758481f3dc360ce9a3277c65f374285e",
-                "sha256:c954d2250168d28797dd4e3ac5cf812a406cd5a92674ee4c8f123c889786aa8e",
-                "sha256:cb55c73c5f4408779d0cf3eef9f762b9c9f147a77de7b258bef0a5628adc85cc",
-                "sha256:cd45e1dc79c835ce60f7404ec8119f2eb06d38b1deba146f07ced3bbc44505ff",
-                "sha256:d3f5614314d758649ab2ab3a62d4f2004c825922f9e370b29416484086b264ec",
-                "sha256:d920f33822747519673ee656a4b6ac33e382eca9d331c87770faa3eef562aeb2",
-                "sha256:db2b95f9de79181805df90bedc5a5ab4c165e6ec3fe99f970d0e302f384ad222",
-                "sha256:e59e304978767a54663af13c07b3d1af22ddee3bb2fb0618ca1593e4f593a106",
-                "sha256:e85e99945e688e32d5a35c1ff38ed0b3f41f43fad8df0bdf79f72b2ba7bc5272",
-                "sha256:ece47d672db52ac607a3d9599a9d48dcb2f2f735c6c2d1f34130085bb12b112a",
-                "sha256:f4039b9cbc3048b2416cc57ab3bda989a6fcf9b36cf8937f01a6e731b64f80d7"
-            ],
-            "markers": "python_version < '3.11'",
-            "version": "==2.2.1"
-        },
         "tomlkit": {
             "hashes": [
                 "sha256:430cf247ee57df2b94ee3fbe588e71d362a941ebb545dec29b53961d61add2a1",
@@ -275,14 +237,6 @@
             ],
             "markers": "python_version >= '3.8'",
             "version": "==0.13.3"
-        },
-        "typing-extensions": {
-            "hashes": [
-                "sha256:38b39f4aeeab64884ce9f74c94263ef78f3c22467c8724005483154c26648d36",
-                "sha256:d1e1e3b58374dc93031d6eda2420a48ea44a36c2b4766a4fdeb3710755731d76"
-            ],
-            "markers": "python_version < '3.11'",
-            "version": "==4.14.1"
         }
     }
 }
--- a/example/lambdas/python-processing/README.md
+++ b/example/lambdas/python-processing/README.md
@@ -11,20 +11,17 @@ This task is intended to be simple - the processing activity takes the input of 
 
 ### Requirements
 
-To develop against this task, you should be using python > 3.6 (CMA compatibility is baselined at 3.6).    Once you have a python env enabled:
-
-```bash
-pip install -r requirements.txt
-```
+To develop against this task, you should be using python > 3.12, preferably using pipenv.   See package.json for build setup and/or refer to [pipenv documentation](https://pipenv.pypa.io/en/latest/) for more on this.
 
 ### Build
 
 To update the container for a PR, you should run:
 
 ```bash
-docker build -t cumuluss/test-python-ingest-processing:{VERSION} .
-docker push cumuluss/test-python-ingest-processing:{VERSION}
+docker build --platform linux/amd64,linux/arm64 -t cumulus-test-ingest-process:{VERSION} .
 ```
+
+Then push to the configured ECR following the AWS console instructions for pushing to ECR for use in your build.
 
 Then update the `python_processing_service` resource in [`python_reference_workflow`](https://github.com/nasa/cumulus/blob/master/example/cumulus-tf/python_reference_workflow.tf) to utilize the correct image reference.
 
@@ -32,4 +29,4 @@ Then update the `python_processing_service` resource in [`python_reference_workf
 
 ## Updates
 
-Updates should generally consist of updates to the included `requirements.txt`, as the purpose of this task is to ensure compatibility with updates to the [`cumulus-message-adapter-python`](https://github.com/nasa/cumulus-message-adapter-python) client library via [`cumulus-process-py`] dependencies.
+Updates should generally consist of updates to the included `Pipfile`, as the purpose of this task is to ensure compatibility with updates to the [`cumulus-message-adapter-python`](https://github.com/nasa/cumulus-message-adapter-python) client library via [`cumulus-process-py`] dependencies.
--- a/example/lambdas/python-reference-activity/Dockerfile
+++ b/example/lambdas/python-reference-activity/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.10-buster
+FROM python:3.12.11-slim-bookworm
 ADD . /source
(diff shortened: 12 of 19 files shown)


packages/db/src/migrations/20250912712638_create_granule_duplicates_table.ts
```diff
@@ -0,0 +1,24 @@
...
+      .integer('granule_cumulus_id')
+      .references('cumulus_id')
+      .inTable('granules')
+      .notNullable();
+    table
+      .specificType('status', 'char(1)')
+      .comment('Granule active status')
+      .notNullable();
+    table
+      .string('group_id')
```
Nnaga1: I think, if 1 active per set is expected, I could change this to being `active_granule_cumulus_id` pointing to the `cumulus_id` of this table (like parent/child)

CHANGELOG.md
```diff
@@ -6,6 +6,21 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
+## Handle Granules with identical file names within a collection
+
+### Migration Notes
+
+#### CUMULUS-4146 Determine how a granule is marked 'hidden’ vs ‘active'
+
+Please follow the instructions before upgrading Cumulus
+
+- The changes in CUMULUS-4146 require updates to the PostgreSQL database
+  in the production environment. Please follow the instructions in []
```
Nnaga1: same comment as last PR, would this be neccessary?

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
```diff
@@ -0,0 +1,31 @@
+import { Knex } from 'knex';
+
+export const up = async (knex: Knex): Promise<void> => {
+  await knex.schema.createTable('granule_duplicates', (table) => {
+    table
+      .increments('cumulus_id')
+      .primary();
+    table
```
Nnaga1: should I add other variations of this (sql, py...)?

packages/db/src/models/granule-duplicates.ts
```diff
@@ -0,0 +1,97 @@
...
+  constructor() {
+    this.tableName = TableNames.granuleDuplicates;
+  }
+
+  async create(
+    knexTransaction: Knex | Knex.Transaction,
+    item: PostgresGranuleDuplicate
+  ) {
+    return await knexTransaction(this.tableName).insert(item);
+  }
```
mason-t-yates: How would you feel about doing some error handling here? Something like whats below

```suggestion
 if (!item) {
    throw new Error('Item is required for create operation');
  }
  
  if (!item.granule_cumulus_id || !item.group_id) {
    throw new Error('granule_cumulus_id and group_id are required');
  }
  
  return await knexTransaction(this.tableName).insert(item);
}
```

CHANGELOG.md
```diff
@@ -6,6 +6,21 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+## Handle Granules with identical file names within a collection
```
npauzenga: Could we add the Cumulus ticket number for this epic?

CHANGELOG.md
npauzenga: It'd be necessary if the migration is large enough to warrant it. Do they need special instructions for this or can the migration be run (against a large database) without disrupting ingest/normal operation?

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
npauzenga: Same question as above. We generally add `sql` files to our migrations dir when we intend users to run those commands manually. Do we want that here? Creating a table without populating anything seems like it wouldn't take very long. Is this safe to run in a production environment where tying up resources would be a problem?

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
```diff
@@ -0,0 +1,31 @@
...
+      .comment('Granule active status')
+      .notNullable();
+    table
+      .string('group_id')
+      .comment('Granule duplicate group id')
+      .notNullable();
+  });
+
+  await knex.raw('CREATE INDEX IF NOT EXISTS granule_duplicates_granule_cumulus_id_index ON granule_duplicates(granule_cumulus_id)');
+  await knex.raw('CREATE INDEX IF NOT EXISTS granule_duplicates_group_id_index ON granule_duplicates(group_id)');
```
npauzenga: What's the thinking around indexing both of these fields?

packages/db/src/models/granule-duplicates.ts
```diff
@@ -0,0 +1,97 @@
...
+  async upsert(
+    knexTransaction: Knex | Knex.Transaction,
+    item: PostgresGranuleDuplicate
+  ) {
+    return await knexTransaction(this.tableName)
+      .insert(item)
+      .onConflict(['granule_cumulus_id', 'group_id'])
+      .merge()
+      .returning('*');
+  }
```
mason-t-yates: Similar to my comment above
```suggestion
} catch (error: any) {
    throw new Error(`Failed to upsert granule duplicate`);
  }
```

packages/db/src/models/granule-duplicates.ts
```diff
@@ -0,0 +1,97 @@
...
+    knexOrTransaction: Knex | Knex.Transaction,
+    items: PostgresGranuleDuplicate[],
+    returningFields: string | string[] = '*'
+  ): Promise<PostgresGranuleDuplicate[]> {
+    return await knexOrTransaction(this.tableName)
+      .insert(items)
+      .returning(returningFields);
+  }
+
+  async delete(
```
mason-t-yates: I'm a TS noob, but this would delete just one at a time right? Do we need to enable bulk deletes?

CHANGELOG.md
```diff
@@ -6,6 +6,21 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
+#### CUMULUS-4146 Determine how a granule is marked 'hidden’ vs ‘active'
+
+Please follow the instructions before upgrading Cumulus
+
+- The changes in CUMULUS-4146 require updates to the PostgreSQL database
+  in the production environment. Please follow the instructions in []
+
+### Added
+
+- **CUMULUS-4146**
```
npauzenga: I think we need additional documentation here. Is there a ticket for it or was that intended to be addressed here?

I was thinking we would come out of this ticket with an ERD or similar showing how this table is intended to operate. The entire feature will obviously need docs but for this, I think it's sufficient to have an ADR unless you can think of anything more user-facing for just schema additions like this.

npauzenga: Realizing I'm throwing two things out there. An ADR with a description of the table and it's intention would probably be sufficient here.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
```diff
@@ -0,0 +1,31 @@
...
+      .inTable('granules')
+      .notNullable();
+    table
+      .specificType('status', 'char(1)')
+      .comment('Granule active status')
+      .notNullable();
+    table
+      .string('group_id')
+      .comment('Granule duplicate group id')
+      .notNullable();
```
npauzenga: So each granule _involved in a duplicate scenario_ would have an entry in this table? e.g. if there are two granules with different granule_cumulus_ids, both would have records here with the same group_id? And `status` defines which of them is active?

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
```diff
@@ -0,0 +1,31 @@
...
+      .increments('cumulus_id')
+      .primary();
+    table
+      .integer('granule_cumulus_id')
+      .references('cumulus_id')
+      .inTable('granules')
+      .notNullable();
+    table
+      .specificType('status', 'char(1)')
+      .comment('Granule active status')
```
npauzenga: What would you think about an enum here? We only have two states as far as I can tell.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
```diff
@@ -0,0 +1,31 @@
+import { Knex } from 'knex';
+
+export const up = async (knex: Knex): Promise<void> => {
+  await knex.schema.createTable('granule_duplicates', (table) => {
+    table
```
npauzenga: How is this table used? Will we need to fetch a `granuleDuplicates` record every time we get a granule to see if that granule is the active one? Or is there a function that would query the table and mark impacted granules active/hidden based on the query from this table?

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
```diff
@@ -0,0 +1,31 @@
...
+      .integer('granule_cumulus_id')
+      .references('cumulus_id')
+      .inTable('granules')
+      .notNullable();
+    table
+      .specificType('status', 'char(1)')
+      .comment('Granule active status')
+      .notNullable();
+    table
+      .string('group_id')
```
npauzenga: Why string?

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
npauzenga: I'm a little concerned... though I don't have great reasoning... about the granule's active status being stored in a table separate from `granules`. Our existing granule fetches wouldn't know if the granule retrieved is active or hidden without this separate query, right? Are we adding an API call to every/most/some granule request?

I think I need a short write-up (ADR?) about how and when this table is intended to be queried.

packages/db/src/models/granule-duplicates.ts
```diff
@@ -0,0 +1,97 @@
...
+import { isRecordDefined } from '../database';
+import { TableNames } from '../tables';
+
+import { PostgresGranuleDuplicate } from '../types/granule-duplicate';
+
+export default class GranuleDuplicatesPgModel {
+  readonly tableName: TableNames;
+
+  // can't extend base class because type for this data doesn't contain
+  // an updated_at property
```
npauzenga: Why _don't_ we include the timestamp fields on the new table?

packages/db/src/models/granule-duplicates.ts
```diff
@@ -0,0 +1,97 @@
...
+  constructor() {
+    this.tableName = TableNames.granuleDuplicates;
+  }
+
+  async create(
+    knexTransaction: Knex | Knex.Transaction,
+    item: PostgresGranuleDuplicate
+  ) {
+    return await knexTransaction(this.tableName).insert(item);
```
npauzenga: Curious if we can include the fields above an extend the base model using `super` calls but, in either case, I think we want to explicitly have it return something? e.g. 
```
super.create(knexOrTransaction, item, '*');
```
Not sure if knex does that automatically or not.

packages/db/src/models/granule-duplicates.ts
```diff
@@ -0,0 +1,97 @@
+import { Knex } from 'knex';
```
npauzenga: I think we need model tests for this. Is this a part of the ticket you're working?

packages/db/src/models/granule-duplicates.ts
```diff
@@ -0,0 +1,97 @@
...
+      .returning(returningFields);
+  }
+
+  async delete(
+    knexTransaction: Knex.Transaction,
+    params: Partial<PostgresGranuleDuplicate>
+  ): Promise<number> {
+    return await knexTransaction(this.tableName)
+      .where(params)
+      .del();
```
npauzenga: This is probably what we want but, just to be obnoxious, what happens if a user deletes a record for the "active" granule?

packages/db/src/models/granule-duplicates.ts
```diff
@@ -0,0 +1,97 @@
...
+  search(
+    knexTransaction: Knex | Knex.Transaction,
+    query: Partial<PostgresGranuleDuplicate>
+  ) {
+    return knexTransaction<PostgresGranuleDuplicate>(this.tableName)
+      .where(query);
+  }
+  async count(
+    knexOrTransaction: Knex | Knex.Transaction,
+    params: ([string, string, string] | [Partial<PostgresGranuleDuplicate>])[]
```
npauzenga: What are these params? Need a docstring.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
npauzenga: Adding the indices might be the catch here

packages/db/src/models/granule-duplicates.ts
npauzenga: The model doesn't necessarily need validation, even if there's a case for it (not sure). The endpoint would probably handle some of the more complex tasks in re-assigning active status, etc.?

packages/db/src/migrations/20250912712638_create_granule_duplicates_table.ts
npauzenga: Was this resolved? Did we get confirmation it's only 1 active?

Nnaga1: The one DAAC that responded to my post asking said they _expected_ 1 active and 1 or more hiddens

CHANGELOG.md
Nnaga1: since its just creating a new empty table and 2 indexes for that new table, I dont expect it to be long-running or performance-heavy at all, I guess I can try on the LP one and see how long it takes maybe?

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: `Do we want that here? ` I guess it couldn't hurt? I dont think this would really take that long or be a load on resources, possibly the index but since its on an empty table? not sure :/ I can try it out on the LP one and see 👀

CHANGELOG.md
Nnaga1: `An ADR with a description of the table and it's intention would probably be sufficient here.` I can make this 👍

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: `if there are two granules with different granule_cumulus_ids, both would have records here with the same group_id? And status defines which of them is active?` yes this exactly, only duplicate granules would be in this table, to abstract out from the granules' table + deal with the multiple active scenario (if that ends up being the case)

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: the other options we had weren't sufficient for the different cases which ended up with this ^

flag: if producerGranuleIds are different, promotion/demotion doesn't work

parent/child in granules table + flag: doesn't work with multiple actives

this is the rainfall solution that supports all of those ^ 

I guess it can be changed if we really dont agree, but i think we kinda exhausted the conversation there and it lead to here ^ :/

Nnaga1: I can make the ADR tho 👀

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: I guess it could be a number, to save storage/space, I can change that 👍

packages/db/src/models/granule-duplicates.ts
Nnaga1: wasn't sure whether it was neccessary but it doesnt seem like it would hurt to add it in, I guess I can change it to it can adapt BasePgModel

packages/db/src/models/granule-duplicates.ts
Nnaga1: I can try to make it so it adopts BasePgModel  👀

packages/db/src/models/granule-duplicates.ts
Nnaga1: I wasn't sure what tests to add, since im introducing this new thing I can add tests for the PgModel 👍

packages/db/src/models/granule-duplicates.ts
Nnaga1: `if a user deletes a record for the "active" granule?` if its deleted in the main granules table I would hope there would be some functionality to also delete it here, if its deleted in this table just randomly then... I'm not a 100% sure, I would assume thered be some guard to ensure only hiddens can be deleted (if that can be implemented?) I can look into that

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
```diff
@@ -0,0 +1,31 @@
+import { Knex } from 'knex';
```
Jkovarik: Note: if we're proposing to merge this, we should update the DB arch diagram - it's a really easy image/command to run, reach out if you'd like to pair.

CHANGELOG.md
npauzenga: Yeah, I think this'll probably be a non-issue but, once we have an understanding of how this table will be used, it may be worth trying the migration against the LP snapshot to ensure it doesn't have any unexpected performance issues. Open to team discussion though. If everyone is confident it's not an issue 🤷 

Migrations are tricky territory and we can put DAACs in a tough position if we're not careful. Just making sure we're doing our diligence.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
npauzenga: Yeah, I don't see a need to include separate SQL files if we can recommend DAACs run the migration in prod without issue.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
npauzenga: Gotchya. So we definitely need a clear answer on whether we need to support multiple 'active' granules as that opens some debate about structure.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
npauzenga: That's fair. Thanks. If we spell out _how/when_ we're using this table, that should alleviate some concern.

CHANGELOG.md
Jkovarik: I.... don't think there's a migration of data, just a table creation with a FK - I'm 95% confident a new table creation shouldn't be performance bound at all, and this shouldn't create any locks or anything like that on the old table.

packages/db/src/models/granule-duplicates.ts
npauzenga: Unless there's a reason not to, having timestamp fields seems like good practice. I'm guessing this won't be a massive table where a few extra fields would be a problem but let me know.

CHANGELOG.md
Jkovarik: Suggestion: Intention and *documented use cases/reasons for it*.     Meaning we have links in the ADR to Nathan's doc and/or the feature docs that led us to this with a narrative on why.

packages/db/src/models/granule-duplicates.ts
npauzenga: yeah, just take a look at packages/db/tests/models for examples.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const up = async (knex: Knex): Promise<void> => {
+  await knex.schema.createTable('granule_duplicates', (table) => {
```
Jkovarik: :ant: I'd like to take a shot at calling it something other than 'granule_duplicate' - the data structure we're tracking here is a granule version grouping - duplicate carries an overloaded connotation that might be unfortunate in user-space, and there's not necessarily  to tie the data structure to the use directly. 

How about granule_group/s table?

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Jkovarik: I think the thing I'm not seeing addressed here is the API object -> database object translation  and/or the API design around hiding/un-hiding (e.g. do we do it as a function of something in the `granule` object (e.g. if it's a PUT/PATCH and a change we do a thing) or a different endpoint and divorce the granule status from the granule object (for the record I'm a fan of the latter).   

I think that's fine here, in that it can be a implementation detail in downstream tickets, we can choose to change it if the approach needs modified, but we need to make sure the work/implications are captured in the epic.     Can we commit to making sure the epic has been reviewed for missing ticket in light of that question work?

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Jkovarik: This falls solidly into the "what states does the  group state machine support" realm - if we're going with a separate table and not a parent/child relationship I don't know if we have to answer that question here.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
```diff
@@ -0,0 +1,31 @@
...
+    table
+      .increments('cumulus_id')
+      .primary();
+    table
+      .integer('granule_cumulus_id')
+      .references('cumulus_id')
+      .inTable('granules')
+      .notNullable();
```
Jkovarik: I think we're going to want `.onDelete('CASCADE');` here, *so long as* there's no issue with the group having only hidden granules being a problem.

packages/db/src/models/granule-duplicates.ts
Jkovarik: See: comment in the migration about ON CASCADE.  

We should definitely have the DB enforce removal here if the granule is being deleted.

Jkovarik (changes requested):
This looks good @Nnaga1 !      Asking for: 

-  Schema update 
-  Consider table name question 
-  Deletion cascade on FK granule_cumulus_id
-  Approach sanity check re: outstanding epic tickets 

Happy to pair, help or comment on demand.      Good work!

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: I thought it would be default adding indexes to the primary/important keys of the table, I can remove it or change them if its deemed extraneous

CHANGELOG.md
Nnaga1: I could also mention how it can be changed based on understanding of requirements (i.e. getting confirmation on if 1 active/multiple actives)

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: `if we're going with a separate table and not a parent/child relationship I don't know if we have to answer that question here.` if we're going with a separate table then the group state machine would support having multiple actives, if parent/child in the granules' table + flag then it wouldn't (the other implementation)

Jkovarik: What I mean is a separate table allows for either option, going with a parent granule and inferring the relationship can *only* allow for 1 active, some number of hidden.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: ` do we do it as a function of something in the granule object` I know users mentioned upon reingest they would want the former granule to be hidden and the newer one to be active, so this part seems inevitable at least in the context of ingest/re-ingest, but for other cases of promotion/demotion I would support having different endpoints that deals with the duplicates table + granules files location changes (since it would be updating the duplicates table + the granules files locations key/path values)

Nnaga1: `we can choose to change it if the approach needs modified, but we need to make sure the work/implications are captured in the epic.` I can go through and confirm/create new tickets as needed for whats missing related to this

CHANGELOG.md
Nnaga1:
```
postgres=> \timing on
Timing is on.
postgres=> CREATE TABLE granule_duplicates (
postgres(>   cumulus_id SERIAL PRIMARY KEY,
  granule_cumulus_id INTEGER NOT NULL REFERENCES granules(cumulus_id),
  status CHAR(1) NOT NULL,
  group_id VARCHAR NOT NULL
);  cumulus_id SERIAL PRIMARY KEY,
postgres(>   granule_cumulus_id INTEGER NOT NULL REFERENCES granules(cumulus_id),
  status CHAR(1) NOT NULL,
  group_id VARCHAR NOT NULL
);  granule_cumulus_id INTEGER NOT NULL REFERENCES granules(cumulus_id),
postgres(>   status CHAR(1) NOT NULL,
  group_id VARCHAR NOT NULL
);  status CHAR(1) NOT NULL,
postgres(>   group_id VARCHAR NOT NULL
);  group_id VARCHAR NOT NULL
postgres(> ););
CREATE TABLE
Time: 2243.190 ms (00:02.243)
postgres=> CREATE INDEX IF NOT EXISTS granule_duplicates_granule_cumulus_id_index 
postgres->   ON granule_duplicates(granule_cumulus_id);  ON granule_duplicates(granule_cumulus_id);
CREATE INDEX
Time: 160.547 ms
postgres=> CREATE INDEX IF NOT EXISTS granule_duplicates_group_id_index 
postgres->   ON granule_duplicates(group_id);  ON granule_duplicates(group_id);
CREATE INDEX
Time: 135.245 ms
postgres=> SELECT COUNT(*) FROM granules;
   count   
-----------
 163328300
(1 row)

Time: 48837.799 ms (00:48.838)
postgres=> SELECT COUNT(*) from granule_duplicates;
 count 
-------
     0
(1 row)

Time: 46.545 ms
postgres=> 
```

off of cumulus-lp just now 🙌

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: I can mention in the ADR this option allows for both and if its not deemed the case, then it can be changed or something

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: `Will we need to fetch a granuleDuplicates record every time we get a granule to see if that granule is the active one?` yes and since the granule_cumulus_id is linked to granules you can just query this new table and then join if neccessary (for file locations) where granule_cumulus_id = granules cumulus_id

CHANGELOG.md
npauzenga: Yup, that tracks! Thanks for checking that out. I know it was fairly obvious that this wouldn't be a concern but seeing proof is really nice 😄 

I don't think we need any separate SQL/manual migration steps here. We can probably clean some of that up.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
npauzenga: So is that what we'll need to do for every GET request of a granule? Do we add a query to this table in every granule endpoint? This line of questioning is the same as the above, I just need to understand how this table is used.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
npauzenga: Honestly I'm not sure what our posture is on adding indices in advance. The speed improvements on primary IDs are likely worth it, especially for new tables where I'm not sure we'd have a drawback. Seems worth doing as a matter of good practice.

I'm curious about the fields we're indexing though. If we make group_id a number, that might be better but, more to the point, are these the fields we'll likely be using in `WHERE` clauses when we look up `granuleDuplicates` records?

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
npauzenga: Thinking about plural v. singular: Can one granule have many groups? Additional questions if 'yes' 😄

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Jkovarik: Thoughts: We definitely want an index on the FK for join query speed alone.        The query on this table is likely going to look something like 

WITH target_group AS (
  SELECT gg.group_id
  FROM granule_groups gg
  JOIN granules g ON g.cumulus_id = gg.granule_cumulus_id
  WHERE g.granule_id = $1         -- ← bind your granule_id here
  LIMIT 1                          -- if each granule belongs to exactly one group
)
SELECT g.*

with this structure, *or* alternately direct code logic given a cumulus_id in a DB return.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Jkovarik: "do we do it as a function of something in the granule object I know users mentioned upon reingest they would want the former granule to be hidden and the newer one to be active,"

Sure, but that's going to be part of a workflow lambda business logic, not a framework thing.       The current high level  plan is to have the duplicate detection lambda/logic detect a collision and immediately 'demote' the collision granule, then do ingest things as normal.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: `We definitely want an index on the FK for join query speed alone.` so on granules(granule_id) and granule_groups(granule_cumulus_id) in one index? should I keep the ones I have too? I wasn't really that sure on these either

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: I can agree with that ^, `I think the thing I'm not seeing addressed here is the API object`: to this I think separate endpoints that deal with promotion/demotion/active status, as you mention, keeping the active status separate from what exists rn

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: `So is that what we'll need to do for every GET request of a granule? Do we add a query to this table in every granule endpoint?` I would assume yes for the ones that need active status, otherwise it wouldnt need to

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: `Can one granule have many groups? Additional questions if 'yes' ` Nathan confirmed the answer to this as no in the slack questions I asked him last week

packages/db/src/models/granule-duplicates.ts
Nnaga1: I think im going to do what Nate's comment said above and try to get it to adopt the BasePgModel class, this model isn't really special so it can just adopt the functions it already has, this isn't finalized yet ^ (need to make those changes)

packages/db/src/models/granule-duplicates.ts
Nnaga1: I don't see why it would hurt to add this to the upsert, I can add this 👍

packages/db/src/models/granule-duplicates.ts
Nnaga1: I think endpoints/api or lambda functionality cover that, these are just mainly small pg helper stuff for units iirc

CHANGELOG.md
Nnaga1: added 👍

packages/db/src/models/granule-duplicates.ts
Nnaga1: added 👍

packages/db/src/models/granule-duplicates.ts
Nnaga1: adopted BasePgModel 👍

packages/db/src/models/granule-duplicates.ts
Nnaga1: added tests 👍

packages/db/src/models/granule-duplicates.ts
Nnaga1: added the on cascade 👍

packages/db/src/migrations/20250915152638_create_granule_groups_table_and_index.ts
```diff
@@ -0,0 +1,36 @@
...
+      .comment('Granule active status')
+      .notNullable();
+    table
+      .integer('group_id')
+      .comment('Granule duplicate group id')
+      .notNullable();
+    table
+      .timestamps(false, true);
+    table
+      .unique(['granule_cumulus_id']);
```
jennyhliu: A unique constraint automatically creates a unique index, so adding another one could result in two unique indexes on the same column with different names.

packages/db/src/migrations/20250915152638_create_granule_groups_table_and_index.ts
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  await knex.raw('CREATE INDEX IF NOT EXISTS granule_groups_granule_cumulus_id_index ON granule_groups(granule_cumulus_id)');
+  await knex.raw('CREATE INDEX IF NOT EXISTS granule_groups_group_id_index ON granule_groups(group_id)');
+};
+
+export const down = async (knex: Knex): Promise<void> => {
+  await knex.raw('DROP INDEX IF EXISTS granule_groups_granule_cumulus_id_index');
+  await knex.raw('DROP INDEX IF EXISTS granule_groups_group_id_index');
```
jennyhliu: No need to drop indexes before dropping table.

packages/db/src/migrations/20250915152638_create_granule_groups_table_and_index.ts
```diff
@@ -0,0 +1,36 @@
+import { Knex } from 'knex';
+
+export const up = async (knex: Knex): Promise<void> => {
+  await knex.schema.createTable('granule_groups', (table) => {
```
jennyhliu, line 4: Should the name `granule_groups` be more specific, like `granule_duplicate_groups`, `granule_replacement_groups`?
How do we query the association of the granules within the same group?

packages/db/src/migrations/20250915152638_create_granule_groups_table_and_index.ts
jennyhliu: Will granule_cumulus_id be unique? Can a granule associated with multiple groups and multiple active granules?
If it's unique, why don't we use it  as a primary key?

Nnaga1: `Will granule_cumulus_id be unique?` it should be unique the same way cumulus_id of granules is unique

`Can a granule associated with multiple groups and multiple active granules?` A granule I think we decided we are not explicitly allowing only 1 active (both cases 1 active + multiple actives, hence the group_id), but its been confirmed a granule cannot be part of multiple groups

Nnaga1: `If it's unique, why don't we use it as a primary key?` I dont see why we cant, how would I do that in the migration file?

packages/db/src/migrations/20250915152638_create_granule_groups_table_and_index.ts
Nnaga1: I originally had it as `granule_duplicates` but in this comment I think we agreed with `granule_groups`: https://github.com/nasa/cumulus/pull/4066#discussion_r2349983225

Jkovarik:
This PR looks to be being continued in https://github.com/nasa/cumulus/pull/4069, FYI for reviewers.

packages/db/src/migrations/20250915152638_create_granule_groups_table_and_index.ts
```diff
@@ -0,0 +1,31 @@
...
+      .comment('Granule active state')
+      .notNullable();
+    table
+      .integer('group_id')
+      .comment('Granule duplicate group id')
+      .notNullable();
+    table
+      .timestamps(false, true);
+    table
+      .unique(['granule_cumulus_id']);
```
jennyhliu: This `unique` constraint should be removed.  A primary key is already: NOT NULL, UNIQUE, Indexed

packages/db/src/migrations/20250915152638_create_granule_groups_table_and_index.ts
```diff
@@ -0,0 +1,31 @@
+import { Knex } from 'knex';
+
+export const up = async (knex: Knex): Promise<void> => {
+  await knex.schema.createTable('granule_groups', (table) => {
+    table
+      .integer('granule_cumulus_id')
+      .references('cumulus_id')
+      .inTable('granules')
+      .notNullable()
```
jennyhliu: notNullable() is redundant

packages/db/src/migrations/20250915152638_create_granule_groups_table_and_index.ts
```diff
@@ -0,0 +1,31 @@
...
+      .onDelete('CASCADE')
+      .primary();
+    table
+      .specificType('state', 'char(1)')
+      .comment('Granule active state')
+      .notNullable();
+    table
+      .integer('group_id')
```
jennyhliu, line 12: I suggest to put group_id as the second column as it's more likely to be queried

Nnaga1: changed 👍

packages/db/src/migrations/20250915152638_create_granule_groups_table_and_index.ts
Nnaga1: removed 👍

packages/db/src/migrations/20250915152638_create_granule_groups_table_and_index.ts
Nnaga1: removed 👍

packages/db/src/migrations/20250912712638_create_granule_duplicates_table.ts
npauzenga: So do we need more DAAC input?

packages/db/src/migrations/20250915152638_create_granule_groups_table_and_index.ts
Jkovarik: Is this removed?  The PR diff still looks like it's there 👀

Jkovarik: Weird, confirmed it's right on the branch.

packages/db/src/models/granule-duplicates.ts
npauzenga: @Nnaga1 I agree with everything before the comma 😄 

That's correct that the endpoint will likely handle the "bulk" portion of "bulk delete". We have a similar structure for Granule Bulk Deletes where the endpoint maps over a list of granules and eventually calls the model's `delete` method (with a bunch of other logic): https://github.com/nasa/cumulus/blob/master/packages/api/src/lib/granule-delete.ts#L93

Whether thanks actually a "bulk" action is debatable but that's likely the same pattern this is intending to follow.

The second part `these are just mainly small pg helper stuff for units iirc`is a little off-base though I think. These models represent the core business logic of adding, getting, removing, and modifying database entries. We should be using these models in our APIs, Tasks, Tests, etc. as they're really database logic layer. So yes, definitely they can be used in units but also anywhere we want to delete a `granuleDuplicates` record.

Jkovarik: Also @mason-t-yates  the intent, at least so far, is that via cascading deletes from the granules table, we don't have an expected use case to bulk remove records from this table (e.g. ungrouping).    The "what happens when I delete to 1 active record in a group" approach still needs to be determined, but optimally that would be a function of promotion/demotion and/or granule deletion logic in the business logic *or* database triggers.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Jkovarik: I think some of that depends on how we decide to handle the db granule->api graunle translation layer and the dashboard presentation updates.    

I'd urge us to avoid presenting this flag status on every granule view, and at *most* show that it's part of a grouping.     

In either case the  data model as proposed here seems to covers the ability to do a join query performant to get that info, either as a specific endpoint, in the translation layer *or* via some other mechanism.

packages/db/src/migrations/20250912712638_create_granule_duplicates_table.ts
Jkovarik: I think we landed on using the model as proposed here because the ideation of the requirement is both in the feature analysis and scenario documents and this allows us to handle whatever it turns into in implementation.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Jkovarik: This is mostly irrelevant as we're now making the FK the primary key (and it will be indexed as a result).

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
npauzenga: >In either case the data model as proposed here seems to covers the ability to do a join query performant to get that info, either as a specific endpoint, in the translation layer or via some other mechanism.

I think that's right and I'm happy to go with the team if everyone is reasonably comfortable here. My point generally is that I'm concerned we don't know how we're planning on actually using this and that gives me pause. If we don't have that planned out, it does seem like the schema may need to be changed as we discover more. But I guess that's also not the end of the world.

packages/db/src/models/granule-duplicates.ts
Nnaga1: `These models represent the core business logic of adding, getting, removing, and modifying database entries. We should be using these models in our APIs, Tasks, Tests, etc. as they're really database logic layer. So yes, definitely they can be used in units but also anywhere we want to delete a granuleDuplicates record.` actually TIL, after doing all the Search class stuff removing ES I thought this was pretty much mainly used for tests, but that makes sense

packages/db/src/migrations/20250912712638_create_granule_duplicates_table.ts
npauzenga: Ok, fair enough. I've pushed on this enough to feel confident 😄

Nnaga1: yea if it needs to be tightened based on changing reqs this definitely can be changed, but rn it allows pretty much all edge cases which I think was the intention (and handles IART-923)

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
npauzenga: Oh I didn't consider the schema changes. @Nnaga1 is that correct?

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: `I think that's right and I'm happy to go with the team if everyone is reasonably comfortable here. My point generally is that I'm concerned we don't know how we're planning on actually using this and that gives me pause.` I agree with you in that, it does seem like this kindof came out of left field how complicated this is (rather than just a flag).

packages/db/src/models/granule-duplicates.ts
npauzenga: There also may be instances where it ISN'T true that the model isn't our mechanism for updating tables but I'd argue that that's bad.

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: `I'd urge us to avoid presenting this flag status on every granule view, and at most show that it's part of a grouping.
In either case the data model as proposed here seems to covers the ability to do a join query performant to get that info, either as a specific endpoint, in the translation layer or via some other mechanism.` I guess this can be scoped out a bit more in the following tickets (or even changed)

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Nnaga1: `we're now making the FK the primary key` yes theres no cumulus_id anymore, just granule_cumulus_id with all of:

```
    table
      .integer('granule_cumulus_id')
      .references('cumulus_id')
      .inTable('granules')
      .onDelete('CASCADE')
      .primary();
```

packages/db/src/migrations/20250915152638_create_granule_duplicates_table_and_index.ts
Jkovarik: That's fair.     It might be worth a huddle to talk more about the expected group state machine representation  if that's a point of discomfort.    It's worth making sure we didn't miss anything in the prior convos. 

For what it's worth, for the most part I think the questions outstanding with respect to the data model/state machine rules are:

*  Are *we* (not and perhaps not importantly users) going to need to have an internal state at any point where there are two active granules in a grouping
* Should there be restrictions on the state, particularly things like there MUST be one granule in an active state in a group. 

It feels likely we don't have certainty re: a simple parent/child for graph cases, particularly at the edges.           @jennyhliu's point that we're likely to need some sort of stored metadata about what put a granule in the grouping is a good example of the sort of schema updates I feel we should expect.

Nnaga1:
```
Are we (not and perhaps not importantly users) going to need to have an internal state at any point where there are two active granules in a grouping
Should there be restrictions on the state, particularly things like there MUST be one granule in an active state in a group.
```
getting this scoped out really affects the active/hidden way of representation based on the content in the ADR for this, which is why its pretty flexible right now. Nathan partially confirmed that 1 active is expected but I guess its kind of up to us if that's **enforced**

Jkovarik: I also want to highlight that there's a difference between the *user* expected states and valid intermediate states. 

For what it's worth, the other state I'm concerned with is no active, all 'hidden' and that sort of thing.

Jkovarik (approved):
I'm satisfied this meets the existing ticket requirements with respect to this PR thanks for addressing my feedback @Nnaga1 , good work.     Approving, but would like a review if other changes are made 👍🏻

npauzenga (approved):
This is great @Nnaga1. Thanks for taking your time and working through all of the conversations here. This PR is really setting the stage for the entire epic so I'm glad we worked out so much of it. Appreciate you hanging in 😄

